PR #139, opened 2014-04-23
Create postgres.js

brianc:
Placeholder file to begin discussion/work for adding node-postgres instrumentation.


jwhitehorn:
:+1: This is literally the only thing keeping me from upgrading my New Relic subscription.

PR #178, opened 2014-10-20
Log the actual error for config not being read

knownasilya:
Basically it's impossible to debug, especially if the error only happens in a Docker container (which it is, for me).


hayes:
I'll try to get this released this week. Sorry for the delay.

wraithan:
1.14.6 contains this patch. Thanks!

PR #185, opened 2014-12-04
Fix domain circular reference

syrnick:
Prevent newrelic agent from trying to serialize process.domain to JSON.


hayes:
I'll try to get this released this week, thanks for the fix!

wraithan:
Redacted list is not where this belongs, that is strictly for things that could contain passwords that we don't want pushed. Next release will contain a fix for this though. Thanks for bringing this to our attention.

wraithan:
1.14.6 contains a fix for this

PR #191, opened 2015-01-15
This fixes a test failure when NODE_ENV was set in the environment when ...

markstos:
...running the tests.

For example:

   NODE_ENV=foo npm test

To get consistent results when quering the state of the environment, the
environment needs to start out in a consistent state.


hayes:
released in 1.14.7

PR #192, opened 2015-01-15
Support Node >= 1.0 and io.js.

markstos:
io.js is out now and although it is version 1.0.1, trying to require
'newrelic' with it was crashing:

     message = "New Relic for Node.js requires a version of Node equal to or\n" +
               "greater than 0.6.0. Not starting!"

There was a bug in a version checking that was only checking that
the minor version was less than 6, but was ignoring the major version,
so it failed to notice that that 1.0.1 > = 0.6.0.

Before and after checking was done by loading the iojs REPL and trying
this:

   var nr = require('./index.js');

This crashes before the change and works after it.

Ref: https://iojs.org/

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,7 +13,9 @@ try {
   logger.debug("Process was running %s seconds before agent was loaded.",
                process.uptime())
 
-  if (process.version && process.version.split('.')[1] < 6) {
+  var major =  parseInt(process.version.split('.')[0]);
+  var minor =  process.version.split('.')[1];
+  if (process.version && (major === 0) && (minor < 6)) {
     message = "New Relic for Node.js requires a version of Node equal to or\n" +
               "greater than 0.6.0. Not starting!"
 
--- a/index.js
+++ b/index.js
@@ -15,9 +15,9 @@ try {
 
   var major =  parseInt(process.version.split('.')[0]);
   var minor =  process.version.split('.')[1];
-  if (process.version && (major === 0) && (minor < 6)) {
+  if (process.version && (major === 0) && (minor < 8)) {
     message = "New Relic for Node.js requires a version of Node equal to or\n" +
-              "greater than 0.6.0. Not starting!"
+              "greater than 0.8.0. Not starting!"
 
     logger.error(message)
     throw new Error(message)


asilvas:
+1

index.js
```diff
@@ -13,7 +13,9 @@ try {
   logger.debug("Process was running %s seconds before agent was loaded.",
                process.uptime())
 
-  if (process.version && process.version.split('.')[1] < 6) {
+  var major =  process.version.split('.')[0];
```
asilvas: parseInt

index.js
```diff
@@ -13,7 +13,9 @@ try {
...
+  var minor =  process.version.split('.')[1];
+  if (process.version && (major === 0) && (minor < 6)) {
```
asilvas: `minor` works because of it's implicit conversion, but `major` is doing a type and value equality, so it'll need to be converted to number.

markstos:
I found a second bug here. The code checks for > 0.6.0, but the README says: "New Relic for Node is only supported on Node.js 0.8 and newer".  I presume the README is correct here.

markstos:
I've amended the first commit to add parseInt-- thanks for that catch. I've added a second commit to bump the version requirement to 0.8.0 to match the README. I presume you wanted them to agree.

hayes:
Thanks for contributing this change.  We are working to properly support io.js (and node 0.12) and we  know there are parts of the agent that will not work yet.  1.0 of io.js is still an alpha (similar to 0.11.15 which is an rc for node 0.12), so we will not officially support it until it has a release that is considered stable by the maintainers of the project. That being said, the agent being limited to node version in the 0.x range is a bug that this PR would fix.  We want to also allow people to run the agent on 0.6 as well, we do not provide support or guarantee functionality on 0.6, but most of the agent should work without any issues, so we do not feel it makes sense to artificially restrict its usage.

I'll try to get this merged and included in today's release.

hayes:
We like to credit community contribution in our release notes, do you have a preferred way for us refer to you (most people usually use their github name, or full name).

markstos:
My preferred reference is "Mark Stosberg". Thanks!

hayes:
released in 1.14.7

markstos:
Thanks!

PR #193, opened 2015-01-28
fallback for empty cursor.collection

amenadiel:
Since I upgraded to node's MongoDB connector v2 (specifically, my package.json says "mongodb": "^2.0.13") I started getting a fatal error

```sh
TypeError: Cannot read property 'collectionName' of undefined
    at mongoCursorOperationProxy (/var/www/apppath/node_modules/newrelic/lib/instrumentation/mongodb.js:125:41)
```

The errors line is

```js
var collection = cursor.collection.collectionName
```

and it seems that `cursor.collection` is empty now. BUT, I saw that cursor.ns has the form `databaseName.collectionName`, so I wrote a fallback to use that property instead if needed.

### DIFF
--- a/lib/instrumentation/mongodb.js
+++ b/lib/instrumentation/mongodb.js
@@ -123,7 +123,7 @@ function wrapCursorOperation(tracer, operationName) {
   return function cls_wrapCursorOperation(operation) {
     return tracer.segmentProxy(function mongoCursorOperationProxy() {
       var cursor = this
-      var collection = cursor.collection.collectionName
+      var collection = cursor.collection ? cursor.collection.collectionName : cursor.ns.split('.')[1];
       var terms = cursor.selector
 
       if (!tracer.getTransaction()) {


seriousben:
Having the same problem

hayes:
Thanks for this PR.  Unfortunately supporting the mondgodb >= 2.0 is a bit more complicated than this simple change.  I am working on a full rewrite of our internal apis including the whole mongo instrumentation.  We are hoping to release this very soon, and it should resolve the issues you are seeing in newer versions of the mongo driver. This is still a couple of weeks from release, but we are actively working on it.

best,
Michael

seriousben:
Thanks a lot for your feedback

seriousben:
Isn't there some way to just not break our apps even without supporting mongodb>=2.0?

hayes:
We are still not quite ready to officially release the version of the node agent that has the fix for mongo 2.0 in it, but if you are interested in getting access to a beta version, please shoot me an email at mhayes@newrelic.com

seriousben:
Is new release including fix for this?

hayes:
Yep, should be fixed. Please let me know if you run into any additional
issues

seriousben:
Works great thanks a lot.
Maybe a small issue, I am seeing 3 different database for mongo while I am only talking to a single one. Might be related to connection pooling or something like that?

hayes:
that sounds odd. If you send me a link (mhayes@newrelic.com) I'll be happy to take a look

PR #194, opened 2015-02-18
Don't crash when using mongodb driver 2.0.16

vkarpov15:
One of the companies that's been helping QA the new version of the mongodb driver uses New Relic in their test environment, and turns out the New Relic agent crashes ungracefully with the new version of the driver. See LearnBoost/mongoose#2652 and mongodb/node-mongodb-native#1253. We added the ability to get the collection name from the cursor back into the driver for v2.0.16. I'm not sure if this is the correct way to handle this on your end, but this tweak seems to be enough to at least make the agent stop crashing.

### DIFF
--- a/lib/instrumentation/mongodb.js
+++ b/lib/instrumentation/mongodb.js
@@ -123,7 +123,9 @@ function wrapCursorOperation(tracer, operationName) {
   return function cls_wrapCursorOperation(operation) {
     return tracer.segmentProxy(function mongoCursorOperationProxy() {
       var cursor = this
-      var collection = cursor.collection.collectionName
+      var collection = cursor.collection ?
+        cursor.collection.collectionName :
+        cursor.namespace.collection
       var terms = cursor.selector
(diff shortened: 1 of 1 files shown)


vkarpov15:
Actually, woops, didn't look at the PRs and so didn't see discussion on #193. Feel free to close if you deem it necessary. But hopefully the fact that you can get collection name from cursor helps :)

hayes:
Thanks for looking in to this, we have an updated version of our mongo instrumentation that we are releasing very soon that supports getting collection names across various versions of mognodb.  I'll add a note here when its released.

vkarpov15:
Much thanks, I'm looking forward to the release. In the meantime, feel free to report any issues you run into on [our JIRA board](https://jira.mongodb.org/browse/NODE).

hayes:
a fix for this is released in 1.17.0

vkarpov15:
Thanks!

PR #196, opened 2015-06-11
Small bugs with custom events.

shezarkhani:
* Fixed the logging for number of dropped custom events.
* Fixed the reservoir sampling size for custom insight events if it's not enabled.


wraithan:
@shezarkhani Bummer, I missed that other call. Fortunately that is only used in the case where custom events were disable after the process had collected some events and would only affect one harvest. 

The logger is a problem as well. I'll try to get this merged into our next release (that I anticipate will be next week)

wraithan:
This has been merged and released in 1.20.2

PR #200, opened 2015-09-12
remove 'bound' from function name to avoid error in node >=4.x

esco:
Newer versions of V8 that support ES2015 automatically prepend the word "bound" to the name of a function that was created with `bind`. Node 4.0.0 uses V8 v4.5 which has this behavior. This breaks express and connect instrumentation in the newest version of node because the following code uses the name value from bound functions:

```javascript
var wrapped = new Function(
      'tracer', '__NR_handle', 'wrappedHandle', 'path', 'template',
      'return function ' + name + arglist + handlerTemplate.toString().substring(11)
    )(tracer, __NR_handle, wrappedHandle, path, template)
```
>https://github.com/newrelic/node-newrelic/blob/6ef82e8dcc0eee2fc3f847b11cd5e8bb41080923/lib/instrumentation/express.js#L279

>Ex of binding: https://github.com/newrelic/node-newrelic/blob/6ef82e8dcc0eee2fc3f847b11cd5e8bb41080923/lib/instrumentation/express.js#L573

When the string is eval'ed it tries to create a function that has two names `function bound foo(){}`.

This fix removes all occurrences of "bound" in the function name unless the name of the function is "bound" in which case it keeps the last occurrence.

### DIFF
--- a/lib/instrumentation/connect.js
+++ b/lib/instrumentation/connect.js
@@ -121,7 +121,7 @@ module.exports = function initialize(agent, connect) {
     // allow for lookup.
     /*eslint-disable no-eval*/
     var wrapped = eval(
-      '(function(){return function ' + name + arglist +
+      '(function(){return function ' + name.replace(/\s|bound(?!$)/g,'') + arglist +
       template.toString().substring(11) + '}())'
     )
     /*eslint-enable no-eval*/
--- a/lib/instrumentation/express.js
+++ b/lib/instrumentation/express.js
@@ -276,7 +276,7 @@ module.exports = function initialize(agent, express) {
     // name to allow for lookup.
     var wrapped = new Function(
       'tracer', '__NR_handle', 'wrappedHandle', 'path', 'template',
-      'return function ' + name + arglist + handlerTemplate.toString().substring(11)
+      'return function ' + name.replace(/\s|bound(?!$)/g,'') + arglist + handlerTemplate.toString().substring(11)
     )(tracer, __NR_handle, wrappedHandle, path, template)
 
     wrapped[ORIGINAL] = __NR_handle


urish:
+1

senorcris:
+1

ChALkeR:
@esco See https://github.com/nodejs/node/issues/2798#issuecomment-139796430

esco:
@ChALkeR thanks

wraithan:
This patch isn't complete enough to fix it for all cases. There are other things (such as `get` and `set`) that can be prepended as well.

Also any mangling of the function name should occur in the `mangle` function. I'll be working with one of our new engineers to get a fix in today, but the release may not happen until later this week to properly test the agent and see if there are other cases where this will fail.

In general, a patch without a test wont be accepted. Because of the huge diverse set of environments we run in, we need to have tests for any new stuff to make sure it all functions properly across that set.

esco:
@wraithan 

Awesome, its great to see newrelic taking node updates so rigorously, thank you for supporting node 4.0. I would've updated tests but there didn't seem to be any relevant tests in place and was hoping you could give guidance on that.

 `get` and `set` bring up another issue. I tried replicating this problem with those before making the fix. Every time you access a get or set, you'll get their return value (as expected). How do you plan on instrumenting accessors (get/set)?

Looking forward to the patch :-) 

Thanks!

esco:
@wraithan 

Looks like get/set don't have "get" or "set" prepended to the function name :-)

Using `getOwnPropertyDescriptor` we can get a reference to the accessor.

```javascript
var obj = {
  get foo() {
    return Math.random() > 0.5 ? 'foo' : 'bar';
  },

  set foo(bar) {
    this.bar = bar;
  }
};
console.log(Object.getOwnPropertyDescriptor(obj, 'foo').get.toString());
console.log(Object.getOwnPropertyDescriptor(obj, 'foo').set.toString());
console.log(Object.getOwnPropertyDescriptor(obj, 'foo').get.bind(this).toString());
```

output
```
function foo() {
    return Math.random() > 0.5 ? 'foo' : 'bar';
  }
function foo(bar) {
    this.bar = bar;
  }
function bound foo() { [native code] }
```

wraithan:
@esco Interesting, that is different than what the spec says, unless the spec has changed since the last time I gave it a look. I'll go through the spec and make sure that changed. In anycase, I'd still prefer to have this in our name mangling function that clears out other strictness problems and such.

wraithan:
Yeah, the spec still says that it should be prepending `set ` and `get `:

http://www.ecma-international.org/ecma-262/6.0/#sec-ecmascript-standard-built-in-objects

Relevant paragraph:

> Every built-in Function object, including constructors, that is not identified as an anonymous function has a name property whose value is a String. Unless otherwise specified, this value is the name that is given to the function in this specification. For functions that are specified as properties of objects, the name value is the property name string used to access the function. Functions that are specified as get or set accessor functions of built-in properties have "get " or "set " prepended to the property name string. The value of the name property is explicitly specified for each built-in functions whose property key is a Symbol value.

wraithan:
I'd rather fix this once and get it right so we don't have another problem when that portion of the spec is implemented in v8

esco:
@wraithan nice catch with the spec. Good point, `mangle` seems like a more appropriate place for this.

kellyrmilligan:
can't find it ,so asking here, when will this module work with express and node >=4.0?

danmactough:
@kellyrmilligan should be good to go as of v1.22.1 I think

wraithan:
@kellyrmilligan @danmactough We do not support 4.0 yet officially. That said we've fixed all known crashers with it, it is more a matter of cleaning up our tests and running it through the gambit to make sure we are confident our agent will perform well on it. Users have been reporting success using it though, and I have it in our benchmarking suite and it hasn't crashed there.

danmactough:
Thanks for the clarification @wraithan. FWIW we haven't seen any issues either.

PR #202, opened 2015-10-17
CAT Documentation

rsteckler:
This documents the ability to enable or disable cross application tracing from the config file.

### DIFF
--- a/lib/config.default.js
+++ b/lib/config.default.js
@@ -494,5 +494,13 @@ exports.config = {
    slow_sql: {
     enabled: false,
     max_samples: 10
-  }
+  },
+  
+  /**
+   * These options control features of New Relic not found elsewhere.
+   * cat enables and disables cross application tracing.  The default is enabled. (https://docs.newrelic.com/docs/apm/transactions/cross-application-traces/cross-application-tracing)
+   */
+   feature_flag: {
+     cat: true
+   }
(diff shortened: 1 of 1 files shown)


rsteckler:
A little context:  After configuring new relic for my node app, I started seeing calls I was making to a third party service start failing.  After a fair amount of debugging, I realized that new relic was adding the x-newrelic-id and x-newrelic-transcation headers to all outbound http requests.  Those headers were making the third party API barf.
I looked through the docs, forum, and StackOverflow looking for a way to disable it and couldn't find anything.  A quick pass through the source showed that the CAT code is checking this undocumented config file setting.

I believe it should be documented because there are cases where it breaks things.

wraithan:
Unfortunately we do not document the feature flags as they are meant to be temporary settings for our users who agree to beta for us to be able to turn on. They are considered private API, so we don't consider it a breaking change when we remove the ability to set them, which happens periodically as we sweep the code base to remove old feature flags.

We have an outstanding feature request to add the ability to toggle CAT headers. But by turning off that feature flag you are also turning off things like synthetics transactions and other things. It is no longer just CAT that is behind that as we've built on top of CAT. Honestly at this point that flag should have already been removed. Your best bet is to contact support and request this as it will let our PMs know this should be a priority.

PR #203, opened 2015-11-16
Add `NEW_RELIC_LOG_ENABLED` as a `BOOLEAN_VARS`

jakecraige:
You could not disable logging via an environment var not set in code because it
would not be converted to a boolean value properly.

This allows it to be set properly from the environment because the
`BOOLEAN_VARS` are converted to booleans.


wraithan:
@jakecraige wow, that's a pretty big whoops! Thanks for surfacing this. I'll chat with our team about getting this into the next release!

wraithan:
@jakecraige this has been merge and released in v1.24.0 Thanks!

PR #204, opened 2015-12-23
Support mysql2

jhollingworth:
We're using [node-mysql2](https://github.com/sidorares/node-mysql2) which is a drop in replacement for [node-mysql](https://github.com/felixge/node-mysql). From testing we've found the existing mysql instrumentation works perfectly for mysql2. This PR just replaces `mysql2` for `mysql` if its loaded.

### DIFF
--- a/lib/shimmer.js
+++ b/lib/shimmer.js
@@ -66,6 +66,8 @@ function _postLoad(agent, nodule, name) {
   if (name === 'pg.js') {
     instrumentation = 'pg'
   } if (name === 'mysql2') {
+    // mysql2 (https://github.com/sidorares/node-mysql2) is a drop in replacement for mysql
+    // which conforms to the existing mysql API. If we see mysql2, treat it as mysql
     instrumentation = 'mysql'
   } else {
     instrumentation = base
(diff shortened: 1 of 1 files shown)


jebu:
+1 to this. We are also dependent on mysql2 and this change works well.

benatkin:
:+1: 

I think it needs versioned tests here: https://github.com/newrelic/node-newrelic/tree/a46631145fd7abb889f6ee9e50dcfcecfe3b0ab1/test/versioned

lykkin:
@jhollingworth thanks for the pr! @benatkin has it right, it needs tests in order to be accepted.  when i get some time i'll add some tests to cover this and roll this into an upcoming release

lykkin:
merged out of band in v1.25.2

terencechow:
I see tests for this but don't see any sql statements showing up in my newrelic apm dashboard

Is mysql2 supposed to work with newrelic? Note I'm using mysql2 promises and using prepared statements exclusively (i.e. pool.execute and not pool.query)

NatalieWolfe:
Hi @terencechow! You should bring that to the support forums: https://support.newrelic.com/. Our TSEs should be able to help debug that.

PR #205, opened 2016-02-10
Return Hapi server object from wrappedConnection

kschat:
This was causing an issue when creating connections with Hapi.

Code to reproduce:

```
require('newrelic');

const Hapi = require('hapi');

const server = new Hapi.Server();
const connection = server.connection();

connection.route({/*...*/}); // Throws NPE
```

I didn't write any tests due to the size of the change. If you'd still like tests let me know and I'll add one.

### DIFF
--- a/lib/instrumentation/hapi.js
+++ b/lib/instrumentation/hapi.js
@@ -286,6 +286,8 @@ module.exports = function initialize(agent, hapi) {
           'connection'
         )
       }
+
+      return plugin
     }
   }
 


wraithan:
@kschat which version of hapi were you on when you saw this issue?

kschat:
Version 13. I also saw the issue with versions 12, and 9.

martinkuba:
@kschat if you can include a test, that would be much appreciated.  Otherwise we cannot accept the PR as is.

martinkuba:
@kschat thanks for the fix.  It has been included in v1.25.5.

PR #208, opened 2016-02-21
Don't reset partialName to original on Express wrappedHandle

itai-codefresh:
The problem:
In Errors page in new relic site, you see an empty url (for example: 'post /').
This happens because 'cleanup' function (line 400) resets the transaction partialName from the full url to the original which is always empty ('/').
The clean up function is being called in case there is an error middleware for a specific handler. so what happens is that the transaction name is being reset and then you don't see the url in the main errors page.

The solution:
remove the line because there is not need to reset the partialName and this fixes the bug.

### DIFF
--- a/lib/instrumentation/express.js
+++ b/lib/instrumentation/express.js
@@ -397,7 +397,6 @@ module.exports = function initialize(agent, express) {
     handle.apply(this, args)
 
     function cleanup() {
-      transaction.partialName = original
       next.apply(this, arguments)
     }
   }


martinkuba:
@itai-codefresh we need to reset the partialName in order to support nested routers.  Can you send me an example Express routing that is giving you this problem?

itai-codefresh:
Sure.
var app = express();
app.use('/api/ok', function(request, response, next){
    response.send("ok!");
    next();
});
app.use('/api/fail', function(request, response, next){
    next(new Error("generated an error"));
});
app.use(function(err, request, response, next){
    return response.status(400).send(err.toString() + " extended error");
});

var server = require('http').createServer(app);
server.listen(30020, function () {
    console.log('Express server listening on 30020');
});

just call /api/fail and then you will see the problem.
cleanUp is being called before the error handler middleware.
you will see that in new relic's error page in column 'URL and type' you will see 'get /'

martinkuba:
@itai-codefresh Thanks for the example.  If you use a route handler instead of a middleware (app.get instead of app.use), then the transaction will be named correctly.  Since any middleware can end the response, arguably we could handle it as well, but it will require more work than the change you suggested.  It's on our list to work on improvements for Express routing, and I will make a note of this.  For now I would suggest that you use route handlers.

itai-codefresh:
@martinkuba Thanks for the reply.
I understand what you are explaining but my example was a simple one.
actually we are using express.Router() which are also not working.
is an example of our usage:
    var app = express();
    var router   = require('express').Router();
    router.get("/second", function(request, response, next){
        next(new Error("my error"));
    });
    app.use('/api/first', router);
    
    app.use(function(err, request, response, next){
        return response.status(400).send(err.toString() + " extended error");
    });

you will also see that also with express.Router it does not work.
express.Router is the very standard way people are writing their application and if this is not supported it is really bad. you can't expect people only using app.get (etc..) when we are talking about one of the bets monitor systems the exists.
Please provide some solution to at least support express.Routers
Thanks.

martinkuba:
@itai-codefresh The wrapHandle() function is there to handle nested routers - when Express enters the sub-router, this is where we grab the mount path.  If the router does not respond (end the response), we remove the path and continue.

With that said, yes, there is an issue here.  When an error handler ends the response, we currently don't know which route/middleware handler the error came from.  Most of the time it's not an issue, but in this case it is - the error comes from the router, but the error handler is outside of the router.

These kinds of things are on our radar, and we will be making improvements soon.  For now you can attach the error handler to the router as well.  The response will then be ended in the router and the transaction name will be correct.

martinkuba:
@itai-codefresh We have just released 1.26.0, which should address this issue.  Please let me know if you still see any problems after upgrade.

itai-codefresh:
@martinkuba Thanks for the response. I have checked the new version and this problem was not resolved in this version.
You can take my previous example and you will see that the problem still exists.
Can you please provide information?
Thanks.

martinkuba:
@itai-codefresh the only thing I can think of is if you are requiring express more than 10 times.  Is that the case?  Did you get a trace that shows where this is coming from?

martinkuba:
@itai-codefresh indeed this would happen if you called require('express') more than 10 times.  We will fix it in the next release.  Thanks for bringing it up to our attention.

itai-codefresh:
Well in our case we split our routers to different files so we must require express to get the Router object.
we use this line: var router = require('express').Router();
so yeah you are correct that we are requiring express more than 10 times, but i believe this is a common use.

The only question i have currently:
Is the current fact going to affect the performance of the server during runtime? i didn't dig into the implementation but having so much listeners sounds not good

martinkuba:
The listener does not do much, and it will be called only as many times as you call require('express').  I ran a quick test, and it has no perceivable effect.  We will release a fix next week at the latest.

martinkuba:
@itai-codefresh The listeners bug is now fixed in the latest release (1.26.1).  I am closing this PR, since I believe we have addressed your issue.

PR #210, opened 2016-03-02
dont assume error middleware is being passed an error

theoephraim:
We use "error-handling" middlewares to do more than just error handling.

Our app uses 2 error handling middlewares - one to handle errors, and another to just handle versioning and response formatting for normal requests.

For example. We just call `next` with a model (or an array of models, or some json including models) and our middleware looks at the requested version and converts the model to a specific formatted output for that model at that specific version.

Anyway, you just need to check if the first arg being passed to the middleware is actually an error!

### DIFF
--- a/lib/instrumentation/express.js
+++ b/lib/instrumentation/express.js
@@ -83,7 +83,7 @@ module.exports = function initialize(agent, express) {
   var interceptor
   // This is the error handler we inject for express4. Yanked from connect support.
   function sentinel(error, req, res, next) {
-    if (error) {
+    if (error && error instanceof Error) {
       var transaction = agent.tracer.getTransaction()
       agent.errors.add(transaction, error)
     }


lykkin:
Hey, thanks for the PR! Unfortunately, there is no guarantee made that the errors being passed into the error handler will be proper errors.  We have a few refactors around express instrumentation coming up, and we will keep this use case in mind when addressing error handling.

theoephraim:
Seems to me that making the assumption that they are errors is just as much of a problem as not capturing some of them that are not proper errors?

I assume I'm not the only person using error handling middlewares this way...
Anyway, we'll continue to use our fork for now, but please keep this in mind during your refactor!

martinkuba:
@theoephraim We have just released 1.26.0, which should address this issue.  The error will not be reported if you have an error handler that handles the error, i.e. it either responds with a non-error HTTP status or calls next() without the error.  Please let me know if you still see any problems after upgrade.

martinkuba:
@theoephraim I believe we have addressed your issue - I am closing this PR.  If you still see an issue, feel free to open another PR or submit a ticket with our support.

PR #212, opened 2016-03-10
Adding ioredis support.

guilhermef:
Ioredis is the top recommended client for Redis, and It's the only one that supports Sentinel and cluster.
Right now a few tests are broken, I'll change the PR as soon as it passes.
I'll just leave this here to get some help from the community on a few issues.

### DIFF
--- a/lib/instrumentation/ioredis.js
+++ b/lib/instrumentation/ioredis.js
@@ -1,0 +1,51 @@
+'use strict'
+
+var stringifySync = require('../util/safe-json').stringifySync
+var shimmer = require('../shimmer')
+var urltils = require('../util/urltils.js')
+var recordRedis = require('../metrics/recorders/redis.js')
+var REDIS = require('../metrics/names').REDIS
+
+
+module.exports = function initialize(agent, redis) {
+  var tracer = agent.tracer
+
+  shimmer.wrapMethod(
+    redis && redis.prototype,
+    'redis.prototype',
+    'sendCommand',
+    function wrapSendCommand(original) {
+      return tracer.wrapFunction(
+        REDIS.OPERATION + 'Unknown',
+        recordRedis,
+        original,
+        wrapper
+      )
+    }
+  )
+
+  function wrapper(segment, args, bind) {
+    segment.name = REDIS.OPERATION + (args[0].name || 'unknown')
+    if (args[0].args && typeof args[0].args !== 'function') {
+      urltils.copyParameters(agent.config,
+        {key: stringifySync(args[0].args[0], 'Unknown')}, segment.parameters)
+    }
+
+    // capture connection info for datastore instance metric
+    segment.port = this.connector.options.port
+    segment.host = this.connector.options.host
+    args[0].promise._then = (function(promise, original){
+      return function(){
+        if (typeof arguments[0] === 'function') {
+          arguments[0] = bind(arguments[0], true, true)
+        }
+        if (typeof arguments[1] === 'function') {
+          arguments[1] = bind(arguments[1], true, true)
+        }
+        return original.apply(promise, arguments)
+      }
+    })(args[0].promise, args[0].promise._then)
+
+    return args
+  }
+}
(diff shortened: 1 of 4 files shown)


guilhermef:
It now adds support do ioredis and has the same integrations tests as the other Redis module.

pedrosnk:
:+1:  This PR is really useful for me and my projects

martinkuba:
@guilhermef thanks for doing this, nice work.  It looks like it is more or less a drop in for the redis module.  We should be able to include it in one of our upcoming releases.

guilhermef:
thanks @martinkuba .

philidem:
Is there a newrelic recommendation on how to utilize this change while we're waiting on an official merge? I could modify `require('newrelic/lib/instrumentations')` and also copy `ioredis.js` to `newrelic/lib/instrumentation/ioredis.js` after running `npm install` but not sure if there is a more preferred approach.

martinkuba:
@philidem yes, that's what you would have to do for now.  Another option would be to fork and install from Github.

philidem:
This is the code that I used to manually instrument ioredis:
```javascript
const newrelic = require('newrelic');

const Redis = require('ioredis');

let orig_sendCommand = Redis.prototype.sendCommand;

let nextId = 0;

Redis.prototype.sendCommand = function(command, stream) {
    let promise = orig_sendCommand.call(this, command, stream);
    if (promise) {
        promise.finally(newrelic.createTracer('redis ' + command.name, () => {
            // nothing to do
        }));
    }

    return promise;
};
```

It seems to be working but if you see anything wrong with this approach then please let me know.

lib/instrumentation/ioredis.js
```diff
@@ -0,0 +1,51 @@
...
+    segment.name = REDIS.OPERATION + (args[0].name || 'unknown')
+    if (args[0].args && typeof args[0].args !== 'function') {
+      urltils.copyParameters(agent.config,
+        {key: stringifySync(args[0].args[0], 'Unknown')}, segment.parameters)
+    }
+
+    // capture connection info for datastore instance metric
+    segment.port = this.connector.options.port
+    segment.host = this.connector.options.host
+    args[0].promise._then = (function(promise, original){
```
martinkuba: We probably do not want to represent these as callback segments (the bind() call is meant specifically for callbacks).  Also, wrapping 'then' here will not work well with chained promises, as far as how they would be displayed in traces.

We are planning to make our instrumentation of promises better, which would take care of this in the future.  For now, I would take this out.  It will make any new segments from inside the 'then' callback siblings of the original call, but at least it will be consistent.

martinkuba:
@guilhermef See my comment about wrapping 'then'.  In addition, ioredis allows using callbacks instead of promises, which would need more work.  We think we can address all of this by instrumenting Bluebird better.  If you take out wrapping 'then', we will accept the PR.

guilhermef:
Sure, I'll change the PR this weekend.
Em 31/03/2016 20:58, "Martin Kuba" <notifications@github.com> escreveu:

teddychan:
is  there any plan to merge it and release?
as we are using ioredis heavily and need to be supported.

pungoyal:
:+1: we need this support as well

pwnsrma:
This is really useful @guilhermef :+1:

lykkin:
This was merged out of band in v1.26.2, thanks so much!

wong2:
What should I do to use this support?

guilhermef:
@wong2 it should work like this: https://github.com/guilhermef/node-newrelic/blob/27fef4d039453cd108dc485fa7749a959ada2748/test/integration/instrumentation/ioredis.tap.js

wong2:
@guilhermef yeah I find it out, it's because I'm using async/await

comtaler:
@wong2 I am using async/await for ioredis as well. is there a way to make it work?

NatalieWolfe:
@wong2 @guilhermef The New Relic Node Agent supports async/await on [Node 8 since 2.3.0](https://github.com/newrelic/node-newrelic/blob/master/NEWS.md#230-2017-10-16). If you're having problems with async/await support in the Node Agent, I recommend checking our [support website](https://support.newrelic.com/) and maybe opening a post on our forums where our support engineers can help you.

PR #213, opened 2016-03-25
API shutdown method

echmykhun:
Added method to API to have possibility to forcibly send data

### DIFF
--- a/api.js
+++ b/api.js
@@ -782,6 +782,10 @@ API.prototype.recordCustomEvent = function recordCustomEvent(eventType, attribut
  * @param {function}[callback]  - callback function, that runs when data are already sent
  */
 API.prototype.sendData = function sendData(callback) {
+  var metric = this.agent.metrics.getOrCreateMetric(
+      NAMES.SUPPORTABILITY.API + '/sendData'
+  );
+  metric.incrementCallCount();
     var self = this;
     self.agent.collector.connect(function (error) {
         if (error) {
--- a/api.js
+++ b/api.js
@@ -777,6 +777,27 @@ API.prototype.recordCustomEvent = function recordCustomEvent(eventType, attribut
   this.agent.customEvents.add([instrinics, attributes])
 }
 
+/**
+ * Forcibly sends data to monitoring
+ * @param {function}[callback]  - callback function, that runs when data are already sent
+ */
+API.prototype.sendData = function sendData(callback) {
+    var self = this;
+    self.agent.collector.connect(function (error) {
+        if (error) {
+            logger.error(error);
+        }
+        self.agent.harvest(function (err) {
+            if (err) {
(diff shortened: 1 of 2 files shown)


martinkuba:
@echmykhun what is your use case for sending data on demand?

echmykhun:
@martinkuba For example, if i have some unhandled domain exception, i will shut down my process, but i really want to have this exception on my dashboard. So, first of all, I'd like to push all data newrelic was able to collect to monitoring, and only after this i will shut the process down

martinkuba:
We have thought about that use case in relation to the uncaughtException event, but did not want to hold up the process.  If that is your only use case, I would prefer adding something like shutdown() with the option to send any pending data before shutting down.  Or, we could add a special config for reporting uncaught exceptions, that users could opt into.

The agent does filtering and sampling of data before sending to the server, so letting users arbitrarily initiate this could result in inconsistent data.  If there is a good use case for it, we would probably want to do more than just adding the API method.

StyleT:
@martinkuba Hi! Also we're using forced `agent.harvest` in our cron tasks because if you want to execute `process.exit()` in the end of the task you should be sure that NR already sent all data.

echmykhun:
@martinkuba Hello again!
Could you, please, provide me more details regarding your thoughts about shutdown() function. Also I like the idea to have special config option for uncaught exceptions reporting.

martinkuba:
@StyleT thanks, so providing something for just uncaughtException is not enough.

@echmykhun something like this, I think:

``` javascript
API.prototype.shutdown = function shutdown(options, callback) {
  if (options && options.collectPendingData) {
    this.agent.harvest(function() {
      this.agent.stop(callback)
    })
  } else {
    this.agent.stop(callback)
  }
}
```

echmykhun:
@martinkuba I see your point now, thank you

api.js
```diff
@@ -777,6 +777,28 @@ API.prototype.recordCustomEvent = function recordCustomEvent(eventType, attribut
...
+ * @param {object} [options=] - object with shut down options
+ * @param {boolean} [options.collectPendingData=] - boolean value that tells if NR agent must forsibly send pending data
+ * @param {function} [callback=new Function()]  - callback function, that runs when agent stopped
+ */
+API.prototype.shutdown = function shutdown(options, cb) {
+    var metric = this.agent.metrics.getOrCreateMetric(
+        NAMES.SUPPORTABILITY.API + '/shutdown'
+    );
+    metric.incrementCallCount();
+    var callback = cb || new Function();
```
martinkuba: If no callback is provided, log and return - see createTracer() for an example.  Also, options is optional, so maybe check if only callback was provided.

lykkin:
This was merged out of band in v1.26.2, thanks for the contribution!

PR #215, opened 2016-04-06
Fix TypeError during instrumentation

DullReferenceException:
We've been receiving errors like the following, which are crashing our processes:

```
TypeError: Cannot read property 'transaction' of null
    at Layer.wrappedLayerHandleRequest [as handle_request] (/src/node_modules/newrelic/lib/instrumentation/express.js:305:35)
    at /src/node_modules/express/lib/router/index.js:277:22
    ...
```

It appears that, for whatever reason, `tracer.getSegment()` is returning null, and execution of the following line of code when this happens causes an unhandled error. This adds a simple null/undefined check to avoid that and continue with the request the same way as with an inactive transaction.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ npm-debug.log
 ca-index*
 ca-serial*
 .DS_Store
+.idea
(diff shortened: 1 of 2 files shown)


lykkin:
Hey @DullReferenceException, thanks for bringing this to our attention, and apologies for the issue!

I'll roll this patch into the next release.

lykkin:
This was merged out of band in v1.26.2, thanks again!

DullReferenceException:
Thanks for the fast response!

PR #216, opened 2016-04-11
add .npmignore to ignore examples and test folder

Sinewyk:
examples and tests are not useful in a npm package

ref https://docs.npmjs.com/files/package.json

To simplify the files array and make it more resilient to change, we could also move all source files into lib and use the main field of the package.json

Also, NEWS.md should probably be renamed CHANGELOG.md to even more strip down the files config.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,2 @@
+test
+examples


martinkuba:
@Sinewyk thanks for the suggestions, agreed that we should at the very least exclude tests and examples.  We might go with .npmignore instead for now - smaller change, and to avoid accidental breakage due to missing files.

Sinewyk:
PR updated to ignore instead.

lykkin:
This was merged out of band into v1.27.0.  Thanks for the contribution!

Sinewyk:
Nah, thank you =).

PR #222, opened 2016-07-14
parse statement with line breaks

evollu:
it can't parse correctly if the sql has a like break after SELECT
for example 
```
SELECT
a
from TABLE_NAME
```
won't be recognized

### DIFF
--- a/lib/db/parse-sql.js
+++ b/lib/db/parse-sql.js
@@ -7,7 +7,7 @@ var stringifySync = require('../util/safe-json').stringifySync
 
 
 var OPERATIONS = [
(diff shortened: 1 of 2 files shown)


lykkin:
Thanks for this contribution as well.  Tests would be very helpful, if possible.

evollu:
@lykkin test added. also I'm replacing line breaks with surrounding white space with single white space so the query looks cleaner

lykkin:
thanks again, i'll make sure this gets into the next release!

evollu:
👍

martinkuba:
This is now fixed in v1.29.0.

PR #223, opened 2016-07-14
support sql-template-strings generated params

evollu:
I'm using `sql-template-strings` to build parameterized query.
the object doesn't have text property but object.text returns sql statement. I guess it implemented it though a getter function.

### DIFF
--- a/lib/instrumentation/pg.js
+++ b/lib/instrumentation/pg.js
@@ -13,7 +13,7 @@ function initializeSegment(tracer, segment, config) {
   var statement
   if (config && (typeof config === 'string' || config instanceof String)) {
     statement = config
-  } else if (config && config.hasOwnProperty('text')) {
+  } else if (config && config.text) {
     statement = config.text
   } else {
     // Won't be matched by parser, but should be handled properly


lykkin:
Hey, thanks for the contribution!  Would you mind adding a test that covers the issue you are running into?  If we end up refactoring our db instrumentation I'd hate to reintroduce this issue.

evollu:
will do

evollu:
there is no test for parsing query in postgresql.test.js at all. is there any example?

lykkin:
sorry for the confusion, the tests are fairly disorganized currently.  the integration tests for pg would be [here](https://github.com/evollu/node-newrelic/blob/0207c96f0219bb2d91297c07773b04cfdfe130a5/test/versioned/pg-4/), and an example can be [here](https://github.com/evollu/node-newrelic/blob/0207c96f0219bb2d91297c07773b04cfdfe130a5/test/versioned/pg-4/pg.common.js#L164).

evollu:
i tried `npm test` but it seem not pickup the test. what command should I run?

lykkin:
the command to run the tests is `make test` or `make integration` if you want to run the integration suite.

i'd be willing to write the tests for this pr if you'd like to skip setting up the testing environment.

evollu:
that will be great! I'm not familiar with this test framework. what is this?

lykkin:
cool! i'll write the tests and roll this into our next release. thanks again for the contribution!

we use tap for the integration tests, but in order to set up the test environment we use [make](https://en.wikipedia.org/wiki/Makefile). so things like standing dbs up in docker containers and installing deps for the tests is done through make.

evollu:
👍

martinkuba:
This is now released in v1.29.0.

PR #224, opened 2016-08-03
WIP: Couchbase bucket operation instrumentation

mwain:
Instrumenting couchbase (couchnode). WIP as i have not yet finished the tests, wanted to get early feedback.

couchbase module doesn't export the `Bucket` object, so having to wrap `cluster.openBucket` which returns a Bucket and from that i get the prototype and wrap. This is only done once. 

I opted to wrap the individual methods of Bucket object and not the private `Bucket._invoke` method as i think it gives a more granular metrics. However i'm happy to update if you deem it better to wrap fewer methods.

### DIFF
--- a/lib/instrumentation/couchbase.js
+++ b/lib/instrumentation/couchbase.js
@@ -1,0 +1,85 @@
+'use strict'
+var shimmer = require('../shimmer')
+var ParsedStatement = require('../db/parsed-statement')
+var COUCHBASE = require('../metrics/names').COUCHBASE
+
+var BUCKET_OPERATIONS = [
+  'get',
+  'query',
+  'getMulti',
+  'getAndTouch',
+  'getAndLock',
+  'getReplica',
+  'touch',
+  'unlock',
+  'remove',
+  'upsert',
+  'insert',
+  'replace',
+  'append',
+  'prepend',
+  'counter'
+]
+
+module.exports = function initialize(agent, couchbase) {
+  var tracer = agent.tracer
+  var bucketInstrumented = false
+
+  if (couchbase.Cluster && couchbase.Cluster.name === 'MockCluster') {
+    return
+  }
+
+  // Couchbase don't export Bucket, can only instrument this once a 
+  // a call to Cluster.openBucket is called for the first time.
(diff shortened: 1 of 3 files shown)


martinkuba:
@mwain Thanks for submitting the PR!  We have just released a beta of the agent with new API that will allow you to write and deploy database instrumentation independently of the agent.  In other words, you could plug in your Couchbase instrumentation without having to wait for a new version of the agent.  The new API also shields you from having to know the internals of the agent.

We will be happy to answer any questions about the new API on our [forum](https://discuss.newrelic.com/c/language-agents/node-js-agent), and we would love to get your feedback as well.  If you are interested in participating in the beta, please follow instructions on this page https://www.npmjs.com/package/@newrelic/beta-agent.

mwain:
@martinkub Didn't see the beta version. Will take a look at the new API, however doubt will roll a beta version into production env.

As for this PR, and adding couchbase instrumentation as standard to newerlic 1.x and 2.x, are you happy for it to be merged once i completed the tests?

martinkuba:
@mwain We are in the middle of changing the mechanism (API) for doing database instrumentations in the agent, which is why we have released the beta.  While that's in progress, we are reluctant to accept new database instrumentations into the v1.x agent because we would have to rewrite them very soon anyway.

The only difference between the 2.0.0 beta and the current 1.x version is the new high-level API, which is implemented on top of the existing internal APIs.  The beta is being supported the same way as the current version of the agent - through our support channel.  We will also be keeping the two branches in sync, so they will not diverge as far as bug fixes and new minor features.

We would also be happy to work with you to update the work that you have already done to the new API.  Once that's done, you will be able to start using it right away, without having to wait for a new version of the agent.  So, with all that said, our recommendation is that you give the beta a try.

psvet:
@mwain We're going to close this out since it's been open for quite some time with no activity. As mentioned above, the API has since been expanded to allow independent [datastore instrumentation](https://newrelic.github.io/node-newrelic/docs/tutorial-Datastore-Simple.html). But feel free to add some tests and submit an updated PR. Thanks!

PR #225, opened 2016-08-19
Stub for "shutdown" API method

StyleT:
This PR fixes work of [`shutdown()`](https://docs.newrelic.com/docs/agents/nodejs-agent/supported-features/nodejs-agent-api#other-api) API method with disabled NewRelic module via `newrelic.js`.

Right now callback that was passed to `shutdown()` won't be called.


StyleT:
Any updates here?

martinkuba:
@StyleT Thanks for making this fix.  We should check that a callback was passed in, and it could also be the first argument (options is optional).  Other than that, we will include this in the next release.

StyleT:
@martinkuba Fixed!

martinkuba:
Included in version 1.30.1.

PR #227, opened 2016-09-19
[instrumentation/express] fix paramware wrapper signature

Gertt:
According to Express' documentation (and implementation) the parameter middleware accepts five arguments, the last being the parameter's name.

```
The parameters of the callback function are:

- req, the request object.
- res, the response object.
- next, indicating the next middleware function.
- The value of the name parameter.
- The name of the parameter.
```

https://expressjs.com/en/4x/api.html#router.param
https://expressjs.com/en/4x/api.html#app.param

I took a look at the unit tests and was not sure where to fit a new one in or if it's necessary for this change. If needed, would be nice if someone can point me in the right direction.

### DIFF
--- a/lib/instrumentation/express.js
+++ b/lib/instrumentation/express.js
@@ -625,7 +625,7 @@ module.exports = function initialize(agent, express) {
       }
       return param.apply(this, arguments)
 
-      function paramwareWrapper(req, res, next, paramValue) {
+      function paramwareWrapper(req, res, next, paramValue, paramName) {
         var transaction = tracer.segment && tracer.segment.transaction
         if (!transaction) {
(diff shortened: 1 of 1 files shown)


NatalieWolfe:
Hi Gerrt, thanks for reporting this to us! We've implemented a fix for this using an `arguments` slice and it will be in our next release.

NatalieWolfe:
We've just published 1.30.4 with fixes for this problem. Please let me know if it has or if there is more we can do. I would like to confirm it has resolved your problem before closing this PR. Thanks!

Gertt:
@NatalieWolfe the issue has been resolved for us in 1.30.4, thanks!

PR #228, opened 2016-09-30
Fix Cannot read property transaction of null

J-Gallo:
When you want to log data after rendering the view to the user, in multiple cases it is logging the error: "Cannot read property transaction of null".

With this fix we check if there is any tracer segment before the transaction, if not we do the same as the transaction is not active.


marianocibelli:
+1

martinkuba:
Thanks for the fix, we will include it in our next release.

J-Gallo:
Awesome, if you need anything, just ask.

PR #230, opened 2016-10-27
Enabling notion of audit log in agent config file, adding ability to filter in only specific transmission events

zperacha:
This change is to support more fine grain logging controls for larger enterprises that are looking to monitor/capture traffic between Agent to Collector.


martinkuba:
This has been included in version 1.34.0.

PR #232, opened 2016-11-08
Stop pools double counting mysql calls

fragglebob:
I noticed that all my database calls were being reported twice, and it had something to do with using a pooled connection. 

I worked out it was due to the pool's query method being wrapped, as well as each of the pooled connection's query method being wrapped too.

Before:
screenshot from 2016-11-08 08-19-51

After:
screenshot from 2016-11-08 08-49-10

### DIFF
--- a/lib/instrumentation/mysql.js
+++ b/lib/instrumentation/mysql.js
@@ -252,9 +252,6 @@ module.exports = function initialize(agent, mysql) {
           return getConnectionHandler(pool, getConnection)
         })
 
-        // patch the pools .query method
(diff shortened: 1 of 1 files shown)


martinkuba:
Thanks for bringing this to our attention.  We definitely don't want to double count.  We might need to continue to wrap the pool in order to maintain context across async boundaries when pool is used.  We'll double check on that and get back to you.

fragglebob:
No problem. :+1:

lykkin:
@fragglebob this issue has been resolved in v1.35.0, thanks for the pr!

PR #233, opened 2016-11-23
Keep correct transaction when calling Bluebird promise.nodeify nodeback

mdlavin:
When using Bluebird's `promise.nodeify` the current transaction is lost. The same support for `asCallback` should also be provied to `nodeify` since they are aliases of each other.

### DIFF
--- a/lib/instrumentation/bluebird.js
+++ b/lib/instrumentation/bluebird.js
@@ -6,7 +6,7 @@ var BLUEBIRD_SPEC = {
   name: 'bluebird',
   constructor: 'Promise',
   $proto: {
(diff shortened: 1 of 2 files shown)


mdlavin:
I don't think the build failures were introduced by my change, but let me know if there is something I need to change.

lykkin:
@mdlavin Thanks for the pr!  No need to worry about the tests, the failures are due to environmental differences and we are going to fix those up shortly.  The changes look good, we'll include this in the next release!

lykkin:
this pr was merged out of band in v1.35.0, thanks again for the pr, @mdlavin !

PR #235, opened 2016-12-08
Include `find` in instrumented MongoDB operations

sgilroy:
I noticed that `find` operations were not showing up in transaction traces and performance breakdowns. Looking at mongodb.js revealed that it was missing. Adding `find` here caused find to show up in the transactions in New Relic. Am I missing something? It seems odd that such a ubiquitous operation as "find" would have been omitted. Perhaps we are doing something else wrong/unusual in our app or how we are integrating with New Relic or MongoDB? Note that we are using Mongoose v4.3.7 at the moment.
image

### DIFF
--- a/lib/instrumentation/mongodb.js
+++ b/lib/instrumentation/mongodb.js
@@ -48,6 +48,7 @@ var COLLECTION_OPS = [
   'dropAllIndexes',
   'dropIndex',
   'ensureIndex',
+  'find',
   'findAndModify',
   'findAndRemove',
   'findOne',


sgilroy:
Contrary to my previous theory, it looks like the absence of MongoDB find operations in New Relic is actually not a result of the omission of `find` from the list of operations to instrument in `COLLECTION_OPS` in `newrelic/lib/instrumentation/mongodb.js`. This list of operations only applies to instrumenting older versions of the mongodb driver. The mongodb 2.x driver used by mongoose 4.x is instrumented by newrelic via mongodb's new Application Performance Monitoring API (APM), and the `find` operation is intentionally excluded from being instrumented by New Relic because `find` itself has no callback (no asynchronous processing) and only creates a query. So for places in our app where we use `find` we should expect to see a corresponding `toArray` operation in New Relic, which is (confusingly) dissimilar to what we will see in New Relic for most other database operations.

Based on the above, I believe this PR is not actually appropriate.

lykkin:
@sgilroy thanks for the heads up! this wouldn't be the expected behavior, the find call should show up in metrics/traces.  we'll take a look into this and keep you updated.

michaelgoin:
@sgilroy closing this PR given its age and looks like the initial intention was not met. 

We appreciate you taking a stab at the problem. Capturing Mongo's chaining API, such as the synchronous `find`, is still an issue we have. We are not sure when it may be fixed, as it has not risen to the top of priorities.

Thanks again and don't hesitate to submit PR's again in the future.

PR #238, opened 2017-01-10
Have default export as a function

arnarthor:
I was looking at the code and found it odd that the `Config.initialize()` does take a config parameter but it's not used in `./index.js`. 

Therefor I'd like propose this change. I do realise that this is a breaking change for the package. Now instead of doing `require('newrelic)`, it would need to be `require('newrelic')()`.

This additionally allows for use of sending a config into the constructor of the package instead of having to rely on a file in the root of the project so projects could do something like `require('newrelic')(config)` which could then be located else where than in the root. I generally don't like having files located in the root due to the folder structure in projects I work on. I don't like having hard dependencies on where files are located as configurations to packages

If there's a specific reason for the other API feel free to close this. Since the config initialiser accepts the argument already this makes I think this could make sense.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,111 +3,113 @@
 var logger = require('./lib/logger.js')
 var semver = require('semver')
 
-var message
-var agent
+module.exports = function init(_config) {
+  var message
+  var agent
 
-var agentVersion = require('./package.json').version
-logger.info(
-  "Using New Relic for Node.js. Agent version: %s; Node version: %s.",
-  agentVersion, process.version
-)
-
-if (require.cache.__NR_cache) {
-  logger.warn(
-    'Attempting to load a second copy of newrelic from %s, using cache instead',
-    __dirname
-  )
-  module.exports = require.cache.__NR_cache
-} else {
-  initialize()
-}
-
-function initialize() {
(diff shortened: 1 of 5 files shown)


NatalieWolfe:
Thanks for the patch, however I do not believe this is necessary. You can specify the path to the directory containing the configuration file using the environment variable `NEW_RELIC_HOME`. So, for instance if you have this project structure:

```
/path/to/project/
  - index.js
  - configs/
    - newrelic.js
```

You could set `NEW_RELIC_HOME=/path/to/project/configs`. In addition, you can set most [configuration values through environment variables](https://docs.newrelic.com/docs/agents/nodejs-agent/installation-configuration/nodejs-agent-configuration), so if you prefer you could use those entirely and set `NEW_RELIC_NO_CONFIG_FILE=true` then you do not need a `newrelic.js` file at all. Please let me know if either of these solutions work for you.

The reason `Config.initialize` accepts an optional configuration object is for our own testing, so that each test can provide its own configuration without needing separate configuration files.

arnarthor:
Ahh ok, fair enough. Didn't notice the setting for the home file. I'll close this then.

PR #240, opened 2017-01-26
add option to wrap Promises to recorderSpec

Gertt:
Is this something you would consider adding?

I was trying to use DatastoreShim to create an isntrumentation for [neo4j/neo4j-javascript-driver](https://github.com/neo4j/neo4j-javascript-driver) which returns a Promise when you want to run a query. Seems to me like an easy way to handle this with the shim api would be very handy.

### DIFF
--- a/lib/shim/datastore-shim.js
+++ b/lib/shim/datastore-shim.js
@@ -228,6 +228,9 @@ DatastoreShim.prototype.captureInstanceAttributes = captureInstanceAttributes
  * @property {bool} [stream=false]
  *  If `true`, the return value will be wrapped as a stream.
  *
+ * @property {bool} [promise=false]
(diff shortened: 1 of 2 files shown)


NatalieWolfe:
Thanks for the contribution! Unfortunately these changes will have the side effect of swallowing unhandled rejections. I'll take this in and see if I can work around that plus add any tests to get it into the agent.

NatalieWolfe:
Good news! I've added tests and updated `_bindPromise` so it wont change the process `unhandledRejection` event. Assuming review goes well this should be in the next release of the v2 agent. Thanks again for your contribution!

NatalieWolfe:
We've released this in 2.5.0-beta. Please let us know how it works for you, and thanks for the contribution!

PR #242, opened 2017-02-21
Change NEW_RELIC_NO_CONFIG to NEW_RELIC_NO_CONFIG_FILE 

Maubic:
If configuration file can not be found, the message shown point to use wrong NEW_RELIC_NO_CONFIG instead NEW_RELIC_NO_CONFIG_FILE environment variable


NatalieWolfe:
Good catch! This should be going out with our next release. Thanks for the contribution.

peterpme:
🎉 came here to do the same thing :)

NatalieWolfe:
This was released in [1.38.0](https://github.com/newrelic/node-newrelic/blob/master/NEWS.md#v1380-2017-03-16). Thanks again for the contribution, Maubic!

PR #248, opened 2017-05-10
Update default config file name

jlegrone:


jlegrone:
Looks like this also needs to be patched in v2.x -- I can submit a separate PR for that if desired.

NatalieWolfe:
Thanks for the contribution! I'll get this merged into the main line and it will be promoted to our v2 beta line next time we release that.

PR #252, opened 2017-06-22
Make shallowCopy work with "clean" objects

mastfissh:
It seems in node 8 it's possible to have objects that don't have Object.prototype methods included. This was causing our application to break. I think by calling Object.prototype explicitly this should solve it.


NatalieWolfe:
Hi @mastfissh,

Thanks for the contribution! We'll try to get this into our next release. I've also taken this opportunity to protect pretty much every call to `hasOwnProperty` so hopefully we wont have this problem going forward.

mastfissh:
Thanks!

PR #254, opened 2017-11-08
Use more secure sha256 hashing

jeneoten:
I ran a security audit against my web application which includes the newrelic agent, and it flagged the use of the md5 hashing algorithm as insecure. Would it be better to use sha256 as the hashing algorithm, because it is much more secure than md5?

### DIFF
--- a/lib/util/hashes.js
+++ b/lib/util/hashes.js
@@ -54,9 +54,9 @@ function calculatePathHash(appName, pathName, referingPathHash) {
 }
 
 function getHash(appName, txName) {
-  var md5sum = crypto.createHash('md5')
-  md5sum.update(appName + ';' + txName, 'utf8')
-  var buf = md5sum.digest()
+  var hash = crypto.createHash('sha256')
+  hash.update(appName + ';' + txName, 'utf8')
+  var buf = hash.digest()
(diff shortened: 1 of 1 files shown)


NatalieWolfe:
Hello! Thanks for the contribution, but unfortunately we will not be changing the hashing algorithm. The reason for this is three part: MD5's vulnerability is not important to our use case; if we did change it, every installed agent in every language we support would need to be updated (see below); and finally MD5 is really fast which means we're using less of our customer's resources.

MD5's insecurity centers around its vulnerability to [hash collisions](https://en.wikipedia.org/wiki/Collision_attack). This is particularly bad when one uses a hashed value to authenticate a user (for example, comparing hashed passwords).

Our agent only uses this hashing to identify cross-application communication by passing a hashed application identifier and hashed transaction name. The risk of a hash collision in this scenario is our service connecting the wrong application and transaction, but even then only if your account has two applications with colliding names which also generate transactions with colliding names. Since the severity and likelihood of this collision are both extremely small, we're willing to accept the collision potential.

While I loathe to use institutional momentum to resist change, in this case the trade off of effort (updating every New Relic agent in existence) versus the benefit (reducing potential for conflated cross-application linkage) is simply not worth it.

PR #255, opened 2017-11-14
Update https-proxy-agent to v2

erikrestificar:
## CHANGE LOG
your version of https-proxy-agent is using an old version of debugger that has a know vulnerability:

https://nodesecurity.io/advisories/534

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
   "dependencies": {
     "async": "^2.1.4",
     "concat-stream": "^1.5.0",
-    "https-proxy-agent": "^0.3.5",
+    "https-proxy-agent": "^2.1.0",
     "json-stringify-safe": "^5.0.0",
     "readable-stream": "^2.1.4",
     "semver": "^5.3.0"


NatalieWolfe:
Thanks for the contribution! Unfortunately newer versions of `agent-base` (a dep of `https-proxy-agent`) drop support for Node <4. Our agent still supports Node 0.10 and 0.12 so we can't take this upgrade directly. We are currently triaging this vulnerability and I will get back to you as soon as we have made a decision.

NatalieWolfe:
After investigating this further we confirmed our agent is not vulnerable to this issue. `https-proxy-agent@0.3.6` has [`"debug": "2"` as their dependency](https://github.com/TooTallNate/node-https-proxy-agent/blob/v0.3.6/package.json#L26), which is satisfied with any v2 release of `debug`. The advisory you linked to cites `>= 2.6.9 < 3.0.0 || >= 3.1.0` as the patched versions. Thus, as long as a different module isn't requiring a bad version of `debug`, our agent will come with a patched version.

As a side note, we [run `nsp` against our agent](https://travis-ci.org/newrelic/node-newrelic/jobs/296413827#L586) as part of our regular CI suite. [`nsp` is the Node Security Platform tool](https://www.npmjs.com/package/nsp) for identifying these vulnerabilities in your application. If it found a vulnerability in our agent, we would know about it the next time we pushed a commit. In addition to that, our Security team uses [Snyk](https://snyk.io/) to keep an eye on this too.

Thank you again for bringing this to our attention, luckily we're safe from this vulnerability. Upgrading `https-proxy-agent` is on our list of things to do when we drop support for 0.10 and 0.12, along with a whole host of other upgrades. In the meantime we must stick with this version though.

alum:
There's another caveat to consider. It seems that versions of `newrelic` between `1.16.2` and `1.40.0` comes with `https-proxy-agent` as a bundled dependency.

`newrelic@1.16.2` has `debug@1.0.4`
... (didn't test all the versions here)
`newrelic@1.34.0` has `debug@2.3.2`
`newrelic@1.35.1` has `debug@2.3.3`
`newrelic@1.36.2` has `debug@2.6.0`
`newrelic@1.37.2` has `debug@2.6.1`
`newrelic@1.38.2` has `debug@2.6.1`
`newrelic@1.39.1` has `debug@2.6.6`
`newrelic@1.40.0` has `debug@2.6.8`

`newrelic@2.0.0` does not use bundled dependencies.

This means that even though `https-proxy-agent@0.3.6` has `"debug": "2"` as their dependency, npm will not install the latest version of 2.x.x, but rather just use the bundled version.

NatalieWolfe:
Thanks for the clarification, I hadn't considered the v1 line's bundled dependencies since we dropped that practice with v2 back in July. As you point out though, that is only an issue with the v1 New Relic Agent. We are no longer supporting the v1 line except for major security vulnerabilities which this is not.

If there is a reason you cannot upgrade to v2, please let us know and we can discuss options. But otherwise this isn't a concern for us.

PR #256, opened 2017-12-05
Mysql adding query values in instrumented mysql query

iassal:
## CHANGE LOG
Changing mysql instrumentation to insert values in sql query

Example: Instead of instrumenting `select * from user where id = ?`, we'll instrument `select * from user where id = 12345`

### DIFF
--- a/lib/instrumentation/mysql.js
+++ b/lib/instrumentation/mysql.js
@@ -2,6 +2,7 @@
 
 var dbutils = require('../db/utils')
 var properties = require('../util/properties')
+var sqlString = require('sqlString');
 
 
 module.exports = function initialize(agent, mysql, moduleName, shim) {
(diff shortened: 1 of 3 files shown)


NatalieWolfe:
Hi @iassal! Thanks for the contribution. The reason we don't do value substitution in the queries is that the query values could be sensitive. It could include email addresses, home addresses, API keys, social security numbers, passwords, names, credit card numbers, insurance information, or any number of other pieces of sensitive information that could be stored in a database.

Some of our customers have strict regulations that limit what can be sent to third parties such as us. Even putting this behind a configuration value could make it legally treacherous for these companies to use the Node Agent.

If you really want to do this, I would recommend you use [`newrelic.recordCustomEvent()`](https://docs.newrelic.com/docs/agents/nodejs-agent/supported-features/nodejs-agent-api#record_custom_event) and put both the original query and your query with substituted values into the attributes along with any other information you are interested in. You can then create a dashboard on Insights for these events.

iassal:
Thanks @NatalieWolfe for sharing your insights! I'll explore the recordCustomEventApproach.

PR #258, opened 2018-01-30
Fixed error if process.config.variables.node_prefix missing

BrighTide:
## CHANGE LOG
- If process.config.variables.node_prefix is falsey (which can happen if using electron, leading to this issue https://discuss.newrelic.com/t/new-relic-on-electron-nodejs/53601)  the getGlobalPackages function in lib/environment.js will give an err when it shouldn't
- Added a missing error check

### DIFF
--- a/lib/collector/facts.js
+++ b/lib/collector/facts.js
@@ -12,6 +12,9 @@ function facts(agent, callback) {
     systemInfo: a.apply(fetchSystemInfo, agent),
     environment: agent.environment.getJSON
   }, function factMapCb(err, data) {
+    if (err) {
+      return callback(err)
+    }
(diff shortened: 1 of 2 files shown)


NatalieWolfe:
Nice catch! I'll pull this into our dev repo and try to get it merged soon. Thanks for contributing!

BrighTide:
No problemo, and cheers 👍

NatalieWolfe:
This has been released as part of [2.7.0](https://github.com/newrelic/node-newrelic/blob/master/NEWS.md#270-2018-02-01). Please let me know if that fixes things for you or if there is anything more we can do!

PR #260, opened 2018-02-27
Add integration test for Promise#bind in combination with catch

papandreou:
## CHANGE LOG

Add an integration test for `Promise#bind` propagating the right context to `catch`. 

## INTERNAL LINKS

https://support.newrelic.com/tickets/285687

## NOTES

This is a follow-up to a support thread. I'm unable to run the integration part of the test suite locally because I don't have access to https://github.com/newrelic/SSL_CA_cert_bundle

### DIFF
--- a/test/integration/instrumentation/promises/methods.js
+++ b/test/integration/instrumentation/promises/methods.js
@@ -150,7 +150,7 @@ module.exports = function(t, library, loadLibrary) {
   testTryBehavior('attempt')
 
   ptap.test('Promise.bind', function(t) {
-    t.plan(2)
+    t.plan(3)
 
     t.test('context', function(t) {
       testPromiseContext(t, function(Promise, name) {
@@ -164,6 +164,19 @@ module.exports = function(t, library, loadLibrary) {
         return Promise.bind(ctx, name).then(function(value) {
           t.equal(this, ctx, 'should have expected `this` value')
           t.equal(value, name, 'should not change passed value')
+        })
+      })
+    })
+
+    t.test('catch', function(t) {
+      var ctx = {}
+      var err = new Error('oh dear')
+      testPromiseClassMethod(t, 2, function(Promise, name) {
+        return new Promise(function(resolve, reject) {
+          reject(err)
+        }).bind(ctx, name).catch(function(reason) {
+          t.equal(this, ctx, 'should have expected `this` value')
+          t.equal(reason, err, 'should not change rejection reason')
         })
       })
     })


papandreou:
Whoa, that's a lot of test failures. Is that because this is a PR without access to Travis secrets? Could someone on the inside try running the tests? :)

NatalieWolfe:
Thanks for the test! Our public Travis is always in a spot of trouble unfortunately, but I'll make sure this gets integrated into our dev repo and released soon.

papandreou:
@NatalieWolfe, thanks for checking it out :)

I'm not even sure the test is correct, I had to create it "blindfolded" because I couldn't run the test suite locally. It's my attempt to get us closer to solve the linked support issue, which originated here: https://github.com/clearbit/clearbit-node/issues/32

NatalieWolfe:
@papandreou Turns out we were fixing this just last night. Didn't realize this was the same ticket at first. :)

For future reference, this would be part of the versioned tests which you can run like this:
```
$ npm run versioned-tests          # Runs all instrumentation tests
$ npm run versioned-tests bluebird # Runs just bluebird
```

papandreou:
Ah, great, you must have made more progress than me, then. It appears that the test I wrote works fine in both bluebird 2 and 3. I hope you manage to track it down :)

NatalieWolfe:
Looks like the problem case is specifically when the bind context is a function.

`Promise#bind` is what we call an "instance cast method." Cast methods are any methods which take some arbitrary value and return a promise. `Promise.resolve` is the simplest example of a static cast method.

This is opposed to a "then method," which is any method that takes a function to execute when the promise finishes (resolves or rejects). The obvious examples of then methods are `Promise#then` and `Promise#catch.`

In the later case we want to wrap the continuation function (the function passed to the then method) to propagate our transaction context. In the former case (a cast method), if a function happens to be passed in, we should not wrap it because it will not be executed as a continuation.

Our error was that we treated all instance cast methods as then methods. Under most circumstances this was okay because we simply ignore non-function arguments to then methods. But in the case where a function is handed in we would wrap it.

We have resolved this error and added tests for this scenario. It should be in our next release.

papandreou:
@NatalieWolfe, thanks a lot for tracking it down and for the thorough explanation. Great detective work 🕵️‍♀️ 

Looking forward to the next release :)

papandreou:
I'll close this PR since even more elaborate tests will land in the next release.

NatalieWolfe:
We just released a fix for this. Please let me know if it resolves your problem or if there is more we can do. Thanks!

papandreou:
@NatalieWolfe, thanks! We've been running the fixed version in production for three days now without any hiccups. The original did only occur in a rare error case, though, but let's assume that it's fixed. Will get back in touch if it pops up again :)

PR #263, opened 2018-04-30
test(travis): add node version 10

gergelyke:


lykkin:
Hey, thanks for the PR!

We currently have our internal repo testing against v10, and it doesn't pass.  Right now the promise instrumentation seems to lose state in a few cases. We looking into shipping support for node 10 in the near future.

gergelyke:
Ahh, I see - thanks for the heads up Bryan!

finnp:
@lykkin Do you have an approx. date when newrelic will support Node 10?

tanuck:
Is there any update on this yet? Node 10 has been out for some time now and it won't be long before it becomes the active LTS version.

josdejong:
`newrelic@1.40.0` seems to work fine on Node 10, though that's not really a solution of course.

NatalieWolfe:
@josdejong There were changes to Promises, Async Hooks, and the HTTP module in Node 10 which all subtly break the agent. The agent will not crash under these conditions, but the data reported is highly likely to be incorrect.

josdejong:
Thanks for the update @NatalieWolfe, that sounds tricky. I hope you can solve it soon!

psvet:
We should now have full Node 10 support as of `4.6.0`. Let us know if you run into any issues!

PR #265, opened 2018-06-08
Explicitly reference package.json as a .json file in require()

guyellis:
## CHANGE LOG

- Make `require()` statements explicitly reference `package.json` as a `.json` file.

## NOTES

- This solves a problem when requiring/importing newrelic from a Typescript file.

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,7 +12,7 @@ var logger = require('./lib/logger')
 var psemver = require('./lib/util/process-version')
 
 
-var agentVersion = require('./package').version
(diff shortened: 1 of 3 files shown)


psvet:
Thanks for the PR! We'll include this in the next release 👍

psvet:
This is included in [4.1.5](https://github.com/newrelic/node-newrelic/blob/master/NEWS.md#415-2018-06-11). Thanks @guyellis!

PR #266, opened 2018-07-03
Allow to pass nonce option to getBrowserTimingHeader.

joaovieira:
## CHANGE LOG
* Allow to pass `nonce` to getBrowserTimingHeader()

## INTERNAL LINKS

https://discuss.newrelic.com/t/content-security-policy-and-browser-injection/2629/42

## NOTES

`getBrowserTimingHeader` generates an inline script tag. This is generally a bad practice if you want to protect your page with HTTP Content-Security-Policy (CSP) header. In the situations you do require it you should generate and pass a `nonce` UUID which tells the browser those scripts are allowed to be executed.

This PR adds an option to add a `nonce` attribute to the generated `<script>` tag.

Currently we're injecting the nonce in the middle of the returned string header with `String.substr()` - which is not ideal.

### DIFF
--- a/api.js
+++ b/api.js
@@ -21,7 +21,7 @@ const MODULE_TYPE = require('./lib/shim/constants').MODULE_TYPE
  * CONSTANTS
  *
  */
-const RUM_STUB = "<script type='text/javascript'>window.NREUM||(NREUM={});" +
+const RUM_STUB = "<script type='text/javascript' %s>window.NREUM||(NREUM={});" +
                 "NREUM.info = %s; %s</script>"
 
 // these messages are used in the _gracefail() method below in getBrowserTimingHeader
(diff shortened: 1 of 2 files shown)


NatalieWolfe:
Thanks for the contribution, @joaovieira! This looks great, so I'll pull it into our dev repo and make sure it gets released soon.

NatalieWolfe:
@joaovieira This was released in 4.3.0. Please try it out and let us know if it works! Thanks again for contributing this feature.

joaovieira:
@NatalieWolfe sorry for the delay. Just tested it and worked perfectly. Thanks for the update!

ccope:
For others who see this, the nonce header is set on the script tag returned by this function, but *that* script generates *another* script tag which loads from New Relic's servers. So you still need to whitelist them in your CSP string.

PR #270, opened 2019-01-25
Should check if isActive is a function first

pegurnee:
## NOTES
If isActive is not a function at this point we have an issue.


lykkin:
Hey, @pegurnee! Is this still an issue? If you ran into any issues while trying to write a test for your case, let us know and we can help you out.

pegurnee:
@lykkin Actually, we were having other problems. I'm going to dismiss this one and call it a day! Sorry for the absolutely delayed response.

PR #271, opened 2019-02-06
Undefined check on segment transaction trace

DullReferenceException:
We're getting some errors when doing a `JSON.stringify`:

```
TypeError: Cannot read property 'root' of undefined
    at TraceSegment.toJSON
(/src/node_modules/newrelic/lib/transaction/trace/segment.js:389:75)
    at JSON.stringify (<anonymous>)
    at <our code...>
```

It appears that `segment.transaction.trace` is `undefined`. Note that this occurs during a connection timeout, in case you want to fix the underlying bug where that is not populated. I'm not sure where the trace segment is getting attached to the object tree we're serializing, but I thought you'd probably want this `TypeError` fixed in any case.

It's unclear to me where a test for this should reside, but I can add one if someone has a pointer to how this condition can be reproduced/tested.

### DIFF
--- a/lib/transaction/trace/segment.js
+++ b/lib/transaction/trace/segment.js
@@ -386,7 +386,12 @@ TraceSegment.prototype.toJSON = function toJSON() {
         segment.getExclusiveDurationInMillis()
     }
 
-    var start = segment.timer.startedRelativeTo(segment.transaction.trace.root.timer)
+    var trace = segment.transaction.trace
+    if (!trace) {
+      continue
+    }
+
+    var start = segment.timer.startedRelativeTo(trace.root.timer)
     var duration = segment.getDurationInMillis()
 
     var segmentChildren = segment.getCollectedChildren()


DullReferenceException:
The more I look at this, the more it seems that our code is mutating the trace object; we do a de-cycling before serializing, so maybe that's deleting some properties. To work around this, we're going to try omitting `__NR_segment` before serializing. 

Have you considered making that a non-enumerable property so it will be hidden from other tools that are inspecting HTTP requests?

lykkin:
Hey, @DullReferenceException!  Apologies for the delay in response! Is this still an issue? Were you able to discover any more details about this issue?

This behavior is definitely unexpected, and the change here would cause some silent data loss in the case where the root has been lost.

I believe I have seen this symptom in cases where the transaction has been serialized and deserialized as well.  Making all of our internal references non-enumerable is  good idea (though there would be some performance considerations)! To make sure a fix would cover your situation, would you be able to write a test that exhibits the issue?

DullReferenceException:
Writing a test case for this would be a bit difficult, but here's how our code causes the issue, in a nutshell:

```js
const decycle = require('smart-circular');

// Outbound node request/response, with the `__NR_segment` property attached
function logRequestFailure(request, response) {
  const logPayload = JSON.stringify(decycle({ request, response })); // exception
}
```

The `decycle` call removes properties of the NR metadata, which then causes the `toJSON` method to fail due to this missing property.

Closing because this "fix" is just a workaround for an application bug, but making `__NR_segment` non-enumerable or a symbol so that it isn't visited during serialization would probably be helpful.

PR #272, opened 2019-02-22
Prevent Memcached undefined error

bwolfe:
I am getting these errors:
> TypeError: Cannot read property 'split' of undefined\n    at Client.commandWrapper (/usr/app/node_modules/newrelic/lib/instrumentation/memcached.js:51:51)

on this line of code: `mycache.del(key)`

The del() call was after a key (probably) timed out or potentially removed by another pod.  I cannot replicate this in dev/staging though! >:O

I attempted to write integration tests, but could not get it to fail in this process:
1. Deleting a key after a delete
2. Deleting a key after previous key timed out
3. Deleting a key after end() entire cache.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,8 +37,7 @@ and not ready for general-purpose use.
 
 ### Setup
 
-To run the tests you need a GNU-compatible make, the openssl command-line
-binary, and some services:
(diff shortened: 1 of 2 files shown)


NatalieWolfe:
Thanks for this fix, Ben! I'll merge it into our internal repo to make sure it gets released ASAP.

PR #275, opened 2019-04-01
catch harvest callback error

joway:
## CHANGE LOG

fix crush when `Error: Callback was already called.`

### DIFF
--- a/lib/harvest.js
+++ b/lib/harvest.js
@@ -2,7 +2,7 @@
 
 const a = require('async')
 const CollectorResponse = require('./collector/response')
-const logger = require('./logger').child({component: 'Harvest'})
+const logger = require('./logger').child({ component: 'Harvest' })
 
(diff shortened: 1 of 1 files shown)


joway:
image

When we enable newrelic `proxy` option, it seems sometimes will have `ECONNRESET` error, and the nodejs precess will crush.

So we need add try catch on `callback` function to prevent it.

michaelgoin:
Hi! 

We pushed a patch in version `5.6.4` that prevents this issue from throwing errors and also attempts to capture where the multiple calls were originating from. Using a version of the agent `^5.6.4`, if you see the new log message indicating the attempt at multiple callback invocations, please let us know as it will help us further track down the underlying issue.

I'll be closing this PR, since we won't be accepting the immediate changes, but we appreciate the help.

Thank you,

Michael

joway:
Hi, @michaelgoin ! Thanks for your great work!

PR #276, opened 2019-04-30
Adding support to Node 11 and Node 12

jhkersul:
## CHANGE LOG
Changed the support version on `package.json` to support Node 11 and Node 12. 

New Relic code is completely compatible with Node 12.

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,7 +101,7 @@
   ],
   "homepage": "http://github.com/newrelic/node-newrelic",
   "engines": {
-    "node": ">=6.0.0 <11.0.0",
+    "node": ">=6.0.0 <13.0.0",
     "npm": ">=3.0.0"
   },
   "directories": {


lykkin:
Hey @jhkersul, thanks for the PR!

Even though the agent may be compatible with the newer versions of node at a syntactic level, we would like to be 100% sure that no incompatibilities were introduced in the new versions (e.g. new features that require instrumentation, subtle functionality changes to existing API, etc) before we claim that the agent supports those versions.  We are going to do a full audit of the new versions in the near future.

All that said, if you are using the agent with no issues, I encourage you to continue to do so!

alexashley:
@lykkin Any update on Node 12 compatibility? We use some dependencies that require Node 12, so right now we're turning off the `engine` check with `yarn` in order to use NR. 

Are there any code or test changes that I could make to help?

lykkin:
@alexashley, thanks for the check in! I'm planning on doing a full API audit for v11 and v12 tomorrow, and I'll have a more concrete update then.

Based on a cursory glance, I would be concerned about how the agent reacts with worker threads and if there are any new async methods in core since v10 (also, the agent will be unable to auto-instrument any modules being loaded through ES6 modules, but that's still behind a flag for the time being).  If you have an enumeration of new async endpoints, that'd help a lot.  If you want to write some tests or instrumentation for the new endpoints, that'd be ideal.  If you end up going down that path, let me know what I can do to help out.

alexashley:
> any new async methods in core since v10 

I'll make a pass through the release notes, but I think these are the only two:

- [`events.once`](https://nodejs.org/docs/latest-v12.x/api/events.html#events_events_once_emitter_name)
- [`rl[Symbol.asyncIterator]()`](https://nodejs.org/api/readline.html#readline_rl_symbol_asynciterator)

I wrote a [script](https://github.com/alexashley/node-docs-compare) to diff the module documentation and those were the only async methods I saw in the [output](https://gist.github.com/alexashley/f081c8f7ffa1107e75b7cedcf67737f6).

callumlocke:
Any news on this? It's a major blocker for our project. I don't want to disable yarn's `engines` enforcement for all our deps just because newrelic still isn't '100% sure' about supporting the current version of Node 🙃

michaelgoin:
Hi,

We've decided to open up the `engines` restriction for those who want to make use of the agent in Node 11 & 12 prior to official support. This has not yet shipped but should land in the next couple releases.

We've created this issue https://github.com/newrelic/node-newrelic/issues/279 to keep track of any known issues discovered prior to official support, which is on our roadmap.

PR #282, opened 2019-06-20
[shim] Fix TypeError: segment.transaction.isActive is not a function

### LINKED ISSUE #281: [Bug] unhandledException TypeError: segment.transaction.isActive is not a function
`newrelic` causes unhandledException and crashes the process

```
TypeError: segment.transaction.isActive is not a function
    at Shim.getActiveSegment (/usr/src/app/node_modules/newrelic/lib/shim/shim.js:1329:61)
    at AsyncHook.afterHook (/usr/src/app/node_modules/newrelic/lib/instrumentation/core/async_hooks.js:88:31)
    at emitAfterNative (internal/async_hooks.js:164:40)
    at <anonymous>
    at process._tickDomainCallback [as _tickCallback] (internal/process/next_tick.js:229:7)
```

ali-essam:
## CHANGE LOG
- [shim] Fix TypeError: segment.transaction.isActive is not a function
## NOTES
- Although this might be a hack, it's required to not crash the process
fixes #281

### DIFF
--- a/lib/shim/shim.js
+++ b/lib/shim/shim.js
@@ -1326,7 +1326,9 @@ function getSegment(obj) {
  */
 function getActiveSegment(obj) {
   var segment = this.getSegment(obj)
-  if (segment && segment.transaction && segment.transaction.isActive()) {
+  if (segment && segment.transaction &&
+    typeof segment.transaction.isActive === 'function' &&
+    segment.transaction.isActive()) {
     return segment
   }
   return null
@@ -2024,7 +2026,7 @@ function _wrap(shim, original, name, spec, args) {
       wrapped = arity.fixArity(original, wrapped)
     }
 
-    // TODO: Once all wrapping is converted to proxies, we won't need to 
+    // TODO: Once all wrapping is converted to proxies, we won't need to
     // set this property as the trap on 'get' will return the original for
     // __NR_original. For now, we have to prevent setting this on original.
     if (!wrapped.__NR_original) {


lasergoat:
I am also getting this error:

```
{ level: 'info',
  source: 'skynet',
  status: 200,
  route: null,
  data: 
   { message: 'get call to https://sandbox-quickbooks.api.intuit.com/v3/company/4620816365008403470/query?query=SELECT%20*%20FROM%20Preferences%20STARTPOSITION%200%20MAXRESULTS%201000&minorversion=37 for QBO realm id 4620816365008403470' },
  timestamp: '2019-08-15T15:01:26.468Z',
  operationId: '65a7e071-cb4e-44bd-9cf1-213c65ab0f86' }
TypeError: segment.transaction.isActive is not a function
    at Shim.getActiveSegment (/var/app/current/node_modules/newrelic/lib/shim/shim.js:1329:61)
    at AsyncHook.afterHook (/var/app/current/node_modules/newrelic/lib/instrumentation/core/async_hooks.js:88:31)
    at emitAfterNative (async_hooks.js:402:40)
    at <anonymous>
    at process._tickDomainCallback (internal/process/next_tick.js:228:7)
    at process.wrappedProcess (/var/app/current/node_modules/newrelic/lib/instrumentation/core/timers.js:15:21)
```

any word on including this? If the agent is going to be throwing unhandled exceptions, I don't think we can use New Relic for these projects.

astormnewrelic:
@lasergoat We're definitely interested in fixing this -- are you able to reproduce this issue on command? With code you can share?  If so we'd love to get a reproducible case -- once we have that we'll be better able to asses the full impact of ali-essam (generous) fix.

lib/shim/shim.js
```diff
@@ -1326,7 +1326,9 @@ function getSegment(obj) {
  */
 function getActiveSegment(obj) {
   var segment = this.getSegment(obj)
-  if (segment && segment.transaction && segment.transaction.isActive()) {
+  if (segment && segment.transaction &&
```
gevertex: You should else this with some instrumentation to send a stack trace and other diagnostics to NewRelic so that all of the instances of this problem can be aggregated and give NewRelic devs more information about how to fix this and the code paths involved.  That way though the data does get dropped in these cases, we learn something each time it happens and it can be quantified.

lasergoat:
@astormnewrelic sorry I didn't reply already. I wish I'd been more of a help for you. I ended up just leaving the agent in as-is, and forgetting about the issue. I haven't seen the error since.

PR #289, opened 2019-07-22
Env variable to customize metric hostname

cesine:
## CHANGE LOG

Add an environment variable which can be used to customize the metric hostname.

## NOTES

This is to workaround hostname explosion if you are using dynamic hostnames in k8s for example:

https://discuss.newrelic.com/t/relic-solution-hostname-explosions/58911

> A Hostname Explosion is when an agent installed on an application generates a large number of unique Real Agent IDs, that when queried, causes issues in the APM UI. Real Agent IDs are used for tying an instance of a New Relic Agent to an application on a given host to the New Relic Collector. Unique Real Agent IDs are a symptom of spinning up an agent on a host with a dynamic name.

We are using k8s, and at times of peak traffic can have more than 1000 pods reporting to Newrelic. We would like to keep the original hostname in our log aggregation, and only customize it for reporting to Newrelic. 

We found the NEW_RELIC_PROCESS_HOST_DISPLAY_NAME environment variable, but that only customizes the display name, not the hostname in the metric. One solution we thought of would be to add an environment variable so that consumers can override the `os.hostname()`. For now, I've named it NEW_RELIC_METRIC_HOST_NAME, but I'm open to suggestions!

### How this was tested

```
total ........................................... 3916/5203


  3916 passing (6m)
  23 pending
  1264 failing


real	6m11.288s
user	0m46.105s
sys	0m9.363s
```

### DIFF
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -702,6 +702,12 @@ exports.config = () => ({
      */
     display_name: '',
     /**
+     * Configurable metric name for hosts
+     *
+     * @env NEW_RELIC_METRIC_HOST_NAME
+     */
+    metric_hostname: '',
+    /**
      * ip address preference when creating hostnames
      *
      * @env NEW_RELIC_IPV_PREFERENCE
--- a/lib/config/env.js
+++ b/lib/config/env.js
@@ -94,6 +94,7 @@ const ENV_MAPPING = {
   },
   process_host: {
     display_name: 'NEW_RELIC_PROCESS_HOST_DISPLAY_NAME',
+    metric_hostname: 'NEW_RELIC_METRIC_HOST_NAME',
     ipv_preference: 'NEW_RELIC_IPV_PREFERENCE'
   },
   api: {
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -736,6 +736,10 @@ function getHostnameSafe() {
   this.getHostnameSafe = function getCachedHostname() {
     return _hostname
   }
+  if (process.env.NEW_RELIC_METRIC_HOST_NAME) {
+    _hostname = process.env.NEW_RELIC_METRIC_HOST_NAME
+    return _hostname
+  }
   try {
     _hostname = os.hostname()
     return _hostname


astormnewrelic:
Thanks for the PR @cesine, that looks like a clever solution to your specific kubernetes problem.  We're glad you were able to use GitHub to solve your own problem.

Regarding landing this PR in the agent -- New Relic's currently considering a number of solutions for hostname explosions -- including configuration based approaches similar to your own.  We won't be able to accept this PR at this time, but it has helped influence our thinking on the subject. Have you noticed any unexpected behavior from the rest of APM/NewRelic from using this hostname override?  If so (and you have the time) we'd love to hear about those details.

cesine:
thanks @astormnewrelic ! Closing this, looking forward to the updates 👍

SocalNick:
Hi @astormnewrelic - is there anywhere we can follow along? We are also experiencing this after moving to kubernetes.

astormnewrelic:
@SocalNick Re: following along, your best bet is to keep an eye on the node agent release notes page -- when we land something to fix this it'll be mentioned there: https://docs.newrelic.com/docs/release-notes/agent-release-notes/nodejs-release-notes

That said -- if you're currently blocked by an issue your best bet for official support is to submit something via our official support channels: https://support.newrelic.com/ Doing this will get your issue in front of the folks best positioned to get you a solution.

dragomir-enachi-ibm:
Thanks @cesine , I've been looking for this in last 20 minutes then found your closed PR. 
We are facing same issues where we would like to set the instance name from config.
I wish your PR was accepted.

astormnewrelic:
@dragomir-enachi-ibm Nice to meet you -- the best path forward for now is to contact [the fine folks in our support organization](https://support.newrelic.com/). They've got the training and resources to help with MGI/hostname-explosion issues, (and all sorts of other issues to boot).  Also, for what it's worth, hearing about these issues through official channels helps New Relic prioritize getting fixes like the above into the agent itself.

dragomir-enachi-ibm:
Hi @cesine, 

Nice to meet you,  I've filed a ticket on NR support system, let's see what they say. Will post updated when I get any.

Thank you,
Best regards,
Dragomir Enachi

Issue #316, opened 2019-12-02
Support http/https APIs for node >= 10


kennethaasan:
Related issue: https://github.com/sindresorhus/got/issues/876

Seems like https://github.com/newrelic/node-newrelic/blob/master/lib/instrumentation/core/http.js#L447-L451 breaks requests. 

Also seems like the dependency _node-agent-base_ that you have through _https-proxy-agent_ doesn't support http/https APIs for node >= 10: https://github.com/TooTallNate/node-agent-base/issues/29

astormnewrelic:
@kennethaasan Nice to meet you, thanks for the bug report.  

First, some official business: The best way to get support for the Node Agent is via our official support channels: https://support.newrelic.com/ -- reporting things via the official channels will ensure your issue gets in front the folks best situated to give you immediate help _and_ make sure the issues routed properly through our org. 

With that out of the way -- do you have some sample code/agent-configuration that shows how things break?  I took a look at the links you provided and there's enough extra parties involved that I'm not 100% clear what the problem you're describing is.  If you could provide a small program that demonstrates how things breaks we'll be in a much better position to help out.

kennethaasan:
@astormnewrelic Thanks, I made a support ticket now.

I made a PR to fix this in https://github.com/newrelic/node-newrelic/pull/317

astormnewrelic:
@kennethaasan 

> do you have some sample code/agent-configuration that shows how things break? 

Would you be able to provide the above so we can better understanding the issue you're facing?

kennethaasan:
@astormnewrelic, 

The error message I get is _TypeError [ERR_INVALID_ARG_TYPE]: The "listener" argument must be of type Function. Received type object_.

The NR agent config I use is:
```js
process.env.NEW_RELIC_NO_CONFIG_FILE = 'true';
process.env.NEW_RELIC_DISTRIBUTED_TRACING_ENABLED = 'true';
process.env.NEW_RELIC_LOG_ENABLED = 'false';
process.env.NEW_RELIC_ERROR_COLLECTOR_IGNORE_ERROR_CODES = '400';
import 'newrelic';
```

The error happens when I use got v10 (https://github.com/sindresorhus/got) to request JSON feeds:
```ts
const agent = new https.Agent({
  keepAlive: true,
  maxSockets: 50,
  maxFreeSockets: 50,
})

export async function gotGet<V extends object>(
  url: string
): Promise<Response<V>> {
  const newrelicHeader = newrelic
[...]

PR #318, opened 2019-12-18
#316: Adds support for http/https APIs for node >= 10

kennethaasan:
## CHANGE LOG
- Adds support for http/https APIs for node >= 10

## NOTES
- Please see https://github.com/newrelic/node-newrelic/issues/316

### DIFF
--- a/lib/instrumentation/core/http-outbound.js
+++ b/lib/instrumentation/core/http-outbound.js
@@ -29,19 +29,27 @@ const NEWRELIC_SYNTHETICS_HEADER = 'x-newrelic-synthetics'
  *
  * @return {http.ClientRequest} The instrumented outbound request.
  */
-module.exports = function instrumentOutbound(agent, opts, makeRequest) {
+module.exports = function instrumentOutbound(agent, _url, opts, makeRequest) {
+  if (typeof _url === 'string') {
+    _url = url.parse(_url)
+  }
+
   if (typeof opts === 'string') {
     opts = url.parse(opts)
   } else {
     opts = copy.shallow(opts)
   }
 
-  let hostname = opts.hostname || opts.host || DEFAULT_HOST
-  let port = opts.port || opts.defaultPort
-  if (!port) {
-    port = (!opts.protocol || opts.protocol === 'http:') ? DEFAULT_PORT : DEFAULT_SSL_PORT
+  if (!_url) {
+    _url = opts
   }
 
+  let port = _url.port || opts.defaultPort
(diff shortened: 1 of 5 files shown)


kennethaasan:
@nijotz, please see my comment in https://github.com/newrelic/node-newrelic/issues/316#issuecomment-567257125

This is a nock issue

nijotz:
Ah, thank you I hadn't seen the latest comments

nijotz:
Hi, so I did some testing, and added a test that passes a URL object as the first and only parameter, and the test fails. Is it okay with you if I continue to build on this PR and add commits?

kennethaasan:
@nijotz, can you paste the test you wrote here? I'm not able to reproduce it.

nijotz:
@kennethaasan Actually, it looks like there's a mix up on my end. I'm still playing with it.

nijotz:
I added a [commit](https://github.com/newrelic/node-newrelic/pull/318/commits/77227e06d0af28c38f4ea393ef470650da5888e1) (hope you don't mind) to test different combinations of function signatures across http/https get/request. And it looks like if only a URL object is passed in as an argument, it fails. It looks like this is because [here](https://github.com/newrelic/node-newrelic/blob/77227e06d0af28c38f4ea393ef470650da5888e1/lib/instrumentation/core/http-outbound.js#L34), it checks for `_url` being a string, but if `_url` is a URL object, it doesn't do anything and `opts` is undefined. So when it gets down [here](https://github.com/newrelic/node-newrelic/blob/77227e06d0af28c38f4ea393ef470650da5888e1/lib/instrumentation/core/http-outbound.js#L135), `request.path` is `undefined` and it errors.

nijotz:
I have time to work on this now and I can keep adding onto this PR if that is okay with you.

kennethaasan:
That makes sense. Feel free to keep adding to this PR. I'm happy as long as it's shipped ASAP :)

nijotz:
Hi @kennethaasan, we have an internal Github that we use. I put your commits that were on this PR on an internal branch. I added some commits to add some tests and tweak some of the code. This was released last week as 6.4.2.

[Your commit](https://github.com/newrelic/node-newrelic/commit/245b8e95532b10092c8b815e58cc361e682e3817)

[The release](https://docs.newrelic.com/docs/release-notes/agent-release-notes/nodejs-release-notes/node-agent-642)

nijotz:
Please let us know if you continue to have issues. Closing PR.

PR #332, opened 2020-02-26
Add core files to NPM file allowlist

JamesPeiris:
## NOTES
This PR aims to utilise NPM's file allowlisting feature to reduce the size of the package and follow best security practices. The core (used-in-production) files and directories from the repo . are referenced in a list against the `files` property of `package.json` to achieve this.

This should directly replace the `.npmignore` file, which is advised against because:
- It is a blacklist, so additional directories and files will have to be manually added to prevent bloat / minimise the risk of leaking sensitive data
- It has caused issues when swapping from `.gitignore` (see [this article](https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d)), although as this repo already contains this it is not a problem, more another reason against the choice to use an `.npmignore` file.

There are no functional changes, so it is possible that a pre-release version with this allowlist should be published on npm and tested to ensure all core functionality remains, then to evaluate the improvement in the package size once the docker/travis/git files are no longer bundled.

### DIFF
--- a/package.json
+++ b/package.json
@@ -186,6 +186,7 @@
   "files": [
     "lib/",
     "api.js",
+    "LICENSE",
     "index.js",
     "stub_api.js",
     "newrelic.js",
--- a/package.json
+++ b/package.json
@@ -182,5 +182,13 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/newrelic/node-newrelic.git"
-  }
+  },
+  "files": [
+    "lib/",
+    "api.js",
+    "index.js",
+    "stub_api.js",
+    "newrelic.js",
+    "bin/tracetractor"
+  ]
 }
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,0 @@
-test
-examples
-out
-Makefile


package.json
```diff
@@ -182,5 +182,13 @@
...
     "url": "https://github.com/newrelic/node-newrelic.git"
-  }
+  },
+  "files": [
+    "lib/",
+    "api.js",
+    "index.js",
+    "stub_api.js",
+    "newrelic.js",
+    "bin/tracetractor"
```
JamesPeiris: Unsure about this one as it looks to be a utility clients can leverage (based on the comment [at the top of the file](https://github.com/newrelic/node-newrelic/blob/master/bin/tracetractor#L4)).

If that is the case might it be worth separating all bin contents related to testing such as:

- `ca-gen.js`
- `cassandra-setup.sh`
- `clean.sh`
- `compare-bench-results.js`
- `docker-env-vars.sh`
- `docker-services.sh`
- `publish-docs.sh`
- `run-bench.js`
- `run-versioned-tests.sh`
- `smoke.sh`
- `ssl.sh`
- `test-naming-rules.js`
- `travis-install-gcc5.sh`
- `travis-install-mongo.sh`
- `travis-node.sh`
- `travis-setup.sh`
- `update-ca-bundle.sh`
- `update-cats.sh`

...Into their own bin folder within the `test/` directory, and all scripts/binaries intended for customer use such as:

- `tracetractor`
- maybe others...

...can remain in the top level bin?

After which the entry in this list can simply be changed to `bin/`.

carlonewrelic:
Hi @JamesPeiris , thank you for your suggestion and contribution!  We currently don't have a way to approve or merge this PR into the repository, but we will take your suggestions into consideration for the next release.

package.json
carlonewrelic: Great suggestion!  I like the idea of separating testing related bin files.

JamesPeiris:
No problem @carlonewrelic, thank you for taking the time to review this.

nijotz:
I changed the base of this PR from `master` -> `main`

michaelgoin:
Hey! Sorry we left this hanging for so long. I think this is a good idea but it also requires us to take some careful consideration.

Given we've let this get so far out of date and it is not somethign we have focused on just yet, I'm going to close this out but create a github issue to track.

michaelgoin:
Here's the GitHub issue to track: https://github.com/newrelic/node-newrelic/issues/618

PR #339, opened 2020-04-30
fix: update deps to address security issue

devfreddy:
## CHANGE LOG

Fixes/updates security vulnerabilities in npm packages for `minimist` and `mkdirp`.


michaelgoin:
Thanks! We'll pull this down and it will get pushed back here with the next release to close out this PR.

PR #354, opened 2020-05-19
Fastify Route Naming

astormnewrelic:
## Proposed Release Notes

- Basic support for Fastify transaction naming

## Details

- Provides transaction naming for fastify routes.  Does not fully address error handling, or the agent's usual perf. issues/trade-offs with instrumentation, or deeper fastify features. Intended as a base to work from

- Behind a `fastify_instrumentation` local feature flag.  Add the following to your `newrelic.js` configuration to enable
  ```
  module.exports = {
    /* the rest of your configuration */
    feature_flag       : {
      fastify_instrumentation: true
    }
    /* the rest of your configuration */
  }
  ``` 
## To Do

- [x] In Node Agent repository or in external repository?
  - Feature flagged here for now
- [x] Perhaps gate this behind a feature-flag?

- [x] Versioned runner


astormnewrelic:
> Some initial thoughts...

Thanks for feedback here -- feature flags and wrapping the integration tests in a versioned runner makes sense.  I've bumped this into draft state and will work on those as snippets of innovation time between meetings allows. 

Re: a separate package -- yeah, a wider team/community discussion is needed there. In addition to what you said, our current 1 package = 1 repo thing might be a little unwieldy with out current open source project.  This may be one more stone in the monorepo-via-[lerna](https://github.com/lerna/lerna) bucket.

mcollina:
Let me know if the current hooks are enough for you folks.

A7med-Ragheb:
After enable this configuration, I can see the routes correctly in the transactions apm but all the transactions prefix with word 'Fastify/...', Why? And, Is it also a configuration that I can diable?

michaelgoin:
@A7med-Ragheb I recently noticed that too. It appears to be a naming bug. While Fastify should exist in the name, it should not be showing by the UI on raw display (but rather shown as the framework in certain displays). I think this is because the raw name appears to have both "webtransaction" and "webframeworkuri" which typically we'd only have one of those.

I can add a GitHub issue for that or you are welcome to if you'd prefer to track.

A7med-Ragheb:
@michaelgoin Thanks for your reply, I've created a new bug, Can you please check it to see If it has enough information? Please, let me know If I can add anything else.
https://github.com/newrelic/node-newrelic/issues/757

michaelgoin:
@A7med-Ragheb looks great, thank you!

PR #355, opened 2020-05-20
Test filtering

nijotz:
## Proposed Release Notes

* More precise filtering of leveraging secrets.


bin/ssl.sh
```diff
@@ -8,7 +8,7 @@ set -x  # be chatty and show the lines we're running
 # that are mysterious and not understood, so let
 # bail early if we detect that's the case
 ENGINE_OPENSSL=`openssl version | awk '{print $(1)}'`
-if [ "$ENGINE_OPENSSL" == "LibreSSL" ]
+if [ "$ENGINE_OPENSSL" = "LibreSSL" ]
```
nijotz, line 11: Noticed some shell errors on github actions when running on ubuntu. Saw that the `==` syntax isn't cross-shell compatible. Same with `-a` vs `-e`

nijotz: These changes fixed those errors

package.json
```diff
@@ -163,7 +162,7 @@
   },
   "devDependencies": {
     "@newrelic/proxy": "^2.0.0",
-    "@newrelic/test-utilities": "^3.1.0",
+    "@newrelic/test-utilities": "^3.3.0",
```
nijotz, line 165: The new test utilities supports skipping using multiple keywords

bin/run-versioned-tests.sh
```diff
@@ -21,5 +21,14 @@ fi
...
+
+# Don't run the aws-sdk tests if we don't have the keys set
+if [[ -z "$AWS_ACCESS_KEY_ID" ]]; then
+  skip="${skip},aws-sdk"
+fi
+
+
 # This is meant to be temporary. Remove once new major version with fixes rolled into agent.
-time ./node_modules/.bin/versioned-tests $VERSIONED_MODE -i 2 -s koa ${directories[@]}
+time ./node_modules/.bin/versioned-tests $VERSIONED_MODE -i 2 -s $skip ${directories[@]}
```
nijotz, line 34: We're able to skip with multiple keywords now

bin/run-versioned-tests.sh
```diff
@@ -21,5 +21,14 @@ fi
 
 export AGENT_PATH=`pwd`
 
+# I think we're skipping koa because a new version was breaking tests
```
michaelgoin, line 24: @koa/router 8.03+ breaks segment naming for nested routes.

We need to prob adjust the versions in koa module and publish to limit until fixed (which unfortunately is not straightforward). We might update the comment to that to help future us. Can be a follow-up.

bin/ssl.sh
michaelgoin: good catch. bash... is not my favorite haha. *watches dev cred decrement by 5*

.github/workflows/ci-workflow.yml
```diff
@@ -75,40 +75,15 @@ jobs:
...
         LASP_SECURE_LICENSE: ${{ secrets.LASP_SECURE_LICENSE }}
-        BENDER_LICENSE: ${{ secrets.BENDER_LICENSE }}
-        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
 
   versioned:
     runs-on: ubuntu-latest
 
-    if: github.repository == 'newrelic/node-newrelic'
-
```
astormnewrelic, line 111: what's the intent behind removing the `if: github.repository == 'newrelic/node-newrelic'` conditional?

.github/workflows/ci-workflow.yml
```diff
@@ -75,40 +75,15 @@ jobs:
...
-    - name: Run Versioned Tests
-      run: npm run secrets
       env:
+        TEST_LICENSE: ${{ secrets.TEST_LICENSE }}
+        COLLECTOR_LICENSE: ${{ secrets.COLLECTOR_LICENSE }}
         LASP_LICENSE: ${{ secrets.LASP_LICENSE }}
         LASP_SECURE_LICENSE: ${{ secrets.LASP_SECURE_LICENSE }}
-        BENDER_LICENSE: ${{ secrets.BENDER_LICENSE }}
-        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
astormnewrelic, line 83: I see we're removing the AWS_ACCESS_KEY_ID variables, which (based on other changes) means we won't run the AWS SDK versioned tests. Presuming this is intentional -- if not see-something/say-something.

bin/ssl.sh
nijotz: Oh this was def a TIL for me

.github/workflows/ci-workflow.yml
michaelgoin: it was only necessary for skipping which this isn't doing anymore. also, it never worked for forked-repo PR's

.github/workflows/ci-workflow.yml
michaelgoin: this is only impacting integration tests. the versioned tests still have the secrets.

PR #360, opened 2020-05-22
Don't skip tests if secrets should be available

nijotz:
## Details
* Another follow up to the CI stuff, adding a failsafe. It forces tests that need secrets to run when the branch being run is master even if the secret they need is not available. This will force tests to fail, we'll be notified, and we can see why our secrets aren't working.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -63,6 +63,7 @@ jobs:
         COLLECTOR_LICENSE: ${{ secrets.COLLECTOR_LICENSE }}
         LASP_LICENSE: ${{ secrets.LASP_LICENSE }}
         LASP_SECURE_LICENSE: ${{ secrets.LASP_SECURE_LICENSE }}
+        FORCE_RUN_TESTS_WITH_SECRETS: ${{ github.repository_owner == 'newrelic' && github.head_ref == 'master' }}
 
   versioned:
     runs-on: ubuntu-latest
--- a/test/helpers/secrets.js
+++ b/test/helpers/secrets.js
@@ -1,0 +1,29 @@
+'use strict'
+
+/**
+ * A helper function to get secrets needed by tests
+ */
+function getTestSecret(secretName) {
+  const envVar = process.env[secretName] || ''
+  return envVar.trim()
+}
+
+/**
+ * Checks whether any of the secrets needed by the test are missing or if the
+ * FORCE_RUN_TESTS_WITH_SECRETS env var is set. This is set by the github
+ * actions workflow when tests are being run on the newrelic repository where
+ * secrets should be available. In this case, even if the license key that a
+ * test is looking for is not present, run the test anyway and let it fail,
+ * because it should be present on the newrelic repo.
+ */
+function shouldSkipTest(...secrets) {
+  const missing = secrets.some(s => !s)  // !s catches empty strings
+  const forceTestRun = process.env.FORCE_RUN_TESTS_WITH_SECRETS === 'true'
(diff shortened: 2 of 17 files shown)


test/integration/lasp_connect.tap.js
```diff
@@ -41,13 +42,13 @@ tap.test('connecting with a LASP token should not error', {skip}, function(t) {
       t.end()
     })
   })
-})
+});
 
-skip = !Boolean(process.env.LASP_SECURE_LICENSE)
-tap.test('missing required policies should result in shutdown', {skip}, function(t) {
+({skip, license} = getTestSecret('LASP_SECURE_LICENSE'))
```
nijotz, line 47: https://gist.github.com/nijotz/90b0e14f87f3c48f2a7fe1b5e5565760

test/helpers/secrets.js
```diff
@@ -0,0 +1,30 @@
+'use strict'
+
+/**
+ * A helper function to get secrets needed by tests
+ */
+function getTestSecret(secretName) {
+  const envVar = process.env[secretName]
```
michaelgoin, line 7: Wouldn't worry about it now but maybe a `.trim()` would be nice here eventually to prevent a certain brand of copy-paste error.

test/helpers/secrets.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Checks whether any of the secrets needed by the test are missing or if the
+ * FORCE_RUN_TESTS_WITH_SECRETS env var is set. This is set by the github
+ * actions workflow when tests are being run on the newrelic repository where
+ * secrets should be available. In this case, even if the license key that a
+ * test is looking for is not present, run the test anyway and let it fail,
+ * because it should be present on the newrelic repo.
+ */
+function shouldSkipTest() {
+  const secrets = Array.from(arguments)
```
michaelgoin, line 20: Not sure I understand the need for this code? Won't there only ever be a single value as we are grabbing from a single ENV var? Why do we need a generic "arguments"?

Beyond that comment, you could use `...args` if you didn't want to use `Array.from()`

.github/workflows/ci-workflow.yml
```diff
@@ -63,6 +63,7 @@ jobs:
         COLLECTOR_LICENSE: ${{ secrets.COLLECTOR_LICENSE }}
         LASP_LICENSE: ${{ secrets.LASP_LICENSE }}
         LASP_SECURE_LICENSE: ${{ secrets.LASP_SECURE_LICENSE }}
+        FORCE_RUN_TESTS_WITH_SECRETS: ${{ github.repository_owner == 'newrelic' && github.head_ref == 'master' }}
```
michaelgoin, line 66: Guessing the specific branch check is to avoid running in PR's from forked repos, since ENV vars aren't limited to the main repo like secrets.

test/helpers/secrets.js
nijotz: Ya, just an attempt at future-proofing. But ya, `...args` is convention, I'll switch that up.

.github/workflows/ci-workflow.yml
nijotz: Ya, ran into an issue where the repository_owner for PRs from forked repos is still `newrelic`. And some people work from the master branch on their forked repo. So had to add both checks.

test/helpers/secrets.js
nijotz: Good call, I'll add that in

nijotz:
Made suggested changes, pushed, waiting on tests.

PR #365, opened 2020-05-29
Add AWS API Gateway V2 Support

kolanos:
AWS released a new version of their API Gateway service. This adds support for V2.

v1: https://github.com/awsdocs/aws-lambda-developer-guide/blob/master/sample-apps/nodejs-apig/event.json
V2: https://github.com/awsdocs/aws-lambda-developer-guide/blob/master/sample-apps/nodejs-apig/event-v2.json

### DIFF
--- a/lib/serverless/event-sources.json
+++ b/lib/serverless/event-sources.json
@@ -21,7 +21,7 @@
             "httpMethod",
             "path",
             "requestContext",
-            "resource"
+            "requestContext.stage"
         ]
     },
     "cloudFront": {


lib/serverless/event-sources.json
```diff
@@ -21,7 +21,7 @@
             "httpMethod",
             "path",
             "requestContext",
-            "resource"
+            "requestContext.stage"
```
michaelgoin, line 24: I see we've removed one key for matching that does not exist in v2 and replaced with a key that exists in both.

kolanos:
@michaelgoin This should be good to go.

Issue #368, opened 2020-06-02
~100MB log is installed using yarn/npm (probably using others also)


bambooCZ:
## Description

`node_modules` are already considered to be the [heaviest object in the universe](https://www.reddit.com/r/ProgrammerHumor/comments/6s0wov/heaviest_objects_in_the_universe/?utm_source=share&utm_medium=web2x). You guys add another dimension to it :). On clean installation there is ~100MB log included.

## Steps to Reproduce

```
mkdir my-new-project
cd my-new-project
npm init
npm i newrelic --only=production
ls -lh node_modules/newrelic
```

## Expected Behavior

Logs shouldn't be installed by any package manager. Especially logs that are this big.

## Relevant Logs / Console output
```
$ ls -lh node_modules/newrelic/
total 107M
-rw-r--r--  1 usr grp  51K 26. říj  1985  api.js
drwxr-xr-x  2 usr grp 4,0K  2. čen 03.35  bin
-rw-r--r--  1 usr grp 3,9K 26. říj  1985  CONTRIBUTING.md
-rw-r--r--  1 usr grp 5,6K 26. říj  1985  index.js
-rw-r--r--  1 usr grp  337 26. říj  1985  jsdoc-conf.json
drwxr-xr-x 19 usr grp 4,0K  2. čen 03.35  lib
-rw-r--r--  1 usr grp  20K 26. říj  1985  LICENSE
-rw-r--r--  1 usr grp  12K 26. říj  1985 'Migration Guide.md'
-rw-r--r--  1 usr grp 106M 26. říj  1985  newrelic_agent.log
-rw-r--r--  1 usr grp 1,5K 26. říj  1985  newrelic.js
-rw-r--r--  1 usr grp 182K 26. říj  1985  NEWS.md
-rw-r--r--  1 usr grp 6,3K  2. čen 03.35  package.json
-rw-r--r--  1 usr grp  31K 26. říj  1985  README.md
-rw-r--r--  1 usr grp  719 26. říj  1985  SECURITY.md
-rw-r--r--  1 usr grp 2,9K 26. říj  1985  stub_api.js
-rw-r--r--  1 usr grp  34K 26. říj  1985  THIRD_PARTY_NOTICES.md
```
Do you see the `newrelic_agent.log` :)

michaelgoin:
Oof. Guessing that is only in the latest version and would be my bad.

Can you confirm this is `v6.8.0`?

[...]

PR #369, opened 2020-06-02
Adds commonly skipped items to npmignore to avoid accidentally publishing

michaelgoin:
## Proposed Release Notes

* Added several items to `.npmignore` to prevent accidental publishing.

## Links

* https://github.com/newrelic/node-newrelic/issues/368

## Details

When publishing from a new location, I caught a few other things but missed that I had a rogue `newrelic_agent.log` hanging out. Our current `.npmignore` doesn't filter that out.

Long-term, using the "files" configuration which gives an allow-list will save us from such things including not needing to update for all possible cases in the future.

There's a backlog item and a community PR hanging-out around using the "files" list. That sort of change is higher-risk, though, as we could break something fundamental when we convert. As such, doing the quick fix now and leaving that for a more dedicated/careful approach.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,3 +2,10 @@ test
 examples
 out
 Makefile
+package-lock.json
+newrelic_agent.log
+**/node_modules/**
+.nyc_output
+.vscode
+.eslint*
+.github


.npmignore
```diff
@@ -2,3 +2,10 @@ test
 examples
 out
 Makefile
+package-lock.json
```
michaelgoin, line 5: i believe npm always ignores this by default (could look up) but might as well make explicit.

PR #370, opened 2020-06-02
Waits for error to trigger on test before ending transaction.

michaelgoin:
## Proposed Release Notes

* Fixed Redis client w/o callback versioned test flicker.

  Doesn't end transaction until error encountered. Increases wait time for first operation which has to complete for the second operation to be successful.

## Details

Should reduce test flickers by not having a hard-set 100ms sleep for the operation. Relies on test framework timeout to end if call hangs. Worst case, will be easier to debug.

Did find second issue. The first key has to exist for the second call to fail. Increasing the timeout will still be somewhat fragile but trying not to have to do major surgery here.

### DIFF
--- a/test/versioned/redis/redis.tap.js
+++ b/test/versioned/redis/redis.tap.js
@@ -120,12 +120,15 @@ test('Redis instrumentation', {timeout : 10000}, function(t) {
   t.test('when called without a callback', function(t) {
     t.plan(4)
 
+    let transaction = null
+
     helper.runInTransaction(agent, function(tx) {
       client.set('testKey', 'testvalue')
       setTimeout(function() {
         // This will generate an error because `testKey` is not a hash.
         client.hset('testKey', 'hashKey', 'foobar')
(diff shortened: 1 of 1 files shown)


test/versioned/redis/redis.tap.js
```diff
@@ -120,12 +120,15 @@ test('Redis instrumentation', {timeout : 10000}, function(t) {
...
+
     helper.runInTransaction(agent, function(tx) {
       client.set('testKey', 'testvalue')
       setTimeout(function() {
         // This will generate an error because `testKey` is not a hash.
         client.hset('testKey', 'hashKey', 'foobar')
-        setTimeout(tx.end.bind(tx), 100)
+
+        transaction = tx
       }, 100) // Redis calls should never take 100 ms
```
nijotz, line 132: Never say never :D

michaelgoin:
OK. 3 straight passes so probably patched the issue at least for a while. Or it will fail once merged :)

PR #371, opened 2020-06-02
Moves error data collection checks up to where errors are added

michaelgoin:
## Proposed Release Notes

* Drops errors earlier during collection when error collection is disabled.

  Error attributes will no longer show up on spans when error collection has been disabled. Other unnecessary work will also be avoided.

### DIFF
--- a/lib/errors/error-collector.js
+++ b/lib/errors/error-collector.js
@@ -193,6 +193,12 @@ class ErrorCollector {
       return
     }
 
+    const shouldCollectErrors = this._shouldCollectErrors()
+    if (!shouldCollectErrors) {
+      logger.trace('error_collector.enabled is false, dropping application error.')
+      return
+    }
+
     if (errorHelper.shouldIgnoreError(transaction, error, this.config)) {
       logger.trace('Ignoring error')
       return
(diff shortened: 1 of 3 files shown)


lib/errors/error-collector.js
```diff
@@ -328,6 +333,17 @@ class ErrorCollector {
     this.seenObjectsByTransaction = Object.create(null)
   }
 
+  _shouldCollectErrors() {
```
astormnewrelic, line 336: +2 for this method.  As agents mature they pickup a lot of these "how should I work under these sets of complex configuration" code chunks.  Getting these into named functions and methods that makes their intent clear is a win.

lib/errors/error-collector.js
```diff
@@ -223,7 +229,15 @@ class ErrorCollector {
    * @param {Exception}  exception The Exception to be traced.
    */
   addUserError(transaction, error, customAttributes) {
-    if (!error) return
+    if (!error) {
```
astormnewrelic, line 232: 👍 Always braces on conditionals is the correct color for this perennial bikeshed.

lib/errors/error-collector.js
```diff
@@ -270,18 +284,6 @@ class ErrorCollector {
...
-    // short-circuit sooner?
-    if (
-      !this.config.collect_errors ||
-      !this.config.error_collector ||
-      !this.config.error_collector.enabled
-    ) {
-      return false
-    }
-
     if (exception.error) {
```
astormnewrelic, line 287: Worth noting that the code between the start of `collect` and this old guard clause ran in the world world, but doesn't run the new world (if the error collector is disable).  This includes a call to `_haveSeen`, which changes the state of the `ErrorCollector`.  However, the state changed by `_haveSeen` appears to only be used by `_haveSeen` itself, and no other methods in the agent call `_haveSeen` (hooray for respecting fake protected/private leading `_`) , so this should be fine.

PR #374, opened 2020-06-03
Add expected error to span

michaelgoin:
## Proposed Release Notes

* Added 'error.expected' attribute to span active at time expected error was noticed.

### DIFF
--- a/lib/transaction/index.js
+++ b/lib/transaction/index.js
@@ -697,6 +697,17 @@ function _linkExceptionToSegment(exception) {
   segment.addAttribute('error.message', details.message)
   segment.addAttribute('error.class', details.type)
 
+  const isExpected = errorHelper.isExpectedException(
+    this,
+    exception.error,
+    this.agent.config,
+    urltils
+  )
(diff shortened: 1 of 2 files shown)


lib/transaction/index.js
```diff
@@ -692,11 +692,21 @@ function _linkExceptionToSegment(exception) {
...
 
+  const config = this.agent.config
+
   // Add error attributes to the span
-  const details = exception.getErrorDetails(this.agent.config)
+  const details = exception.getErrorDetails(config)
   segment.addAttribute('error.message', details.message)
   segment.addAttribute('error.class', details.type)
 
+  const isExpected =
```
michaelgoin, line 702: This was a tradeoff between some code duplication or some work/error extraction duplication. So I chose a little bit of code duplication.

Ultimately, the expected/ignored and error processing pipeline could use a fair amount of streamlining but that is also a fair bit of work. As such, not biting off with this PR. May or may not approach further depending on how current work effort progresses.

PR #377, opened 2020-06-05
Adds request headers map key-value paris to Infinite Tracing gRPC metadata

michaelgoin:
## Proposed Release Notes

* Added connect response request_headers_map key-value pairs to Infinite Tracing gRPC connection metadata.

## Details

This PR is based off the same branch as https://github.com/newrelic/node-newrelic/pull/376 and should be reviewed/merged afterwards. This is to avoid painful conflicts with modifying parts of certain tests.

The keys MUST be sent as lowercase via the metadata per the spec.

### DIFF
--- a/lib/grpc/connection.js
+++ b/lib/grpc/connection.js
@@ -52,6 +52,7 @@ class GrpcConnection extends EventEmitter {
     this._endpoint = this.getTraceObserverEndpoint(traceObserverConfig)
     this._licensekey = null
     this._runId = null
+    this._requestHeadersMap = null
 
     this.stream = null
   }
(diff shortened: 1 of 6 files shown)


lib/grpc/connection.js
```diff
@@ -277,7 +288,7 @@ class GrpcConnection extends EventEmitter {
    * this makes an http2 request with the metadata, and then returns
    * a stream for further writing.
    */
-  _connectSpans(endpoint, license_key, run_id) {
+  _connectSpans() {
     const packageDefinition = protoLoader.loadSync(pathProtoDefinition, protoOptions)
```
astormnewrelic, line 292: 🧀 move acknowledged.

PR #384, opened 2020-06-12
Custom transaction atts to span

michaelgoin:
## Proposed Release Notes

* Custom transaction attributes added via `API.addCustomAttribute` or `API.addCustomAttributes` will now be propagated to the currently active span, if available.

  **Security Recommendation:** Review your Transaction Event attributes configuration. Any attribute include or exclude setting specific to Transaction Events should be applied to your Span Attributes configuration or global attributes configuration. Please see [Node.js agent attributes](https://docs.newrelic.com/docs/agents/nodejs-agent/attributes/nodejs-agent-attributes#configure-attributes) for more on how to configure.

## Details

* Added a priority attribute system. Much of the built-in attribute stuff isn't *really* required for where this is used but leaves open in case we want this to apply to other or prior usages of attributes. Kept intentionally separate for now to limit risk of introducing the new system. See prioritized attributes unit tests for business logic here. 

* Moved custom span attributes off of segments onto SpanContext.

Prioritized attributes allow for prioritization of custom span attributes over transaction attributes. The rough business logic looks like: 
  * attribute w/ same key from transaction should override from transaction
  * attribute w/ same key from span should override from transaction
  * attribute w/ same key from span should override from span
  * attribute w/ same key from transaction should not override from span
  * new attribute from span should replace from transaction when over the limit
  * new attribute from transaction should not replace from transaction or from span, when over the limit

Did some benchmarking of a worst-case scenario around span adding/dropping and optimized a little bit. This made me lose a little confidence in our benchmarking, based on what I was seeing for certain scenarios, but current solution against optimized was verified both with the include benchmark in this PR as well as using `benchmarks` for an iteration based benchmarking approach. 

Due to awkward business logic situation, had to change the optimization a bit. There's ways this could be faster but trying not to make the code too complicated for a "please don't do this" scenario and also avoid introducing a new data structure behind the scenes. That being said, the little bit of added complication here helps keeps too much damage from happening in the case where attribute maximums are hit and we are determining if we can or cannot drop. It will perform the worst when all attributes start out as a lower priority. Either way is much faster than converting to an array and iterating backwards. Def open if others know of a better way of resolving the reverse iteration problem. It also shortens the scan length which is nice even if the reverse iteration wasn't an issue.

### DIFF
--- a/api.js
+++ b/api.js
@@ -272,7 +272,7 @@ API.prototype.setControllerName = function setControllerName(name, action) {
  * @param {string} value The value you want displayed. Must be serializable.
  */
 API.prototype.addCustomAttribute = function addCustomAttribute(key, value) {
-  var metric = this.agent.metrics.getOrCreateMetric(
+  const metric = this.agent.metrics.getOrCreateMetric(
     NAMES.SUPPORTABILITY.API + '/addCustomAttribute'
   )
   metric.incrementCallCount()
@@ -291,24 +291,35 @@ API.prototype.addCustomAttribute = function addCustomAttribute(key, value) {
     return false
   }
 
-  var transaction = this.agent.tracer.getTransaction()
+  const transaction = this.agent.tracer.getTransaction()
   if (!transaction) {
-    return logger.warn('No transaction found for custom attributes.')
+    logger.warn('No transaction found for custom attributes.')
+    return false
   }
 
-  var trace = transaction.trace
+  const trace = transaction.trace
   if (!trace.custom) {
-    return logger.warn(
+    logger.warn(
       'Could not add attribute %s to nonexistent custom attributes.',
       key
     )
+    return false
   }
 
   if (CUSTOM_DENYLIST.has(key)) {
-    return logger.warn('Not overwriting value of NR-only attribute %s.', key)
+    logger.warn('Not overwriting value of NR-only attribute %s.', key)
+    return false
   }
 
   trace.addCustomAttribute(key, value)
+
+  const spanContext = this.agent.tracer.getSpanContext()
+  if (!spanContext) {
+    logger.debug('No span found for custom attributes.')
+    return false
+  }
+
+  spanContext.addCustomAttribute(key, value, spanContext.ATTRIBUTE_PRIORITY.LOW)
 }
 
 /**
@@ -398,14 +409,16 @@ API.prototype.addCustomSpanAttribute = function addCustomSpanAttribute(key, valu
   const spanContext = this.agent.tracer.getSpanContext()
 
   if (!spanContext) {
-    return logger.debug(
+    logger.debug(
       'Could not add attribute %s. No available span.',
       key
     )
+    return false
   }
 
   if (CUSTOM_DENYLIST.has(key)) {
-    return logger.warn('Not overwriting value of NR-only attribute %s.', key)
+    logger.warn('Not overwriting value of NR-only attribute %s.', key)
+    return false
   }
 
   spanContext.addCustomAttribute(key, value)
--- a/api.js
+++ b/api.js
@@ -395,11 +395,11 @@ API.prototype.addCustomSpanAttribute = function addCustomSpanAttribute(key, valu
     return false
   }
 
-  const segment = this.agent.tracer.getSegment()
+  const spanContext = this.agent.tracer.getSpanContext()
 
-  if (!segment) {
+  if (!spanContext) {
     return logger.debug(
-      'Could not add attribute %s. No available span/segment.',
+      'Could not add attribute %s. No available span.',
       key
     )
   }
@@ -408,7 +408,7 @@ API.prototype.addCustomSpanAttribute = function addCustomSpanAttribute(key, valu
     return logger.warn('Not overwriting value of NR-only attribute %s.', key)
   }
 
-  segment.addCustomSpanAttribute(key, value)
+  spanContext.addCustomAttribute(key, value)
 }
 
 API.prototype.setIgnoreTransaction = util.deprecate(
--- a/api.js
+++ b/api.js
@@ -316,7 +316,8 @@ API.prototype.addCustomAttribute = function addCustomAttribute(key, value) {
   const spanContext = this.agent.tracer.getSpanContext()
   if (!spanContext) {
     logger.debug('No span found for custom attributes.')
-    return false
+    // success/failure is ambiguous here. since at least 1 attempt tried, not returning false
+    return
   }
 
   spanContext.addCustomAttribute(key, value, spanContext.ATTRIBUTE_PRIORITY.LOW)
--- a/lib/prioritized-attributes.js
+++ b/lib/prioritized-attributes.js
@@ -1,0 +1,157 @@
+'use strict'
+
+const Config = require('./config')
+const logger = require('./logger').child({component: 'attributes'})
+const isValidType = require('./util/attribute-types')
+const byteUtils = require('./util/byte-limit')
+const properties = require('./util/properties')
+
+const ATTRIBUTE_PRIORITY = {
+  HIGH: Infinity,
+  LOW: -Infinity
+}
+
+class PrioritizedAttributes {
+  constructor(scope, limit = Infinity) {
+    this.filter = makeFilter(scope)
+    this.limit = limit
+
+    this.attributes = new Map()
+  }
+
+
+  isValidLength(str) {
+    return typeof str === 'number' || byteUtils.isValidLength(str, 255)
+  }
+
+  _set(destinations, key, value, truncateExempt, priority) {
+    this.attributes.set(key, {value, destinations, truncateExempt, priority})
+  }
+
+  get(dest) {
+    const attrs = Object.create(null)
+
+    for (let [key, attr] of this.attributes) {
+      if (!(attr.destinations & dest)) {
+        continue
+      }
+
+      attrs[key] = typeof attr.value === 'string' && !attr.truncateExempt
+        ? byteUtils.truncate(attr.value, 255)
+        : attr.value
+    }
+
+    return attrs
+  }
+
+  has(key) {
+    this.attributes.has(key)
+  }
+
+  reset() {
+    this.attributes = new Map()
+  }
+
+  addAttribute(
+    destinations,
+    key,
+    value,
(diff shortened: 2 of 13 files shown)


test/unit/attributes.test.js
```diff
@@ -1,163 +1,233 @@
...
-const {expect} = require('chai')
 const helper = require('../lib/agent_helper')
 const {Attributes} = require('../../lib/attributes')
 const AttributeFilter = require('../../lib/config/attribute-filter')
 
 const DESTINATIONS = AttributeFilter.DESTINATIONS
 const TRANSACTION_SCOPE = 'transaction'
 
-describe('Attributes', () => {
+tap.test('#addAttribute', (t) => {
```
michaelgoin, line 12: Since these tests were going to serve as the start/basis of the prioritized attributes test, I converted this file to use the tap API prior to moving onto that.

test/unit/high-security.test.js
```diff
@@ -243,9 +243,11 @@ describe('high security mode', function() {
     })
 
     it('should not affect addCustomAttribute if high_security is off', function() {
-      agent.config.high_security = false
-      var success = api.addCustomAttribute('key', 'value')
-      should.not.exist(success)
+      helper.runInTransaction(agent, () => {
```
michaelgoin, line 246: I changed other failure cases to return `false` which then failed this test which was relying on those silently failing. Running in a transaction makes everything legit for the test.

api.js
```diff
@@ -291,24 +291,36 @@ API.prototype.addCustomAttribute = function addCustomAttribute(key, value) {
...
   }
 
   if (CUSTOM_DENYLIST.has(key)) {
-    return logger.warn('Not overwriting value of NR-only attribute %s.', key)
+    logger.warn('Not overwriting value of NR-only attribute %s.', key)
+    return false
   }
 
   trace.addCustomAttribute(key, value)
+
```
nijotz, line 315: Got caught up on SpanContexts via https://github.com/newrelic/node-newrelic/pull/372

But curious, why are we keeping the `addCustomAttribute` on the trace, but also adding to the SpanContext?

lib/transaction/trace/segment.js
```diff
@@ -53,7 +53,6 @@ function TraceSegment(transaction, name, recorder) {
   }
 
   this.attributes = new Attributes(ATTRIBUTE_SCOPE)
-  this.customAttributes = new Attributes(ATTRIBUTE_SCOPE, MAXIMUM_CUSTOM_ATTRIBUTES)
```
nijotz, line 56: Removing custom attributes from the trace segment..

lib/spans/span-context.js
```diff
@@ -1,13 +1,36 @@
...
+
+// Scoping impacts memoization. We could decide to add a scope instead of including
+// spans in segment scope in the future.
+const ATTRIBUTE_SCOPE = 'segment'
+
 class SpanContext {
-  constructor(intrinsicAttributes) {
+  constructor(intrinsicAttributes, customAttributes) {
     this.intrinsicAttributes = intrinsicAttributes || Object.create(null)
+
```
nijotz, line 14: And adding custom attributes to the span context. But using Prioritized attributes.

lib/prioritized-attributes.js
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  _hasDroppableAttribute(incomingPriority) {
+    // There will never be anything lower priority to drop
+    if (incomingPriority === ATTRIBUTE_PRIORITY.LOW) {
+      return
+    }
+
+    // Currently only support HIGH/LOW
+    if (!this._lowPriorityAttributeNames) {
```
nijotz, line 145: It looks like this is called every time `addAttribute` is called. Once `_lowPriorityAttributeNames` is created, it won't get updated again. I think the first time an attribute is added, it will create this array, but not add anything to it. After that, this array will never be updated, and the length will always be zero, so this will always return false. Is that right? Maybe the for loop below is supposed to be outside of the conditional above? Confused here..

lib/prioritized-attributes.js
```diff
@@ -0,0 +1,171 @@
...
+      if (properties.hasOwn(attrs, key)) {
+        this.addAttribute(destinations, key, attrs[key])
+      }
+    }
+  }
+
+  _hasDroppableAttribute(incomingPriority) {
+    // There will never be anything lower priority to drop
+    if (incomingPriority === ATTRIBUTE_PRIORITY.LOW) {
+      return
```
nijotz, line 141: I'm guessing this is the same as return false, low priority items will never having something that should be dropped in their favor.

api.js
michaelgoin: Transaction attributes *also* go to Spans We are not preventing/removing the attributes from going to transactions.

lib/prioritized-attributes.js
michaelgoin: This will only be called when we are at the attribute maximum and there is no existing attribute to attempt to update. See the if statement on line 65. Once you are at the attribute maximum, it will never be possible for a new low-priority attribute to be added.

If we ever introduce a "medium" or whatever priority, then we'll have to tweak this optimization.

As such, the first time we look for low priority items we'll find all the low priority items that could ever exist. We'll then slowly pop them out until we know there are only high priority items.

I introduced this because there's no way to reverse-iterate the keys. If you want to do that, you first have to create an array from the collection, which iterates the collection anyways. A less optimal approach, but faster than doing that, is to just always iterate forward but at best that will be O(n) for each subsequent add. This only eats that cost once.

lib/prioritized-attributes.js
michaelgoin: yeah. originally it wasn't a true/false function. i ideally should return false. maybe i'll do that and comment the other function a little more.

lib/prioritized-attributes.js
michaelgoin: oof, you did make me realize a gap where this could break tho. if a high priority attribute replaced a low priority attribute. i'll see if i want to fix this optimization or just ship with the slower method at first

michaelgoin: actually, i take it back. what i was worried about is not possible. but it does make me want a test.

lib/prioritized-attributes.js
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+
+    // Only set the attribute if at least one destination passed
+    const validDestinations = this.filter(destinations, key)
+    if (!validDestinations) {
+      return
+    }
+
+    if (droppableAttributeKey) {
```
carlonewrelic, line 115: Nice, drop the key after all other validations have passed.

lib/spans/span-context.js
```diff
@@ -1,13 +1,36 @@
 'use strict'
 
+const {MAXIMUM_CUSTOM_ATTRIBUTES} = require('../attributes')
+const {PrioritizedAttributes, ATTRIBUTE_PRIORITY} = require('../prioritized-attributes')
+const {DESTINATIONS} = require('../config/attribute-filter')
+
+// Scoping impacts memoization. We could decide to add a scope instead of including
+// spans in segment scope in the future.
+const ATTRIBUTE_SCOPE = 'segment'
```
carlonewrelic, line 9: Do you think it's worth adding 'segment' and 'transaction' scope to DESTINATIONS?

lib/prioritized-attributes.js
```diff
@@ -0,0 +1,196 @@
...
+
+    // Item may not get dropped, so we simply store the index as
+    // an upper maximum and allow a future pass to clear out.
+    this.lastFoundIndexCache[incomingPriority] = foundIndex
+
+    return lowerPriorityAttributeName
+  }
+}
+
+function makeFilter(scope) {
```
carlonewrelic, line 184: Here's where we are using current attribute filtering system 👍

test/integration/api/custom-attribute-span-propogation.tap.js
```diff
@@ -0,0 +1,225 @@
...
+function getCustomSpanAttributes(agent) {
+  const spanContext = agent.tracer.getSpanContext()
+  return spanContext && spanContext.customAttributes.get(DESTINATIONS.SPAN_EVENT)
+}
+
+function getCustomTransactionAttributes(transaction) {
+  return transaction.trace.custom.get(DESTINATIONS.TRANS_SCOPE)
+}
+
+function batchAddCustomAttributes(api, attributeCount) {
```
carlonewrelic, line 211: nice helper function for adding multiple attributes

test/integration/api/custom-attribute-span-propogation.tap.js
```diff
@@ -0,0 +1,225 @@
...
+
+        const customSpanAttributes = getCustomSpanAttributes(agent)
+        t.equal(customSpanAttributes.key1, 'last-wins')
+
+        t.end()
+      })
+    })
+  })
+
+  t.test('should not overwrite for same key added via addCustomSpanAttribute', (t) => {
```
carlonewrelic, line 59: minor: It wasn't clear to me that this was testing adding a transaction custom attribute with same key as custom span attribute.  A more clear description might be `should not overwrite span attribute for same key added via addCustomAttribute`.

lib/spans/span-context.js
michaelgoin: i considered but its sort of a different type of "destination".  i'm not sure. honestly, its sort of awkward and just a means of choosing a filter. there's prob a better approach separate from both of these.

test/integration/api/custom-attribute-span-propogation.tap.js
michaelgoin: fair. i was prob lazily trying to squish into one line.

test/unit/spans/span-event.test.js
```diff
@@ -63,9 +63,11 @@ tap.test('fromSegment()', (t) => {
 
       setTimeout(() => {
         const segment = agent.tracer.getTransaction().trace.root.children[0]
-        segment.addCustomSpanAttribute('Span Lee', 'no prize')
         segment.addSpanAttribute('SpiderSpan', 'web')
```
michaelgoin, line 66: fixed a merge conflict here. pretty straightforward.

carlonewrelic: I'm glad SpiderSpan lives!

PR #386, opened 2020-06-17
Parent span event attrs

nijotz:
## Proposed Release Notes
* Added the following parent attributes to span events:
  * 'parent.type'
  * 'parent.app'
  * 'parent.account'
  * 'parent.transportType'
  * 'parent.transportDuration'

### DIFF
--- a/lib/transaction/trace/index.js
+++ b/lib/transaction/trace/index.js
@@ -84,14 +84,24 @@ Trace.prototype.generateSpanEvents = function generateSpanEvents() {
     const spanAggregator = this.transaction.agent.spanEventAggregator
 
     const children = this.root.getChildren()
-    
-    // At the point where 'host.displayName' is available we only 
-    // have a root span.  Adding attribute to first non-root span here.
-    if (children.length > 0 && this.displayName) {
-      children[0].addSpanAttribute(
-        'host.displayName',
-        this.displayName
-      )
+
+    if (children.length > 0) {
+      // At the point where these attributes are available, we only have a
+      // root span. Adding attributes to first non-root span here.
+      const attributeMap = {
+        'host.displayName': this.displayName,
+        'parent.type': this.transaction.parentType,
+        'parent.app': this.transaction.parentApp,
+        'parent.account': this.transaction.parentAcct,
+        'parent.transportType': this.transaction.parentTransportType,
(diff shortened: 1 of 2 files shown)


lib/transaction/trace/index.js
```diff
@@ -84,14 +84,24 @@ Trace.prototype.generateSpanEvents = function generateSpanEvents() {
...
+      const attributeMap = {
+        'host.displayName': this.displayName,
+        'parent.type': this.transaction.parentType,
+        'parent.app': this.transaction.parentApp,
+        'parent.account': this.transaction.parentAcct,
+        'parent.transportType': this.transaction.parentTransportType,
+        'parent.transportDuration': this.transaction.parentTransportDuration
+      }
+
+      for (const [key, value] of Object.entries(attributeMap)) {
```
michaelgoin, line 100: I suppose it cleans up the code a bit here but feels a bit unnecessary to generate the entries for iterating when we already have to provide a key and provide a value via accessor for the key/value pairs. 

should only happen once per transaction, so impact is low. not going to impact PR approval.

test/unit/trace.test.js
```diff
@@ -351,6 +354,53 @@ describe('Trace', function() {
     expect(events.length).to.equal(0)
   })
 
+  it('parent.* attributes should be present on generated spans', function() {
```
michaelgoin, line 357: In the debates between unit tests and integration tests... this is leaning more on the integration test realm to me. Given the usage of DT mechanisms to trace/span impacts. But many of our unit tests blur this line. As such, won't get too picky.

test/unit/trace.test.js
```diff
@@ -351,6 +354,53 @@ describe('Trace', function() {
...
+    const trace = new Trace(transaction)
+    const child = transaction.baseSegment = trace.add('test')
+    child.start()
+    child.end()
+
+    // This should add the parent attributes onto a child span event
+    trace.generateSpanEvents()
+
+    // Test that a child span event has the attributes
+    const attrs = child.attributes.get(DESTINATIONS.SPAN_EVENT)
```
michaelgoin, line 394: Given this PR is introducing these attributes being added to spans, I'd recommend verifying it made it to span that was generated via `trace.generateSpanEvents()` instead of grabbing off the segment. While we know that destination *should* result in the right thing... we should have the Span available so might as well fully check working.

PR #388, opened 2020-06-18
Segment attributes intended for Spans should be explicitly added to Spans

carlonewrelic:
## Proposed Release Notes
Migrated the following attributes to Spans only
http.statusCode
http.statusText
error.message
error.class
error.expected

## Details
We should not be sending attributes intended for spans, and not segments, on segments.  This work is to make sure no attributes added to a segment intended for spans were properly scoped to the attribute filter SPAN_EVENT scope.

### DIFF
--- a/test/unit/error_events.test.js
+++ b/test/unit/error_events.test.js
@@ -1,49 +1,57 @@
 'use strict'
 
-// TODO: convert to normal tap style.
-// Below allows use of mocha DSL with tap runner.
-require('tap').mochaGlobals()
+const tap = require('tap')
+const test = tap.test
 
 const helper = require('../lib/agent_helper')
-const chai = require('chai')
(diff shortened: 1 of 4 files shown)


test/unit/error_events.test.js
```diff
@@ -56,24 +64,34 @@ describe('Error events', function() {
         const timestamp = 0
         const exception = new Exception({error, customAttributes, timestamp})
         tx.addException(exception)
+
+        const segment = tx.agent.tracer.getSegment()
+
+        const spanAttributes = segment.attributes.get(DESTINATIONS.SPAN_EVENT)
```
michaelgoin, line 70: ideally, we'd prob actually verify on span since we'll be ending the transaction anyways but tis miinor.

PR #390, opened 2020-06-18
TypeError: this._resetInterval.unref is not a function

jonescj:
## Proposed Release Notes
Electron application workaround for to overcome interval setting conflicts

## Links
- Initial Question: https://discuss.newrelic.com/t/typeerror-this-resetinterval-unref-is-not-a-function/104242
- Reviewed Question: https://github.com/electron/electron/issues/21162

## Details
As can be seen in the above, Electron application users commonly run into the above situation fairly often. It appears as though there is simple a timer conflict with New Relic and Electron applications. The above change allows the opportunity to pick and choose which timer is relevant based on what is available at the time.

### DIFF
--- a/lib/adaptive-sampler.js
+++ b/lib/adaptive-sampler.js
@@ -1,7 +1,7 @@
 'use strict'
 
 class AdaptiveSampler {
-  constructor(opts) {
+  constructor (opts) {
     this._serverless = opts.serverless
     this._seen = 0
     this._sampled = 0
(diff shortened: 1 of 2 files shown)


michaelgoin:
Hi  @jonescj!

Thank you for the contribution. While we don't officially support Electron, we are definitely open for moving things in that direction via public contribution.

This is an interesting problem. I am surprised by the intentional global timer conflict with incompatible API's. Do you know if this impacts `setTimeout` and `setImmediate` as well?

For resolving, it would be nice to avoid planning on throwing and to not need a try/catch guard when this is expected to just work in Node. More importantly, we'll still be subject to potentially reintroducing down the road and would be reapplying this sort of protection around all usages (base-aggregator is one missing place right now). If that expands to the other timer types, that becomes even more likely.

I'm new to the problem, so I probably need to come up to speed a bit more, but we should consider a more future-proof change. Such a change could be more involved, such as grabbing timers via a safer mechanism, instead of relying on the globals. That still has some risks though, which is why that conflict between regular Node/Electron is a bummer.

nijotz:
I changed the base of this PR from `master` -> `main`

michaelgoin:
Sorry we left this hanging for so-long. While I appreciate wanting a work-around, I'd like to see us find a more future-proof solution than swallowing errors as we are just as likely to reintroduce in other areas (and this doesn't seem to capture all anyways... although that may be from code evolving in the meantime).

Its a shame electron overwrites global timers. There are probably several electron considerations to work through if official support to were to be added. Timer-wise, we might need to a means of grabbing safe timers somehow.

Given the age of this PR and not officially supporting Electron, I'm going to close this PR out. But feel free to re-open with a different approach or open a new PR if you want to keep poking at things.

PR #393, opened 2020-06-23
Update newrelic.js

rr0214:
Adding in Distributed Tracing option and setting it to false.  We want to make this option prominent in the config file used for first time customers. See getting started documentation: https://docs.newrelic.com/docs/agents/nodejs-agent/installation-configuration/nodejs-agent-configuration

### DIFF
--- a/newrelic.js
+++ b/newrelic.js
@@ -14,6 +14,21 @@ exports.config = {
    * Your New Relic license key.
    */
   license_key: 'license key here',
+  /**
+   * Controls the method of cross agent tracing in the agent.
+   * Distributed tracing lets you see the path that a request takes through your
+   * distributed system. Enabling distributed tracing changes the behavior of some
+   * New Relic features, so carefully consult the transition guide before you enable
+   * this feature: https://docs.newrelic.com/docs/transition-guide-distributed-tracing
+   * Default is false.
+   */
+  distributed_tracing: {
+    /**
+     * Enables/disables distributed tracing.
+     *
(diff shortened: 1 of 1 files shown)


newrelic.js
```diff
@@ -14,6 +14,21 @@ exports.config = {
...
+   * this feature: https://docs.newrelic.com/docs/transition-guide-distributed-tracing
+   * Default is false.
+   */
+  distributed_tracing: {
+    /**
+     * Enables/disables distributed tracing.
+     *
+     * @env NEW_RELIC_DISTRIBUTED_TRACING_ENABLED
+     */
+    enabled: false,
```
michaelgoin, line 31:
```suggestion
    enabled: false
  },
```

newrelic.js
```diff
@@ -14,6 +14,21 @@ exports.config = {
    * Your New Relic license key.
    */
   license_key: 'license key here',
+  /**
+   * Controls the method of cross agent tracing in the agent.
```
michaelgoin, line 18: I see this took the language from the `default.js` file and moved it here. I think we probably should improve that language (and tweak the default as well but that can be done separate from this). We generally don't use the terms "cross agent tracing" AKA CAT when referring to Distributed Tracing. Seems like this could be confusing, particularly since there's other settings for CAT. I'd probably omit this line.

rishimukhop1:
Tested this change locally in addition to the automated tests, agent seems to start up without any exceptions.

PR #395, opened 2020-06-25
Tap test conversions

nijotz:
## Proposed Release Notes
* Converted some tests to Node-tap

### DIFF
--- a/test/unit/util/hashes.test.js
+++ b/test/unit/util/hashes.test.js
@@ -1,37 +1,37 @@
 'use strict'
 
-// TODO: convert to normal tap style.
-// Below allows use of mocha DSL with tap runner.
-require('tap').mochaGlobals()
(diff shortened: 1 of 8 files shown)


test/unit/util/hashes.test.js
```diff
@@ -1,37 +1,37 @@
...
-const expect = require('chai').expect
+const tap = require('tap')
 const hashes = require('../../../lib/util/hashes')
 
-describe('hashes', () => {
-  describe('#makeId', () => {
-    it('always returns the correct length', () => {
+tap.test('hashes', (t) => {
+  t.test('#makeId', (t) => {
+    t.test('always returns the correct length', (t) => {
```
carlonewrelic, line 8: minor:  maybe we don't need the nested tests here.

PR #413, opened 2020-06-29
Add source headers

michaelgoin:
## Details

Adds header to what should be all code files except the config template. It is possible some got missed but we should be covered with JS files and bash files.

Noticed while I was in here there's some extra old Travis bash files. Didn't remove with this PR since it is so beefy anyways.

Figuring this falls under the license release note, so not bothering to put something specific.

### DIFF
--- a/bin/test-naming-rules.js
+++ b/bin/test-naming-rules.js
@@ -1,4 +1,10 @@
 #! /usr/bin/env node
+
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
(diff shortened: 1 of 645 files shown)


bin/test-naming-rules.js
```diff
@@ -1,4 +1,10 @@
 #! /usr/bin/env node
```
michaelgoin, line 1: These directives have to be first in the file or cause issues, so that's why the license comes after.

PR #417, opened 2020-07-02
Cleanup

nijotz:
## Details
* Cleaned up a few comments

### DIFF
--- a/test/unit/instrumentation/http/outbound.test.js
+++ b/test/unit/instrumentation/http/outbound.test.js
@@ -340,7 +340,7 @@ tap.test('should add data from cat header to segment', (t) => {
       }
       return emit.apply(this, arguments)
     }
-    // This is really fucking gross.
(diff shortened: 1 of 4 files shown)


test/unit/instrumentation/http/outbound.test.js
```diff
@@ -340,7 +340,7 @@ tap.test('should add data from cat header to segment', (t) => {
       }
       return emit.apply(this, arguments)
     }
-    // This is really fucking gross.
+    // This is really freakin gross.
```
michaelgoin, line 343: lol. i mean you could just remove this. i don't think it provides actionable guidance :)

nijotz: lol true, one second

nijotz: removed the line

PR #450, opened 2020-07-31
On failed instrumentation, prevent multiple requires from re-wrapping shims

RyanCopley:
## Proposed Release Notes

* Adjusted instrumentation error handling to only make one attempt

## Links

https://support.newrelic.com/tickets/416443 - There is a 10 minute video in this ticket that shows the problem in our codebase. I highly suggest watching it.

## Details

While debugging a very confusing memory leak, I uncovered the problem was with our newrelic instrumentation. I made a change elsewhere in our codebase, which caused the instrumentation to fail halfway through, after already recording several shims. When the onRequire function fails, it never sets the flag to tell itself that the code has already been (or at least has already attempted to) wrapped. If you require that same module with failed instrumentation several times, you end up re-shimming the code over and over. This causes excessively high CPU% and memory usage, which quickly spiked our app past what we allocate. By marking the nodule as instrumented, this prevents the constant re-recording of shims-on-shims.  

Unfortunately, this bug bit me when I didn't realize it was related to anything New Relic, and my organization has the logging turned off. This led to a really confusing path to even discovering the root cause, as I was not aware I should be debugging New Relic instrumentation, as the docs mention. Perhaps fatal errors should log, regardless of level? 

Without this change: ~1.9 GB and a pegged CPU. With the change: ~400MB and much less CPU usage. Perhaps a better approach could be to ensure there is a console log if this occurs. 

In my head, it makes sense for it to only want to instrument once, and only once, even if it fails. In the current master branch, it will continue to attempt to instrument on proceeding requires, because it only marks it as instrumented if it was successful.

I'm not entirely sure this is the correct way to tackle this problem, so I'm hoping for this to be a starting point to explain this problem. There is probably many other cases for how error handling should be considered, which I do not have the domain knowledge for. If this is agreed upon to be a bug, I can most certainly help bring this pull request up to spec with a test, or make other adjustments as necessary.

As a New Relic customer (Gannett Inc.), I would be more than happy to hop on a screen share to show the problem in our local code base and how it impacts our application. 

Thanks,
Ryan

### DIFF
--- a/lib/shimmer.js
+++ b/lib/shimmer.js
@@ -83,6 +83,7 @@ function applyDebugState(shimmer, shim, nodule) {
     shim.enableDebug()
     instrumented.push(shim)
     instrumented.push({__NR_unwrap: function unwrapNodule() {
+      delete nodule.__NR_instrumented_errored
       delete nodule.__NR_instrumented
       delete nodule.__NR_shim
     }})
@@ -97,7 +98,7 @@ function applyDebugState(shimmer, shim, nodule) {
  */
 function instrument(agent, nodule, moduleName, resolvedName) {
   var instrumentation = shimmer.registeredInstrumentations[moduleName]
-  if (properties.hasOwn(nodule, '__NR_instrumented')) {
+  if (properties.hasOwn(nodule, '__NR_instrumented') || properties.hasOwn(nodule, '__NR_instrumented_errored')) {
     logger.trace(
       'Already instrumented %s, skipping redundant instrumentation',
       moduleName
@@ -121,6 +122,7 @@ function instrument(agent, nodule, moduleName, resolvedName) {
       nodule.__NR_instrumented = true
     }
   } catch (instrumentationError) {
+    nodule.__NR_instrumented_errored = true
     if (instrumentation.onError) {
       try {
         instrumentation.onError(instrumentationError)
--- a/lib/shimmer.js
+++ b/lib/shimmer.js
@@ -98,7 +98,8 @@ function applyDebugState(shimmer, shim, nodule) {
  */
 function instrument(agent, nodule, moduleName, resolvedName) {
   var instrumentation = shimmer.registeredInstrumentations[moduleName]
-  if (properties.hasOwn(nodule, '__NR_instrumented') || properties.hasOwn(nodule, '__NR_instrumented_errored')) {
+  if (properties.hasOwn(nodule, '__NR_instrumented')
+      || properties.hasOwn(nodule, '__NR_instrumented_errored')) {
     logger.trace(
       'Already instrumented %s, skipping redundant instrumentation',
       moduleName
--- a/lib/shimmer.js
+++ b/lib/shimmer.js
@@ -101,7 +101,7 @@ function instrument(agent, nodule, moduleName, resolvedName) {
   if (properties.hasOwn(nodule, '__NR_instrumented')
       || properties.hasOwn(nodule, '__NR_instrumented_errored')) {
     logger.trace(
-      'Already instrumented %s, skipping redundant instrumentation',
+      'Already instrumented or failed to instrument %s, skipping redundant instrumentation',
       moduleName
     )
     return nodule
--- a/test/integration/shimmer/module-load-fixture-errored.js
+++ b/test/integration/shimmer/module-load-fixture-errored.js
@@ -1,0 +1,4 @@
+'use strict'
+module.exports = () => {
+  return 'hello world'
+}
--- a/test/unit/shimmer.test.js
+++ b/test/unit/shimmer.test.js
@@ -32,10 +32,11 @@ describe('shimmer', function() {
   describe('custom instrumentation', function() {
     describe('of relative modules', makeModuleTests('../helpers/module'))
     describe('of modules', makeModuleTests('chai'))
+    describe('of modules, where instrumentation fails', makeModuleTests('chai', true))
     describe('of deep modules', makeModuleTests('chai/lib/chai'))
   })
 
-  function makeModuleTests(moduleName) {
+  function makeModuleTests(moduleName, throwsError) {
     return function moduleTests() {
       var agent = null
       var onRequireArgs = null
@@ -51,6 +52,9 @@ describe('shimmer', function() {
             instrumentedModule = module
             ++counter
             onRequireArgs = arguments
+            if (throwsError) {
+                throw new Error('This threw an error! Oh no!')
+            }
           },
           onError: function() {}
         }
@@ -92,7 +96,6 @@ describe('shimmer', function() {
         require(moduleName)
         expect(counter).to.equal(1)
       })
-
 
       it('should clean up NR added properties', () => {
         const nrKeys =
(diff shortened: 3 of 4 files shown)


RyanCopley:
Another thought I just had, it may be valuable to not try-catch the onRequire call, so that the error bubbles up and crashes the application rather than silently proceeding.

astormnewrelic:
Great to meet you @RyanCopley.  I worked for a Gannett paper a long time ago -- it's _almost_ like we're coworkers! :)

That's a very cogent explanation -- just to repeat it back, 

1. It sounds like you have an application/system that attempts to load a NodeJS module multiple times, and this module is one that the Node Agent instruments.   

2. However, the instrumentation code itself throws an exception and the module never ends up instrumented. 

3. The next time your application/system attempts to load the module, the same thing happens

4. This repeated half loading of the module instrumentation creates a situation where resource consumption is higher than normal and prevents your application/service from doing its job

Definitely a bug, and definitely something we'll want to fix. 

### Next Steps

Re: letting an exception bubble up and crash the application fast -- we try (as best we can) to avoid unhandled exceptions whenever possible, so that's not a good solution here. 

Regarding your proposted solution -- it's a great start, but I think we need to expand it a bit before landing anything -- specifically. 

1. Setting `__NR_instrumented` to true creates a false impression that the module is fully instrumented when that may/may-not be the case.  I think moving the `__NR_instrumented` code back where is was and introducing a new property `__NR_instrumented_errored` would be appropriate

2. Then add a check for `__NR_instrumented_errored` when we load the instrumentation (along side the existing `__NR_instrumented`)

3. Make sure we delete the `__NR_instrumented_errored` property in `applyDebugState`, same as we do `__NR_instrumented`

4. Make sure we get some tests to exercise this new code and the `__NR_instrumented_errored` property in `test/unit/api/api-instrument.js`

Does that make sense?  If you have any questions please let us know, we're still fleshing out our contributors guide :)

RyanCopley:
@astormnewrelic Howdy Coworker ;) 

Yeah, I anticipated this PR needed some work. I think your suggestions are very reasonable, and make a lot of sense. I appreciate your direction and will incorporate this feedback. I really appreciate your time.

I have opened a newrelic support ticket (#416443) -- Could you please view it? I am uploading a video -- that may or may not explain the problem better. I do not wish this video to be made publicly available. You do already understand the problem, but some additional context couldn't hurt :)

RyanCopley:
@astormnewrelic  -- hopefully this code is a bit more polished! Have a good weekend!

test/unit/shimmer.test.js
```diff
@@ -51,6 +52,9 @@ describe('shimmer', function() {
             instrumentedModule = module
             ++counter
             onRequireArgs = arguments
+            if (throwsError) {
+              throw new Error('This threw an error! Oh no!')
+            }
           },
```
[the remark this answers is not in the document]
RyanCopley: There is an existing test that verifies the module is only loaded once, which should behave the same if there is no error.  If you remove the implementation, this test will fail. I'm not quite sure what you mean by `but does not appear to directly test the new functionality.` -- Do you mean it's not directly testing that the variable is set? This is testing that the instrumentation only occurs once via existing tests. 

https://github.com/newrelic/node-newrelic/pull/450/files#diff-4f9be5e4eb3107fbdf0d69a962cbd634R89 -- This is the test case that, by luck, is already able to test the new functionality.

RyanCopley:
> No rush though (unless there is?)

There is no rush :) Have a great weekend

test/unit/shimmer.test.js
```diff
@@ -32,10 +32,11 @@ describe('shimmer', function() {
   describe('custom instrumentation', function() {
     describe('of relative modules', makeModuleTests('../helpers/module'))
     describe('of modules', makeModuleTests('chai'))
+    describe('of modules, where instrumentation fails', makeModuleTests('chai', true))
```
RyanCopley, line 35: This is the line that is "adding" a new test

test/integration/shimmer/module-load-fixture-errored.js
```diff
@@ -0,0 +1,4 @@
+'use strict'
+module.exports = () => {
```
RyanCopley, line 2: I created a separate module to avoid hitting the require cache. Typically, I use mockery to reset the require cache, but this is an alternate solution.

RyanCopley:
# **I have incorporated #451 and added tests for the errored variant.**

test/unit/shimmer.test.js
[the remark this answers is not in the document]
astormnewrelic: > Do you mean it's not directly testing that the variable is set?

_nod_ yup -- that's what I meant. It can be a little tedious at times, but in addition to some basic "does this thing even work" tests, we _try_ to test that any new state we're setting gets set/unset.  Since you've gone through your tests you probably see sometimes we're better at this than others ;)

lib/shimmer.js
```diff
@@ -97,7 +98,8 @@ function applyDebugState(shimmer, shim, nodule) {
  */
 function instrument(agent, nodule, moduleName, resolvedName) {
   var instrumentation = shimmer.registeredInstrumentations[moduleName]
-  if (properties.hasOwn(nodule, '__NR_instrumented')) {
+  if (properties.hasOwn(nodule, '__NR_instrumented')
+      || properties.hasOwn(nodule, '__NR_instrumented_errored')) {
```
michaelgoin, line 102: Minor: would be nice for the log message to indicate it may have skipped the errored state.

RyanCopley: This would technically log (assuming logging is enabled) in the catch statement prior, right after we set `__NR_instrumented_errored ` to true. In that case, would you still want a log statement added here?

RyanCopley: Specifically, here: https://github.com/newrelic/node-newrelic/pull/450/files#diff-a73cf9db6d0ef10539108450179052caR138

RyanCopley: Or, are you asking me to reword the log?

RyanCopley: If you want it reworded, would you want two separate logs? One for skipping already instrumented -- and one for prior instrumentation failed? -- or would you rather me just include both in the same log message?

michaelgoin: Was thinking would be nice to acknowledge here, too. For the case where someone (or us accidentally in the future) tries over and over again as another hint of what is happening without deep-diving. We'd likely track down where the message was written anyways, which is why I think it is a minor issue.

I think it would be adequate to add some additional language "Already instrumented or failed to instrument...". If wanted to be super helpful could output which sate or log different messages.

But also, I think the PR is good enough to land if you do not wish to take that on.

RyanCopley: No worries dude, I can change this real quick 👍 it's been wonderful working with you all on this and I appreciate all the feedback

astormnewrelic:
Thanks again for the PR @RyanCopley -- now that this is landed in `main` it will go out with the next agent release.  We don't have an set-in-stone release schedule for the Node Agent, but we're aiming for roughly twice a month releases.  If it's taking longer than that or that's too long don't hesitate to push the usual support buttons you already know how to push :) See you next time. 👋

PR #451, opened 2020-08-01
Adding test for __NR_instrumented

astormnewrelic:
## Proposed Release Notes

- Test added to check for pretense of `__NR_instrumented` in loaded modules

## Details

This came out of some work [over in a different PR](https://github.com/newrelic/node-newrelic/pull/450/) -- I had asked our potential contributor to add a test for their new property before I realized we had no good example of testing these properties.

This integration tests exercises the instrumentation loading code to ensure it sets the `__NR_instrumented` property.  Once this lands we should be able to add a test for the new property in #450.

### DIFF
--- a/test/integration/shimmer/module-load-fixture.js
+++ b/test/integration/shimmer/module-load-fixture.js
@@ -1,0 +1,4 @@
+'use strict'
+module.exports = () => {
+  return 'hello world'
+}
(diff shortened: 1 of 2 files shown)


astormnewrelic:
closing, this has been incorporated into https://github.com/newrelic/node-newrelic/pull/450

Issue #446, opened 2020-07-28
Obfluscation causes RangeError: Maximum call stack size exceeded error on big queries


tomashanacek:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

Obfluscation RegExp causes `RangeError: Maximum call stack size exceeded` error on big queries

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT share sensitive information, whether personal, proprietary, or otherwise! )

## Expected Behavior

[NOTE]: # ( Tell us what you expected to happen. )

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results

[NOTE]: # ( Provide any other relevant log data. )
[TIP]:  # ( Scrub logs and diagnostic information for sensitive information )

## Steps to Reproduce

I'm able to replicate the issue by running

```javascript
("'"+(new Array(8400000).fill('a')).join('')+"'").replace(/'(?:[^']|'')*?(?:\\'.*|'(?!'))/gi, '?')
```

I think that`|''` part of single quote regexp is causing the issue

https://github.com/newrelic/node-newrelic/blob/main/lib/util/sql/obfuscate.js#L10

[NOTE]: # ( Please be as specific as possible. )
[TIP]:  # ( Link a sample application that demonstrates the issue. )

## Your Environment

[TIP]:  # ( Include as many relevant details about your environment as possible. )

* ex: Browser name and version:
* ex: Node version:
* ex: Operating System and version:

## Additional context

[TIP]:  # ( Add any other context about the problem here. For example, relevant community posts or support tickets. )

astormnewrelic:
Nice to meet you @tomashanacek and thanks for the bug report -- we can confirm the regular expression

    ("'"+(new Array(8400000).fill('a')).join('')+"'").replace(/'(?:[^']|'')*?(?:\\'.*|'(?!'))/gi, '?')

does seem to destroy the call stack using stock settings. 

    $ node
    > ("'"+(new Array(8400000).fill('a')).join('')+"'").replace(/'(?:[^']|'')*?(?:\\'.*|'(?!'))/gi, '?')
    Thrown:
    RangeError: Maximum call stack size exceeded
        at String.replace (<anonymous>)

[...]

PR #456, opened 2020-08-05
Fix large JSON inserts obfuscation

tomashanacek:
Based on discussion from https://github.com/newrelic/node-newrelic/issues/446
creating a PR to fix `RangeError: Maximum call stack size exceeded error on big queries` caused by obfuscation RegEx

## Proposed Release Notes

Fixes obfuscation of SQL queries with large data inserts

## Links

https://github.com/newrelic/node-newrelic/issues/446

## Details

Fixes `Obfuscation causes RangeError: Maximum call stack size exceeded error on big queries`

### DIFF
--- a/lib/util/sql/obfuscate.js
+++ b/lib/util/sql/obfuscate.js
@@ -7,7 +7,7 @@
 
 module.exports = obfuscate
 
-var singleQuote = /'(?:[^']|'')*?(?:\\'.*|'(?!'))/
+var singleQuote = /'(?:''|[^'])*?(?:\\'.*|'(?!'))/
 var doubleQuote = /"(?:[^"]|"")*?(?:\\".*|"(?!"))/
 var dollarQuote = /(\$(?!\d)[^$]*?\$).*?(?:\1|$)/
 var oracleQuote = /q'\[.*?(?:\]'|$)|q'\{.*?(?:\}'|$)|q'\<.*?(?:\>'|$)|q'\(.*?(?:\)'|$)/
--- a/test/unit/util/obfuscate-sql.test.js
+++ b/test/unit/util/obfuscate-sql.test.js
@@ -37,5 +37,13 @@ tap.test('sql obfuscation', (t) => {
     t.end()
   })
 
+  t.test('should handle large JSON inserts', (t) => {
+    const JSONData = '{"data": "' + new Array(8400000).fill('a').join('') + '"}'
+    const result = obfuscate(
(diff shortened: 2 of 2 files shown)


lib/util/sql/obfuscate.js
```diff
@@ -7,7 +7,7 @@
 
 module.exports = obfuscate
 
-var singleQuote = /'(?:[^']|'')*?(?:\\'.*|'(?!'))/
+var singleQuote = /'(?:''|[^'])*?(?:\\'.*|'(?!'))/
 var doubleQuote = /"(?:[^"]|"")*?(?:\\".*|"(?!"))/
```
astormnewrelic, line 11: Repeating comments left in the issue -- 

I see in `'(?:''|[^'])*?(?:\\'.*|'(?!'))` you've swapped the two conditions so that (in "PCRE comment format") instead of 

```
    /
        ...
        (?:
            [^'] | ''
        )
        ...
    /x
```

we have 

```
    /
        ...
        (?:
            '' | [^'] 
        )
        ...
    /x
```

That seems reasonable and the group still appears to capture "every single character that **isn't a quote** or _is two single quote characters_". We also have additional confidence that current agent behavior is maintained because it also passes our tests.

test/unit/util/obfuscate-sql.test.js
```diff
@@ -37,5 +37,13 @@ tap.test('sql obfuscation', (t) => {
     t.end()
   })
 
+  t.test('should handle large JSON inserts', (t) => {
+    const JSONData = '{"data": "' + new Array(8400000).fill('a').join('') + '"}'
+    const result = obfuscate(
```
astormnewrelic, line 42: 👍 This is a good test. In the current `main` branch this test fails -- with your changes is passes. 

Thinking out loud -- you mentioned 

> I don't think we are actually running this big queries,

Do you think we need a query that's more representative of the queries in your own system?

tomashanacek: I did some more exploration and it is actually very similar to what we run, only difference is what we have in JSON data, but that I can't share since that content is confidential

tomashanacek:
Hi @astormnewrelic, do you have any ETA when the PR could be merged and released?

astormnewrelic:
Thanks for the contribution @tomashanacek merged. We're (informally) shooting for two releases a month, and it looks like this week is release week. If we don't have an agent release out by the end of the week don't hesitate to loop back around and ask us what's up :)

PR #458, opened 2020-08-06
Bumps @newrelic/native-metrics to ^5.3.0.

michaelgoin:
## Proposed Release Notes

* Bumped @newrelic/native-metrics to ^5.3.0.

### DIFF
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -27,6 +27,10 @@ code, the source code can be found at [https://github.com/newrelic/node-newrelic
 * [readable-stream](#readable-stream)
 * [semver](#semver)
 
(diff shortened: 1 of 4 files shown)


THIRD_PARTY_NOTICES.md
```diff
@@ -1317,6 +1321,215 @@ IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 ```
 
+## optionalDependencies
```
michaelgoin, line 1324: This is manually added for now, so we'll want to make sure doesn't get deleted by the tool run in the future. Similar to the footer.

As such, when updating licenses I recommend selectively choosing lines to stage and tossing away errant deletes.

PR #459, opened 2020-08-07
Update package.json

RamyaPayyavula:
older versions of lodash has injection vulnerability. An attacker can inject malicious code via `sourceURL` since it is not sanitized for the user-provided code that leads to the `eval()` function.

### DIFF
--- a/package.json
+++ b/package.json
@@ -174,7 +174,7 @@
     "got": "^8.0.1",
     "http-errors": "^1.7.3",
     "jsdoc": "^3.6.3",
-    "lodash": "^4.17.14",
+    "lodash": "^4.17.19",
     "memcached": ">=0.2.8",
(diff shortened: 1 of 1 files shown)


nijotz:
Just putting information here, it looks like we only use lodash here:

test/unit/high-security.test.js
test/integration/config/attribute-cat.tap.js

It's possible we could remove it as a dep as well.

michaelgoin:
yeah, long-term i think removal will be nice to get away from the non-expoitable disclosure churn. but i'm fine with this PR just bumping deps.

nijotz:
Double-checked on this, it looks like the tests are failing because the lockfile still has the old lodash version.

nijotz:
#490 removes lodash as a dependency. Closing this in favor of that PR.

PR #464, opened 2020-08-17
feat: Adding ability to write to named pipe instead of stdout.

mrickard:
On some occasions, the serverless observability pipeline may benefit from the ability to write output to a named pipe instead of stdout; this adds that, with a fallback to stdout if the pipe isn't found.

## Proposed Release Notes

This commit adds the ability for a serverless application using the Node.js agent to write to a named pipe instead of stdout, if the `/tmp/newrelic-telemetry` pipe is present, or if the agent or application process has defined a `pipePath`, and the pipe is present. 

## Details

This PR includes changes to the serverless collector (`lib/collector/serverless.js`), the serverless collector unit test (`test/unit/collector/serverless.test.js`), and the exceptions test (`test/integration/core/exceptions.tap.js`).

### DIFF
--- a/lib/collector/serverless.js
+++ b/lib/collector/serverless.js
@@ -35,6 +35,10 @@ class ServerlessCollector {
       agent_version: agent.version
     }
     this.payload = {}
+    this.pipePath = (agent.pipePath || process.env.pipePath) ?
+      (agent.pipePath || process.env.pipePath) :
+      '/tmp/newrelic-telemetry'
+    this.pipeError = false
   }
 
   /**
@@ -244,6 +248,29 @@ class ServerlessCollector {
   }
 
   /**
+   * Tests for writing to named pipe
+   */
+  shouldUsePipe() {
+    if (this.pipeError) {
+      return false
+    }
+    return fs.existsSync(this.pipePath)
+  }
+
+  /**
+   * Writes payload to pipe
+   */
+  flushToPipe(payload) {
+    try {
+      fs.writeFileSync(this.pipePath, payload)
+    } catch (e) {
+      // cache error and try again.
+      this.pipeError = e
+      this._doFlush(payload, true)
+    }
+  }
+
+  /**
    * Internal method to handle flushing to stdout.
    *
    * @private
@@ -257,6 +284,10 @@ class ServerlessCollector {
       PAYLOAD_MARKER,
       payload
     ]) + '\n'
+
+    if (this.shouldUsePipe()) {
+      return this.flushToPipe(serializedPayload)
+    }
 
     if (sync) {
       // Long log lines have been truncated at 65538
--- a/test/integration/core/exceptions.tap.js
+++ b/test/integration/core/exceptions.tap.js
@@ -10,6 +10,9 @@ const cp = require('child_process')
 const path = require('path')
 const helper = require('../../lib/agent_helper')
 const helpersDir = path.join(path.resolve(__dirname, '../../'), 'helpers')
+const fs = require('fs')
+const pipePath = '/tmp/newrelic-telemetry'
+const shouldUsePipe = fs.existsSync(pipePath)
 
 tap.test('Uncaught exceptions', (t) => {
   var proc = startProc()
@@ -83,13 +86,30 @@ tap.test('Triggers harvest while in serverless mode', (t) => {
   var message = 'I am a test error'
   var messageReceived = false
   var payload = ''
-  proc.stdout.on('data', function bufferData(data) {
-    payload += data.toString('utf8')
-  })
+  if (!shouldUsePipe) {
+    proc.stdout.on('data', function bufferData(data) {
+      payload += data.toString('utf8')
+    })
+  }
 
-  proc.on('message', function(errors) {
+  const getPayloadFromPipe = function() {
+    return fs.promises.readFile(pipePath)
+      .then(data => {
+        payload += data.toString('utf8')
+        return payload
+      })
+      .catch(err => {
+        throw err
+      }).finally(() => payload)
+  }
+
+  proc.on('message', async function(errors) {
     messageReceived = true
     t.equal(errors.count, 0, 'should have harvested the error')
+
+    if (shouldUsePipe) {
+      await getPayloadFromPipe()
+    }
 
     const lambdaPayload = findLambdaPayload(payload)
     t.ok(lambdaPayload, 'should find lambda payload log line')
--- a/test/unit/collector/serverless.test.js
+++ b/test/unit/collector/serverless.test.js
@@ -14,9 +14,13 @@ const nock = require('nock')
 const chai = require('chai')
 const expect = chai.expect
 const sinon = require('sinon')
+const fs = require('fs')
 const helper = require('../../lib/agent_helper')
 const API = require('../../../lib/collector/serverless')
 const serverfulAPI = require('../../../lib/collector/api')
+
+const pipePath = '/tmp/newrelic-telemetry'
+const shouldUsePipe = fs.existsSync(pipePath)
 
 describe('ServerlessCollector API', () => {
   let api = null
@@ -164,8 +168,13 @@ describe('ServerlessCollector API', () => {
     let logStub = null
 
     beforeEach(() => {
-      logStub = sinon.stub(process.stdout, 'write').callsFake(() => {
-      })
+      if (shouldUsePipe) {
+        logStub = sinon.stub(fs, 'writeFileSync').callsFake(() => {
+        })
+      } else {
+        logStub = sinon.stub(process.stdout, 'write').callsFake(() => {
+        })
+      }
     })
 
     afterEach(() => {
@@ -175,11 +184,19 @@ describe('ServerlessCollector API', () => {
     it('compresses full payload and writes formatted to stdout', (done) => {
       api.payload = {type: 'test payload'}
       api.flushPayload(() => {
-        const logPayload = JSON.parse(logStub.args[0][0])
+        let logPayload = null
+
+        if (shouldUsePipe) {
+          logPayload = JSON.parse(logStub.args[0][1])
+        } else {
+          logPayload = JSON.parse(logStub.args[0][0])
+        }
         expect(logPayload).to.be.an('array')
         expect(logPayload[0]).to.be.a('number')
+
         expect(logPayload[1]).to.equal('NR_LAMBDA_MONITORING')
         expect(logPayload[2]).to.be.a('string')
+
         done()
       })
     })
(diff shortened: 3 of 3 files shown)


nijotz:
Hi Maurice! Quick question, is `/tmp/newrelic-telemetry` something defined outside of the agent? Or a just a default for this?

lib/collector/serverless.js
```diff
@@ -35,6 +35,10 @@ class ServerlessCollector {
       agent_version: agent.version
     }
     this.payload = {}
+    this.pipePath = (agent.pipePath || process.env.pipePath) ?
```
carlonewrelic, line 38: Should we use a better named env variable here like NEWRELIC_PIPE_PATH?

mrickard: I'd be in favor of that.

mrickard:
@nijotz It would be defined outside of the agent.

lib/collector/serverless.js
```diff
@@ -258,6 +285,10 @@ class ServerlessCollector {
       payload
     ]) + '\n'
 
+    if (this.shouldUsePipe()) {
+      return this.flushToPipe(serializedPayload)
+    }
```
michaelgoin, line 290: On cases where the pipe fails (should only be once due to error caching) we'll serialize the payload twice.

Prob not a big deal but we could potential restructure things to avoid that. Fine if you want to leave things.

This prob isn't the best code but a quick idea of what could be done...

```js
const didFlushToPipe = this.shouldUsePipe() && this.flushToPipe(serializedPayload) // make flushToPipe return bool / not execute retry itself

if (!didFlushToPipe) {
  this.flushToStdOut(serializedPayload) // move stdout code into helper function
}
```

lib/collector/serverless.js
```diff
@@ -243,6 +247,29 @@ class ServerlessCollector {
...
+    return fs.existsSync(this.pipePath)
+  }
+
+  /**
+   * Writes payload to pipe
+   */
+  flushToPipe(payload) {
+    try {
+      fs.writeFileSync(this.pipePath, payload)
```
michaelgoin, line 265: I'm assuming we won't have any truncation concerns with this approach like we have had previously.

test/integration/core/exceptions.tap.js
```diff
@@ -10,6 +10,9 @@ const cp = require('child_process')
 const path = require('path')
 const helper = require('../../lib/agent_helper')
 const helpersDir = path.join(path.resolve(__dirname, '../../'), 'helpers')
+const fs = require('fs')
+const pipePath = '/tmp/newrelic-telemetry'
+const shouldUsePipe = fs.existsSync(pipePath)
```
michaelgoin, line 15: Introduce variable paths into the test seems dangerous and likely to increase chance of test flickers or confusion trying to debug locally VS CI, etc.

We should be able to guarantee one state or the other for the sake of the test. Perhaps we need two tests here to exercise both paths? Or do we really only need to execute one style?

test/unit/collector/serverless.test.js
```diff
@@ -14,10 +14,14 @@ const nock = require('nock')
 const chai = require('chai')
 const expect = chai.expect
 const sinon = require('sinon')
+const fs = require('fs')
 const helper = require('../../lib/agent_helper')
 const API = require('../../../lib/collector/serverless')
 const serverfulAPI = require('../../../lib/collector/api')
 
+const pipePath = '/tmp/newrelic-telemetry'
+const shouldUsePipe = fs.existsSync(pipePath)
```
michaelgoin, line 23: Same as above. Introducing variable execution into a test seems dangerous. I'd prefer we isolate the different paths, if necessary to cover both, by controlling the state for the tests.

test/unit/collector/serverless.test.js
```diff
@@ -175,11 +287,12 @@ describe('ServerlessCollector API', () => {
     it('compresses full payload and writes formatted to stdout', (done) => {
       api.payload = {type: 'test payload'}
       api.flushPayload(() => {
-        const logPayload = JSON.parse(logStub.args[0][0])
+        const logPayload = JSON.parse(logStub.args[0][1])
```
michaelgoin, line 290: I see the argument here is changing because `writeFileSync` is now being stubbed, where the payload is the second argument.

It is a bit confusing with reusing `logStub`. I'd prob rename to `writeFileSyncStub`. While it would require further renaming if we change the mechanism, I think the clarity of understanding the pieces in play would be worth it.

lib/collector/serverless.js
```diff
@@ -243,6 +247,29 @@ class ServerlessCollector {
...
+    if (this.pipeError) {
+      return false
+    }
+    return fs.existsSync(this.pipePath)
+  }
+
+  /**
+   * Writes payload to pipe
+   */
+  flushToPipe(payload) {
```
michaelgoin, line 263: Just to be extra clear in javascript land... lets rename this to `flushToPipeSync` to indicate it is always going to be sync

lib/collector/serverless.js
```diff
@@ -35,6 +35,10 @@ class ServerlessCollector {
       agent_version: agent.version
     }
     this.payload = {}
+    this.pipePath = (agent.pipePath || process.env.pipePath) ?
+      (agent.pipePath || process.env.pipePath) :
+      '/tmp/newrelic-telemetry'
```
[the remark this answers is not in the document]
mrickard: I wasn't terribly fond of that either, particularly if we rename the pipe path variable.

lib/collector/serverless.js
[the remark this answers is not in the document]
mrickard: @astormnewrelic had suggested making the path an injectable constructor property.  I'd prepped for that, but then bumped into what you mention, about the agent not having a pipePath. I'd be open to making it a constructor property...but I'm not quite sure where the value would be defined for that.

lib/collector/serverless.js
```diff
@@ -243,6 +247,29 @@ class ServerlessCollector {
...
   }
 
+  /**
+   * Tests for writing to named pipe
+   */
+  shouldUsePipe() {
+    if (this.pipeError) {
+      return false
+    }
+    return fs.existsSync(this.pipePath)
```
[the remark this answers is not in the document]
mrickard: @michaelgoin I wouldn't expect its availability or presence/absence to change during the lambda lifespan. More that it's either there at the beginning of the invocation, or it isn't. 

Are we doing one payload write per invocation? Or potentially multiple writes?

lib/collector/serverless.js
mrickard: I'd think we wouldn't, if the write's synchronous.

lib/collector/serverless.js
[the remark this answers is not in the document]
michaelgoin: ah. you could have a constructor variable (makes testing easier) that defaults to the values you have here w/o the agent part.

so...

```js
constructor(agent, pipePath) {
  // ...
  this.pipePath = pipePath || ...
```

could also do param defaulting but guessing taht would be a bit much with the multiple assigned values.

Or, you could inject these values from the agent side. In which case, the grabbing of the pipe path *could* come from the config (either file or env var), if that's something we potentially want... instead of a one-off ENV var or default only. `this.collector = new ServerlessCollector(this, this.config.lambda_pipe_path)` or some such.

This can be done in future PR's (or wait for need) if don't want to do with this. But I don't think we want the unused agent.pipePath in there for this PR.

lib/collector/serverless.js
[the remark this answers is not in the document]
michaelgoin: Right now, one payload per invocation. 1:1 transaction to invocation and then we immediately write on transaction end.

But we'll obviously do many writes per lifespan of the lambda instance... if the value won't change between invocations (I'd assume wouldn't) then I think we should cache the value.

test/integration/core/exceptions.tap.js
mrickard: Ultimately, I'm happy to call this an artifact of my local testing.  I was running into issues with the exceptions test failing when the pipe was present, but a refactoring has fixed that, so...I don't think a change to exceptions is necessary.

test/unit/collector/serverless.test.js
mrickard: @michaelgoin I think the diff is confusing here--this is within the new `ServerlessCollector with constructor-injected pipe`. 

I'm doing some branching in the original logStub, though, that I'll take out for clarity. The pipe writing is tested here, so that should be enough.

lib/collector/serverless.js
```diff
@@ -24,7 +27,7 @@ class ServerlessCollector {
    *
    * @param {Agent} agent - The agent this collector will use
    */
-  constructor(agent) {
+  constructor(agent, nrPipePath) {
```
michaelgoin, line 30: nit: new relic (nr) is implied, it is our class.

lib/collector/serverless.js
```diff
@@ -35,10 +38,8 @@ class ServerlessCollector {
       agent_version: agent.version
     }
     this.payload = {}
-    this.pipePath = (agent.pipePath || process.env.pipePath) ?
-      (agent.pipePath || process.env.pipePath) :
-      '/tmp/newrelic-telemetry'
-    this.pipeError = false
+    this.NEWRELIC_PIPE_PATH = nrPipePath || process.env.NEWRELIC_PIPE_PATH || defaultPipePath
```
michaelgoin, line 41: nit: stylistically, we typically only use SNAKE_CASE for constants / env vars.

but yeah, that's something better handled via automated tools and not PR nit picks.

similarly to prior, the "new relic" part on the member variable is somewhat redundant. (unlike the ENV var which gets set outside of the agent).

michaelgoin, line 41: +1 ENV var naming

test/unit/collector/serverless.test.js
```diff
@@ -18,9 +18,9 @@ const fs = require('fs')
 const helper = require('../../lib/agent_helper')
 const API = require('../../../lib/collector/serverless')
 const serverfulAPI = require('../../../lib/collector/api')
-
-const pipePath = '/tmp/newrelic-telemetry'
-const shouldUsePipe = fs.existsSync(pipePath)
+const path = require('path')
+const NEWRELIC_PIPE_PATH = path.resolve('/tmp', 'newrelic-telemetry')
+const shouldUsePipe = fs.existsSync(NEWRELIC_PIPE_PATH)
```
michaelgoin, line 23: Seems like more of the dynamic situation where the test will run differently depending on the environment it is running in. Would prefer for this to be deterministic. Either setup to always use a pipe/certain path or always setup to use the old school way, or one of each.

test/integration/core/exceptions.tap.js
```diff
@@ -87,7 +87,7 @@ tap.test('Triggers harvest while in serverless mode', (t) => {
     payload += data.toString('utf8')
   })
 
-  proc.on('message', function(errors) {
+  proc.on('message', async function(errors) {
```
michaelgoin, line 90: i don't see anything async/promise related in here... so the `async` keyword seems like it will create an unnecessary promise?

test/unit/collector/serverless.test.js
michaelgoin: i guess i'm trying to determine why we'd need to index [1] instead of [0]. i assumed that was because `logStub` is not actually a log stub anymore but a `writeFileSync` stub, which has different positional arguments.

test/unit/collector/serverless.test.js
```diff
@@ -175,11 +179,112 @@ describe('ServerlessCollector API', () => {
...
+      api.payload = {type: 'test payload'}
+      for (let i = 0; i < 4096; i++) {
+        api.payload[`customMetric${i}`] = Math.floor(Math.random() * 100000)
+      }
+
+      api.flushPayload(() => {
+        let logPayload = null
+
+        if (shouldUsePipe) {
+          logPayload = JSON.parse(logStub.getCall(0).args[1])
```
michaelgoin, line 205: Seems like this path might not work but tis related to the forking which I'm hoping we can get rid off.

lib/collector/serverless.js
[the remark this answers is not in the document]
mrickard: I've removed the unused agent property.  I'm thinking that `lib/agent.js` would be the right place to set that path on the collector--would that be right?

test/unit/collector/serverless.test.js
mrickard: Agreed--I can remove these from the test file, since we're doing a one-off custom path for testing the same functionality.

test/unit/collector/serverless.test.js
mrickard: I've refactored that out.

test/unit/collector/serverless.test.js
mrickard: Yes--there's a different argument position. (I've updated the names of the stub and payload vars/consts to reflect this.)

test/integration/core/exceptions.tap.js
mrickard: That was an artifact of an edit/revision, though I'd overlooked removing the async here. Thanks!

mrickard:
You're welcome! Thank you for the comments!

PR #479, opened 2020-09-17
fix: Check for named pipe before each payload write

mrickard:

The addition of writing to a named pipe included a check for the pipe from the serverless collector's constructor. In practice, this is susceptible to a race condition, so there would be times when the pipe's been created after the constructor exits. Given that the agent currently writes one payload (and any future additional payloads are likely to be limited in number), this solves the problem of the race condition with minimal effect on performance. 

Tagging @michaelgoin for visibility.

### DIFF
--- a/lib/collector/serverless.js
+++ b/lib/collector/serverless.js
@@ -39,7 +39,6 @@ class ServerlessCollector {
     }
     this.payload = {}
     this.pipePath = pipePath || process.env.NEWRELIC_PIPE_PATH || defaultPipePath
-    this.shouldUsePipe = fs.existsSync(this.pipePath)
   }
 
   /**
@@ -256,7 +255,6 @@ class ServerlessCollector {
       fs.writeFileSync(this.pipePath, payload)
       return true
     } catch (e) {
-      this.shouldUsePipe = false
       logger.warn('Error attempting to write to pipe, falling back to stdout', e)
       return false
     }
@@ -294,7 +292,7 @@ class ServerlessCollector {
       payload
     ]) + '\n'
 
-    const didUsePipe = this.shouldUsePipe && this.flushToPipeSync(serializedPayload)
+    const didUsePipe = fs.existsSync(this.pipePath) && this.flushToPipeSync(serializedPayload)
 
     if (!didUsePipe) {
       this.flushToStdOut(serializedPayload, payload.length, sync)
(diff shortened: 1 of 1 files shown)


lib/collector/serverless.js
```diff
@@ -294,7 +292,7 @@ class ServerlessCollector {
       payload
     ]) + '\n'
 
-    const didUsePipe = this.shouldUsePipe && this.flushToPipeSync(serializedPayload)
+    const didUsePipe = fs.existsSync(this.pipePath) && this.flushToPipeSync(serializedPayload)
```
michaelgoin, line 295: Could you just cache the value at this point? 

something like

```js
// cache value upon first attempted send once we know a pipe should exist 
if (this.shouldUsePipe === undefined) {
  this.shouldUsePipe = fs.existsSync(this.pipePath)
}

const didUsePipe = this.shouldUsePipe && this.flushToPipeSync(serializedPayload)
```

Could extract the memoization to a helper, if wanted.

I'm sure the impact is minimal but I still don't love doing a file read per function invocation.

Or will it sometimes not exist by this point in invocation too? If that's the case, I guess we could stop checking once true but still end up impacting all cases w/o pipe.

mrickard: I'd think that once it's in place, it's not going to be removed. But on first run after a deployment, there's a possibility for the pipe not to exist during the constructor. I'd think that doing it this way could be fine, leaving it undefined until the first flush.

mrickard: Any objections to doing something like this?
```
    if (typeof this.shouldUsePipe === 'undefined') {
      this.shouldUsePipe = fs.existsSync(this.pipePath)
    }
```

michaelgoin: guess i'm curious why we'd want to do a `typeof` operation?

mrickard: I'm not insistent on it. It just feels a bit odd to check for equality on undefined.

michaelgoin: ah 🤷  guess i'm used to it lately via Node... but what we typically would do is account for initializing to `null` via something like:

`this.shouldUsePipe == null` (will coerce and check both null and undefined) OR
`this.shouldUsePipe === null || this.shouldUsePipe === undefined` (arguably faster but this is not a hot path)

both are pretty standard patterns i see in Node (admittedly not an expert). doing a `typeof` seems unnecessary unless we actually cared about the type (especially since there are oddities such as the fact `typeof null` is 'object').

michaelgoin (approved):
👍

mrickard:
Thank you, @michaelgoin !

Issue #483, opened 2020-09-24
Better Node 14, EOL Node 8


rr0214:
### **Problem**

The maintenance cycle for Node.js 8 ended in [<ins>December 2019</ins>](https://blog.risingstack.com/update-nodejs-8-end-of-life-no-support/). This means since December 2019, Node.js 8 has not received updates, bug fixes or security patches, so it is in our customers’ best interest to upgrade their Node.js version to not be open to vulnerabilities. We there do not want to be responsible for fulfilling support for this version.

Also the newest version of Node.js, [<ins>Node.js 14</ins>](https://nodejs.org/en/about/releases/) was released in April 2020 and we want to ensure customers do not see _ **degradations** _ in their experience between Node 12 and Node 14

### **Solution**

Newer versions of the agent will no longer support Node 8 and we will fix any degradation between Node 12 and Node 14. Our documentation will be updated appropriately to reflect Node 14 support and the removal of Node 8 support

jgeurts:
What is the timeline for full node 14 support?

jgeurts:
Never mind.... I saw https://github.com/newrelic/node-newrelic/issues/329#issuecomment-692180970. I wish this project didn't feel so neglected/abandoned.

nijotz:
Hey Jim, we're starting to work on Node v14 support now. You can track progress here: https://github.com/newrelic/node-newrelic/projects/5

nijotz:
Closing in favor of #329

nijotz:
Got that backwards.. reopening

nijotz:
[v7.0.0](https://github.com/newrelic/node-newrelic/releases/tag/v7.0.0) was released which adds Node v14 support

Issue #329, opened 2020-02-20
Node 14 Support


michaelgoin:
This issue will serve as a place to add any discovered issues/incompatibilities with Node14 (once released) prior to official support for Node 14. Please keep in mind we do not officially support odd-numbered releases. 

It should help inform if there are any risks too large to take on in your own project and will also help inform our official support as we tackle that.

For our official support policy, please see: https://docs.newrelic.com/docs/agents/nodejs-agent/getting-started/compatibility-requirements-nodejs-agent#version

paul-pro:
Hello! Any activity in this direction? Maybe support date announcements?

michaelgoin:
Hi @paul-pro,

No immediate activity here. Based on current priorities (always subject to change) I'm guessing we'll add official support closer to the end of our promised window in the link above (closer to LTS start date).

If you are feeling more adventurous, we don't have the 14 engine restricted so you can test out if you wish. So far we've had one 14 specific issue reported and PR'd on the native metrics side but no comments here or PR's on the main agent side, yet.

harrykimpel:
Hi team, any updates on this?

rr0214:
Hi @harrykimpel. We are committed to supporting Node 14 and are targeting completion by end of October.  I would love to know if there are specific parts of Node 14 that you are particularly looking forward to having supported.

harrykimpel:
Hi @rr0214, just generic support for v14. Most of the applications in scope run v14 already. Thanks!

nijotz:
Starting work to support v14 now: https://github.com/newrelic/node-newrelic/projects/5

nijotz:
Closing in favor of #483

PR #487, opened 2020-09-29
Add Node.js 14 to CI jobs

jgeurts:
## Proposed Release Notes
* Add Node.js 14 support

## Links
#483
#329

## Details
* Adds Node.js 14 to CI jobs

### DIFF
--- a/lib/collector/remote-method.js
+++ b/lib/collector/remote-method.js
@@ -107,6 +107,7 @@ RemoteMethod.prototype._post = function _post(data, nrHeaders, callback) {
   }
 
   if (options.compressed) {
+    // NOTE: gzip and deflate throw immediately in Node 14+ with an invalid argument
     try {
       var useGzip = this._config.compressed_content_encoding === 'gzip'
       var compressor = useGzip ? zlib.gzip : zlib.deflate
--- a/lib/collector/remote-method.js
+++ b/lib/collector/remote-method.js
@@ -107,17 +107,22 @@ RemoteMethod.prototype._post = function _post(data, nrHeaders, callback) {
   }
 
   if (options.compressed) {
-    var useGzip = this._config.compressed_content_encoding === 'gzip'
-    var compressor = useGzip ? zlib.gzip : zlib.deflate
-    compressor(data, function onCompress(err, compressed) {
-      if (err) {
-        logger.warn(err, 'Error compressing JSON for delivery. Not sending.')
-        return callback(err)
-      }
+    try {
+      var useGzip = this._config.compressed_content_encoding === 'gzip'
+      var compressor = useGzip ? zlib.gzip : zlib.deflate
+      compressor(data, function onCompress(err, compressed) {
+        if (err) {
+          logger.warn(err, 'Error compressing JSON for delivery. Not sending.')
+          return callback(err)
+        }
 
-      options.body = compressed
-      makeRequest()
-    })
+        options.body = compressed
+        makeRequest()
+      })
+    } catch (err) {
+      logger.warn(err, 'Error compressing JSON for delivery. Not sending.')
+      return callback(err)
+    }
   } else {
     options.body = data
     makeRequest()
--- a/package.json
+++ b/package.json
@@ -142,7 +142,7 @@
     "newrelic-naming-rules": "./bin/test-naming-rules.js"
   },
   "dependencies": {
-    "@grpc/grpc-js": "^1.1.7",
+    "@grpc/grpc-js": "1.0.5",
     "@grpc/proto-loader": "^0.5.4",
     "@newrelic/aws-sdk": "^2.0.0",
     "@newrelic/koa": "^4.0.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "newrelic",
-  "version": "6.13.1",
+  "version": "6.13.2",
   "author": "New Relic Node.js agent team <nodejs@newrelic.com>",
   "license": "Apache-2.0",
   "contributors": [
@@ -142,8 +142,8 @@
     "newrelic-naming-rules": "./bin/test-naming-rules.js"
   },
   "dependencies": {
-    "@grpc/grpc-js": "1.0.5",
-    "@grpc/proto-loader": "^0.5.4",
+    "@grpc/grpc-js": "1.1.7",
+    "@grpc/proto-loader": "^0.5.5",
     "@newrelic/aws-sdk": "^2.0.0",
     "@newrelic/koa": "^4.0.0",
     "@newrelic/superagent": "^3.0.0",
@@ -174,7 +174,6 @@
     "got": "^8.0.1",
     "http-errors": "^1.7.3",
     "jsdoc": "^3.6.3",
-    "lodash": "^4.17.14",
     "memcached": ">=0.2.8",
     "minami": "^1.1.1",
     "mongodb": "^3.3.3",
--- a/package.json
+++ b/package.json
@@ -142,7 +142,7 @@
     "newrelic-naming-rules": "./bin/test-naming-rules.js"
   },
   "dependencies": {
-    "@grpc/grpc-js": "1.0.5",
+    "@grpc/grpc-js": "^1.1.7",
     "@grpc/proto-loader": "^0.5.4",
     "@newrelic/aws-sdk": "^2.0.0",
     "@newrelic/koa": "^4.0.0",
--- a/test/versioned/director/package.json
+++ b/test/versioned/director/package.json
@@ -5,7 +5,7 @@
   "tests": [
     {
       "engines": {
-        "node": ">=8"
+        "node": ">=8 <14"
       },
       "dependencies": {
         "director": ">=1.2",
--- a/test/versioned/hapi/hapi-pre-17/package.json
+++ b/test/versioned/hapi/hapi-pre-17/package.json
@@ -5,7 +5,7 @@
   "tests": [
     {
       "engines": {
-        "node": ">=8"
+        "node": ">=8 <14"
       },
       "dependencies": {
         "hapi": ">=2.0.0 <6",
@@ -20,7 +20,7 @@
     },
     {
       "engines": {
-        "node": ">=8"
+        "node": ">=8 <14"
       },
       "dependencies": {
         "hapi": ">=6.0.0 <7"
@@ -35,7 +35,7 @@
     },
     {
       "engines": {
-        "node": ">=6"
+        "node": ">=6 <14"
       },
       "dependencies": {
         "hapi": ">=7.0.0 <8.3 || >=8.8 <9"
@@ -50,7 +50,7 @@
     },
     {
       "engines": {
-        "node": ">=6"
+        "node": ">=6 <14"
       },
       "dependencies": {
         "hapi": ">=9.0 <10.2 || >=10.4 <16",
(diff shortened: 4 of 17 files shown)


jgeurts:
I would love if someone could help me with the integration test:

```
not ok 47 - test/integration/infinite-tracing-connection.tap.js # time=2971.048ms
  ---
  env: {}
  file: test/integration/infinite-tracing-connection.tap.js
  timeout: 180000
  command: /opt/hostedtoolcache/node/14.11.0/x64/bin/node
  args:
    - test/integration/infinite-tracing-connection.tap.js
  stdio:
    - 0
    - pipe
    - 2
  cwd: /home/runner/work/node-newrelic/node-newrelic
  exitCode: 1
  ...
{
    # Subtest: Inifinite tracing - Connection Handling
        # Subtest
            ok 1 - should not error
            ok 2 - should not error
            ok 3 - requested preconnect
            ok 4 - requested connect
            ok 5 - requested settings
            ok 6 - expect truthy value
            ok 7 - should be equivalent
            ok 8 - should be equal
            ok 9 - should be equal
            ok 10 - should persist request_header_map keys on metadata
            1..10
        # Subtest: should successfully send span after startup
        ok 1 - should successfully send span after startup # time=575.063ms
        
        # Subtest
            ok 1 - should not error
            ok 2 - should not error
            ok 3 - requested preconnect
            ok 4 - requested connect
            ok 5 - requested settings
            ok 6 - should be equal
            ok 7 - requested preconnect
            ok 8 - requested connect
            ok 9 - requested settings
            ok 10 - expect truthy value
            ok 11 - expect truthy value
            ok 12 - should be equivalent
            ok 13 - should be equal
            ok 14 - should not be equal
            ok 15 - should be equal
            ok 16 - should not be equal
            ok 17 - should persist new request_headers_map on metadata
            1..17
        # Subtest: should suceed, with updated run_id and request header metadata, after restart
        ok 2 - should suceed, with updated run_id and request header metadata, after restart # time=833.888ms
        
        # Subtest
            ok 1 - should not error
            ok 2 - should not error
            ok 3 - should have connected prior to initial harvest
            ok 4 - requested preconnect
            ok 5 - requested connect
            ok 6 - requested settings
            ok 7 - expect truthy value
            1..7
        # Subtest: should start immediately not wait for harvest when immediate harvest true
        ok 3 - should start immediately not wait for harvest when immediate harvest true # time=1404.28ms
        
        not ok 4 - 14 UNAVAILABLE: Connection dropped
          ---
          stack: >
            Object.callErrorFromStatus (node_modules/@grpc/grpc-js/src/call.ts:81:24)
          
            Object.onReceiveStatus (node_modules/@grpc/grpc-js/src/client.ts:666:32)
          
            Object.onReceiveStatus (node_modules/@grpc/grpc-js/src/client-interceptors.ts:397:48)
          
            node_modules/@grpc/grpc-js/src/call-stream.ts:237:24
          autoend: true
          at:
            line: 81
            column: 24
            file: node_modules/@grpc/grpc-js/src/call.ts
oracle driver not installed
            function: Object.callErrorFromStatus
          code: 14
          details: Connection dropped
          metadata:
            internalRepr:
              !!omap []
            options: {}
          tapCaught: uncaughtException
          test: Inifinite tracing - Connection Handling
          source: |2
              const message = `${status.code} ${Status[status.code]}: ${status.details}`;
              return Object.assign(new Error(message), status);
            -----------------------^
            }
          ...
        
        1..4
        # failed 1 of 4 tests
    not ok 1 - Inifinite tracing - Connection Handling # time=2965.917ms
      ---
      autoend: true
      ...
    
    not ok 2 - 14 UNAVAILABLE: Connection dropped
      ---
      at:
        line: 81
        column: 24
        file: node_modules/@grpc/grpc-js/src/call.ts
        function: Object.callErrorFromStatus
      stack: >
        Object.callErrorFromStatus (node_modules/@grpc/grpc-js/src/call.ts:81:24)
      
        Object.onReceiveStatus (node_modules/@grpc/grpc-js/src/client.ts:666:32)
      
        Object.onReceiveStatus (node_modules/@grpc/grpc-js/src/client-interceptors.ts:397:48)
      
        node_modules/@grpc/grpc-js/src/call-stream.ts:237:24
      code: 14
      details: Connection dropped
      metadata:
        internalRepr:
          !!omap []
        options: {}
      tapCaught: uncaughtException
      test: TAP
      source: |2
          const message = `${status.code} ${Status[status.code]}: ${status.details}`;
          return Object.assign(new Error(message), status);
        -----------------------^
        }
      ...
    
    1..2
    # failed 2 of 2 tests
    # time=2971.048ms
}
```

nijotz:
It looks like you're running the integration tests. These rely on some services to be running, did you run the services with `npm run services`?

nijotz:
Oh sorry, now I see that the error is happening in the CI, not locally. Let me re-run tests and see if it happened to be an intermittent networking issue.

nijotz:
Nope, still failed. But it looks like it's failing in a way similar to the PR opened by Snyk that upgrades the grpc library: https://github.com/newrelic/node-newrelic/pull/427/checks?check_run_id=856807666

They're both failing in `test/integration/infinite-tracing-connection.tap.js`, although the Snyk PR fails with `This node-oracle driver is obsolete`. But this one fails with `UNAVAILABLE: Connection dropped`.

package.json
```diff
@@ -142,7 +142,7 @@
     "newrelic-naming-rules": "./bin/test-naming-rules.js"
   },
   "dependencies": {
-    "@grpc/grpc-js": "1.0.5",
+    "@grpc/grpc-js": "^1.1.7",
```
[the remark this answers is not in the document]
jgeurts: hmm, I thought so based on https://github.com/grpc/grpc-node/releases/tag/grpc%401.24.3. But it doesn't seem like grpc-js uses that as a dependency, so it could possibly be dropped back to 1.0.5.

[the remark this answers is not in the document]
jgeurts: Same error with the version reverted back to 1.0.5, unfortunately

nijotz:
I spent some time on this hoping to figure it out. I did not. 

The exception appears to originate from [here](https://github.com/grpc/grpc-node/blob/%40grpc/grpc-js%401.0.5/packages/grpc-js/src/subchannel.ts#L497)

The `Connection Dropped` string appears to come from [here](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/call-stream.ts#L214) where the disconnect listener is setup. It bubbles up to [here] (https://github.com/grpc/grpc-node/blob/%40grpc/grpc-js%401.0.5/packages/grpc-js/src/client.ts#L670) where it checks the status and throws an error because it's `UNAVAILABLE` instead of `OK`

Hooking up the debugger, I couldn't get into that stack trace. It seems to happen after the tests are tearing down. My wild guess is that [here](https://github.com/newrelic/node-newrelic/blob/main/lib/grpc/connection.js#L203) we end() the stream which I'm guessing is an async operation?? And some objects are left over 🤷 

At first I thought it may be the server that is setup in the tests. But it looks like the stack trace has a

lib/collector/remote-method.js
```diff
@@ -107,17 +107,22 @@ RemoteMethod.prototype._post = function _post(data, nrHeaders, callback) {
...
+        if (err) {
+          logger.warn(err, 'Error compressing JSON for delivery. Not sending.')
+          return callback(err)
+        }
+
+        options.body = compressed
+        makeRequest()
+      })
+    } catch (err) {
+      logger.warn(err, 'Error compressing JSON for delivery. Not sending.')
```
nijotz, line 123: It appears that if makeRequest() throws an exception, the logs will say that the error was in compressing, but could it be that the compression worked find and it was an error in making the request?

nijotz:
Also, I see the `<14` additions to the versioned tests. Was that because there were some failures? Do those packages not support v14?

jgeurts:
> Also, I see the `<14` additions to the versioned tests. Was that because there were some failures? Do those packages not support v14?

Right, those packages had issues with node 14. All of the packages with `<14` were pretty old. I don't remember specific errors for each package anymore, unfortunately

test/versioned/director/package.json
```diff
@@ -5,7 +5,7 @@
   "tests": [
     {
       "engines": {
-        "node": ">=8"
+        "node": ">=8 <14"
```
michaelgoin, line 8: 👍 Looks like director hasn't been updated in 6 years.

test/versioned/hapi/hapi-pre-17/package.json
```diff
@@ -5,7 +5,7 @@
   "tests": [
     {
       "engines": {
-        "node": ">=8"
+        "node": ">=8 <14"
```
michaelgoin, line 8: 👍 Not surprising breakages may occur pre-17.

test/versioned/pg/pg-post-7/package.json
```diff
@@ -5,7 +5,18 @@
   "tests": [
     {
       "engines": {
-        "node": ">=8"
+        "node": ">=10"
+      },
+      "dependencies": {
+        "pg": ">=8"
```
michaelgoin, line 11: I see here' we've added a new 10+ configuration for 8+ versions of `pg`.

Given `pg-native` hasn't been published in a couple years, I'm guessing it does not yet have a `v14` compatible version.

test/versioned/pg/pg-pre-7/package.json
```diff
@@ -5,7 +5,7 @@
   "tests": [
     {
       "engines": {
-        "node": ">=8"
+        "node": ">=8 <14"
```
michaelgoin, line 8: 👍

lib/collector/remote-method.js
```diff
@@ -107,17 +107,22 @@ RemoteMethod.prototype._post = function _post(data, nrHeaders, callback) {
...
+      return callback(err)
+    }
```
michaelgoin, line 125: Why the addition of the try/catch block? 

Did we have an errant throw? Where was it originating?

michaelgoin (commented):
Thank you for kicking this off. It helps get us started on research.

I left some additional comments on the versioned test so we have record that these changes seem legit.

I did have one question about an introduced try/catch that wasn't clear what value it was adding. Only potential throwing case that seemed possible was if no callback was passed-in... which I think would have invalidated a bunch of prior assumptions (but also could be checked prior to invocation) but I may just be unaware.

As far as gRPC, I'm going to enter a separate issue for that specifically to keep track of and solve to unblock this PR and eventual Node 14 support.

lib/collector/remote-method.js
michaelgoin: I see through local experimentation that the `compressor` call is now throwing immediately with the invalid arg, instead of via the callback, when executed as a part of the remote method test "'should pass through error when compression fails."

This should work fine.

I might recommend leaving a comment as to why ('gzip and deflate now throw immediately in Node 14+ with an invalid argument' or similar).

michaelgoin:
As far as our gRPC issue... this appears to be test only. 

Based on what I've observed, it appears we make the call to disconnect (via unloading the agent) and then the test teardown calls `tryShutdown()`. While this worked fine previously, it looks like now the `tryShutdown` gets invoked too quickly and results in the unavailable 14 status code returned. Even just introducing a `setImmediate` before invoking `tryShutdown` appears to resolve the issue locally.

I'll look to get a proper resolution (and version upgrade while at it) in with issue #493.

michaelgoin:
@jgeurts if you merge in `main` at this point, things should improve a bit.

jgeurts:
> @jgeurts if you merge in `main` at this point, things should improve a bit.

🎉  Looks good!

nijotz (approved):

PR #488, opened 2020-10-01
Updated README for the repo consistency project

paperclypse:
For context:
https://docs.google.com/document/d/1JIog8rJL9MT6_AbIji8VK_RT3mOuIxC2Xb_xQsVksKk/edit#

https://newrelic.atlassian.net/browse/DOC-6121

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,8 @@ For more information on getting started, [check the Node.js docs](https://docs.n
 
 ## Usage
 
+### Using the API
+
 The `newrelic` module returns an object with the Node agent's API methods attached.
 
 ```js
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ In order to take full advantage of this package, make sure you have a [New Relic
 
 As with any instrumentation tool, please test before using in production.
 
-## Installation and getting started
+## Installation
 
 To use New Relic's Node.js agent entails these three steps, which are described in detail below:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,9 +57,11 @@ If you're compiling your JavaScript and can't control the final `require` order,
     $ node -r newrelic your-program.js
     $ node --require newrelic your-program.js
 
+## Getting Started
+
 For more information on getting started, [check the Node.js docs](https://docs.newrelic.com/docs/agents/nodejs-agent/getting-started/introduction-new-relic-nodejs).
 
-## Using the API
```
michaelgoin, line 62: This section is specific to using the agent API. As such, perhaps it should be nested under "Usage" if it is important to have that as a header but I would not remove reference to the API for the content below it.

michaelgoin (changes requested):
Mostly looks good.

One comment on the "using the api" section being specific to usage of the API and not using the agent in general. As such, we want the specific header (or something similar) whether or not you want to add an additional "Usage" section.

README.md
paperclypse: I added a "Using the API" section under the Usage header. I think your point is super valid, but I do think it's important to keep the Usage header for consistency with other repos.

michaelgoin (approved):
👍

PR #489, opened 2020-10-02
fix: Check for named pipe before each flush

mrickard:

There may be some situations in which a named pipe's presence (or absence) changes during an invocation; checking for that before flushing output would guarantee that the payload is written to the correct output. (Submitted in advance of a discussion on best approach.)

### DIFF
--- a/lib/collector/serverless.js
+++ b/lib/collector/serverless.js
@@ -292,10 +292,7 @@ class ServerlessCollector {
       payload
     ]) + '\n'
 
-    if (this.shouldUsePipe === undefined) {
-      this.shouldUsePipe = fs.existsSync(this.pipePath)
-    }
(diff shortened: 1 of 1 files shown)


nijotz:
Hi Maurice, thanks for the PR!

This seems pretty straightforward, do the check every time vs just once. Does what it says.

What about the performance of existsSync being run every flush? Do you see that being an issue at all?

Also, I'm not that familiar with named pipes changing during invocation. Would you mind linking some information about that?

nijotz (approved):
Out-of-channel conversation about this PR: perf hit shouldn't be that bad because the agent only writes one payload per invocation. This was my one concern, so I'm okay with it 👍 

Also, here's an interesting change to existsSync with regards to performance, figured I'd place here if others find it relevant: https://github.com/nodejs/node/pull/7455

PR #491, opened 2020-10-13
Remove lodash from third party files

nijotz:
## Details
* Updates the third party notice files after removing lodash as a dependency.


carlo-808 (approved):
👍

Issue #493, opened 2020-10-15
Update grpc-js, Node 14 compatibility


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

When upgrading CI to run against Node 14, some gRPC/infinite tracing issues were discovered.

Prior, automated PR's to update gRPC discovered test breakages with the upgrades.

We should update grpc-js (and relevant deps) to be up to date with recent fixes and to a version compatible with Node 14 (if exists). If a compatible version does not exist, we'll have to disable infinite tracing for Node 14 and determine a timeline for functionality. I see grpc has a version that added Node 14 support. The grpc-js package likely did not need changes to support 14 being JS.

* Node 14 CI PR: https://github.com/newrelic/node-newrelic/pull/487
* @grpc/grpc-js upgrade PR: https://github.com/newrelic/node-newrelic/pull/427
* @grpc/proto-loader upgrade PR: https://github.com/newrelic/node-newrelic/pull/452

michaelgoin:
Based on what I've observed, it appears we make the call to disconnect (via unloading the agent) and then the test teardown calls tryShutdown(). While this worked fine previously, it looks like now the tryShutdown gets invoked too quickly and results in the unavailable 14 status code returned. Even just introducing a setImmediate before invoking tryShutdown appears to resolve the issue locally.

michaelgoin:
For consideration: on disconnect we remove all listeners and then null out the stream. While this does clean things up, we no-longer have any handling on the stream that isn't immediately cleaned up on the gRPC end. As such, the error is actually thrown for this disconnected case.

PR #495, opened 2020-10-19
Fix gRPC disconnect error

michaelgoin:
## Proposed Release Notes

* Fixed issue where gRPC connection used for infinite tracing could throw if the server shutdown during disconnect of an existing connection.

* Bumped @grpc/grpc-js to 1.1.7.

* Bumped @grpc/proto-loader to ^0.5.5.

## Links

* Fixes: https://github.com/newrelic/node-newrelic/issues/493
* Discovered via: https://github.com/newrelic/node-newrelic/pull/487

## Details

Node 14 testing showed an increased likelihood of an existing race condition in our gRPC/infinite tracing tests. In some cases, the server would be shut down while the `stream.end()` had not yet responded. In this scenario, a disconnected/error status code would be returned. Since there was no error handler (we removed in _disconnect), this error would throw and crash/break the test.

Instead of restructuring all the things, this PR takes a point-fix approach and applies special handlers just for the end case. Once a status is received, these handlers will be cleaned up.

**NOTE:** I'm not sure if it is possible to have no status reported on a `end()` call. I don't *think* so but if there is a chance, that could result in leaking memory (the handlers & stream capture). The combination of things seem like such an edge case I didn't want to bog down but if we encounter this situation we can set a timeout or such to cleanup. 

This PR also bumps our library versions to keep us from falling too far behind, get bug fixes, etc.

While I would normally add a test, getting the timing right gets pretty complicated outside of what we are already encountering in our infinite tracing connection test setup so I opted not to hold this up for that. Should we encounter needing further tweaks, then I think that is likely the right time to take that on.

### DIFF
--- a/lib/grpc/connection.js
+++ b/lib/grpc/connection.js
@@ -198,9 +198,26 @@ class GrpcConnection extends EventEmitter {
     if (this.stream) {
       this.stream.removeAllListeners()
 
+      const oldStream = this.stream
+      this.stream.on('status', (grpcStatus) => {
+        logger.trace('End stream status received [%s]: %s', grpcStatus.code, grpcStatus.details)
+
+        // Cleanup the final end stream listeners.
+        oldStream.removeAllListeners()
+      })
+
+      // Listen to any final errors to prevent throwing.
+      // This is unlikely but if the server closes post
+      // removing listeners and prior to response it could
+      // happen. We noticed this via tests on Node 14.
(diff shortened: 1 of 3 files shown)


carlo-808 (approved):
I see that we added a listener for 'error' and 'status' events on the grpc connection `_disconnect` method to handle a case where the server would be shut down while the `stream.end()` had not yet responded.  👍

PR #499, opened 2020-10-21
Infinite tracing logging modifications

michaelgoin:
## Proposed Release Notes

* Increased logging level of certain infinite tracing / gRPC errors.

* Decreased logging interval of dropped span warning for infinite tracing.

### DIFF
--- a/lib/grpc/connection.js
+++ b/lib/grpc/connection.js
@@ -121,7 +121,7 @@ class GrpcConnection extends EventEmitter {
       // immediately.
       this._setState(connectionStates.connected, this.stream)
     } catch (err) {
-      logger.trace(
+      logger.warn(
         err,
         'Unexpected error establishing gRPC stream, will not attempt reconnect.'
       )
@@ -255,6 +255,9 @@ class GrpcConnection extends EventEmitter {
         // per the spec, An UNIMPLEMENTED status code from gRPC indicates
         // that the versioned Trace Observer is no longer available. Agents
         // MUST NOT attempt to reconnect in this case
+        logger.info(
+          '[UNIMPLEMENTED]: Trace Obserserver is no longer available. Shutting down connection.'
+        )
         this._disconnect()
       } else if (grpc.status[grpc.status.OK] === grpcStatusName) {
         this._reconnect()
@@ -273,7 +276,11 @@ class GrpcConnection extends EventEmitter {
       this._metrics.getOrCreateMetric(NAMES.INFINITE_TRACING.SPAN_RESPONSE_ERROR)
         .incrementCallCount()
 
-      logger.trace('span stream error. Code: [%s]: %s', err.code, err.details)
+      // For errors, the status will either result in a disconnect or a reconnect
+      // delay that should prevent too frequent spamming. Unless the app is idle
+      // and regularly getting Status 13 reconnects from the server, in which case
+      // this will be almost the only logging.
+      logger.warn('Span stream error. Code: [%s]: %s', err.code, err.details)
     })
   }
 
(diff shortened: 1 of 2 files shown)


lib/grpc/connection.js
```diff
@@ -121,7 +121,7 @@ class GrpcConnection extends EventEmitter {
       // immediately.
       this._setState(connectionStates.connected, this.stream)
     } catch (err) {
-      logger.trace(
+      logger.warn(
         err,
         'Unexpected error establishing gRPC stream, will not attempt reconnect.'
       )
```
michaelgoin, line 127: Since there will be no reconnect... there's no reason to be stingy with the log level. This indicates a fundamental issue with the infinite tracing setup.

lib/grpc/connection.js
```diff
@@ -255,6 +255,9 @@ class GrpcConnection extends EventEmitter {
         // per the spec, An UNIMPLEMENTED status code from gRPC indicates
         // that the versioned Trace Observer is no longer available. Agents
         // MUST NOT attempt to reconnect in this case
+        logger.info(
+          '[UNIMPLEMENTED]: Trace Obserserver is no longer available. Shutting down connection.'
+        )
```
michaelgoin, line 260: This could be inferred before but being explicit may help speed up troubleshooting.

lib/grpc/connection.js
```diff
@@ -273,7 +276,11 @@ class GrpcConnection extends EventEmitter {
       this._metrics.getOrCreateMetric(NAMES.INFINITE_TRACING.SPAN_RESPONSE_ERROR)
         .incrementCallCount()
 
-      logger.trace('span stream error. Code: [%s]: %s', err.code, err.details)
+      // For errors, the status will either result in a disconnect or a reconnect
+      // delay that should prevent too frequent spamming. Unless the app is idle
+      // and regularly getting Status 13 reconnects from the server, in which case
+      // this will be almost the only logging.
+      logger.warn('Span stream error. Code: [%s]: %s', err.code, err.details)
```
michaelgoin, line 283: This is the only increased risk part of the PR. This will help with troubleshooting as these issues are not always reproducible after the fact. We have a delay before retrying, so it won't be super heavy spamming of logs but sometimes customers still don't like that. 

Also left a comment about what I learned recently with regards to idle-applications and the server sending Code 13 errors.

lib/spans/span-streamer.js
```diff
@@ -54,14 +54,9 @@ class SpanStreamer {
...
-        'SPAN_DROP_MSG',
-        // interval in ms
-        SPAN_DROP_INTERVAL * 1000,
-        // message
-        SPAN_DROP_MSG + ' Will not warn again for %s seconds.',
-        // variables to put in log message
-        SPAN_DROP_INTERVAL
+        'SPAN_DROP_MSG', // key for the OncePer
+        SPAN_DROP_MSG_INTERVAL_MS,
+        SPAN_DROP_MSG
```
michaelgoin, line 59: Some cleanup to reduce recalculating certain pieces each time.

lib/spans/span-streamer.js
```diff
@@ -8,10 +8,10 @@
 const logger = require('../logger').child({component: 'span-streamer'})
 const NAMES = require('../metrics/names').INFINITE_TRACING
 
+const SPAN_DROP_MSG_INTERVAL_MS = 30000
```
michaelgoin, line 11: In practice, 5 seconds is more than we need. For a situation where the queue struggles to drain, this gets real noisy and the increased interval doesn't add much value given the other metrics we can check.

lib/grpc/connection.js
nijotz: Seems pretty minor still, nice knowledge share in the comment 👍

lib/spans/span-streamer.js
nijotz: I like the cleaner comments and the message not being appended to

nijotz (approved):

carlo-808 (approved):
👍

Issue #500, opened 2020-10-23
Status code ignoring doesn't work for transaction errors if collected too early


michaelgoin:
## Description

If the `transaction.statusCode` property has not yet been set prior to collecting an error the error will not be ignored even if the transaction's ending status code is in the `ignored_status_codes` configuration.

## Expected Behavior

Transaction errors should not be collected when the transaction ends with a status code in the ignored list.

## Additional context

This was introduced in the 6.x line of agent versions. 5.13.1 does not show this behavior. I believe it was with the batch of changes to add error context to spans.

Looking at the differences lead me to believe it is an order of operations problem. We used to filter at collect. Looking at how the `statusCode` property gets set on the transaction, this is often not until the name has been finalized. Name finalization commonly occurs when a transaction is ending.

The current approach aims not to collect to transactions unnecessarily (if I remember right to avoid adding the span context incorrectly) which is nice for not having extra data but we still need to check the ignored codes at the end.

Adding an additional check prior to processing transaction errors on transaction end should resolve the error collection piece of this. However, we might have already added context to spans that should not be there? Or maybe it is OK to have the attributes w/ an error? We'll want to make sure we are handling this case appropriately.

https://newrelic.slack.com/archives/C0511B24H/p1603445856212300

michaelgoin:
While we don't have the appropriate span filtering, it actually looks like we do a transaction error status code check... so will need to debug what is going wrong there.

michaelgoin:
In `onTransactionFinished`, we use the ignored to know if we should collect a status code error but do not use that/the ignored state to limit proper errors that were collected.

So we need to solve the following:
* Do not collect any errors on transaction finished if the transaction has an ignored status code.
* Do not keep span error attributes if the transaction ends with an ignored status code (the harder problem).

PR #503, opened 2020-10-26
Fix ignored status codes

michaelgoin:
## Proposed Release Notes

* Fixed bug where errors would still be collected for transactions with ignored error status codes in certain situations.

* Converted errors ignore unit tests to tap API.

## Links

* Fixes: https://github.com/newrelic/node-newrelic/issues/500

## Details

Errors should now be properly ignored regardless of when we notice the status code for the transaction.

In order to avoid capturing error attributes on span events for errors that should be ignored due to the later noticed status code, span error attributes are now created when the span is created. Error details for these attributes are added to the span context at error-time (last wins) and then used when creating a span from the segment.

The error code really could use some TLC to further simplify and de-duplicate. It has many moving pieces. Span events could use some additional love wrt overlap with streaming spans too.

### DIFF
--- a/lib/errors/error-collector.js
+++ b/lib/errors/error-collector.js
@@ -142,13 +142,18 @@ class ErrorCollector {
     }
 
     const isErroredTransaction = urltils.isError(this.config, transaction.statusCode)
+    const isIgnoredErrorStatusCode = urltils.isIgnoredError(
+      this.config,
+      transaction.statusCode
+    )
+
     const isExpectedErrorStatusCode = urltils.isExpectedError(
       this.config,
       transaction.statusCode
     )
 
     // collect other exceptions only if status code is not ignored
-    if (transaction.exceptions.length) {
+    if (transaction.exceptions.length && !isIgnoredErrorStatusCode) {
       for (let i = 0; i < transaction.exceptions.length; i++) {
         const exception = transaction.exceptions[i]
         if (this.collect(transaction, exception)) {
--- a/lib/errors/index.js
+++ b/lib/errors/index.js
@@ -20,7 +20,20 @@ class Exception {
   }
 
   getErrorDetails(config) {
-    return errorHelper.extractErrorInformation(null, this.error, config)
+    const errorDetails = errorHelper.extractErrorInformation(null, this.error, config)
+    errorDetails.expected = this.isExpected(config, errorDetails)
+
+    return errorDetails
+  }
+
+  isExpected(config, {type, message}) {
+    if (!this._expected) {
+      this._expected =
+        errorHelper.isExpectedErrorClass(config, type) ||
+        errorHelper.isExpectedErrorMessage(config, type, message)
+    }
+
+    return this._expected
   }
 }
 
--- a/lib/spans/span-context.js
+++ b/lib/spans/span-context.js
@@ -21,6 +21,9 @@ class SpanContext {
       customAttributes || new PrioritizedAttributes(ATTRIBUTE_SCOPE, MAXIMUM_CUSTOM_ATTRIBUTES)
 
     this.ATTRIBUTE_PRIORITY = ATTRIBUTE_PRIORITY
+
+    this.hasError = false
+    this.errorDetails = null
   }
 
   addIntrinsicAttribute(key, value) {
@@ -35,6 +38,21 @@ class SpanContext {
       false,
       priority
     )
+  }
+
+  /**
+   * Set error details to be potentially be used to create span
+   * attributes. Attributes will be created unless the transaction
+   * ends with an ignored error status code.
+   *
+   * Last error wins.
+   */
+  setError(details) {
+    this.hasError = true
+
+    // Error details will be used to create attributes unless the transaction ends
+    // with an ignored status code.
+    this.errorDetails = details
   }
 }
 
--- a/lib/spans/span-event.js
+++ b/lib/spans/span-event.js
@@ -84,9 +84,21 @@ class SpanEvent {
    * @return {SpanEvent} The constructed event.
    */
   static fromSegment(segment, parentId = null, isRoot = false) {
+    const spanContext = segment.getSpanContext()
+
+    // Since segments already hold span agent attributes and we want to leverage
+    // filtering, we add to the segment attributes prior to processing.
+    if (spanContext.hasError && !segment.transaction.hasIgnoredErrorStatusCode()) {
+      const details = spanContext.errorDetails
+      segment.addSpanAttribute('error.message', details.message)
+      segment.addSpanAttribute('error.class', details.type)
+      if (details.expected) {
+        segment.addSpanAttribute('error.expected', details.expected)
+      }
+    }
+
     const attributes = segment.attributes.get(DESTINATIONS.SPAN_EVENT)
 
-    const spanContext = segment.getSpanContext()
     const customAttributes = spanContext.customAttributes.get(DESTINATIONS.SPAN_EVENT)
 
     let span = null
@@ -97,7 +109,6 @@ class SpanEvent {
     } else {
       span = new SpanEvent(attributes, customAttributes)
     }
-
 
     for (const [key, value] of Object.entries(spanContext.intrinsicAttributes)) {
       span.intrinsics[key] = value
--- a/lib/spans/streaming-span-event.js
+++ b/lib/spans/streaming-span-event.js
@@ -102,9 +102,21 @@ class StreamingSpanEvent {
   }
 
   static fromSegment(segment, parentId = null, isRoot = false) {
+    const spanContext = segment.getSpanContext()
+
+    // Since segments already hold span agent attributes and we want to leverage
+    // filtering, we add to the segment attributes prior to processing.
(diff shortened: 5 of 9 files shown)


lib/errors/error-collector.js
```diff
@@ -142,13 +142,18 @@ class ErrorCollector {
...
+    )
+
     const isExpectedErrorStatusCode = urltils.isExpectedError(
       this.config,
       transaction.statusCode
     )
 
     // collect other exceptions only if status code is not ignored
-    if (transaction.exceptions.length) {
+    if (transaction.exceptions.length && !isIgnoredErrorStatusCode) {
```
michaelgoin, line 156: This was the main part of the bug. Basically restored some code prior to the later refactors.

lib/spans/span-event.js
```diff
@@ -84,9 +84,21 @@ class SpanEvent {
...
+    // filtering, we add to the segment attributes prior to processing.
+    if (spanContext.hasError && !segment.transaction.hasIgnoredErrorStatusCode()) {
+      const details = spanContext.errorDetails
+      segment.addSpanAttribute('error.message', details.message)
+      segment.addSpanAttribute('error.class', details.type)
+      if (details.expected) {
+        segment.addSpanAttribute('error.expected', details.expected)
+      }
+    }
+
```
michaelgoin, line 99: Don't particularly love this... but since we already are attaching span agent attributes to segments it avoids introducing a bunch of extra stuff for now.

lib/transaction/index.js
```diff
@@ -747,19 +746,13 @@ function _linkExceptionToSegment(exception) {
...
-  const details = exception.getErrorDetails(config)
-  segment.addSpanAttribute('error.message', details.message)
-  segment.addSpanAttribute('error.class', details.type)
-
-  const isExpected =
-    errorHelper.isExpectedErrorClass(config, details.type) ||
-    errorHelper.isExpectedErrorMessage(config, details.type, details.message)
-
-  if (isExpected) {
-    segment.addSpanAttribute('error.expected', isExpected)
```
michaelgoin, line 762: This work got pushed to the span event creation. Now we are just setting the details on the span context. If the transaction has an errored status code, these details will not be used when the span is created.

test/unit/error_events.test.js
```diff
@@ -70,13 +69,6 @@ test('Error events', (t) => {
         const exception = new Exception({error, customAttributes, timestamp})
         tx.addException(exception)
 
-        const segment = tx.agent.tracer.getSegment()
-
-        const spanAttributes = segment.attributes.get(DESTINATIONS.SPAN_EVENT)
-
-        t.equal(spanAttributes['error.class'], 'Error')
-        t.equal(spanAttributes['error.message'], 'some error')
-
```
michaelgoin, line 79: These made no sense to test here and would have required some serious replumbing of the test.

test/unit/error_events.test.js
```diff
@@ -96,6 +88,33 @@ test('Error events', (t) => {
       })
     })
 
+    t.test('should include spanId agent attribute', (t) => {
```
michaelgoin, line 91: But while I was in here... figured I'd add a check for the spanId attribute.

test/unit/errors/ignore.test.js
```diff
@@ -5,223 +5,270 @@
...
     })
+  })
 
-    it('Ignore status code should result in 0 errors reported', function() {
-      helper.runInTransaction(agent, function(tx) {
-        const errorAggr = agent.errors
-        agent.config.error_collector.capture_events = true
-        agent.config.error_collector.ignore_status_codes = [500]
-        tx.statusCode = 500
+  t.test('Ignore status code should ignore when status set after collecting errors', (t) => {
```
michaelgoin, line 201: The unified diff makes it hard to see... but this is the new test. It will be easier to see by looking at the commit diffs instead of the full one. I separated out the tap conversion into its own commit to keep cleaner.

nijotz:
Looking into this code, and the history of it, figured I'd drop the (NR internal) link here if anyone else is curious:
https://source.datanerd.us/NodeJS-agent/nodejs_agent/pull/1782/files

nijotz:
Looks like that PR (1782 internal) moved the logic for ignoring status codes to the collect method: https://github.com/newrelic/node-newrelic/blob/defab4a7d69e5ddd3199208f1a37de4d8ba8af13/lib/errors/aggregator.js#L234

nijotz:
Another (NR internal) PR moved that ignoring logic from `collect()` to `add()`:
https://source.datanerd.us/NodeJS-agent/nodejs_agent/pull/1966/

A fun quote from that PR review:
>I may be missing something through the following changes but it looks like this change will break user errors for the ignored/expected cases.

nijotz:
Relevant line: https://github.com/newrelic/node-newrelic/blob/bfb2e70e59ec4660691e7d17716271058db0781b/lib/errors/error-collector.js#L196

lib/errors/error-collector.js
nijotz: Alright, I think I've grokked it. It looks like the `shouldIgnoreError()` function (called when an error is added in `add()`) looks at the transaction's `statusCode` to test whether it should be ignored. But by the time the transaction has ended the `statusCode` may have changed. When `onTransactionEnd()` is called, the `statusCode` will be set to its final value. It calls `collect()`, which use to do ignore checks, but due to refactors was moved out. So, this will stop the collection of all add exceptions if the status code should be ignored.

lib/errors/error-collector.js
```diff
@@ -142,13 +142,18 @@ class ErrorCollector {
     }
 
     const isErroredTransaction = urltils.isError(this.config, transaction.statusCode)
+    const isIgnoredErrorStatusCode = urltils.isIgnoredError(
```
nijotz, line 145: There's `urltils.isIgnoredError()` and `helpers.shouldIgnoreError()`. Should these be consolidated? Not saying that should be part of this PR, just came across that when reviewing this section of code.

michaelgoin: yeah, the error helpers got introduced along the way but the code wasn't really consolidated in meaningful ways.

this all could use a really thoughtful redesign but certainly some of that could be simplified.

lib/transaction/index.js
nijotz: Did you mean ignored here instead of errored?

michaelgoin: yeah... i meant "ignored error status code"

lib/spans/span-event.js
nijotz: Ah, ya, looks like we're adding span attributes to the segment right before getting the attributes off the segment for creating a span.

lib/errors/index.js
```diff
@@ -20,7 +20,20 @@ class Exception {
...
+  }
+
+  isExpected(config, {type, message}) {
+    if (!this._expected) {
+      this._expected =
+        errorHelper.isExpectedErrorClass(config, type) ||
+        errorHelper.isExpectedErrorMessage(config, type, message)
+    }
+
+    return this._expected
```
nijotz, line 36: Looks like this was moved from `_linkExceptionToSegment()` which appears to be the only function that calls `getErrorDetails()`. Nice refactor/consolidation.

lib/spans/span-context.js
```diff
@@ -36,6 +39,21 @@ class SpanContext {
       priority
     )
   }
+
+  /**
+   * Set error details to be potentially be used to create span
+   * attributes. Attributes will be created unless the transaction
+   * ends with an ignored error status code.
+   *
+   * Last error wins.
```
nijotz, line 48: There can only be one.

lib/spans/streaming-span-event.js
```diff
@@ -102,9 +102,21 @@ class StreamingSpanEvent {
...
+    // filtering, we add to the segment attributes prior to processing.
+    if (spanContext.hasError && !segment.transaction.hasIgnoredErrorStatusCode()) {
+      const details = spanContext.errorDetails
+      segment.addSpanAttribute('error.message', details.message)
+      segment.addSpanAttribute('error.class', details.type)
+      if (details.expected) {
+        segment.addSpanAttribute('error.expected', details.expected)
+      }
+    }
+
```
nijotz, line 117: hmm, just realized how similar these `fromSegment()` functions are. Why not abstract?

test/integration/distributed-tracing/span-error-attributes.tap.js
```diff
@@ -198,6 +198,56 @@ tap.test('span error attributes', (t) => {
     t.end()
   })
 
+  t.test("Span error attributes aren't added with ignored status errors", (t) => {
```
nijotz, line 201: Lovely

test/unit/error_events.test.js
nijotz: 👍

test/unit/error_events.test.js
nijotz: so extra! :D

nijotz (approved):

carlo-808 (approved):

PR #504, opened 2020-10-27
Adds additional metrics/logging to better enable infinite tracing troubleshooting

michaelgoin:
## Proposed Release Notes

* Added additional support metrics and logging for infinite tracing.

## Details

Queue size and queue capacity will be sent up once a minute based on the metric harvest interval. We may find sending up the size more frequently will be useful but I figured this will be good enough to see if the queue is staying consistently full/close to full at first.

Dropped will give us a more specific actually dropped VS just stuck in the queue count.


lib/spans/span-streamer.js
```diff
@@ -47,11 +47,15 @@ class SpanStreamer {
 
     // If not writeable (because of backpressure) queue the span
     if (!this._writable) {
-      if (this.spans.length <= this.queue_size) {
+      if (this.spans.length < this.queue_size) {
```
michaelgoin, line 50: Off by 1...

carlo-808 (approved):
👍 
I see we have added and will populate metrics for the streaming span event aggregator aroudn queue size, queue capacity and dropped spans.  There's a off by 1 bug fix and added logging.

Issue #505, opened 2020-10-27
Remove setIgnoreTransaction


nijotz:
Remove API.prototype.setIgnoreTransaction which has been deprecated.

This allows us to NOT collect information for ignored transactions, improving performance overhead. This is captured in #506 

https://github.com/newrelic/node-newrelic/blob/main/api.js#L433-L439

PR #527, opened 2020-11-03
Remove setIgnoreTransaction API method

nijotz:
## Proposed Release Notes
* Removed deprecated setIgnoreTransaction API method

## Links
#505


carlo-808 (approved):
👍  I see setIgnoreTransaction API method is removed.  Nice job converting some tests to use the addIgnoringRule API method.

Issue #508, opened 2020-10-27
Remove deprecated HTTP response attributes


nijotz:
`httpResponseCode`, `response.status`, `httpResponseMessage` have been deprecated. Remove them.
* https://github.com/newrelic/node-newrelic/blob/main/lib/instrumentation/core/http.js#L175-L184
* https://github.com/newrelic/node-newrelic/blob/main/lib/instrumentation/core/http.js#L195-L203
* https://github.com/newrelic/node-newrelic/blob/main/lib/instrumentation/core/http.js#L216-L224
* https://github.com/newrelic/node-newrelic/blob/main/lib/serverless/aws-lambda.js#L392-L400
* https://github.com/newrelic/node-newrelic/blob/main/lib/serverless/aws-lambda.js#L403-L410

PR #528, opened 2020-11-03
Remove deprecated http response attributes

carlo-808:
## Proposed Release Notes
* Removed deprecated httpResponseCode, response.status and httpResponseMessage http response attributes.

## Links

Closes: #508

### DIFF
--- a/test/lib/fixtures.js
+++ b/test/lib/fixtures.js
@@ -10,8 +10,8 @@ module.exports = {
   httpAttributes: [
     'request.headers.host',
     'request.method',
-    'response.status',
-    'httpResponseCode'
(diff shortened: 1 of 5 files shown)


test/lib/fixtures.js
```diff
@@ -10,8 +10,8 @@ module.exports = {
   httpAttributes: [
     'request.headers.host',
     'request.method',
-    'response.status',
-    'httpResponseCode'
+    'http.statusCode',
```
nijotz, line 13: oh, interesting that the new attributes weren't added to the tests. Good catch

nijotz (approved):

Issue #511, opened 2020-10-27
Clean up version checks in HTTP instrumentation


nijotz:
Some of the HTTP instrumentation tries to be compatible with function signatures in version of Node prior to v10. Since we no longer support those version, this code can be cleaned up:
* [semvar checks that will always be true](https://github.com/newrelic/node-newrelic/blob/main/lib/instrumentation/core/http.js#L30-L31)
* [checking if URL is a global](https://github.com/newrelic/node-newrelic/blob/main/lib/instrumentation/core/http.js#L446)
* https://github.com/newrelic/node-newrelic/blob/main/lib/instrumentation/core/http.js#L561
* https://github.com/newrelic/node-newrelic/blob/main/lib/instrumentation/core/http.js#L450

PR #529, opened 2020-11-04
Remove node <=9.x.x version checks on http instrumentation

carlo-808:
## Proposed Release Notes
* **BREAKING** strictly uses `global.URL` class in http core instrumentation
* **BREAKING** removes Nodejs 8.x.x - 9.x.x checks
## Links
Closes: #511

### DIFF
--- a/lib/instrumentation/core/http.js
+++ b/lib/instrumentation/core/http.js
@@ -27,8 +27,6 @@ const DESTS = require('../../config/attribute-filter').DESTINATIONS
  * CONSTANTS
  *
  */
-const SHOULD_WRAP_HTTPS = psemver.satisfies('>=9.0.0 || 8.9.0')
-const SHOULD_FORMAT_ARGS = psemver.satisfies('>=10.0.0')
 const NR_CONNECTION_PROP = '__NR__connection'
 const NEWRELIC_ID_HEADER = 'x-newrelic-id'
 const NEWRELIC_APP_DATA_HEADER = 'x-newrelic-app-data'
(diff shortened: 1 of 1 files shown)


lib/instrumentation/core/http.js
```diff
@@ -413,37 +410,34 @@ function urlToOptions(_url) {
 }
 
 function wrapRequest(agent, request) {
-  // URL as a global was not added until Node v10
-  const URLClass = global.URL ? global.URL : url.URL
+  const URLClass = global.URL
+
   return function wrappedRequest(input, options, cb) {
     let reqArgs = arguments
```
nijotz, line 416: I think reqArgs was defined here and then in the if block, redefined. Now that the if block is gone, I think reqArgs can just be defined below on 438

lib/instrumentation/core/http.js
```diff
@@ -413,37 +410,34 @@ function urlToOptions(_url) {
...
-      } else {
-        options = Object.assign(input || {}, options)
-      }
-
-      reqArgs = [options, cb]
+    // If the first argument is a URL, merge it into the options object.
+    // This code is copied from Node internals.
+    if (typeof input === 'string') {
+      const urlStr = input
+      input = urlToOptions(new URLClass(urlStr))
```
nijotz, line 422: Seems like this could be simplified to just `new URL(urlStr)` here and the URLClass definition removed.

lib/instrumentation/core/http.js
carlo-808: Good catch!  I'll update.

nijotz (approved):
👍

Issue #510, opened 2020-10-27
Drop deprecated Distributed Tracing methods


nijotz:
* https://github.com/newrelic/node-newrelic/blob/main/lib/transaction/handle.js#L101-L114
* https://github.com/newrelic/node-newrelic/blob/main/lib/transaction/index.js#L1091-L1094

PR #537, opened 2020-11-06
Remove dt methods

nijotz:
## Proposed Release Notes
* **BREAKING** Removed deprecated Distributed Tracing API methods

## Links
#510 

## Details
`createDistributedTracePayload()` and `acceptDistributedTracePayload()` were on the `TransactionHandle` class and intended to be used by end users. Methods by the same name were also on the `Transaction` class itself, and called by both the methods on the `TransactionHandle` class and by the `insertDistributedTraceHeaders()` and `acceptDistributedTraceHeaders()` on the `Transaction` class. These "headers" methods can accept W3C Trace Context headers and New Relic DT headers.

The "payload" methods were deprecated and are now removed from the API/`TransactionHandle`. End users should use the "headers" methods.

When processing New Relic DT headers, the "headers" methods use the the "payload" methods internally. This PR adds an underscore to the "payload" methods to indicate that they are only internally used by the new "headers" methods on the `Transaction` class. As such, many tests that used the "payload" methods have been updated. Some were updated to use the "headers" methods where possible.

### DIFF
--- a/lib/transaction/index.js
+++ b/lib/transaction/index.js
@@ -929,7 +929,7 @@ function insertDistributedTraceHeaders(headers) {
   }
 
   try {
-    const newrelicFormatData = this.createDistributedTracePayload().httpSafe()
+    const newrelicFormatData = this._createDistributedTracePayload().httpSafe()
     headers[NEWRELIC_TRACE_HEADER] = newrelicFormatData
     logger.trace('Added outbound request distributed tracing headers in transaction %s', this.id)
   } catch (error) {
@@ -993,8 +993,8 @@ function acceptTraceContextPayload(traceparent, tracestate, transport) {
  * @param {object} payload                - The distributed trace payload to accept.
  * @param {string} [transport='Unknown']  - The transport type that delivered the payload.
  */
-Transaction.prototype.acceptDistributedTracePayload = acceptDistributedTracePayload
-function acceptDistributedTracePayload(payload, transport) {
+Transaction.prototype._acceptDistributedTracePayload = _acceptDistributedTracePayload
+function _acceptDistributedTracePayload(payload, transport) {
   if (!payload) {
     this.agent.recordSupportability('DistributedTrace/AcceptPayload/Ignored/Null')
     return
@@ -1089,11 +1089,6 @@ function acceptDistributedTracePayload(payload, transport) {
     return
   }
 
-  // TODO: This should be removable / covered by acceptDistributedTraceHeaders
-  // once the Transaction#acceptDistributedTracePayload API that directly invokes
-  // this acceptDistributedTracePayload is fully removed in a future Major version.
-  transport = TRANSPORT_TYPES_SET[transport] ? transport : TRANSPORT_TYPES.UNKNOWN
-
   this.parentType = data.ty
   this.parentApp = data.ap
   this.parentAcct = data.ac
@@ -1163,9 +1158,9 @@ Transaction.prototype._getParsedPayload = function _getParsedPayload(payload) {
 /**
  * Creates a distributed trace payload.
  */
-Transaction.prototype.createDistributedTracePayload = createDistributedTracePayload
+Transaction.prototype._createDistributedTracePayload = _createDistributedTracePayload
 
-function createDistributedTracePayload() {
+function _createDistributedTracePayload() {
   const config = this.agent.config
   const accountId = config.account_id
   const appId = config.primary_application_id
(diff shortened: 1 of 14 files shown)


test/unit/distributed_tracing/dt-cats.test.js
```diff
@@ -64,7 +68,10 @@ describe('distributed tracing', function() {
 
         if (testCase.outbound_payloads) {
           testCase.outbound_payloads.forEach((outbound) => {
-            const created = JSON.parse(tx.createDistributedTracePayload().text())
+            let headers = {}
+            tx.insertDistributedTraceHeaders(headers)
+            const payload = headers.newrelic
+            const created = tx._getParsedPayload(payload)
```
nijotz, line 74: I update these cross agent tests to use the "headers" methods. One effect of this was that the `acceptDistributedTraceHeaders()` technically had a bug. If the `newrelic` header was present, but was an empty string, the agent should send up a `DistributedTrace/AcceptPayload/Ignored/Null` metric. But the way the if statement was setup, if the newrelic "payload" was empty it would skip processing all together.

lib/transaction/index.js
```diff
@@ -886,19 +886,14 @@ function acceptDistributedTraceHeaders(transportType, headers) {
...
   if (traceparent) {
     logger.trace('Accepting trace context DT payload for transaction %s', this.id)
-
     // assumes header keys already lowercase
     const tracestate = headers[TRACE_CONTEXT_STATE_HEADER]
-
     this.acceptTraceContextPayload(traceparent, tracestate, transport)
-  } else {
+  } else if (NEWRELIC_TRACE_HEADER in headers) {
+    logger.trace('Accepting newrelic DT payload for transaction %s', this.id)
```
nijotz, line 893: This change allows empty string newrelic headers through to the payload method for processing, so that `DistributedTrace/AcceptPayload/Ignored/Null` metrics can be processed.

lib/transaction/index.js
```diff
@@ -1185,11 +1175,6 @@ function createDistributedTracePayload() {
     return new DTPayloadStub()
   }
 
-  // TODO: This should be removable / covered by insertDistributedTraceHeaders
-  // once the Transaction#createDistributedTracePayload API that directly invokes
-  // this createDistributedTracePayload is fully removed in a future Major version.
-  this._calculatePriority()
-
```
nijotz, line 1192: This method is never called directly, only via the "headers" method, which calls `_calculatePriority()`

test/unit/transaction.test.js
```diff
@@ -985,42 +985,33 @@ describe('Transaction', function() {
...
-    it('generates a priority for entry-point transactions', () => {
-      expect(tx.priority).to.equal(null)
-      expect(tx.sampled).to.equal(null)
-
-      tx.createDistributedTracePayload()
-
-      expect(tx.priority).to.be.a('number')
-      expect(tx.sampled).to.be.a('boolean')
-    })
-
```
nijotz, line 1003: This method should no longer generate priorities since it is not the entry-point. `insertDistributedTraceHeaders()` is and it will calculate priority.

test/unit/transaction.test.js
```diff
@@ -985,42 +985,33 @@ describe('Transaction', function() {
...
+      const payload = JSON.parse(tx._createDistributedTracePayload().text())
       expect(payload.d.sa).to.equal(tx.sampled)
       expect(payload.d.pr).to.equal(tx.priority)
     })
 
     it('adds the current span id as the parent span id', function() {
       agent.config.span_events.enabled = true
       agent.tracer.segment = tx.trace.root
-      const payload = JSON.parse(tx.createDistributedTracePayload().text())
+      tx.sampled = true
```
nijotz, line 1013: I had to add `sample = true` here because before the "payload" method would call `_calculatePriority()`. And this would set `sampled` to `true` on the transaction. And when sampled was true, `createDistributedTracePayload()` [would add an `id` to the payload](https://github.com/newrelic/node-newrelic/blob/75729ded806c336d52bf685e43dbfa133c084b97/lib/transaction/index.js#L1205). Since `_calculatePriority()` was removed, the test below for `payload.d.id` would fail. So I had to manually add `sampled = true` to this test.

test/unit/distributed_tracing/dt-cats.test.js
carlo-808: Nice catch and update!

carlo-808 (approved):
👍 
I see `createDistributedTracePayload` and `acceptDistributedTracePayload` methods removed from `API/TransactionHandle`.

PR #541, opened 2020-11-09
Release v7.0.0

nijotz:
## Proposed Release Notes

* Added official parity support for Node 14

* Dropped Node v8.x support. For further information on our support policy,
  see: https://docs.newrelic.com/docs/agents/nodejs-agent/getting-started/compatibility-requirements-nodejs-agent.
  * Removed Node v8.x from CI
  * Adds check that minimum Node version is >=10 and warns if >=15
  * Sets Node engine to >=10
  * **BREAKING** Dropped support for Node v8.x HTTP get() function signature
    * strictly uses global.URL class in http core instrumentation
    * removes Nodejs 8.x - 9.x checks
  * Update New Relic Dependencies to versions with updated Node version support
    * @newrelic/aws-sdk v3.0.0
    * @newrelic/koa v5.0.0
    * @newrelic/native-metrics v6.0.0
    * @newrelic/superagent v4.0.0
    * @newrelic/test-utilities v5.0.0

* **BREAKING** Removed deprecated setIgnoreTransaction API method

* **BREAKING** Removed deprecated httpResponseCode, response.status and
  httpResponseMessage http response attributes

* **BREAKING** Removed the api.custom_parameters_enabled configuration item and
  associated environment variable NEW_RELIC_API_CUSTOM_PARAMETERS. Please use
  api.custom_attributes_enabled instead

* **BREAKING** Removed deprecated Distributed Tracing API methods,
  createDistributedTracePayload() and acceptDistributedTracePayload()

* Finalized removal of ignored_params and capture_params

* Added additional logging to W3C Trace Context header creation

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,37 @@
+### 7.0.0 (2020-11-09):
+
+* Drop Node v8.x support and add Node v14.x
+  * Removed Node v8.x and added v14.x to CI
+  * Adds check that minimum Node version is >= 10
+  * Sets Node engine to >= 10 and < 15
+  * BREAKING Dropped support for Node v8.x HTTP get() function signature
+    * strictly uses global.URL class in http core instrumentation
+    * removes Nodejs 8.x - 9.x checks
+  * Update New Relic Dependencies to versions with updated Node version support
+    * @newrelic/aws-sdk v3.0.0
+    * @newrelic/koa v5.0.0
+    * @newrelic/native-metrics v6.0.0
+    * @newrelic/superagent v4.0.0
+    * @newrelic/test-utilities v5.0.0
+
+* BREAKING Removed deprecated setIgnoreTransaction API method
+
+* BREAKING Removed deprecated httpResponseCode, response.status and
+  httpResponseMessage http response attributes
+
+* BREAKING Removed the api.custom_parameters_enabled configuration item and
+  associated environment variable NEW_RELIC_API_CUSTOM_PARAMETERS. Please use
+  api.custom_attributes_enabled instead
+
+* BREAKING Removed deprecated Distributed Tracing API methods
+
+* Removed deprecation warning for invalid config items, ignored_params and
+  capture_params. Use attributes.exclude and attributes.enabled instead
+
+* Added additional logging to W3C Trace Context header creation.
+
+* Upgrade @newrelic/koa from 4.0.0 to 4.1.0
+
 ### 6.14.0 (2020-10-28):
 
 * Updated README for consistency.
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,16 +14,16 @@
     * @newrelic/superagent v4.0.0
     * @newrelic/test-utilities v5.0.0
 
-* BREAKING Removed deprecated setIgnoreTransaction API method
+* **BREAKING** Removed deprecated setIgnoreTransaction API method
 
-* BREAKING Removed deprecated httpResponseCode, response.status and
+* **BREAKING** Removed deprecated httpResponseCode, response.status and
   httpResponseMessage http response attributes
 
-* BREAKING Removed the api.custom_parameters_enabled configuration item and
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,3 +1,37 @@
+### 7.0.0 (2020-11-09):
```
michaelgoin, line 1: I would separate out the items for adding Node 14 support and dropping Node 8. Just make easily visible/searchable. Prob something like 'Added official parity support for Node 14' as the first item.

I'd recommend we make the "BREAKING" lines bold.

Finally, with the dropping of Node 8... I'd prob include a link to the supported versions policy (of which we prob need to update for Node 16 now).

```
For further information on our
  support policy, see:
  https://docs.newrelic.com/docs/agents/nodejs-agent/getting-started/compatibility-requirements-nodejs-agent.
```

NEWS.md
```diff
@@ -1,3 +1,37 @@
...
+
+* Drop Node v8.x support and add Node v14.x
+  * Removed Node v8.x and added v14.x to CI
+  * Adds check that minimum Node version is >= 10
+  * Sets Node engine to >= 10 and < 15
```
michaelgoin, line 6: I hope we aren't doing this? I asked that we not set the engine version to < 15 because it prevents folks from trying the agent early on future versions such as when 16 is released.

NEWS.md
```diff
@@ -1,3 +1,37 @@
...
+* BREAKING Removed deprecated setIgnoreTransaction API method
+
+* BREAKING Removed deprecated httpResponseCode, response.status and
+  httpResponseMessage http response attributes
+
+* BREAKING Removed the api.custom_parameters_enabled configuration item and
+  associated environment variable NEW_RELIC_API_CUSTOM_PARAMETERS. Please use
+  api.custom_attributes_enabled instead
+
+* BREAKING Removed deprecated Distributed Tracing API methods
```
michaelgoin, line 26: I'd list these two methods out specifically.

NEWS.md
```diff
@@ -1,3 +1,37 @@
...
+
+* Removed deprecation warning for invalid config items, ignored_params and
```
michaelgoin, line 28: I could see this maybe being slightly confusing to an end user. I think (technically) these were already removed but only deprecation warnings remained... but I could see it also reading like "we just stopped deprecating." 

Might reword to "Finalized removal of ignored_params and capture_params" or something similar.

NEWS.md
```diff
@@ -1,3 +1,37 @@
...
+  capture_params. Use attributes.exclude and attributes.enabled instead
+
+* Added additional logging to W3C Trace Context header creation.
+
+* Upgrade @newrelic/koa from 4.0.0 to 4.1.0
```
michaelgoin, line 33: looks like this ends up at v5.0.0 per your notes above so prob could remove this line.

NEWS.md
nijotz: It doesn't prevent them, just [warns them](https://github.com/newrelic/node-newrelic/pull/524/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R60)

michaelgoin: OK good. This release message is wrong then. Would remove the `<15` from this line.

NEWS.md
nijotz: pushed fix

NEWS.md
nijotz: pushed fix

NEWS.md
nijotz: pushed fix

NEWS.md
nijotz: Split out and bolded BREAKING

NEWS.md
nijotz: fixed

nijotz:
alright I think I addressed all feedback

michaelgoin (approved):
Looks good 👍

PR #550, opened 2020-11-17
Update grpc-js version to v1.2.0

carlo-808:
## Proposed Release Notes
* Update `@grpc/grpc-js` to version v1.2.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,11 +142,10 @@
     "newrelic-naming-rules": "bin/test-naming-rules.js"
   },
   "dependencies": {
-    "@grpc/grpc-js": "^1.2.0",
+    "@grpc/grpc-js": "1.2.0",
     "@grpc/proto-loader": "^0.5.5",
     "@newrelic/aws-sdk": "^3.0.0",
     "@newrelic/koa": "^5.0.0",
-    "@newrelic/native-metrics": "^6.0.0",
     "@newrelic/superagent": "^4.0.0",
     "@tyriar/fibonacci-heap": "^2.0.7",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -139,13 +139,14 @@
     "versioned": "npm run prepare-test && time ./bin/run-versioned-tests.sh"
   },
   "bin": {
-    "newrelic-naming-rules": "./bin/test-naming-rules.js"
+    "newrelic-naming-rules": "bin/test-naming-rules.js"
```
michaelgoin, line 142: what's up with this change?

package.json
```diff
@@ -139,13 +139,14 @@
...
   },
   "dependencies": {
-    "@grpc/grpc-js": "1.1.8",
+    "@grpc/grpc-js": "^1.2.0",
```
michaelgoin, line 145: Are you intentionally not pinning the version now? We havent' really discussed if we are ready to do that yet or not.

package.json
```diff
@@ -139,13 +139,14 @@
...
     "@grpc/proto-loader": "^0.5.5",
     "@newrelic/aws-sdk": "^3.0.0",
     "@newrelic/koa": "^5.0.0",
+    "@newrelic/native-metrics": "^6.0.0",
```
michaelgoin, line 149: ???

carlo-808: ooof!  I'll fix that.

package.json
carlo-808: Good catch.  I'll pin that.

michaelgoin (approved):

Issue #554, opened 2020-11-23
AWS `http.statusCode` not captured in async lambdas


carlo-808:
## Description
`http.statusCode` attributes are not being captured for a lambda web response when lambda is invoked asynchronously.

Code where async lambda response is processed:
https://github.com/newrelic/node-newrelic/blob/main/lib/serverless/aws-lambda.js#L196-L214

## Steps to Reproduce
- Execute async lambda with ALB event or API Gateway Lambda Proxy event.
- View NR payload and see `http.statusCode` not sent in span and trace events.

## Your Environment
AWS Lambda with Node v12

PR #556, opened 2020-11-24
Capture http.statusCode on web response of async invoked lambda

carlo-808:
## Proposed Release Notes
* Fixes a bug where the `http.statusCode` attribute was not being captured for an async invoked lambda.
## Links
* Closes: #554

### DIFF
--- a/lib/serverless/aws-lambda.js
+++ b/lib/serverless/aws-lambda.js
@@ -129,7 +129,6 @@ class AwsLambda {
       const segmentRecorder = isApiGatewayLambdaProxy ? recordWeb : recordBackground
       const segment = shim.createSegment(functionName, segmentRecorder)
       transaction.baseSegment = segment
-      
       // resultProcessor is used to execute additional logic based on the
       // payload supplied to the callback.
       let resultProcessor
@@ -180,7 +179,6 @@ class AwsLambda {
       shim.agent.setLambdaArn(context.invokedFunctionArn)
 
       shim.agent.setLambdaFunctionVersion(context.functionVersion)
-      
       segment.addSpanAttributes(awsAttributes)
 
       segment.start()
@@ -194,7 +192,7 @@ class AwsLambda {
         throw err
       }
       if (shim.isPromise(res)) {
-        res = lambdaInterceptPromise(res, txnEnder)
+        res = lambdaInterceptPromise(res, resultProcessor, txnEnder)
       }
       return res
     }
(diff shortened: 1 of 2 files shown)


lib/serverless/aws-lambda.js
```diff
@@ -194,16 +192,20 @@ class AwsLambda {
...
       return res
     }
 
     // In order to capture error events
     // we need to store the error in uncaughtException
     // otherwise the transaction will end before they are captured
-    function lambdaInterceptPromise(prom, cb) {
+    function lambdaInterceptPromise(prom, resultProcessor, cb) {
       return prom.then(function onThen(arg) {
+        // check if we are in web transaction
```
michaelgoin, line 205: Seems this should only occur for an uncaught exception, so I believe it is fine to only process the result in the 'then'.

lib/serverless/aws-lambda.js
```diff
@@ -397,7 +399,7 @@ function setWebResponse(transaction, response) {
       responseCode
     )
 
-    const segment = transaction.agent.tracer.getSegment()
+    const segment = transaction.baseSegment
```
michaelgoin, line 402: Curious why baseSegment is necessary here? Can we leave a comment to indicate why?

I'm assuming this changes the behavior of which span the status code attribute is assigned to. What span was it on before and which one is it on now? Or is this just a more convenient way to grab the same one?

lib/serverless/aws-lambda.js
carlo-808: OK, do you mean I can remove the comment here?

lib/serverless/aws-lambda.js
carlo-808: > Curious why baseSegment is necessary here? Can we leave a comment to indicate why?
> 
> I'm assuming this changes the behavior of which span the status code attribute is assigned to. What span was it on before and which one is it on now? Or is this just a more convenient way to grab the same one?

When testing an async invoked lambda we did not have a current active segment at this point in the code and I noticed that in the synchronous tests the `http.statusCode` was added to the base segment. Granted, these are simple lambda operations. Should I check if we have a current segment first and if not, add it to base segement?

lib/serverless/aws-lambda.js
michaelgoin: I was just commenting out loud that I *think* this is fine, IF the catch is only called on uncaught exceptions like I think it is.

That being said, I don't think the comment is necessary and could potentially be confusing in the future if we added a result processor for non-web case for some reason. (Really though, this instrumentation could prob use some restructuring at some point now that it has evolved a few times). So I'd be fine if you removed the comment.

lib/serverless/aws-lambda.js
michaelgoin: OK. So in theory, these will be the same segment but in the case of the async version... we've lost state in the async hop.

I think it is fine to comment as such. I'm a little surprised we are losing state. Maybe we've already set back to the starting state at this point.

michaelgoin (approved):

PR #570, opened 2020-12-29
Fix empty log message

nijotz:
## Proposed Release Notes
* Fixed empty log message

## Details
`warnOnce()` needs a key as the first argument, and the logging message as the second argument


carlo-808 (approved):
I see usage of warnOnce fixed for the streaming span event aggregator and the query-sample.test.js converted to use tap. 🦖

PR #573, opened 2020-12-31
Fix Node minimum version message (fixes #557)

### LINKED ISSUE #557: Incorrect version in error message
## Description

Error message [here](https://github.com/newrelic/node-newrelic/blob/main/index.js#L53) has version 8.0.0 hard coded in the message string but the conditional is checking for `'<10.0.0'`.

Seems minor, but much time was wasted trying to figure this out before I looked at the source code.

## Expected Behavior

Error message should accurately reflect the version that is being checked.

## Steps to Reproduce

Try running with node version < 10 but >= 8.0.0 and you will see the confusing message.

nijotz:
## Proposed Release Notes
* Fixed log message regarding Node minimum version

## Links
* #557

### DIFF
--- a/index.js
+++ b/index.js
@@ -51,13 +51,13 @@ function initialize() {
     // TODO: Update this check when Node v10 is deprecated.
     if (psemver.satisfies('<10.0.0')) {
       message = 'New Relic for Node.js requires a version of Node equal to or\n' +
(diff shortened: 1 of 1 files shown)


carlo-808 (approved):
👍

Issue #567, opened 2020-12-16
SQL Traces can be generated with invalid SQL ID


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT share sensitive information, whether personal, proprietary, or otherwise! )

When the number of digits to generate a SQL ID was increased to reduce collisions, it intended to rely on `parseInt` to trim down to a valid Integer. At the time, we validated it resulted in a valid JavaScript number but it turns out `parseInt()` can return a value that is larger than `Number.MAX_SAFE_INTEGER` while mostly seeming fine.

When we send this value to the server, the trace gets dropped because the number is not a valid `long`.

Example query for reproduction: 

```
'select `createdAt`, `updatedAt`, `id`, `firstName`, `lastName` from `Employee` limit ?'.
```

Example of parseInt behavior:

```
> const intVal = parseInt('3785db4c90b40f8a', 16)
undefined
> intVal > Number.MAX_SAFE_INTEGER
true
```

We'll lose some precision but we can probably trim the hex string down to 13 digits to ensure the largest possible hex number is within the valid range of values.

nijotz:
Fixed in latest release (7.1.0)

PR #574, opened 2021-01-04
Use 15 char hex to guarantee signed long fit

nijotz:
## Proposed Release Notes
* Modified SQL trace ID generation to fit in a signed long

## Links
#567 

## Details
I converted the test file for query traces to tap (first commit). Then I updated the tests to have a query that would generate an ID that is ```> 2^63 - 1``` and added a test that would fail for this larger ID (second commit). Then the fix is simply taking 15 hex characters instead of 16 (third commit). I believe we're limiting IDs to 2^60, but still plenty of ID space I think.

### DIFF
--- a/lib/db/slow-query.js
+++ b/lib/db/slow-query.js
@@ -22,9 +22,8 @@ function SlowQuery(segment, type, query, trace) {
 }
 
 function normalizedHash(value) {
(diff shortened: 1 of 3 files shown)


carlo-808 (approved):
I see query trace hash is updated to create a value that will fit within a signed long int and with supporting tests added (plus bonus test converted to tap). 👍

Issue #583, opened 2021-01-20
Bump grpc-js to latest version (v1.2.3)


carlo-808:
Update grpc-js to v1.2.3

PR #594, opened 2021-01-20
Bumped grpc-js version to ^1.2.4

michaelgoin:
## Proposed Release Notes

* Bumped grpc-js version to ^1.2.4. 

  Now allows minor and patch auto-updates.

## Links

* Closes: https://github.com/newrelic/node-newrelic/issues/583

## Details

We've yet to encounter any fundamental issues with grpc-js updates past the 1.x version. Seems time to loosen that restriction and reduce some toil. We can revisit if an issue occurs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -220,9 +220,9 @@
       }
     },
     "@grpc/grpc-js": {
-      "version": "1.2.2",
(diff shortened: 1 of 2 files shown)


michaelgoin:
Broken test resolved via: https://github.com/newrelic/node-newrelic/pull/595

carlo-808 (approved):
👍

PR #595, opened 2021-01-20
Fixes resolveMx test by using example.com for a valid exchange.

michaelgoin:
## Proposed Release Notes

* Fixes resolveMx test by using example.com for a valid exchange.

## Details

Updated the test to point to example.com which returns an exchange. The test previously pointed at a personal domain which expired.


carlo-808 (approved):
👍

PR #602, opened 2021-01-26
Adds golden-path release note generation script.

michaelgoin:
## Details

Added golden-path release note generation script intended for manual run.

Grabs PR's merged since the last release in GitHub and then inserts new release notes into News.md.

Last release date determined from the date of the commit associated with the tag of the release, instead of date release was 'published' to avoid missing changes that may happen mid-process.

Long-term, will probably add ability to manually set a release or tag to handle non-standard cases such as having most recently published an emergency backport.

To run: `node ./bin/generate-release-notes.js`

As this evolves and potentially is made more generic, more of the hard-coded details should be parameterized.

### DIFF
--- a/bin/generate-release-notes.js
+++ b/bin/generate-release-notes.js
@@ -1,0 +1,94 @@
+'use strict'
+
+const fs = require('fs')
+const github = require('./github')
+
+const FILE_NAME = 'NEWS.md'
+const PROPOSED_NOTES_HEADER = 'Proposed Release Notes'
+const NEXT_VERSION_HEADER = '### vNext (TBD):'
+
+async function generateReleaseNotes() {
+  try {
+    const latestRelease = await github.getLatestRelease()
+    console.log(`The latest release is: ${latestRelease.name} published: ${latestRelease.published_at}`)
+    console.log(`Tag: ${latestRelease.tag_name}, Target: ${latestRelease.target_commitish}`)
+
+    const tag = await github.getTagByName(latestRelease.tag_name)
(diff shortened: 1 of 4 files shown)


bin/generate-release-notes.js
```diff
@@ -0,0 +1,96 @@
...
+    const commitDate = commit.commit.committer.date
+
+    console.log(`Finding merged pull requests since: ${commitDate}`)
+
+    const mergedPullRequests = await github.getMergedPullRequestsSince(commitDate)
+    console.log(`Found ${mergedPullRequests.length}`)
+
+    const releaseNoteData = mergedPullRequests.map((pr) => {
+      const parts = pr.body.split(/(?:^|\n)##\s*/g)
+      const {1: proposedReleaseNotes} = parts
```
carlo-808, line 29: Is there a chance that the 2nd part of the parts array may not exist?

bin/github.js
```diff
@@ -0,0 +1,106 @@
+'use strict'
+
+const { Octokit } = require("@octokit/rest")
```
carlo-808, line 3: Nice kit! Makes github interactions easy.

carlo-808 (approved):
👍  Just one minor comment if you want to address now or work on as we build the script out.

bin/generate-release-notes.js
michaelgoin: It is certainly possible this would blow up with a malformed PR template. Was just aiming for golden path here.

We can tackle more robustness as we iterate.

PR #603, opened 2021-01-26
Adds script to auto-update NEWS.md version header from vNext.

michaelgoin:
## Details

When npm version is ran, the package.json version gets updated and then executes our 'version' script in package.json.

Example when running `npm version patch` w/ some release notes:

```
Updating vNext header to:  ### v7.1.1 (2021-01-26)
*** [SUCCESS] ***
```

### DIFF
--- a/bin/update-changelog-version.js
+++ b/bin/update-changelog-version.js
@@ -1,0 +1,66 @@
+'use strict'
+
+/**
+ * Executed as a part of the "version" step of npm version.
+ * Updates the placeholder release note header with the incremented version
+ * from running npm version
+ */
+
+const fs = require('fs')
+const packageInfo = require('../package.json')
+
+const FILE_NAME = 'NEWS.md'
+const NEXT_VERSION_HEADER = '### vNext (TBD):'
+
+const SUCCESS_MSG = '*** [SUCCESS] ***'
+const FAIL_MSG = '! [FAILURE] !'
+
+updateChangelogVersion()
+
+async function updateChangelogVersion() {
+  try {
+    await updateHeader(FILE_NAME)
+
+    console.log(SUCCESS_MSG)
+  } catch (err) {
+    console.log(FAIL_MSG)
+    console.error(err)
+  }
+}
+
+function updateHeader(file) {
+  const promise = new Promise((resolve, reject) => {
+    fs.readFile(file, 'utf8', (err, data) => {
+      if (err) {
+        return reject(err)
+      }
+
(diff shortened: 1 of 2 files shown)


bin/update-changelog-version.js
```diff
@@ -0,0 +1,66 @@
...
+      }
+
+      // toISOString() will always return UTC time
+      const todayFormatted = new Date().toISOString().split('T')[0]
+      const version = `v${packageInfo.version}`
+      const newChangelogHeader = `### ${version} (${todayFormatted})`
+
+      console.log('Updating vNext header to: ', newChangelogHeader)
+
+      if (!data.startsWith(NEXT_VERSION_HEADER)) {
```
carlo-808, line 45: Maybe we should add a note that the `NEXT_VERSION_HEADER` is added to News.md from the generate-release-notes.js script?

package.json
```diff
@@ -136,7 +136,9 @@
     "unit": "rm -f newrelic_agent.log && time tap --test-regex='(\\/|^test\\/unit\\/.*\\.test\\.js)$' --timeout=180 --no-coverage",
     "update-cross-agent-tests": "./bin/update-cats.sh",
     "versioned-tests": "./bin/run-versioned-tests.sh",
-    "versioned": "npm run prepare-test && time ./bin/run-versioned-tests.sh"
+    "update-changelog-version": "node ./bin/update-changelog-version",
+    "versioned": "npm run prepare-test && time ./bin/run-versioned-tests.sh",
```
carlo-808, line 140: Should we differentiate between the `versioned` and `version` script a bit more? Maybe `version` could be `release-version`?

carlo-808 (approved):
Code looks good! Just added a couple minor comments that you might want to address.

bin/update-changelog-version.js
michaelgoin: Makes sense. Could also add manually but that is the intent for sure.

package.json
michaelgoin: I don't love how similar they are either.

That being said, I can't rename 'version' because that is a predefined script name by npm to get ran at the right time when `npm version` is called. There are pre, mid and post hooks. This one is mid.

We could rename the versioned test script but since a bunch of things are reliant on that I punted for now. Longer-term, perhaps we tweak the test side of things.

carlo-808: That makes sense. Yeah, maybe we can rename versioned test script next time we need to do work on that.

carlo-808 (approved):
👍

PR #604, opened 2021-01-27
Adds workflow for publishing to npm when a v* tag is pushed to the repo.

michaelgoin:
## Proposed Release Notes

* Added workflow for publishing to npm when a v* tag is pushed to the repo.

## Details

Auto-publishes when a new tag is pushed in the form v*. 

Can also trigger manually via workflow dispatch but there's one big limitation. Via the UI, only branches are available as ref. So if we ever need to manually publish we'll need to do from a branch. Could do main if clean/same version but may want ot create a branch from the tag being released and then publish that... if comes up. Hopefully longer-term GitHub will expand this (I think they may have already via the workflow dispatch API).

Could potentially modify the checkout action to take a param for ref to enable setting to a tag or such.

### DIFF
--- a/.github/workflows/publish-release-tag.yml
+++ b/.github/workflows/publish-release-tag.yml
@@ -6,7 +6,8 @@ on:
       - 'v*' # Push events to matching v*, e.g. v1.0, v20.15.10
   workflow_dispatch:
     # Allow manual trigger for non-standard cases.
-    # Unfortunately only allows branches are refs, currently (Jan 2021).
+    # (Jan 2021): Currently only allows branches as refs,
+    # so cannot specify a tag (or arbitrary sha).
 
 jobs:
   npm-publish:
--- a/.github/workflows/publish-release-tag.yml
+++ b/.github/workflows/publish-release-tag.yml
@@ -1,0 +1,30 @@
+name: Publish Release Tag
+
+on:
+  push:
+    tags:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-release-tag.yml
```diff
@@ -0,0 +1,30 @@
+name: Publish Release Tag
+
+on:
+  push:
+    tags:
+      - 'v*' # Push events to matching v*, e.g. v1.0, v20.15.10
+  workflow_dispatch:
+    # Allow manual trigger for non-standard cases.
+    # Unfortunately only allows branches are refs, currently (Jan 2021).
```
carlo-808, line 9: This means you must select a branch to run the workflow on?
minor: should it be '.. allows branches as refs'?

carlo-808 (commented):
Looks pretty straight forward. 👍 
I added a minor comment.

.github/workflows/publish-release-tag.yml
michaelgoin: oops yeah.

michaelgoin: for manual (worker-dispatch) runs you can only select a branch, you cannot select a tag or input an arbitrary sha.

there's prob a way to work around that by doing a custom param that inputs into the checkout action but not taking that on for now.

carlo-808 (approved):
Looks good! 👍

PR #609, opened 2021-02-01
revert from failed npm version run on node v15

carlo-808:
Revert changes from failed `npm run version` run under Node version 15


michaelgoin (approved):
👍

Issue #585, opened 2021-01-20
Automate release tagging


michaelgoin:
Ideally, handles backports... but can defer to add support later.

PR #619, opened 2021-02-12
Adds create-release workflow and supporting scripts

carlo-808:
## Proposed Release Notes
* Adds `Create Release` github actions workflow to create a release for `patch` or `minor` releases
* Updates agent version and adds release tags
## Links
Closes: #585

### DIFF
--- a/bin/check-workflow-run.js
+++ b/bin/check-workflow-run.js
@@ -1,0 +1,77 @@
+'use strict'
+
+const Github = require('./github')
+
+const SUCCESS_MSG = '*** [OK] ***'
+const FAIL_MSG = '! [ERROR] !'
+
+const formatRun = (run) => {
+  return {
+    name: run.name,
+    repository: run.repository.name,
+    repoOwner: run.repository.owner.login,
+    branch: run.head_branch,
+    status: run.status,
+    url: run.url,
+    workflow_id: run.workflow_id,
+    event: run.event
+  }
+}
+
+async function checkWorkflowRun(repoOwner, branch) {
+  const github = new Github(repoOwner)
+
+  try {
+    const results = {
+      ci: false,
+      smokeTest: false
+    }
+
+    const latestRun = await github.getLatestWorkflowRun('ci-workflow.yml', branch)
+
+    if (latestRun === undefined) {
+      console.log('No ci workflow run found.')
+    } else {
+      console.log('CI workflow run details: ', JSON.stringify(formatRun(latestRun)))
+      if (latestRun.status === 'completed' && latestRun.conclusion === 'success') {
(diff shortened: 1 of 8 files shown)


bin/create-release.js
```diff
@@ -0,0 +1,71 @@
...
+    if (!passesWorkflowChecks) {
+      process.exit(1)
+    }
+  } catch (err) {
+    console.log(err)
+
+    process.exit(1)
+  }
+
+  exec(`npm version ${releaseType} && git push origin ${branch} && git push --tags`, (err, stdout) => {
```
michaelgoin, line 54: Not sure what the output looks like but if nothing obvious, we may want to write a message before executing this step.

bin/github.js
```diff
@@ -13,94 +10,110 @@ const octokit = new Octokit({
...
-
-  return result.data
-}
-
-async function getTagByName(name) {
-  const perPage = 100
+class Github {
+  constructor(repoOwner, repository) {
+    // Default to node agent repo for now
+    this.repoOwner = repoOwner ? repoOwner : 'newrelic'
```
michaelgoin, line 16: could be simplified to: `this.repoOwner = repoOwner || 'newrelic'`

bin/github.js
```diff
@@ -13,94 +10,110 @@ const octokit = new Octokit({
...
-    owner: repoOwner,
-    repo: repository
-  })
-
-  return result.data
-}
-
-async function getTagByName(name) {
-  const perPage = 100
+class Github {
```
michaelgoin, line 13: Given the change to be a class now... I believe generate-release-notes.js will be broken as it has not been updated to use the class.

michaelgoin (changes requested):
Since generate-release-notes will be broken, requesting changes on this PR.

Minor comments don't need to be addressed unless desired.

An additional note (can be handled separately): we are getting to enough related scripts that it likely is starting to make sense to organize these into an additional folder at some point.

bin/check-workflow-run.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    const latestRun = await github.getLatestWorkflowRun('ci-workflow.yml', branch)
+
+    if (latestRun === undefined) {
+      console.log('No ci workflow run found.')
+    } else {
+      console.log('CI workflow run details: ', JSON.stringify(formatRun(latestRun)))
+      if (latestRun.status === 'completed' && latestRun.conclusion === 'success') {
+        results.ci = true
```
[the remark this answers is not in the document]
carlo-808: Sounds good, not changing this at this time.

michaelgoin (approved):

Issue #620, opened 2021-02-16
Failure to reconnect to trace observer endpoint on 200 OK at higher log levels


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

It appears that currently, when the `stream.on('data', ...)` event is not subscribed to, the `stream.on('status', ...)` event is not fired (or at least not hitting our handler) by grpc-js as of `1.2.5` (and likely earlier given related issues we've had). This results in us failing to receive indication that we have received a 200 OK status and need to reconnect. This then triggers `ERR_STREAM_WRITE_AFTER_END` errors as it seems the stream has been handled internally.

Without knowing to reconnect (hitting status), we stay on this bad stream and do not recover. The queue fills up and spans get dropped.

To work around in the near-term, we should subscribe to the data event but still limit logging to higher levels (to avoid filling logs).

**NOTE:** to reproduce, information must be written to the stream from the server. If no data written, everything works as expected.

It also appears `ERR_STREAM_WRITE_AFTER_END` errors do not trigger the 'status' event like status-code based errors do (makes sense) so if for some reason we were to hit this error again due to other grpc-js situations, we would never recover. As such, we should handle this case and reconnect. Given it is not status-code based, I'm guessing we should do an immediate reconnect but prob requires further thought.

michaelgoin:
Update: this turns out to be standard Node streams behavior.

See conversation here: https://github.com/grpc/grpc-node/issues/1696

And the text blurb here: https://nodejs.org/api/stream.html#stream_event_end

PR #621, opened 2021-02-17
Fixes infinite tracing reconnect handling

michaelgoin:
## Proposed Release Notes

* Fixed bug where the agent failed to reconnect to Infinite Tracing grpc streams on Status OK at higher log levels.
 
  Node streams require all data be consumed for the end/status events to fire. We were only reading data at lower log levels where we'd use/log the data. This resulted in a failure to reconnect and 'ERR_STREAM_WRITE_AFTER_END' errors. The agent now always listens to the 'data' event, even if not logging, and will also reconnect (with 15 second delay) on any 'ERR_STREAM_WRITE_AFTER_END' error.

## Links

* Fixes: https://github.com/newrelic/node-newrelic/issues/620
* https://nodejs.org/api/stream.html#stream_event_end

### DIFF
--- a/test/integration/grpc/reconnect.tap.js
+++ b/test/integration/grpc/reconnect.tap.js
@@ -24,6 +24,7 @@ const protoLoader = require('@grpc/proto-loader')
 const MetricAggregator = require('../../../lib/metrics/metric-aggregator')
 const MetricMapper = require('../../../lib/metrics/mapper')
 const MetricNormalizer = require('../../../lib/metrics/normalizer')
+const StreamingSpanEvent = require('../../../lib/spans/streaming-span-event')
 
 const helper = require('../../lib/agent_helper')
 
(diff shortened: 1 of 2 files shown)


test/integration/grpc/reconnect.tap.js
```diff
@@ -114,6 +115,93 @@ tap.test(
...
+      connection.on('disconnected', () => {
+        countDisconnects++
+        t.ok(true, 'disconnected')
+
+        // if we've disconnected twice, the test is done
+        // mark the state as permanantly closed in order to
+        // avoid further automatic reconnects (skipping
+        // _setState to avoid an additional disconnect event)
+        if (countDisconnects > 1) {
+          connection._state = 3 // replace with actual fake enum
```
carlo-808, line 192: minor: maybe add TODO here? or even just comment here that 3 is permanently closed. Comments above are great so, no really urgency to update here.

carlo-808 (approved):
I see grpc connection now reading response from the `data` event. 👍

test/integration/grpc/reconnect.tap.js
michaelgoin: Ah yeah, copy-paste from Alan's original test case. Prob leave for now. Eventually, we can prob extract some of this to some shared logic anyways.

PR #622, opened 2021-02-19
Stops calling send() on streaming span aggregator on initial harvest.

michaelgoin:
## Proposed Release Notes

* Removed initial harvest send() call on streaming span event aggregator to prevent warning in logs.

## Details

Longer-term, would be nice if the code was restructured so we didn't need to check for streaming VS non streaming but going with the minimal effort for now.


carlo-808 (approved):
👍 Added a check for streaming span aggregator to gate to aggregator.send().

PR #630, opened 2021-02-23
Fix create-release action to use branch reference

carlo-808:
* Update create-release action to use github.ref for branch input


michaelgoin (approved):
👍

Issue #637, opened 2021-02-25
Error thrown when transction.acceptDistributedTraceHeaders does not receive headers object


carlo-808:
## Description
When a string is passed to the transaction.acceptDistributedTraceHeaders as the headers parameter, an error will be thrown here: 
https://github.com/newrelic/node-newrelic/blob/main/lib/transaction/index.js#L892

When a string is passed in as `headers`, an error like the following is thrown:
`Uncaught TypeError: Cannot use 'in' operator to search for 'newrelic' in string`

## Expected Behavior
Error should not be thrown in this case and a trace level log should indicate that the headers parameter was not an object.

## Steps to Reproduce
call `transaction.acceptDistributedTraceHeaders('anything', 'not a header object')`

PR #638, opened 2021-02-25
Do not throw error when headers is string passed into acceptDistributedTraceHeaders

carlo-808:
## Proposed Release Notes
* Fixed issue where a call to `transaction.acceptDistributedTraceHeaders` would throw an error when the `headers` parameter is a string.

## Links
Closes: #637

### DIFF
--- a/lib/transaction/index.js
+++ b/lib/transaction/index.js
@@ -879,14 +879,14 @@ Transaction.prototype.getIntrinsicAttributes = function getIntrinsicAttributes()
  */
 Transaction.prototype.acceptDistributedTraceHeaders = acceptDistributedTraceHeaders
 function acceptDistributedTraceHeaders(transportType, headers) {
-  const transport = TRANSPORT_TYPES_SET[transportType] ? transportType : TRANSPORT_TYPES.UNKNOWN
-
-  if (typeof headers !== 'object') {
+  if (headers == null || typeof headers !== 'object') {
     logger.trace(
       'Ignoring distributed trace headers for transaction %s. Headers not passed in as object.',
       this.id)
     return
   }
+
+  const transport = TRANSPORT_TYPES_SET[transportType] ? transportType : TRANSPORT_TYPES.UNKNOWN
 
   // assumes header keys already lowercase
   const traceparent = headers[TRACE_CONTEXT_PARENT_HEADER]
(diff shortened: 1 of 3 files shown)


lib/transaction/index.js
```diff
@@ -889,11 +889,14 @@ function acceptDistributedTraceHeaders(transportType, headers) {
...
     const tracestate = headers[TRACE_CONTEXT_STATE_HEADER]
     this.acceptTraceContextPayload(traceparent, tracestate, transport)
-  } else if (NEWRELIC_TRACE_HEADER in headers) {
+  } else if (headers[NEWRELIC_TRACE_HEADER]) {
     logger.trace('Accepting newrelic DT payload for transaction %s', this.id)
     // assumes header keys already lowercase
     const payload = headers[NEWRELIC_TRACE_HEADER]
     this._acceptDistributedTracePayload(payload, transport)
+  } else {
+    this.agent.recordSupportability('DistributedTrace/AcceptPayload/Ignored/Null')
```
michaelgoin, line 898: ah, i think now i see why the 'in' was used. this metric you are adding here is intended if the payload is null, just for the New Relic format. so if the 'newrelic' header exists, but there's nothing in the payload itself. 

so we aren't quite doing the right thing here if it is a string, this location handles both formats, and we prob should protect against it being null/undefined as well.

so now looking at this closer, we prob wanna just add some checking at the top. you could prob continue to use the `in` operator in that case or you could use `Object.hasOwnProperty`. hasOwn would be slightly more restrictive, as won't walk prototype chain but that should also be rare.

carlo-808: Thanks! I'll add a check at the top instead and keep the `in` operator.

lib/transaction/index.js
```diff
@@ -881,6 +881,13 @@ Transaction.prototype.acceptDistributedTraceHeaders = acceptDistributedTraceHead
 function acceptDistributedTraceHeaders(transportType, headers) {
   const transport = TRANSPORT_TYPES_SET[transportType] ? transportType : TRANSPORT_TYPES.UNKNOWN
 
+  if (typeof headers !== 'object') {
+    logger.trace(
+      'Ignoring distributed trace headers for transaction %s. Headers not passed in as object.',
+      this.id)
+    return
+  }
```
michaelgoin, line 889: Fun fact... in JavaScript typeof null is an 'object'.

```
> let blah = null
undefined
> typeof blah === 'object'
true
```

So you'll want to check null/undefined first. Would do something like `if (headers == null || typeof headers !== 'object')`.

Also, might as well put first prior to the transport code.

michaelgoin (approved):
👍

Issue #462, opened 2020-08-11
Agent Should Not use Redirect Host When Reconnecting to the Agent


astormnewrelic:
When connecting to New Relic, the Node.js Agent uses the value in `agent.config.host` to post a request to the New Relic `preconnect` endpoint.  This endpoint returns a "redirect host", which (at the time of this writing) often looks like one of the following

    collector-00n.newrelic.com
    gov-collector-00n.newrelic.com
    
This URL becomes the URL that agents will send data to.

There is a bug in the Node Agent related to this functionality.  If the agent needs to _reconnect_ to the collector, it will **incorrectly** use this redirect host to call the `preconnect` endpoint, when instead it should use the original configured value in `agent.config.host`.  

It appears the root cause of this bug is the agent rewriting the `agent.config.host` value with the redirect host, and never resetting it when the agent resets.

We should probably _not_ rewrite this value. Instead we should 

1. Store the redirect host separately
2. Have the `RemoteMethod`s (save `preconnect`) use this stored redirect host value)

astorm:
📆 📆 📆 📆 📆 📆 👍

PR #640, opened 2021-02-26
Do not redirect preconnect

michaelgoin:
## Proposed Release Notes

* Fixed bug where agent would attempt to call the 'preconnect' endpoint on the redirect host returned by the previous 'preconnect' call when reconnecting to the New Relic servers.

  The 'preconnect' calls will now always use the original agent configuration value. Subsequent endpoints (connect, harvest endpoints, etc.) will continue to leverage the new redirect host value returned by 'preconnect.' The original config values are no-longer overridden.

## Links

* Fixes: https://github.com/newrelic/node-newrelic/issues/462

## Details

Given this impacts our initial handshake and continued functioning, this is slightly higher risk than say instrumentation changes.

When the new redirect-host comes in, all RemoteMethod instances except 'preconnect' are now updated with the new host and port endpoint values. The endpoint is not stored separately (currently) than the RemoteMethods.

Long-term, I'd like to see potentially having a single place to update. I don't think we need instances per method but rather can have a RemoteMethods like class that has all the shared logic and when you call the appropriate function in passes in the appropriate path details, etc.

### DIFF
--- a/lib/collector/api.js
+++ b/lib/collector/api.js
@@ -12,6 +12,8 @@ const RemoteMethod = require('./remote-method')
 
 const NAMES = require('../metrics/names')
 
+const DEFAULT_PORT = 443
+
 // just to make clear what's going on
 const TO_MILLIS = 1e3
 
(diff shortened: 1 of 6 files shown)


carlo-808 (approved):
Looks good. I see api remote methods updated to accept initial collector endpoint and new method for updating endpoint with host/port from preconnect new endpoint. I agree with your comment that it would be better if we didn't have to update each remote method. 👍

Issue #579, opened 2021-01-12
setImmediate instrumentation should check for active transaction to reduce propagation after transaction end


michaelgoin:
## Is your feature request related to a problem? Please describe.

similar to nextTick, etc.

PR #643, opened 2021-03-01
setImmediate no longer binds/propagates segments for an ended transaction

michaelgoin:
## Proposed Release Notes

* Stopped binding/propagating segments via `setImmediate` for ended transactions.

## Links

* Closes: https://github.com/newrelic/node-newrelic/issues/579

## Details

Straightforward spot-fix for setImmediate. Going to set up a follow-up story to look at doing this in general for our various binding/wrapping methods. Seems if the associated transaction has ended we should be doing as little work as possible. Handling these cases obviously are a bit more edge-case but we don't control the environments we run in and we want to clean up our memory usage quickly and not do unnecessary work.

### DIFF
--- a/lib/instrumentation/core/timers.js
+++ b/lib/instrumentation/core/timers.js
@@ -61,7 +61,12 @@ function initialize(agent, timers, moduleName, shim) {
 
   function wrapSetImmediate(shim, fn) {
     return function wrappedSetImmediate() {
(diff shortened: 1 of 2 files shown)


carlo-808 (approved):
Nice catch! Wrapped setImmediate does not propagate segment for an already ended transaction. 👍

Issue #477, opened 2020-09-14
Requirement of NEW_RELIC_APP_NAME environment variable in serverless mode (Lambda)


superrafek:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

Node.js agent appears to still require `NEW_RELIC_APP_NAME` when running in Lambda as per [documentation](https://docs.newrelic.com/docs/serverless-function-monitoring/aws-lambda-monitoring/get-started/enable-serverless-monitoring-aws-lambda#node). This value does not seem to be used anywhere in the UI or in any Event but without it the agent does not start. The other agents do not seem to share the same requirement.

## Expected Behavior

Remove the requirement of this environment variable for Lambda.

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results

[NOTE]: # ( Provide any other relevant log data. )
[TIP]:  # ( Scrub logs and diagnostic information for sensitive information )

## Steps to Reproduce

Remove `NEW_RELIC_APP_NAME` in Node.js agent instrumented Lambda function to see it fail.

## Your Environment

AWS Lambda with Node.js agent using Manual Instrumentation.

carlonewrelic:
Thanks for submitting the issue @superrafek , we'll get this work prioritized with our PM.

nijotz:
The NEW_RELIC_APP_NAME environment variable is used to name the application. In APM, the applications are named based on this environment variable for Node apps. Some other agents provide a default like 'My Application' or 'PHP Application'.

https://docs.newrelic.com/docs/agents/manage-apm-agents/app-naming/name-your-application#app-name

Are you asking for a default name to be used if one is not provided?

michaelgoin:
We definitely don't need the requirement for Lambda. We should also consider a default for standard cases to ease the process of getting up and running. For example, we could name based off of the package.json 'name' property and log a message saying that is what we are using. If nothing there either, we could default to a generic name.

PR #645, opened 2021-03-02
Defaults the application name when running in serverless mode.

michaelgoin:
## Proposed Release Notes

* Removed requirement to configure application name when running in AWS Lambda (serverless mode).

  Application name is not currently leveraged by New Relic for Lambda invocations. The agent now defaults the application name in serverless mode to remove the requirement of end-user configuration while handling cases if it were to be leveraged in the future. 

## Links

* Closes: https://github.com/newrelic/node-newrelic/issues/477


carlo-808 (approved):
Sets app name to env AWS_LAMBDA_FUNCTION_NAME in serverless mode (lambda). 👍

Issue #642, opened 2021-03-01
Add certificate usage tracking


michaelgoin:
## Is your feature request related to a problem? Please describe.

As we've gotten closer to removing usage of the bundled certs, there have been more concerns around potential impacts.

I'm not sure if we can track *intentional* usage of the bundled certs but we should see what we can track via support metrics. At a minimum, we can probably detect if certificates were used at all.

Also, given this is old software... should prob double-check we actually still are bundling the certs and it wasn't dropped at some point prior to our knowledge.

PR #646, opened 2021-03-03
Adds support metric for cert usage.

michaelgoin:
## Proposed Release Notes

* Added support metric to be able to track usage of cert bundle via usage of custom certificates.

## Links

* Closes: https://github.com/newrelic/node-newrelic/issues/642

## Details

We always include the cert bundle if the certificates configuration item is populated. As such, tracking that as a proxy for the bundle.

Introduces a naming convention for tracking feature usage.


carlo-808 (approved):
👍  I see metric to track cert bundle usage and a new Feature supportability metric namespace. 👾

PR #648, opened 2021-03-03
Index bad config

michaelgoin:
## Proposed Release Notes

* Removed index-bad-config test which tested a no-longer possible use-case.

* Removed license-key test logic from serverless-harvest test.
  
  Serverless mode does not require a license key as data transfer is handled by the integration.


carlo-808 (approved):
Removed unnecessary skip test for license in serverless harvest integration test. Removed unnecessary index-bad-config integration test. 👍

Issue #587, opened 2021-01-20
Automate release creation in GitHub


michaelgoin:
We may want to use this as a trigger, in which case it will be manual... but entering this issue to ensure it does not get forgotten as the design emerges.

PR #653, opened 2021-03-05
Automate GitHub release creation

michaelgoin:
## Links

* Closes: https://github.com/newrelic/node-newrelic/issues/587

## Details

* Adds script which will take a tag and then automatically create a release in GitHub.
* Body of release is pulled from the release notes matching the tag.
* Adds execution of script to publish workflow which executes after npm publish.

Unfortunately, since workflow_dispatch doesn't allow setting a tag instead of a branch... that workflow will fail on manual execution for the create release. We could consider eventually adding a tag input on workflow dispatch and then executing against that for code checkout and release creation but not tackling yet.

### DIFF
--- a/.github/workflows/publish-release-tag.yml
+++ b/.github/workflows/publish-release-tag.yml
@@ -10,7 +10,7 @@ on:
     # so cannot specify a tag (or arbitrary sha).
 
 jobs:
-  npm-publish:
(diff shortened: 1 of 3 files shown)


carlo-808 (approved):
Release creation action and script look 👍

PR #656, opened 2021-03-09
Prepare release automation

michaelgoin:
## Details

* Creating release notes and bumping version now handled by prepare-release script.
* Former 'create-release' script now 'create-release-tag' script which only handles creating and pushing version tags for a branch. Pushing will trigger the deployment workflow. Also added some additional validation.
* Removes update release note script that used to run as part of 'npm version' command as no-longer needed.
* Adds workflow to kick-off tagging script. Only supports manual run with this script.
* Resolves issue where empty release notes would still insert newlines into formatting.

I changed too much in the tag script for it to keep track of the rename so looks like a delete/add.


bin/create-release-tag.js
```diff
@@ -0,0 +1,95 @@
...
+
+async function createReleaseTag() {
+  // Parse commandline options inputs
+  program.parse()
+
+  const options = program.opts()
+
+  console.log('Script running with following options: ', JSON.stringify(options))
+
+  const branch = options.branch ? options.branch.replace('refs/heads/', '') : null
```
carlo-808, line 21: Will the null ever be reached in these checks? Seems like it will be defaulted to 'main' and 'newrelic', but seems like good safety check anyway.

bin/git-commands.js
```diff
@@ -2,127 +2,95 @@
...
+async function pushTags() {
+  const stdout = await execAsPromise('git push --tags')
+  const output = stdout.trim()
 
-  return promise
+  return output
 }
 
-function pushToRemote(remote, branchName) {
+function execAsPromise(command) {
```
carlo-808, line 84: Nice helper function.

bin/npm-commands.js
```diff
@@ -0,0 +1,31 @@
+'use strict'
```
carlo-808, line 1: Nice breaking npm commands out into it's own file.

carlo-808 (approved):
Nice updates to the release process! 👍

PR #657, opened 2021-03-09
Minor release script tweaks

michaelgoin:
## Details

* Updates prepare-release script name missed in previous update..
* Removes unnecessary null checks in create-release-tag script.


carlo-808 (approved):
👍

Issue #439, opened 2020-07-21
Truncated outbound requests are labeled as category generic rather than http


carlo-808:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description
When an outbound http request is made resulting in a truncated segment, the resulting span is labeled with the attribute category: generic.  It should have been labeled category: http.  This is happening because the the truncated segment name is prefixed with 'Truncate' before the segment is converted to a span.  The span category is set to http by looking for the 'External' prefix in the segment name attribute.

Relevant code:
https://github.com/newrelic/node-newrelic/blob/main/lib/transaction/index.js#L214-L224

[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT share sensitive information, whether personal, proprietary, or otherwise! )

## Expected Behavior
Spans resulting from an outbound http request should be labeled category http regardless of whether the originating segment was truncated or not.

[NOTE]: # ( Tell us what you expected to happen. )

## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results

[NOTE]: # ( Provide any other relevant log data. )
[TIP]:  # ( Scrub logs and diagnostic information for sensitive information )

## Steps to Reproduce

[NOTE]: # ( Please be as specific as possible. )
[TIP]:  # ( Link a sample application that demonstrates the issue. )

## Your Environment
Node.js agent version 6.11.0

## Additional context

[TIP]:  # ( Add any other context about the problem here. For example, relevant community posts or support tickets. )

astormnewrelic:
Some good first steps here would be a test case that demonstrates the behavior.

PR #662, opened 2021-03-10
Fixes bug where truncated segments/spans would not have appropriate category

michaelgoin:
## Proposed Release Notes

* Fixed bug where truncated http (external) or datastore segments would generate generic spans instead of appropriate http or datastore spans.

## Links

* fixes: https://github.com/newrelic/node-newrelic/issues/439

## Details

Don't know for sure if datastore operations are impacted (harder to repro) but seemed possible so have a somewhat manufactured test case.

A digression: these tests aren't really unit tests and should be moved to integration tests at some point.

I think a more ideal way to handle this would be for a segment's span context to have the category set when we determine a segment is an external or a datastore. That being said, the datastore instrumentation didn't really lend itself to this in a straightforward manner. So still looking at segment names which is a bit brittle.

On the plus side, the regex approach is faster than the string search even with the extra truncated check. So not making things worse.

```
Truncated startswith OR x 9,128,685 ops/sec ±0.49% (93 runs sampled)
Truncated Regex test x 30,493,176 ops/sec ±0.26% (93 runs sampled)
Regular Regex test x 31,840,454 ops/sec ±3.18% (88 runs sampled)
Regular startswith x 17,677,031 ops/sec ±1.91% (83 runs sampled)
Fastest is Regular Regex test
```

### DIFF
--- a/lib/spans/span-event.js
+++ b/lib/spans/span-event.js
@@ -9,7 +9,7 @@ const Config = require('../config')
 const {truncate} = require('../util/byte-limit')
 
 const {DESTINATIONS} = require('../config/attribute-filter')
-const NAMES = require('../metrics/names')
+
 const HTTP_LIBRARY = 'http'
 const CLIENT_KIND = 'client'
 const CATEGORIES = {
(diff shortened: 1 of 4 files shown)


carlo-808 (approved):
I see we are using regex to test segment name is truncated for span events with validating tests in place. 👍

PR #663, opened 2021-03-10
Adds workflow_dispatch to CI workflows

michaelgoin:
## Details

* Added workflow_dispatch to CI and Smoke Test Workflows
* Fixed Node versions on Smoke Test workflow


carlo-808 (approved):
Good catch! 🦅

PR #668, opened 2021-03-10
Add certificate_bundle feature flag to enable deprecation and removal efforts

michaelgoin:
## Proposed Release Notes

* Added feature flag to allow disabling of certificate bundle usage.

  **Deprecation Warning:** The certificate bundle included by New Relic will be disabled by default and then fully removed in later major versions. We recommend testing with the certificate_bundle feature flag set to `false` to determine if you will need to modify your environment or setup your own appropriate bundle. Example configuration: `feature_flag: { certificate_bundle: false }`

## Links

* Closes: https://github.com/newrelic/node-newrelic/issues/664

### DIFF
--- a/lib/collector/http-agents.js
+++ b/lib/collector/http-agents.js
@@ -103,14 +103,27 @@ function proxyOptions(config) {
     proxy_url: proxy_url
   }
 
-  // merge user certificates with built-in certs
-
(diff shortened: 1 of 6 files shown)


carlo-808 (approved):
Cert bundle deprecation begins 📦

PR #670, opened 2021-03-11
Port New Developer Documentation

michaelgoin:
## Proposed Release Notes

* Added new developer documentation to /docs folder.

  This information is ported over from private GHE wiki used prior to going open source. S/O @astorm for original versions of the function wrapping and module instrumentation docs. 

## Details

Made some tweaks to remove some internal-only deets, make code links work, etc. Hopefully I caught everything or didn't miss anything not true anymore.

Added some additional "how we work" notes to the developer setup doc just to give a high level summary. Would read through that and make sure nothing misleading/confusing there. We may eventually want a more robust writeup in its own doc.

### DIFF
--- a/docs/function-wrapping.md
+++ b/docs/function-wrapping.md
@@ -1,0 +1,284 @@
+# Function Wrapping
+
+New Relic's agents work largely (but not exclusively) via [code instrumentation](https://en.wikipedia.org/wiki/Instrumentation_(computer_programming)). Agents replace function and method definitions with their own (i.e. monkey patch, wrap, etc.). These new functions and methods will call the original functions and methods they're replacing, but also take steps to time that original function or method call and create New Relic data products (metrics, events, traces, segments, etc) as needed and specified.
+
+The Node.js agent has a leg up on other New Relic agents, in that replacing method definitions is a built in feature of the language.
+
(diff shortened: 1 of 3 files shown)


docs/function-wrapping.md
```diff
@@ -0,0 +1,284 @@
...
+
+The Node.js agent has a leg up on other New Relic agents, in that replacing method definitions is a built in feature of the language.
+
+This document will examine the agent's [Shim.wrap](https://github.com/newrelic/node-newrelic/blob/979994fd7250cad4f19ca4fe65a9a1b2d7ca9311/lib/shim/shim.js#L577-L635) method -- this is the method that other wrappers rely on for swapping out a javascript method definition with their own. Once you understand this method, you'll have the core building block you need to understand all the Shim helper methods related to function wrapping.
+
+## Our Test Program
+
+Normally, in order to use a shim helper you need to configure and setup an [instrumentation](./module-instrumentations.md). You can do that if you like, but this document will use the following test program.
```
carlo-808, line 11:
```suggestion
Normally, in order to use a shim helper you need to configure and setup an [instrumentation](./docs/module-instrumentations.md). You can do that if you like, but this document will use the following test program.
```

carlo-808 (changes requested):
Just one link needs fixing.

docs/function-wrapping.md
michaelgoin: Curious why you are suggesting this change? The link works for me testing directly from branch and open the file via this PR. Will it behave different once merged?

carlo-808: Ah, I see what's up, I tried link in preview mode on PR review. Link looks good on second look because referencing file is in same folder. Removing suggestion. 😸

carlo-808 (approved):
All good. 📜  👍

Issue #654, opened 2021-03-06
Remove host from identifier field in collector handshake


carlo-808:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description
We are noticing slowness for MTQW queries from node agent data with a `connectPayload.identifier` value populated. Node agent is generating an identifier based on `nodejs` + `host` + `appName`.
https://github.com/newrelic/node-newrelic/blob/6f631ba9de995a9d3fb2dce4fed3282431c94599/lib/collector/facts.js#L70
The `identifier` field is not required and the backend will generate a usable identifier for the collector from other data sent in handshake.

Alternatively we can remove `host` from the identifier as well.

From TDP: 
>what happens in this particular instance is that we're detecting that the connection is originating from a host with a "dynamic hostname" (that looks like a UUID pattern) and instead of creating a separate agent instance for each connection, we treat them as one - this will mean that the customer loses the ability to see per-host breakdowns, but it's intentional - we do this to protect ourselves from inundating the system with rapidly churning hostnames 
by putting the hostname in the identifier field, you're circumventing that protection

michaelgoin:
It looks like to avoid the customer situation that prompted this change, we'll still need to keep the 'identifier' field. But we do need to remove the host name so the New Relic server won't do bad things and slow down things for customers.

Customer background on problem: https://github.com/newrelic/node-newrelic/commit/c0901e6807a50ac3969d79ab48c31c8e0232a6b5#r18254962

> So, it sounds like the previous behavior was to assume that if there were multiple comma-separated app names, then the first app name given would be unique per-host.
>In our case we were using multiple agents per host and our first app name happened to be the /same/ for all the agents.
>This change creates a new identifier that uses all the app names as part of the unique identifier, not just the first one.

And from our internal server comments...

```
// Note that by using the first app name, and not all of the app names,
// agents cannot connect "a;b" and "a;c" at the same time.
//
// Agents that wish to avoid this limitation should provide an
// 'identifierOverride' that contains all of the app names.
//
```

Chatting internally, it sounds like the host will still be used as a part of the key that is used to determine if two agent connections are the same real agent or a separate one. As such, we should be safe to remove the host but we'll want to keep the full app names / send the identifier to keep from breaking at least the one customer.

See internal slack thread here: https://newrelic.slack.com/archives/C0511B24H/p1614618798048500

PR #671, opened 2021-03-11
Removes host from identifier override sent to connect endpoint

michaelgoin:
## Proposed Release Notes

* Fixed bug where applications with multiple names on a dynamically named host (UUID like) would have instances consolidated, losing per-host breakdowns.

  Removed 'host' from agent 'identifier' override to prevent server safety mechanism from kicking in. Host will still be used to identify unique agent instances, so was unnecessary to include as part of the identifier. This also resulted in additional processing overhead on the back-end. The identifier override is still kept in place with multiple application names to continue to allow uniquely identifying instances on the same host with multiple application names where the first name may be identical. For example `app_name['myName', 'unique1']` and `app_name['myName', 'unique2']`. These names would consolidate down into a single instance on the same host without the identifier override.

## Links

* Fixes: https://github.com/newrelic/node-newrelic/issues/654
* https://github.com/newrelic/node-newrelic/commit/c0901e6807a50ac3969d79ab48c31c8e0232a6b5#r18254962
* https://newrelic.slack.com/archives/C0511B24H/p1614618798048500

## Details

While researching the various pieces involved, I found the default identifier also uses the 'dispatcher' from the environment data instead of 'nodejs'. For a web app, this would be 'http'. 

So no custom identifier would look like 'http:MyExpressApp' where our override will look like: 'nodejs:MyExpressApp'.

Given we don't know of any additional issues, I'm going to leave our deviation for now to limit what changes and (hopefully) reduce surface area of issues resulting from this change.

Going to flag this as a risk: medium given we are changing an area that was implemented due to a previous issue. All signs seem to point toward this continuing to solve the problem but changing potential instance identification mechanisms is more dangerous than usual and we should be aware.

See the linked GitHub issue for further details.

### DIFF
--- a/lib/collector/facts.js
+++ b/lib/collector/facts.js
@@ -64,14 +64,7 @@ function facts(agent, callback) {
       }
     }
 
-    // TODO:  After reconfiguring agent startup to wait for the server to start
-    //        or for the first transaction, add the `port` for the server too.
-    // NOTE: The concat is necessary to prevent sort from happening in-place.
-    results.identifier = [
-      'nodejs',
-      results.host,
-      results.app_name.concat([]).sort().join(',')
-    ].join(':')
+    results.identifier = getIdentifierOverride(results.app_name)
 
     const ipAddresses = getAllIPAddresses()
     if (ipAddresses.length) {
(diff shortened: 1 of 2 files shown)


carlo-808 (approved):
Good comments around the change. I see we are removing host from the identifier property we send to collector to alleviate backend slowness. 🤞

carlo-808:
This PR was merged in and it appears that a github bug has held the PR open. Leaving open for now to see if resolved by github.

PR #673, opened 2021-03-12
Consolidate tagging and deployment into single workflow

michaelgoin:
## Details

Modifies create release tag workflow to just be a 'create release' workflow. Tags, publishes and creates a github release.

Removes tag triggering & github release creation from publish workflow to allow for non-standard manual deploys on branches.


carlo-808 (approved):
Create Release workflow now creates release tags, publishes to npm and creates github release off tags. Added publish branch workflow to manually npm publish

PR #674, opened 2021-03-12
Misc. Cleanup

michaelgoin:
## Details

Cleaned up a few things I saw looking at something else.

* Clean script was deleting docs and package-lock.
  * package-lock was legacy from when not included, so removed that clear
  * docs was because of our api docs generation script thing. instead of reworking that just renamed the /docs folder to /documentation
* There was an errant import VS code had thrown in there that was missed.
* Needed copyright headers on the new scripts.


carlo-808 (approved):
🧹

PR #675, opened 2021-03-12
Adds linting to scripts in bin folder and cleans up existing.

michaelgoin:
## Proposed Release Notes

* Added linting to scripts in /bin folder.

  Linting rules added are slightly more permissive than production rules and allow full ecma 8.


carlo-808 (approved):
🌬️ linter 🧹

PR #676, opened 2021-03-12
Removes major release check from prepare script

michaelgoin:
…review.

## Details

Given the prepare script generates a PR for review, we don't need the extra safety-check up front like we did when we were tagging and pushing right after setting the version.

The additional human eyes are a better safety mechanism.


carlo-808 (approved):
Makes sense 💭  👍

Issue #627, opened 2021-02-23
Remove requirement to have a config file unless a special ENV var set


michaelgoin:
## Is your feature request related to a problem? Please describe.

Using ENV vars is pretty typically the standard for Docker, K8s, Lambda, etc. Currently, we require an ENV var to be set to allow only using ENV var w/o a config file. While not a big barrier, it is not always obvious this needs to be set until failure on run. While I'm sure this was in place as some sort of early safeguard, I think the ENV case is common enough we should work to remove this barrier. I also believe this is unique to the Node agent.

Instead, let's ensure ENV + config precedence is working correctly and remove the requirement on this ENV var.

PR #679, opened 2021-03-17
The NEW_RELIC_NO_CONFIG_FILE env var is now not needed to run the agent without config file

carlo-808:
## Proposed Release Notes
* The `NEW_RELIC_NO_CONFIG_FILE` environment variable is no longer needed to run the agent without a configuration file.
  * If a configuration file is used with agent configuration environment variables, the environment variables will override the corresponding configuration file settings.
## Links
Closes: #627 
## Details
There are a few environment variables that do not have a corresponding configuration file setting, but those are set by the environment they run in (ie AWS Lambda).

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -1635,6 +1635,8 @@ function initialize(config) {
 
   var filepath = _findConfigFile()
   if (!filepath) {
+    logger.info('No configuration file found, deferring to environment variables.')
+
     config = new Config(Object.create(null))
     if (config.newrelic_home) delete config.newrelic_home
     return config
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -1616,11 +1616,9 @@ Config.prototype.getAggregatorConfig = function getAggregatorConfig(method) {
  *   5. If this module is installed as a dependency, the directory above the
  *      node_modules folder in which newrelic is installed.
  *
- * For configuration via environment (useful on Joyent, Azure, Heroku, or
- * other PaaS offerings), set NEW_RELIC_NO_CONFIG_FILE to something truthy
- * and read README.md for details on what configuration variables are
- * necessary, as well as a complete enumeration of the other available
- * variables.
+ * When node process environment varaibles and a config file are used,
+ * the environment variables will override their corresponding
+ * configuration file settings.
  *
  * @param {object} config Optional configuration to be used in place of a
  *                        config file.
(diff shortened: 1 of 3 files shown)


michaelgoin (dismissed):
👍 Looks good.

A couple things before mashing that merge button...

If you haven't already, I'd recommend testing with a config file that exists but is not readable in case it behaves any differently (throws for example) and make sure we handle that situation OK. As we'll now be attempting to load the config file regardless, when we would not before.

On that note, we've completed removed the variable. Just stating out loud we could have (or could still) keep it for preventing even attempting to load the file. In theory, attempting to load that file should not be expensive... so guessing not critical.

And to these points, there is a potential change in behavior where customers with the ENV var set, that also had a local config file, may get some settings from that local config file they were not before. Seems pretty low risk but it is a possibility.

carlo-808:
Good point @michaelgoin , I'll test out the case you mentioned regarding an unreadable config file.

> On that note, we've completed removed the variable. Just stating out loud we could have (or could still) keep it for preventing even attempting to load the file. In theory, attempting to load that file should not be expensive... so guessing not critical.

Maybe I should leave the variable `NEW_RELIC_NO_CONFIG_FILE` read in there and log something about it being ignored?

> And to these points, there is a potential change in behavior where customers with the ENV var set, that also had a local config file, may get some settings from that local config file they were not before. Seems pretty low risk but it is a possibility.

I could add more information to the log message to this point. What do you think?

carlo-808:
Good call, I'm going to bring the NEW_RELIC_NO_CONFIG_FILE check back and have it only enforce no config file. 👍

carlo-808:
@michaelgoin It looks like we handle the error from a bad config file gracefully and disable the agent even if environment variables are set, so I think we are good there.
Error message from stdout is:
```
New Relic for Node.js is disabled due to an error:
Error: Unable to read configuration file "/location/of/newrelic.js".
```

michaelgoin:
@carlo-808 that actually seems bad to me that we disable the agent given we are trying to move away from it being required if you don't set a special ENV var.

carlo-808:
Do you think we should just log a warning that the config file is unreadable in this case and try to start the agent?

michaelgoin:
@carlo-808 yeah, i think logging a warning that we were unable to read the config file and that we are attempting to fall back to ENV vars.

michaelgoin (approved):
👍 looks good.

PR #700, opened 2021-03-17
Adds module root to shim.require() logging to make issues more clear.

michaelgoin:
## Proposed Release Notes

* Added module root to shim.require() logging to aid debugging.


carlo-808 (approved):
We are now logging the module root when an error occurs during shimRequire.
⚡

PR #703, opened 2021-03-18
Update warning log message when cannot find config or bad config file

carlo-808:
## Details
Update log info/warn message when config file is not found vs unreadable.


michaelgoin (approved):
👍

PR #705, opened 2021-03-18
Change prepare workflow working to allow major releases.

michaelgoin:
## Details

Missed updating the workflow in the prepare-script update. Updates wording to not flag major as unsupported (but still somewhat show not standard i.e. for breaking changes.)


carlo-808 (approved):
👍

Issue #704, opened 2021-03-18
Update configuration template to enable Distributed Tracing


coreyarnold:
## Is your feature request related to a problem? Please describe.
New customers may use the configuration template to start their instrumentation. Currently Distributed Tracing (DT) is not enabled through that configuration and the customer will have explicitly choose to update the configuration in the future to enable it.

## Feature Description
Update the configuration template to enable DT

## Describe Alternatives
leave as is

## Additional context
DT is a feature many customers find value in, but is not enabled by default. Customers can still choose to turn the feature off, but we should lead with the example config showing it turned on.

## Priority
Really Want

PR #706, opened 2021-03-18
Distributed tracing is set to enabled in newrelic.js template configuration file

carlo-808:
## Proposed Release Notes
* Distributed tracing is set to enabled in the `newrelic.js` template configuration file supplied with the agent.
## Links
Closes: #704

### DIFF
--- a/test/unit/api/api-get-linking-metadata.test.js
+++ b/test/unit/api/api-get-linking-metadata.test.js
@@ -48,6 +48,8 @@ tap.test('Agent API - getLinkingMetadata', (t) => {
   })
 
   t.test('should return available fields, no DT data, when DT enabled - no transaction', (t) => {
+    agent.config.distributed_tracing.enabled = true
+
     const metadata = api.getLinkingMetadata()
 
     // Trace and span id are omitted when there is no active transaction
--- a/test/unit/api/api-get-linking-metadata.test.js
+++ b/test/unit/api/api-get-linking-metadata.test.js
@@ -16,12 +16,7 @@ tap.test('Agent API - getLinkingMetadata', (t) => {
   let api = null
 
   t.beforeEach((done) => {
-    // implicitly disabling distributed tracing to match original config base settings
-    agent = helper.instrumentMockedAgent({
-      distributed_tracing: {
-        enabled: false
-      }
-    })
+    agent = helper.instrumentMockedAgent()
     api = new API(agent)
 
     done()
@@ -34,20 +29,9 @@ tap.test('Agent API - getLinkingMetadata', (t) => {
     done()
   })
 
-  t.test('should return available fields, no DT data, when DT disabled - no transaction', (t) => {
-    const metadata = api.getLinkingMetadata()
-
-    t.notOk(metadata['trace.id'])
-    t.notOk(metadata['span.id'])
-    t.equal(metadata['entity.name'], 'New Relic for Node.js tests')
-    t.equal(metadata['entity.type'], 'SERVICE')
-    t.notOk(metadata['entity.guid'])
-    t.equal(metadata.hostname, agent.config.getHostnameSafe())
-
-    t.end()
-  })
-
   t.test('should return available fields, no DT data, when DT disabled in transaction', (t) => {
+    agent.config.distributed_tracing.enabled = false
+
     helper.runInTransaction(agent, function() {
       const metadata = api.getLinkingMetadata()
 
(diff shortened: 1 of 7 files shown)


test/unit/api/api-get-linking-metadata.test.js
```diff
@@ -16,7 +16,12 @@ tap.test('Agent API - getLinkingMetadata', (t) => {
   let api = null
 
   t.beforeEach((done) => {
-    agent = helper.loadMockedAgent()
+    // implicitly disabling distributed tracing to match original config base settings
+    agent = helper.instrumentMockedAgent({
+      distributed_tracing: {
+        enabled: false
+      }
+    })
```
michaelgoin, line 24: Minor: but in a few months, etc. we won't understand what "original config base settings" were and why they would matter here.

I might have left as-is and updated the individual test cases to disable DT when they wanted DT disabled. Whether per test or special setup.

test/unit/instrumentation/http/no-parallel/queue-time.test.js
```diff
@@ -22,7 +22,12 @@ tap.test('built-in http queueTime', (t) => {
   let THRESHOLD = null
 
   t.beforeEach((done) => {
-    agent = helper.instrumentMockedAgent()
+    // implicitly disabling distributed tracing to match original config base settings
+    agent = helper.instrumentMockedAgent({
+      distributed_tracing: {
+        enabled: false
+      }
+    })
```
michaelgoin, line 30: I'm kinda surprised this broke something here... unless it is testing something Cross Application Tracing related in here? Although, I'm not sure what that has to do with queue time?

Curious to learn more here.

Assuming there's a valid reason and not just an unnecessary assertion in there... minor, but I don't think you need that comment about 'original config base settings.' It won't mean anything to anyone in the future as the significance of actually why it was disabled is not clear.

michaelgoin (dismissed):
Looks good / is straightforward enough.

I didn't go dig through each one but general concerns around why a couple of the tests even need this set. Beyond that, some minor comments on the comment referencing 'original config base settings' not likely being all that useful long term.

Approving.

test/unit/instrumentation/http/no-parallel/queue-time.test.js
carlo-808: I'll go ahead and take a quick pass and rework the tests you mentioned above and remove the comments on the test setup.

test/unit/api/api-get-linking-metadata.test.js
```diff
@@ -29,20 +29,9 @@ tap.test('Agent API - getLinkingMetadata', (t) => {
     done()
   })
 
-  t.test('should return available fields, no DT data, when DT disabled - no transaction', (t) => {
```
carlo-808, line 32: removed because duplicate of test at line 50

test/unit/instrumentation/http/no-parallel/queue-time.test.js
carlo-808: You're right, enabling DT did not affect this test. Must have updated on accident.

test/unit/api/api-get-linking-metadata.test.js
```diff
@@ -59,8 +48,6 @@ tap.test('Agent API - getLinkingMetadata', (t) => {
   })
 
   t.test('should return available fields, no DT data, when DT enabled - no transaction', (t) => {
-    agent.config.distributed_tracing.enabled = true
```
michaelgoin, line 62: Minor: Given test is for DT enabled specifically, could have left explicit set in here which would avoid issues if the default ever changed again and makes test setup obvious.

test/unit/metrics-recorder/http.test.js
```diff
@@ -383,6 +393,18 @@ describe("recordWeb", function() {
...
+        ],
+        [{name: 'ErrorsByCaller/Unknown/Unknown/Unknown/Unknown/all'},
+          [1, 0.001, 0.001, 0.001, 0.001, 0.000001]
+        ],
+        [{name: 'DurationByCaller/Unknown/Unknown/Unknown/Unknown/allWeb'},
+          [1, 0.001, 0.001, 0.001, 0.001, 0.000001]
+        ],
+        [{name: 'ErrorsByCaller/Unknown/Unknown/Unknown/Unknown/allWeb'},
+          [1, 0.001, 0.001, 0.001, 0.001, 0.000001]
+        ],
```
michaelgoin, line 407: Minor: Test has nothing specific to do with DT so feels a bit superfluous to add these here (not the intent of the test).

michaelgoin (approved):
Looks good. 

A couple minor comments to keep in mind with future changes.

Approving.

michaelgoin (approved):
👍

Issue #710, opened 2021-03-22
Agent introduces memory leak when axios errors noticed


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

When axios errors are noticed by the agent (at least in some situations) the agent introduces a memory leak.

Axios errors store the request on their error objects for situations such as a 500 status code returned. These request objects can have a promise handle held by `handleRequestError`.

In these situations, the agent has a TraceSegment stored in the segment map used to keep track of state for promises via async-hooks. A transaction will have a reference to the promise via the error it has stored, keeping the promise from getting GC'd and the resulting 'destroy' event for async-hooks. Since 'destroy' is never called, the TraceSegment is never freed to release the transaction or segment tree. As such, all of these items are held indefinitely in memory.

## Expected Behavior

We should allow promises (and all memory) to free up in these situations

## Steps to Reproduce

1. Call an endpoint that returns a 500 error.
2. Ensure error captured via `noticeError` (or potentially by not handling at all).

```js
app.get('/error', function middlewareOne(req, res, next) {
  makeAxiosRequest('http://localhost:3000/error').then((result) => {
    console.log('.then')
    console.log(result.data)
    res.send('we did it!')
  }).catch((err) => {
    nr.noticeError(err) // This results in the memory leak
    console.log('.error')
    console.error(err.message)
    res.status(500)
    res.send()
  })
})

function makeAxiosRequest(url) {
  console.log('making axios.request return promise')
  return axios.request(url)
}
```

## Additional context

* https://newrelic.atlassian.net/browse/GTSE-7862
* https://newrelic.zendesk.com/agent/tickets/447727

I believe our usage of 'destroy' as a means of cleaning up promise references is problematic. Not just from memory-leak potential but for increased latency cleaning up memory in customer apps. I've been pondering this for a bit but this issue really nails it home. That being said, I'm guessing there's legacy reasons this was used and migrating to something like resolvePromise may be have a higher risk (unless resolvePromise just didn't exist at the beginning? will have to research)

To get a quick resolution out there, we can like likely see about extract error information faster to avoid holding onto unwanted objects when capturing errors. That should solve this with very minimal risk.  (Obv up to implementer to decide best path and not inclusive of all ideas).

Then we can start exploring bigger ideas such as evolving our async-hooks implementation. I'll be entering a GitHub issue specifically around our promise tracking implementation.

PR #712, opened 2021-03-23
Clears error collections on transaction end after processing.

michaelgoin:
## Proposed Release Notes

* Fixed issue where capturing axios request errors could result in a memory leak.

  The agent now clears error references on transaction end, which are not used for later processing. Errors returned from 'axios' requests contain a reference to the request object which deeper down has a handle to a promise in `handleRequestError`. The TraceSegment associated with that promise has a handle to the transaction, which through the error capture ultimately kept the promise in memory and prevented it from being destroyed to free-up the TraceSegment from the segment map. This change also has the benefit of  freeing up some memory early for transactions held onto for transaction traces.

* Added active transaction check to `wrappedResEnd` to prevent unecessary work for ended transactions in the case of multiple `Response.prototype.end()` invocations.

## Links

* Closes: https://github.com/newrelic/node-newrelic/issues/710

## Details

Ideally, I'd like to avoid holding onto user errors passed-in at all. Unfortunately, this requires touching several place and doesn't play well with our duplicate error detection. The duplicate detection uses the object itself as a key to avoid situations such as an end user calling `noticeError(error)` on the same instance twice on accident. 

It may be worth considering if that safety mechanism isn't something we want/need to keep around. I'm not sure if any other agents do similar but I don't remember anything like that from my previous team. 

I'm going to submit a separate issue to dig into this further.

While this change has the benefit of also just freeing up some memory sooner... there is some risk in clearing things off the transaction as we might in the future attempt to do something with this information later after transaction end. Unlikely at this point but calling out. More likely would be haphazardly adding something that is used by transaction traces in here to clear out as well.

### DIFF
--- a/lib/instrumentation/core/http.js
+++ b/lib/instrumentation/core/http.js
@@ -267,6 +267,11 @@ function wrapResponseEnd(agent, proto) {
         return end.apply(this, arguments)
       }
 
+      if (!txInfo.transaction.isActive()) {
+        logger.trace('wrappedResEnd invoked for ended transaction implying multiple invocations.')
+        return end.apply(this, arguments)
+      }
+
       // If an error happened, add it to the aggregator.
       if (txInfo.error) {
         if (!txInfo.errorHandled || urltils.isError(agent.config, this.statusCode)) {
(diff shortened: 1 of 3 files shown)


carlo-808 (approved):
Looks good, we are clearing transaction user errors and exceptions on transaction end. 👍

PR #718, opened 2021-03-26
Adds active transaction check for parentSegment to promise tracking

michaelgoin:
## Proposed Release Notes

* Agent no longer propagates segments for promises via async-hooks when the transaction associated with the parentSegment has ended.

  This change reduces the amount of context tracking work needed for certain rare edge-case scenarios involving promises.

## Details

Also moved type check up earlier for faster short-circuiting (as well as just checking agent.getTransaction() in the if statement). These are unlikely to have a noticeable impact but given the frequency of execution of the hooks and added check, seemed good to go ahead and do.

### DIFF
--- a/test/integration/core/async_hooks.js
+++ b/test/integration/core/async_hooks.js
@@ -211,6 +211,30 @@ test("the agent's async hook", function(t) {
     })
   })
 
+  t.test('stops propagation on transaction end', function(t) {
+    const agent = setupAgent(t)
+
+    helper.runInTransaction(agent, function(txn) {
+      t.ok(txn, 'transaction should not be null')
+      const segment = txn.trace.root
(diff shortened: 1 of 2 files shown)


test/integration/core/async_hooks.js
```diff
@@ -211,6 +211,30 @@ test("the agent's async hook", function(t) {
...
+      function one() {
+        return new Promise((done) => {
+          const currentSegment = agent.tracer.segment
+          t.ok(currentSegment, 'should have propagated a segment')
+          txn.end()
+
+          done()
+        }).then(() => {
+          const currentSegment = agent.tracer.segment
+          t.notOk(currentSegment, 'should not have a propagated segment')
```
carlo-808, line 231: nice test

carlo-808 (approved):
👍

Issue #725, opened 2021-03-31
Infinite tracing memory leak on reconnects


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

Each time the server tells our gRPC/infinite tracing connection to reconnect, we leak a few grpc-js objects.

It looks like ChannelImplementation (and some child objects) get created when we create a new service from a service definition. These seem to be held in memory by timers indefinitely, even after we've closed the related stream and let the client go out of memory.

The good news is we do not need to be creating a new client each time. We can hold onto the client instance and just call `recordSpan` on it. This appears to resolve the issue for both idle reconnects and the active every 30-second OK status reconnects.

## Steps to Reproduce

The easiest way to detect (which is probably why we have not noticed prior) is to start a service with infinite tracing and then do nothing. If you capture heap snapshots every 30 seconds or so, you'll roughly see ChannelImplementation and related child objects (DnsResolver, promises, etc.) grow by 1 each snapshot. This is because the server has started sending idle disconnects every 20 seconds (if I remember right). So we will then reconnect after about 15 seconds.

PR #726, opened 2021-03-31
Fixes infinite tracing reconnect memory leak

michaelgoin:
## Proposed Release Notes

* Fixed memory leak introduced when Infinite Tracing is enabled.

  When Infinite Tracing endpoints reconnected they would instantiate a new gRPC client prior to calling `client.recordSpan()`. It appears several objects created by grpc-js (`ChannelImplementation` and child objects, promises, etc.) are held in memory indefinitely due to scheduled timers even when the client is no-longer referenced and the associated stream closed. We now avoid this situation by only creating the client once and then reusing it to establish new stream connections. 

## Links

* Fixes: https://github.com/newrelic/node-newrelic/issues/725

## Details

I could think of any meaningful tests that would catch/prevent reintroducing the issue in a meaningful way. So relying on code comments for now.

Fixed a couple tests that when they would throw would impact downstream tests due to not reverting stubs.

Also added some comments of places to remove checks when we drop Node 10 support later this year.

### DIFF
--- a/lib/grpc/connection.js
+++ b/lib/grpc/connection.js
@@ -14,14 +14,14 @@ const util = require('util')
 
 const connectionStates = require('./connection/states')
 
-const protoOptions = {keepCase: true,
+const PROTO_OPTIONS = {keepCase: true,
   longs: String,
   enums: String,
   defaults: true,
   oneofs: true
 }
 
-const pathProtoDefinition = __dirname +
+const PROTO_DEFINITION_PATH = __dirname +
   '../../../lib/grpc/endpoints/infinite-tracing/v1.proto'
 
 const DEFAULT_RECONNECT_DELAY_MS = 15 * 1000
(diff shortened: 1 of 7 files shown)


lib/grpc/connection.js
```diff
@@ -320,26 +322,48 @@ class GrpcConnection extends EventEmitter {
...
+   * ChannelImplementation and related objects will stay in memory even after
+   * the stream is closed and we do not have a handle to the client. Currently
+   * impacting grpc-js@1.2.11 and several earlier versions.
+   */
+  _createClient(endpoint) {
+    logger.trace('Creating gRPC client for: ', endpoint)
+
+    const packageDefinition = protoLoader.loadSync(PROTO_DEFINITION_PATH, PROTO_OPTIONS)
+
+    const protoDescriptor = grpc.loadPackageDefinition(
```
carlo-808, line 353: nice change with better var name

carlo-808 (approved):
Looks like we don't create an new client everytime we call client.recordSpan(). 👍

Issue #711, opened 2021-03-22
Look into evolving async-hooks promise tracking away from 'destroy' event


michaelgoin:
## Is your feature request related to a problem? Please describe.

The 'destroy' event is not fired until a promise is GC'd. That means we'll never free the associated segment, it's associated transaction and full segment tree, until a promise has been fully cleaned-up.

This can result in keeping a large amount of objects in memory for a longer-time than necessary. Both increasing overhead and also giving the illusion of memory leaks in certain situations.

Further, in cases where we accidentally end up with a handle on a promise we can introduce a memory leak since the transaction will not go out of memory until the promise is cleaned-up.

For example: https://github.com/newrelic/node-newrelic/issues/710

Ideally, I think we could use the `promiseResolve` event for promises. 

There may be legacy reasons this was not used that could break existing assumptions (or perhaps this event just didn't exist at the time of implementation). Some research needs to be done as to how the agent behaves.

Regardless, this will be slightly higher risk and may warrant a feature-flag and testing with some willing clients or <insert mitigation strategy here>. Unless we find our confidence is super high.

PR #727, opened 2021-04-05
feature-flag: clean up segment references on promise resolve

michaelgoin:
## Proposed Release Notes

* Added new feature-flag 'new_promise_tracking' which enables cleaning up of segment references on native promise resolve instead of destroy. Includes usage of async-await. This can be enabled via `feature_flag: { new_promise_tracking: true }` in the config file or `NEW_RELIC_FEATURE_FLAG_NEW_PROMISE_TRACKING=1` in your ENV vars.

  Applications with heavy promise usage or high-throughput applications with some promise usage should see moderate to high reduction in memory usage and may see a slight reduction in CPU usage. A bump in throughput may also be noticed in some cases. Results will vary by application.
  
  If you are experiencing high overhead levels with your promise usage and the agent attached, we recommend testing your application with  'new_promise_tracking' set to true to see if overhead is reduced. You'll also want to verify your data is still being captured correctly in case it falls into a known or unknown limitation of this approach.  **NOTE: chaining of promise continuations onto an already resolved promise across an async hop (scheduled timer) will result in state-loss with this new functionality turned on. This is a less-common use-case but worth considering with your applications.**

## Links

* Closes: https://github.com/newrelic/node-newrelic/issues/711

## Details

The 'destroy' event for promises is not invoked until a promise has been garbage collected. This means customers holding onto promises, the agent indirectly or other means may keep our segments in the segmentMap for an extra long amount of time. Even when references are quickly removed, we wait for the next GC (not necessarily immediate) to remove references our segment tree/transaction/etc. so they can be cleaned up.

When the feature flag is enabled, we now remove segments from the segment map when they are resolved. While customer's could still trigger us to leak by not resolving a promise, this greatly speeds up our clean-up of transaction related objects from memory and reduces our chance to grow an existing memory leak or introduce a new leak via indirect reference.

Unfortunately, since we will have cleaned out a promise on resolve... if you schedule a continuation on a resolved promise, it won't be able to pull up the parent segment associated with the resolved parent promise. As such, if you've made a hop w/o state you may lose transaction state. Theoretically, you could also end up with a different parent than the current racking, in this scenario, if a segment is in context but different than the parent would have been.

This seem pretty niche and may be best solved via targeted instrumentation VS trying to solve the world via async-hooks.

Also, if we were all in on async-hooks. Especially, using the executionAsyncResource now available... we may support more of this style of scenario while maybe even reducing the amount of hooks (and thus overhead) needed. That's a bigger effort as some state consolidation refactoring is likely needed first, etc.

This hopefully is a step towards seeing some wins in some of our existing scenarios. If we see consistent enough success without unknown limitations that are deemed critical, we can look to make this the default functionality in a future major release. 

-----

It's worth noting I tried to optimize for ripping code out in the future for a single implementation, so some opportunities for code re-use were not leveraged as they would have hindered the cleanup instead of help.

-----

### Stress Test

Here are some stress test numbers from my local machine. These are in no-way 'official' nor do they represent a particular customer's real-world scenario. They should be a good indicator for potential wins, however.

* koa
* koa-router
* axios (external calls)
* async-await

Multiple middleware doing nothing 
|          | Response Time | Throughput | CPU Utilization | Memory Usage |
|----------|---------------|------------|-----------------|--------------|
| Original | 0.171 ms      | 117k rpm   | 88.45%          | 119.7 MB     |
| New      | 0.155 ms      | 124k rpm   | 84.45%          | 114.5 MB     |

Multiple middleware, small timer delays 
|          | Response Time | Throughput | CPU Utilization | Memory Usage |
|----------|---------------|------------|-----------------|--------------|
| Original | 3.59 ms       | 78.3k rpm   | 76.13%          | 154.5 MB    |
| New      | 3.26 ms       | 88.7k rpm   | 73.13%          | 114.2 MB    |

Multiple middleware, external call 
|          | Response Time | Throughput | CPU Utilization | Memory Usage |
|----------|---------------|------------|-----------------|--------------|
| Original | 30.7 ms       | 11.6k rpm   | 36.49%         | 984.2 MB     |
| New      | 28.6 ms       | 12.5k rpm   | 35.93%         | 105.7 MB     |

Multiple middleware, external call that errors 
|          | Response Time | Throughput | CPU Utilization | Memory Usage |
|----------|---------------|------------|-----------------|--------------|
| Original | 30.7 ms       | 11.6k rpm   | 42.05%         | 1.1 GB       |
| New      | 30.4 ms       | 11.7k rpm   | 39.17%         | 113.4 MB     |

### DIFF
--- a/test/integration/core/async_hooks-new-promise.js
+++ b/test/integration/core/async_hooks-new-promise.js
@@ -1,0 +1,566 @@
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const test = require('tap').test
+const helper = require('../../lib/agent_helper')
+const asyncHooks = require('async_hooks')
+
+test('await', function(t) {
+  const agent = setupAgent(t)
+
+  helper.runInTransaction(agent, async function(txn) {
+    let transaction = agent.getTransaction()
+    t.equal(
+      transaction && transaction.id,
+      txn.id,
+      'should start in a transaction'
+    )
+
+    // await Promise.resolve("i'll be back")
+
+    const segmentMap = require('../../../lib/instrumentation/core/async_hooks').segmentMap
+
+    const promise = new Promise((resolve) => {
+      // don't immediately resolve so logic can kick in.
+      setImmediate(resolve)
+    })
+
+    // There may be extra promises in play
+    const promiseId = [...segmentMap.keys()].pop()
+
+    await promise
+
+    t.notOk(segmentMap.has(promiseId), 'should have removed segment for promise after resolve')
+
+    transaction = agent.getTransaction()
+    t.equal(
+      transaction && transaction.id,
+      txn.id,
+      'should resume in the same transaction after await'
+    )
+
+    txn.end()
+
+    // Let the loop iterate to clear the microtask queue
+    setImmediate(() => {
+      t.equal(segmentMap.size, 0, 'should clear segments after all promises resolved')
+      t.end()
+    })
+  })
+})
+
+test("the agent's async hook", function(t) {
+  class TestResource extends asyncHooks.AsyncResource {
+    constructor(id) {
+      super('PROMISE', id)
+    }
+
+    doStuff(callback) {
+      process.nextTick(() => {
+        if (this.runInAsyncScope) {
+          this.runInAsyncScope(callback)
+        } else {
+          this.emitBefore()
+          callback()
+          this.emitAfter()
+        }
+      })
+    }
+  }
+
+  t.autoend()
+  t.test('does not crash on multiple resolve calls', function(t) {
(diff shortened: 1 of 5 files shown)


test/integration/core/async_hooks-new-promise.js
```diff
@@ -0,0 +1,566 @@
...
+   * This specific use case is not currently supported with the implementation that clears
+   * segment references on promise resolve.
+   */
+  t.test(
+    'parent promises that are already resolved DO NOT persist to continuations ' +
+    'scheduled after a timer async hop.',
+    function(t) {
+      const agent = setupAgent(t)
+      const tasks = []
+      const intervalId = setInterval(() => {
```
carlo-808, line 142: interesting cleanup code

carlo-808 (approved):
👍  We now have a feature flag to enable removing segment references on promise resolve rather than the promise destroy async hook. 💥

Issue #729, opened 2021-04-07
Wrong APM data after enabling new promise tracking


artur-ma:
Looks like the agent sends wrong APM metrics after enabling new promise tracking

## Description

After enabling the feature NEW_RELIC_FEATURE_FLAG_NEW_PROMISE_TRACKING
we see a drop in throughput from 1.5k RPM to ~10 RPM
which is wrong since the overall system stayed with the same throughput 
we are not doing anything special with the "post resolve" on the promise

## Your Environment

Node 14
inside k8s pod
**Fastify** as web framework

michaelgoin:
@artur-ma thank you for trying the feature-flag so quickly and sorry for the troubles.

Unfortunately, a change like this can change can have a bit of a cascading effect due to a change in assumptions. We tried to test out several scenarios but looks like we missed some, which we knew was a bit of a risk. Hopefully, we can track down this scenario and find a straightforward resolution.

1.5k to 10 is a huge drop. I wonder if there ended up being some state conflation in this scenario where somehow the same transactions were used over and over again.

Anything you can share about how you are setup using fastify and your promise usage would be great to help us reproduce. Particularly, code samples.

If you are willing to capture trace-level logs, that would also be extremely helpful. In that case, I'd recommend opening a support ticket and sharing via the ticket as the level of detail we capture in trace logs can expose details such as license key, route params, etc. that wouldn't be good to have publicly visible.

artur-ma:
> 1.5k to 10 is a huge drop. I wonder if there ended up being some state conflation in this scenario where somehow the same transactions were used over and over again.

Just to clarify, there was no real drop in throughput, 
I also ran autocannon, to see if it grows, but it didn't the throughput in autocannon was about 1k requests/min, but on APM dashboard the values didn't change

So actually there was no degradation in performance, only metric report.

> Anything you can share about how you are setup using fastify and your promise usage would be great to help us reproduce. Particularly, code samples.

I will try to have some reproducible example for you, basically we don't have 
any plugin on our fastify except fastify-compress 

Each request:
* Gets the data from redisDB (using ioredis) and returns it as a response 
* Sends Kafka message(KafkaJs) withing non-awaited promise

michaelgoin:
@artur-ma The latest release v7.3.1 *may* resolve your issue. Don't have enough details to know for sure but seems similar based on what you have shared. Would be worth testing out again.

artur-ma:
[...]

PR #731, opened 2021-04-13
Only propagates null segments on resolve to prevent reuse of transaction

michaelgoin:
## Proposed Release Notes

* Fixed issue with 'new_promise_tracking' feature flag functionality where segments for ended transactions would get propagated in certain cases by promises that had no continuations scheduled (via await or manually).

## Links

* Likely same problem as: https://github.com/newrelic/node-newrelic/issues/729

## Details

~While the code to reproduce this locally is pretty straightforward, I cannot get a test to reproduce this. I'm wondering if the underlying test framework implementation is having an impact. I'll continue to poke at it but I figured it was better get something out there to resolve than hold-off given this is being a feature-flag anyways. I'm hoping this resolves the one issue we've seen (linked above) but it may not as haven't seen all the details there yet.~

Was able to get tests going by using an existing framework (fastify in this case). Koa would have worked as well, but that is in an external module.

### DIFF
--- a/lib/instrumentation/core/async_hooks.js
+++ b/lib/instrumentation/core/async_hooks.js
@@ -185,7 +185,12 @@ function getPromiseResolveStyleHooks(segmentMap, agent, shim) {
         return
       }
 
-      shim.setActiveSegment(hookSegment)
+      // Because the ID will no-longer be in memory until dispose to propagate the null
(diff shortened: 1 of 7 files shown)


carlo-808 (approved):
👍

PR #733, opened 2021-04-14
Filters out PR that created release tag commit on merge.

michaelgoin:
## Details

In some cases, appears the commit date can be earlier than the same PR's merge date which resulted in the PR coming back for the 'merged after' query. This filters that out.


carlo-808 (approved):
👍

Issue #344, opened 2020-05-13
How to use newrelic with esm "import" syntax?


AdityaAnand1:
https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_enabling

When enabling esm mode, either import is recognized or require is recognized. Both are not recognized together, so is there a way to the import syntax instead of require? (I'm unwilling to convert my entire app back to the require format just for this)

I'd like to use this:
```
import 'newrelic.js';
```

I tried this:

```
import 'newrelic.cjs';
```

(explicitly making it cjs = use common module for one file)

Which currently gives an error:
```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: C:\Projects\backend\src\bin\www.js
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1163:13)
```

carlonewrelic:
Hi @AdityaAnand1 

You should be able to import newrelic this way

`import newrelic from 'newrelic'`

The newrelic.js config file will not load properly in esm mode.  You will need to setup the agent configuration with environment variables.  That being said, we currently do not support auto-instrumenting modules loaded via esm mode.  We do plan to add support in the future.  If you would like to help prioritize this work you can submit a feature request through [New Relic's explorer hub](https://discuss.newrelic.com/c/feature-ideas) or by [submitting a support ticket](https://support.newrelic.com/).

AdityaAnand1:
Thank you, can you provide the exact instructions for setting up this way? Or maybe a link to the docs?

1. Which environment variables are required for minimal setup? (The docs give the whole list but will it work with just app_name + license key environment variables?)
2. Is there a way to enable to debug logging for newrelic? Right now, I get no visibility whether or not it is working or not? (Something I can set to `.isDebugMode(true)` on my local environment so that I know that it'll work when I push it to prod

AdityaAnand1:
Instead of 

```
import newrelic from 'newrelic'
```

the following

```
import 'newrelic'
```

should work too, right? (unless the newrelic import needs to be used in some way)

AdityaAnand1:
I tried the environment variable route, however it doesn't seem to be detecting them (In the following image I print the env variable right after newrelic says it doesn't exist)

![image](https://user-images.githubusercontent.com/13274079/81881797-b878e400-95ae-11ea-95a3-4467d04be78d.png)

Any ideas?

carlonewrelic:
[...]

PR #739, opened 2021-04-28
Support .js .cjs and custom config filename

Maddemacher:
## Proposed Release Notes

- default support for config files named `newrelic.js` and `newrelic.cjs`. Config file name can also be specified with the NEW_RELIC_CONFIG_FILENAME environment variable.

## Links

Fixes: https://github.com/newrelic/node-newrelic/issues/344

## Details

We need to have the newrelic config file end with `.cjs` since our package using newrelic is of type module. This will also give users the ability to specify an arbitrary config file name instead of being limited to only `newrelic.js`

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -32,18 +32,6 @@ const DEFAULT_MAX_PAYLOAD_SIZE_IN_BYTES = 1000000
 const DEFAULT_CONFIG_PATH = require.resolve('./default')
 const SPAN_EVENT_LIMIT = 1000
 const BASE_CONFIG_PATH = require.resolve('../../newrelic')
-const CONFIG_FILENAMES = [
-  process.env.NEW_RELIC_CONFIG_FILENAME,
-  'newrelic.js',
-  'newrelic.cjs'
-].filter(Boolean)
-const CONFIG_FILE_LOCATIONS = [
-  process.env.NEW_RELIC_HOME,
-  process.cwd(),
-  process.env.HOME,
-  path.join(__dirname, '../../../..') // above node_modules
-].filter(Boolean)
-
 const HAS_ARBITRARY_KEYS = new Set([
   'ignore_messages',
   'expected_messages',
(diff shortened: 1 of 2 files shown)


carlo-808:
Hi @Maddemacher , could you add a test for this functionality. It might not be super straight forward to test. 

Could you give us a better sense of how you are using cjs and/or es modules? In our early experiments, cjs and esm were not working with our agent.

Maddemacher:
We've also realised that esm is not really supported. So in the end we will be forced to go back to babel to have everything working the way we need it.

The change to be able to specify a filename for the config file is imo still relevant but the .cjs default can be removed if you want.

Ill await your guidance before continuing

@carlo-808

carlo-808:
Thank you for the response @Maddemacher . I will go ahead an approve, but we would appreciate it if you could add a sanity test around loading the cjs config. Here's a good example to follow:
https://github.com/newrelic/node-newrelic/blob/main/test/unit/config/config.test.js#L1226-L1284
The test should just verify it loads the file with no issues.

Maddemacher:
@carlo-808 will do,

lib/config/index.js
```diff
@@ -54,9 +46,22 @@ const exists = fs.existsSync || path.existsSync
 let logger = null // Lazy-loaded in `initialize`.
 let _configInstance = null
 
+const getConfigFileNames = () => [
+  process.env.NEW_RELIC_CONFIG_FILENAME,
```
carlo-808, line 50: You could keep the file name config array and file locations arrays as constants at the top and do something like `CONFIG_FILE_LOCATIONS.filter(Boolean).splice` rather than using a function.

test/unit/config/config.test.js
```diff
@@ -1306,6 +1306,81 @@ describe('the agent configuration', function() {
...
   })
 
+  describe('Selecting config file path', () => {
+    let origHome
+    let originalWorkingDirectory
+    const DESTDIR = path.join(__dirname, 'test_NEW_RELIC_CONFIG_FILENAME')
+    const NOPLACEDIR = path.join(__dirname, 'test_NEW_RELIC_CONFIG_FILENAME_dummy')
+    let CONFIG_PATH
+
+    beforeEach(() => {
```
carlo-808, line 1316: Thank you for adding these tests! 👍

lib/config/index.js
Maddemacher: as I mentioned, this was changed for testing purposes. Since this was constructed at load time I was unable to modify these in my tests without changing them to functions instead of just constant lists

carlo-808: Were there other tests failing? I just tried the `config.test.js` tests with the arrays directly, rather than from a function and the tests pass. Sorry to be a pain, but could you try the tests again with `CONFIG_FILE_LOCATIONS` and a `CONFIG_FILE_NAMES` array?

Maddemacher: with the file names as an array instead of a function:

image

The old tests work fine, my new ones are the ones that are failing since the ENV-vars are read when the module is loaded which make them impossible to override in my tests.

Maddemacher: The thing is, if they are not functions, they will default to the current working directory and search for newrelic.js, that file exists which will make them pass since the lists are constructed att module load time.

I changed the assertions so we have the supposed filename within the config so we can assert against that. 

All in all, they need to be functions since we need to change the values of those lists in the tests otherwise we would just search for newrelic.js in the root of the project. Now they will fail the way they are supposed to since I added something better to assert against within the config.

carlo-808 (approved):
👍

d1ffuz0r:
@carlo-808 hi!
Is there an ETA when this fix will be available on npm?
thank you

carlo-808:
hi @d1ffuz0r , we'll probably get a release out with this change early next week.

PR #749, opened 2021-05-07
Bumps redis versioned test timeout to 20s.

michaelgoin:
## Proposed Release Notes

* Bumped timeout for redis versioned tests.

## Details

Looks like a few failures I could track down were hitting the 10K timeout while only having attempted the polling retry once. As such bumping the timeout. I doubled to just be done with it. If it still fails at this timeout, will try a different approach altogether.

I added some additional commenting for when the hset is actually executed to confirm nothing unexpected is happening.


carlo-808 (approved):
Timeout bumped

PR #752, opened 2021-05-10
Uses unique DB index for redis tests to avoid conflicts

michaelgoin:
## Proposed Release Notes

* Updated redis versioned tests to use unique DB indexes per file to avoid collisions and flushing of in-progress tests.

## Details

All of the redis tests use the same DB index and regularly the same keys. At the start of each test file, the redis DB index is cleared for the test, instead of say cleaning up after each key operation. While currently (just for versioned tests) this would be fine as I believe we run the individual files in series for a single suite, the causes issues when ioredis and redis tests are both running because they run in parallel. We'd also have issues if these were integration tests all all files run in parallel.

I discovered this while debugging the current flicker by accidentally making things worse. Turns out one of the ioredis test files was clearing the DB mid-test for one of the redis tests.

Given the current setup, the easy/quick fix was to give each file a unique DB index (essentially a unique namespace). 0-15 are valid. If we had a large suite of redis test files, we may instead want to consider unique keys and cleaning up just the key/operation after each test.

I also cleaned up some error handling (for examplew hen your redis docker container is not up) so that each file will display this in a meaningful way and not show a bunch of random failures.

Lastly, I ran into an odd situation where the cleanup code for ioredis-3 to remove the ioredis ref wasn't good enough. Loading ioredis was throwing due to trying to redifine Promise. Unsure if this is a specific version issue or what but would only occur when manually running the file in a certain state (would pass as a full versioned run). Don't know why my changes might impact that or if maybe just bad luck with where versions ended up during my testing, but I updated the clearing to clear all ioredis references from the require cache and that did the trick. I print out the number cleared via tap comments in case there's more weirdness in the future.

### DIFF
--- a/test/lib/agent_helper.js
+++ b/test/lib/agent_helper.js
@@ -246,9 +246,9 @@ const helper = module.exports = {
    * @param {function} callback
    *  The operations to be performed while the server is running.
    */
-  bootstrapRedis: (redis, dbIndex, callback) => {
+  flushRedisDb: (redis, dbIndex, callback) => {
     if (!callback) {
-      // bootstrapRedis(dbIndex, callback)
+      // flushRedisDb(dbIndex, callback)
       callback = dbIndex
       dbIndex = redis
       redis = require('redis')
--- a/test/versioned/ioredis/ioredis-3.tap.js
+++ b/test/versioned/ioredis/ioredis-3.tap.js
@@ -17,16 +17,20 @@ tap.test('ioredis instrumentation', function(t) {
   var agent, redisClient
 
   t.beforeEach(function(done) {
-    setup(t, function(a, client) {
-      agent = a
-      redisClient = client
+    setup(t, function(error, result) {
+      if (error) {
+        return done(error)
+      }
+
+      agent = result.agent
+      redisClient = result.client
       done()
     })
   })
 
   t.afterEach(function(done) {
-    helper.unloadAgent(agent)
-    redisClient.disconnect()
+    agent && helper.unloadAgent(agent)
+    redisClient && redisClient.disconnect()
     done()
   })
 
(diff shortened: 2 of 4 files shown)


test/lib/agent_helper.js
```diff
@@ -246,9 +246,9 @@ const helper = module.exports = {
    * @param {function} callback
    *  The operations to be performed while the server is running.
    */
-  bootstrapRedis: (redis, dbIndex, callback) => {
+  flushRedisDb: (redis, dbIndex, callback) => {
```
michaelgoin, line 249: Since this basically did like one thing tangibly (although I do think the selection of the index persists), I updated the function name to be clear it is flushing the redis DB.

test/versioned/ioredis/ioredis-3.tap.js
```diff
@@ -5,30 +5,32 @@
...
-
-// CONSTANTS
-var DB_INDEX = 2
+const tap = require('tap')
+const helper = require('../../lib/agent_helper')
+const assertMetrics = require('../../lib/metrics_helper').assertMetrics
+const params = require('../../lib/params')
 
+// Indicates unique database in Redis. 0-15 supported.
+const DB_INDEX = 4
```
michaelgoin, line 14: This is the primary part for fixing the conflicts.

The rest of the code updates below are primarily to handle errors better for the tap framework.

test/versioned/ioredis/ioredis-3.tap.js
```diff
@@ -101,18 +103,46 @@ tap.test('ioredis instrumentation', function(t) {
...
+  let deletedCount = 0
+  Object.keys(require.cache).forEach((key) => {
+    if (key.indexOf('/ioredis/node_modules/ioredis/') >= 0) {
+      delete require.cache[key]
+      deletedCount++
+    }
+  })
+
+  t.comment(`Cleared ${deletedCount} modules matching '*/ioredis/node_modules/ioredis/*'`)
+}
```
michaelgoin, line 148: Walks the require cache and deletes the ioredis related items.

carlo-808 (approved):
👍  I see DB_INDEX updated to different values for each test to get rid of redis ci test flickers.

Issue #745, opened 2021-05-05
New promise tracking could leak for non-resolved promises


michaelgoin:
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description

If a customer were to have a promise that never resolves (resolve/reject), but still cleans-up all references out of memory, we might leak a segment reference due to no-longer using the 'destroy' hook. As such, we likely want to delete from both the 'promiseResolve' and destroy hooks. This is an edge case as typically, the resolve/reject handlers would still be in memory and keep the promise in memory but is possible, so we probably want to be defensive to prevent introducing new issues into customer apps.

Adding the destroy hook back will add some overhead but hopefully will still be enough of a win given the large drop in memory usage that already exists via the feature flagged implementation that removes on 'promiseResolve'. Impact should be measured to understand how noticeable it will be for a few promise scenarios.

PR #758, opened 2021-05-17
Adds destroy hook segment removal to handle non-resolved promise cleanup.

michaelgoin:
## Proposed Release Notes

* Fixed issue when using the 'new_promise_tracking' feature flag where segment mapping may not get cleaned up for promises which never resolve but have all references removed (and thus get cleaned up by GC).

  Adds segment cleanup on 'destroy' when using 'new_promise_tracking' feature flag in addition to the existing 'promiseResolve' hook. Unfortunately, preventing leaks for this edge-case does come with a small amount of additional overhead due to adding another hook. Memory gains from feature flag usage should still be worth the trade-off and reduced garbage collection may offset perf/CPU impacts or event still result in net gain, depending on the application.

## Links

* Fixes: https://github.com/newrelic/node-newrelic/issues/745

## Details

I'm guessing we can't get away with getting rid of the before/after hooks with the current design but if we could get rid of one of those, we'd remove the additional impact.

### DIFF
--- a/lib/instrumentation/core/async_hooks.js
+++ b/lib/instrumentation/core/async_hooks.js
@@ -191,6 +191,10 @@ function getPromiseResolveStyleHooks(segmentMap, agent, shim) {
       if (hookSegment === null) {
         shim.setActiveSegment(hookSegment)
       }
+    },
+    destroy: function destroyHandler(id) {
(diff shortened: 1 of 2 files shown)


carlo-808 (approved):
👍

PR #785, opened 2021-06-17
Fixes loading config from the main module's directory

alexpls:
## Proposed Release Notes

- Fix loading config from the main module's directory.

## Links

N/A

## Details

Hey 👋  - thanks for your work on this package!

Prior to v7.5.0 config files stored at the same filepath as the main module's filename would be correctly loaded, however [this commit](https://github.com/newrelic/node-newrelic/commit/3931f69cbc66629d65423b6419d69f5f7f3cdf0d) introduced a regression on the behaviour by no longer including the `path.dirname(process.mainModule.filename)` in the config file location candidates.

The effect of this is that if the New Relic config file is stored in the same directory as the main module, it may not be loaded in. My workaround for the time being will be to manually set the `NEW_RELIC_HOME` env var to my main module's directory.

This fix addresses the regression, and adds some additional test coverage.

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -56,13 +56,12 @@ const getConfigFileLocations = () => [
   process.env.NEW_RELIC_HOME,
   process.cwd(),
   process.env.HOME,
-  path.join(__dirname, '../../../..') // above node_modules
+  path.join(__dirname, '../../../..'), // above node_modules
+  // the REPL has no main module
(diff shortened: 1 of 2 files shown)


carlo-808 (approved):
Thank you for @alexpls for the fix and associated tests. This looks great and I verified the fix locally as well.

PR #805, opened 2021-07-19
Pin mongodb versioned tests <4

carlo-808:
## Proposed Release Notes
* Pinned mongodb versioned tests to <4.0.0.
## Details
Mongodb 4.0.0 was released on 7/13/2021 and is breaking one of the versioned tests. There seems to be several breaking changes in the [changelog](https://github.com/mongodb/node-mongodb-native/blob/v4.0.0/HISTORY.md) that may have to be addressed with a focused effort.

### DIFF
--- a/test/versioned/mongodb/package.json
+++ b/test/versioned/mongodb/package.json
@@ -42,7 +42,7 @@
         "node": ">=10"
       },
       "dependencies": {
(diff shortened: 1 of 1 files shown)


bizob2828 (approved):
Verified pinning version to `<4.0.0` fixes CI.

PR #811, opened 2021-07-20
set engines to node 16 and update node vers check

carlo-808:
## Proposed Release Notes
* Added check for minimum node version >= 12.
* Set package.json engines to >= 12.
## Links
Closes: #797

### DIFF
--- a/index.js
+++ b/index.js
@@ -48,16 +48,16 @@ function initialize() {
       preAgentTime
     )
 
-    // TODO: Update this check when Node v10 is deprecated.
-    if (psemver.satisfies('<10.0.0')) {
+    // TODO: Update this check when Node v12 is deprecated.
+    if (psemver.satisfies('<12.0.0')) {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -48,16 +48,16 @@ function initialize() {
...
-                'greater than 10.0.0. Not starting!'
+                'greater than 12.0.0. Not starting!'
 
       logger.error(message)
       throw new Error(message)
 
-      // TODO: Update this check when Node v16 support is added
-    } else if (!psemver.satisfies(pkgJSON.engines.node) || psemver.satisfies('>=15.0.0')) {
+      // TODO: Update this check when Node v18 support is added
+    } else if (!psemver.satisfies(pkgJSON.engines.node) || psemver.satisfies('>=18.0.0')) {
```
michaelgoin, line 60: think you want 17 here so we start notifying with the next odd release.

PR #812, opened 2021-07-20
Remove Node v10 from ci version matrix.

carlo-808:
## Proposed Release Notes
* Removed Node v10 from ci workflow and smoke-test version matrix.
* Set ci workflow linting check to use Node v16.


michaelgoin:
Don't forget about the smoke test workflow.

michaelgoin (approved):

PR #813, opened 2021-07-21
Linting always uses latest LTS. Bumped to setup-node@v2.

michaelgoin:
## Proposed Release Notes

* Updated linting to always use latest LTS Node version.

* CI and Smoke Test scripts now use setup-node@v2.


bizob2828 (approved):
Nice find.  I see it's using `14.17.3` which is latest LTS

PR #822, opened 2021-07-26
Updates third party notices for recent tap and external module updates.

michaelgoin:
## Details

Not leaving a specific release note as this is inferred with the tap upgrade and we haven't shipped yet


bizob2828 (approved):
👍🏻

carlo-808 (approved):
👍  3rd party notices updated

PR #824, opened 2021-07-28
Change to https protocol in homepage of package.json

pzrq:
## Proposed Release Notes
 * Fixed the `homepage` field in package.json to use `https` in the link to the github repo

## Details

I think this will fix the `URL` field display when downstream services, such as `yarn outdated` consume this, newrelic was the odd one out of a sample in a project I was looking at:

Screen Shot 2021-07-28 at 23 31 13

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "debugging",
     "profiling"
   ],
-  "homepage": "http://github.com/newrelic/node-newrelic",
+  "homepage": "https://github.com/newrelic/node-newrelic",
   "engines": {
     "node": ">=12.0.0",
     "npm": ">=6.0.0"


bizob2828 (commented):
@pzrq thanks for your contribution! I agree this should be fixed but can you describe the issue you're seeing when this homepage is using `http` vs `https`? I ran `yarn outdated` with newrelic as a dep and everything works as expected

bizob2828 (approved):

pzrq:
> @pzrq thanks for your contribution! I agree this should be fixed but can you describe the issue you're seeing when this homepage is using http vs https? I ran yarn outdated with newrelic as a dep and everything works as expected

@bizob2828 I guess you figured it out, though in a downstream repo, if an older version of newrelic, such as v7.5.5 is installed, then `yarn outdated` will report a URL which contains `http` as per my screenshot.

PR #834, opened 2021-08-05
update third party notices for release

carlo-808:
Update third party notices for release


bizob2828 (changes requested):
looks like http-errors is in both manifest and notices twice

carlo-808:
> looks like http-errors is in both manifest and notices twice

Good catch! I'll remove one of them.

bizob2828 (approved):

Issue #852, opened 2021-08-24
DT enabled - allow for configuring span event limit


carlo-808:
Users should be allowed to configure the span event reservoir size (`SPAN_EVENT_LIMIT`).

Configuration value spec is [here](https://source.datanerd.us/agents/agent-specs/blob/master/Span-Events.md#configuration-and-transmission). 

Allow for users to set the span event limit via configuration file. The minimum allowed value should be 1000 and the maximum allowed value should be 10,000.

Add an environment variable to set the `SPAN_EVENT_LIMIT`.

PR #863, opened 2021-08-31
Add span event max samples config

michaelgoin:
## Proposed Release Notes

* Added ability to configure the maximum number of spans that can be collected per minute via `span_events.max_samples_stored`.

## Links

* https://github.com/newrelic/node-newrelic/issues/852

## Details

I decided not to keep a lower-limit with this PR. Other max_samples_stored configuration items do not have a locally enforced minimum. Should a minimum become a hard requirement, enforced on the agent-side, we can resolve with a follow-up PR.

I also noticed there was inconsistent logging between the appropriate child logger and the base logger. This was due to some lazy initialization via constructor to facilitate log testing. This had some downsides with not having the appropriate feature identified but also risk of using a `null` logger depending on future order operation (low risk but existed). For now, I changed to just instantiate the child logger on module load and ripped out the tests as they did not seem critical. Longer-term, injecting an appropriate logger or similar and having obviously only usable via instance or propagation may be preferred but I did not bite that off with this effort.


carlo-808 (approved):
Looks good. 👍  Not enforcing lower bound of span event limit is fine.

PR #869, opened 2021-09-03
log SpanEvent/Limit

garbados:
## Proposed Release Notes

- Added supportability metric SpanEvent/Limit

## Links

- Closes https://github.com/newrelic/node-newrelic/issues/862

### DIFF
--- a/lib/spans/span-event-aggregator.js
+++ b/lib/spans/span-event-aggregator.js
@@ -92,6 +92,9 @@ class SpanEventAggregator extends EventAggregator {
 
     this.periodMs = periodMs
     this.limit = limit
+    this._metrics
+      .getOrCreateMetric(`${NAMES.SUPPORTABILITY.PREFIX}SpanEvent/Limit`)
+      .recordValue(this.limit)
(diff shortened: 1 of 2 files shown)


bizob2828 (approved):
Congrats on your first PR 🎉 . LGTM

lib/spans/span-event-aggregator.js
```diff
@@ -92,6 +92,9 @@ class SpanEventAggregator extends EventAggregator {
 
     this.periodMs = periodMs
     this.limit = limit
+    this._metrics
+      .getOrCreateMetric(`${NAMES.SUPPORTABILITY.PREFIX}SpanEvent/Limit`)
+      .recordValue(this.limit)
```
michaelgoin, line 97: Minor (and this is already merged) but we do have a collection of 'SpanEvent' support metrics organized in the 'names' file that has our various metric names/prefixes/etc.

https://github.com/newrelic/node-newrelic/blob/main/lib/metrics/names.js#L233-L237

For future similar changes would prob organize in there.

PR #875, opened 2021-09-07
transition priority-queue.test.js from mocha to tap

garbados:
## Proposed Release Notes

- Transitioned priority-queue.test.js to use tap rather than Mocha globals.


carlo-808 (approved):
👍  priority-queue tests converted to tap

PR #877, opened 2021-09-07
transition parse-sql.test.js to use tap over mocha globals

garbados:
## Proposed Release Notes

- Transitioned parse-sql.test.js to use tap over Mocha globals.


carlo-808 (approved):
👍  parse-sql tests converted to tap

PR #878, opened 2021-09-07
transition analytics_events.js to tap from mocha globals

garbados:
## Proposed Release Notes

- Transitioned analytics_events.js to Tap from using Mocha globals.


garbados:
I updated this PR so that the agent is more reliably constructed and destructed. This avoided restructuring the tests much, as the beforeEach hooks were actually pretty sensitive.

bizob2828 (approved):

Issue #854, opened 2021-08-24
DT enabled - Set new default span event limit


carlo-808:
Set the span event limit to the optimal default value determined by OATS testing.

Based on testing, we should set the span event limit to 2000. (double check this value next week 9/7)

PR #879, opened 2021-09-07
Update default span event limit to 2000

carlo-808:
## Proposed Release Notes
* Updated default span event limit to 2000.
## Links
Closes: #854


bizob2828 (approved):
1000 -> 2000 👍

PR #889, opened 2021-09-09
update eslint config

garbados:
## Proposed Release Notes

- Applied new lint rules barring the use of `var` and preferring the use of `const` wherever possible.

## Links

- eslint config issue https://github.com/newrelic/eslint-config-newrelic/issues/3

## Details

This PR makes extensive changes to the repo in applying these new rules. Only a few of these changes were done by hand. This may result in changed test output; locally I experience five test failures, but I'm not sure if CI will experience the same.

### DIFF
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -1335,7 +1335,7 @@ IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
 ### @newrelic/eslint-config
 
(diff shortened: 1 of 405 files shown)


garbados:
I am able to reproduce the single test failure but only when running `npm run unit` -- the test file that fails succeeds when run alone. This suggests some kind of leftover state inbetween test files. Converting to draft while I investigate...

garbados:
Ready for review!

bizob2828 (approved):
It's nice this auto-fixed most. Always good to keep code base up to modern constructs.

Issue #900, opened 2021-09-14
transaction-shim handleCATHeaders does not handle DT if CAT disabled.


michaelgoin:
The not-entirely accurately named (anymore) handleCATHeaders function processes CAT and DT headers. It does the check for CAT first, so if that is disabled it won't ever get to executing the DT logic. Other places that do both check DT first.

This will be more likely to happen with CAT set to disabled by default (PR which will be landing soon).

For these tests, if you set CAT to disabled (the beforeEach setup is setting it to enabled), you'll see the tests fail:
https://github.com/newrelic/node-newrelic/blob/main/test/unit/shim/transaction-shim.test.js#L589-L647

These tests fail if CAT not explicitly enabled, should be updated to *NOT* do that as part of fix. Good for red/green:
* https://github.com/newrelic/node-newrelic/blob/main/test/versioned/amqplib/promises.tap.js#L312
* https://github.com/newrelic/node-newrelic/blob/main/test/versioned/amqplib/callback.tap.js#L302

Here's the offending code:
https://github.com/newrelic/node-newrelic/blob/main/lib/shim/transaction-shim.js#L256-L282

Correct ordering of logic:
https://github.com/newrelic/node-newrelic/blob/main/lib/instrumentation/core/http.js#L141-L154

There's prob more consolidation to be had around some of this code which may or may not make sense given CAT will be fully removed. Keep that in mind for changes that could make removal easier (or avoid making it harder).

This function is also not named well. If we want to change it, it will have to be with a major release though. I'm partially wondering if there's a better way than a special shim method. We have transaction handle methods that then make calls on transactions. Perhaps when we rip out CAT we can rip this function out and have obvious replacement more similar to the transaction handle style. I don't love that our APIs users use and we use are so divergent. Any chance for leveraging same underlying API? This probably warrants its own story flagged for the semver:major to go with the CAT deprecation or added as a comment to that issue to be generated when we plan that. Just getting thoughts out somewhere.

michaelgoin:
https://github.com/newrelic/node-newrelic/issues/903

PR #902, opened 2021-09-14
Fixes transaction-shim.handleCATHeaders() handling of DT headers

michaelgoin:
## Proposed Release Notes

* Fixed issue where DT headers would not be processed by `transaction-shim.handleCATHeaders()` when CAT was explicitly disabled. This primarily impacts `amqplib` instrumentation.

## Links

* Fixes: https://github.com/newrelic/node-newrelic/issues/900

## Details

Given our intent to rip out CAT, I opted for a pretty simple reorganization of the logic and added some TODO reminders.

### DIFF
--- a/lib/shim/transaction-shim.js
+++ b/lib/shim/transaction-shim.js
@@ -254,30 +254,38 @@ function setTransactionName(name) {
  *  The transport type that brought the headers. Usually `HTTP` or `HTTPS`.
  */
 function handleCATHeaders(headers, segment, transportType) {
-  // Is CAT enabled?
-  if (!this.agent.config.cross_application_tracer.enabled) {
-    this.logger.trace('CAT disabled, not extracting header.')
+  // TODO: rename function or replace functionality when CAT fully removed.
+
+  if (!headers) {
+    this.logger.debug('No headers for CAT or DT processing.')
     return
   }
-  if (!this.agent.config.encoding_key) {
-    this.logger.warn('Missing encoding key, not extracting CAT headers!')
-    return
-  } else if (!headers) {
-    this.logger.debug('No headers to search for CAT within')
+
+  const config = this.agent.config
+
+  if (!config.cross_application_tracer.enabled && !config.distributed_tracing.enabled) {
+    this.logger.trace('CAT and DT disabled, not extracting headers.')
     return
   }
 
   // Check that we're in an active transaction.
-  segment = segment || this.getSegment()
-  if (!segment || !segment.transaction.isActive()) {
-    this.logger.trace('Not adding CAT reply header, not in an active transaction.')
+  const currentSegment = segment || this.getSegment()
+  if (!currentSegment || !currentSegment.transaction.isActive()) {
+    this.logger.trace('Not processing headers for CAT or DT, not in an active transaction.')
     return
   }
 
-  const transaction = segment.transaction
+  const transaction = currentSegment.transaction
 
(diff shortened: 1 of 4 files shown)


lib/shim/transaction-shim.js
```diff
@@ -254,33 +254,41 @@ function setTransactionName(name) {
...
 
   // Check that we're in an active transaction.
-  segment = segment || this.getSegment()
-  if (!segment || !segment.transaction.isActive()) {
-    this.logger.trace('Not adding CAT reply header, not in an active transaction.')
+  const currentSegment = segment || this.getSegment()
+  if (!currentSegment || !currentSegment.transaction.isActive()) {
+    this.logger.trace('Not processing headers for CAT or DT, not in an active transaction.')
     return
   }
```
michaelgoin, line 276: Didn't love re-assigning of the incoming function argument as that can sometimes lead to future bugs when trying to check against the original value if you don't notice the reassignment. While the risk was low here, wanted to clean up the pattern.

lib/shim/transaction-shim.js
```diff
@@ -254,33 +254,41 @@ function setTransactionName(name) {
...
 
-  const transaction = segment.transaction
+  const transaction = currentSegment.transaction
 
-  if (this.agent.config.distributed_tracing.enabled) {
+  if (config.distributed_tracing.enabled) {
```
michaelgoin, line 280: The crux of this bug fix is checking/handling DT prior to CAT if DT is enabled.

lib/shim/transaction-shim.js
```diff
@@ -254,33 +254,41 @@ function setTransactionName(name) {
...
   }
-  if (!this.agent.config.encoding_key) {
-    this.logger.warn('Missing encoding key, not extracting CAT headers!')
-    return
-  } else if (!headers) {
-    this.logger.debug('No headers to search for CAT within')
+
+  const config = this.agent.config
+
+  if (!config.cross_application_tracer.enabled && !config.distributed_tracing.enabled) {
```
michaelgoin, line 266: Short circuiting if both aren't enabled to avoid a small amount of work and reduce some nesting which should aid in the future code rip out. This can be reduced and the later DT check removed when we rip out CAT.

lib/shim/transaction-shim.js
```diff
@@ -254,33 +254,41 @@ function setTransactionName(name) {
  *  The transport type that brought the headers. Usually `HTTP` or `HTTPS`.
  */
 function handleCATHeaders(headers, segment, transportType) {
-  // Is CAT enabled?
-  if (!this.agent.config.cross_application_tracer.enabled) {
-    this.logger.trace('CAT disabled, not extracting header.')
+  // TODO: rename function or replace functionality when CAT fully removed.
```
michaelgoin, line 257: I will be entering a GitHub issue for this shortly.

lib/shim/transaction-shim.js
carlo-808: Nice clean up.

test/unit/shim/transaction-shim.test.js
```diff
@@ -650,6 +650,27 @@ tap.test('TransactionShim', function (t) {
       }
     )
 
+    t.test('should propagate w3c headers when CAT expicitly disabled', (t) => {
```
carlo-808, line 653: Nice additional test for. trace context headers.

carlo-808 (approved):
🐱  👍

PR #906, opened 2021-09-17
Resolve http outbound test flicker

michaelgoin:
## Proposed Release Notes

* Replaced http-outbound test call to use example.com to avoid unpredictable connection resets.

## Details

My spellcheck was also unhappy so resolved some misspellings.


bizob2828 (approved):
Withstood my loop over running suite 50 times in a row, 0 failures. good enough for me 🚢 🇮🇹

Issue #909, opened 2021-09-17
Cassandra-Driver v4 and above not included in tests


garbados:
In the versioned test suite, only versions [3.4.0 up to but not including 4.0.0](https://github.com/newrelic/node-newrelic/commit/a3cf8ad2f0fafc5663d32c07d1e0aedc23cd3dc6) of `cassandra-driver` are tested. As the inner workings of the Cassandra client have [changed](https://github.com/newrelic/node-newrelic/issues/564) since then, we should test against more recent versions of the driver.

Issue #564, opened 2020-12-09
DOC ISSUE: Datastore Example Not Accurate


icpenguins:
## Description
When following the documentation for the `Datastore` example using Cassandra, the reference to `_innerExecute` ([link](https://github.com/newrelic/node-newrelic/blame/main/examples/shim/Datastore-Simple.md#L20)) is not accurate. The documentation goes on later to provide a source link to the Cassandra line where `_innerExecute` should be found ([link](https://github.com/datastax/nodejs-driver/blob/master/lib/client.js#L355)). However, the code base has changed to the point that `_innerExecute` no longer exists making this documentation less effective at showing how best to instrument a datastore.

carlo-808:
Good catch @icpenguins ! Thank you for bringing this up with us. We'll get this prioritized and work on it as soon as we can.

PR #910, opened 2021-09-17
support cassandra-driver >4

garbados:
## Proposed Release Notes

- Support Cassandra driver v4.0.0 and above.

## Links

- Closes https://github.com/newrelic/node-newrelic/issues/909

## Details

This PR adds support for Cassandra v4. In v4.4.0, the private method `_innerExecute` was replaced with `_execute`. Because our tests were pinned to test only *up to but not including* v4, we only detected this failure once it was [reported to us](https://github.com/newrelic/node-newrelic/issues/564). This PR creates query hooks on `_execute` if `_innerExecute` does not exist. As well, the test suite has been updated

### DIFF
--- a/test/versioned/cassandra-driver/package.json
+++ b/test/versioned/cassandra-driver/package.json
@@ -8,7 +8,7 @@
         "node": ">=12"
       },
       "dependencies": {
-        "cassandra-driver": ">=3.4.0 <4"
+        "cassandra-driver": ">=3.4.0 <5"
       },
       "files": [
         "query.tap.js"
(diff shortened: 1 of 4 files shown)


test/versioned/cassandra-driver/package.json
```diff
@@ -8,7 +8,7 @@
         "node": ">=12"
       },
       "dependencies": {
-        "cassandra-driver": ">=3.4.0 <4"
+        "cassandra-driver": ">=3.4.0 <5"
```
michaelgoin, line 11: Do we need to cap at 5? Or can we leave that open-ended?

garbados: We could certainly leave it open ended. I'll make the change.

garbados:
Ready for final review!

michaelgoin (approved):
👍

PR #911, opened 2021-09-20
fix cassandra example

garbados:
## Links

- Closes https://github.com/newrelic/node-newrelic/issues/564
- See also: https://github.com/newrelic/node-newrelic/pull/910

## Details

Updates the datastore example to reflect the current state of instrumentation. A different issue should cover a rearchitecture of our docs away from implementation-sensitive examples.


bizob2828 (approved):
LGTM

Issue #922, opened 2021-09-23
Timer clear can ignore incorrect segment resulting in loss of MongoDB segments/spans


michaelgoin:
Our timer clear logic flags the segment created for the timer as 'ignored' so it does not get reported, as it did not have a chance to later execute the handler. It does not, however, seem to take into account the case of 'opaque' and maybe even 'internal' segments.

MongoDB is executing a timer behind the scenes in context of a transaction. Since this is happening while one of the MongoDB segments is active, and those segments are 'opaque', no new segment is created for the timer. When MongoDB calls 'clearTimeout' on that timer, the MongoDB segment is retrieved from the callback since no new segment was created and then flagged as 'ignored'. This ignoring results in the segment/span being dropped on the floor at the end of transaction processing. It did not, however, prevent it from counting against the DB counts communicated on the transaction event itself.

For example, when the following are captured:

```
Datastore/statement/MongoDB/testCollection/count
Datastore/statement/MongoDB/testCollection/toArray
Datastore/statement/MongoDB/testCollection/aggregate
```

Only  'aggregate' will be sent up as a segment or span.

PR #932, opened 2021-09-27
Fix opaque timer issue

michaelgoin:
## Proposed Release Notes

* Fixes issue with `clearTimeout` that could result in dropping parent segments or spans.

  This bug resulted in some MongoDB calls being dropped from Transaction Traces and Distributed Traces (spans): https://github.com/newrelic/node-newrelic/issues/922.

## Links

* Fixes: https://github.com/newrelic/node-newrelic/issues/922

## Details

I looked at the pre-shim code and it had the same issue. I think ideally we could avoid doing work when the parent is opaque for the timers but we currently have to based on the existing context propagation mechanisms. If we successfully switch to newer context mechanisms, perhaps we can have a more obvious means of detecting we are not in this context. We could potentially use another mechanism for identifying it is in fact a timer span.

For now, I went with matching against "opaque" which is the case that results in this so there's at least an obvious cause/effect check.

Internal also has the case to trigger this but it seems, at least on the segment creation side, we check for botht he child being flagged internal and the parent as to when it won't create the child segment. Timers are not 'internal' so seem to create children. I added a test case to catch if this behavior ever changes.

### DIFF
--- a/lib/instrumentation/core/timers.js
+++ b/lib/instrumentation/core/timers.js
@@ -85,9 +85,9 @@ function initialize(agent, timers, moduleName, shim) {
 }
 
 function makeWrappedPromisifyCompatible(shim, timers) {
-  const originalSetTimout = shim.getOriginal(timers.setTimeout)
-  Object.getOwnPropertySymbols(originalSetTimout).forEach((symbol) => {
-    timers.setTimeout[symbol] = originalSetTimout[symbol]
+  const originalSetTimeout = shim.getOriginal(timers.setTimeout)
+  Object.getOwnPropertySymbols(originalSetTimeout).forEach((symbol) => {
+    timers.setTimeout[symbol] = originalSetTimeout[symbol]
   })
 
   const originalSetInterval = shim.getOriginal(timers.setInterval)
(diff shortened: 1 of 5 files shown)


lib/instrumentation/core/timers.js
```diff
@@ -70,7 +70,7 @@ function initialize(agent, timers, moduleName, shim) {
     return function wrappedClearTimeout(timer) {
       if (timer && timer._onTimeout) {
         const segment = timer._onTimeout.__NR_segment
-        if (segment) {
+        if (segment && !segment.opaque) {
```
michaelgoin, line 73: This is essentially the whole fix.

test/integration/core/timers.tap.js
```diff
@@ -185,66 +185,99 @@ tap.test('nextTick with extra args', function testNextTick(t) {
   }
 })
 
-tap.test('clearTimeout', function testNextTick(t) {
```
michaelgoin, line 188: There' was a duplicate test in here I cleared out. That makes this diff look weird unfortunately.

test/integration/core/timers.tap.js
```diff
@@ -185,66 +185,99 @@ tap.test('nextTick with extra args', function testNextTick(t) {
...
-    t.fail()
-  }
+tap.test('clearTimeout should function outside of transaction context', (t) => {
+  setupAgent(t)
+
+  const timer = setTimeout(t.fail)
+
+  clearTimeout(timer)
+
+  setImmediate(t.end)
```
michaelgoin, line 211: I split this out of the existing setTimeout test so each concept had own test.

bizob2828 (approved):
nice test additions to ensure this logic continues as we change our codebase.  I agree with your long term explanation but this feels like it's good enough to fix this one case

PR #936, opened 2021-09-28
add addHook handler to fastify instrumentation

garbados:
## Proposed Release Notes

- Instrument Fastify routes by wrapping `addHook`.

## Links

- Closes https://github.com/newrelic/node-newrelic/issues/925

## Details

As of this writing, this PR does not add additional tests. (I'm working on it...!) Any pointers about testing this would be appreciated, cc @bizob2828


test/versioned/fastify/new-state-tracking.tap.js
```diff
@@ -83,6 +83,34 @@ tap.test('fastify with new state tracking', (t) => {
 
     t.equal(transactions.length, 2)
   })
+
+  t.test('should wrap request/response hooks', async (t) => {
+    let ok = false
+
+    fastify.addHook('onRequest', (request, reply, done) => {
```
bizob2828, line 90: personally I would put these kind of tests in the naming suites.  I'd abstract away the creation of the fastify server and add hooks for every single request lifecycle event.  Then assert the segment names which means it's prob best to name each plugin function instead of arrow functions

lib/instrumentation/fastify.js
```diff
@@ -40,6 +51,22 @@ const setupRouteHandler = (shim, fastify) => {
...
   })
+
+  shim.wrap(fastify, 'addHook', function addWrapHook(shim, fn) {
+    return function wrappedAddHook() {
+      const args = shim.argsToArray.apply(shim, arguments)
+      if (REQUEST_HOOKS.includes(args[0])) {
+        const middlewareFunction = args[1]
+        const newMiddlewareFunction = shim.recordMiddleware(
+          middlewareFunction,
+          buildMiddlewareSpecForMiddlewareFunction(shim, args[0])
```
bizob2828, line 62: you may need to rebase but the 2nd are expects the segment name. so like this https://github.com/newrelic/node-newrelic/blob/main/lib/instrumentation/fastify.js#L52 but instead of hardcoding `onRequest` in that string you'd use `args[0] which is the lifecycle name

test/versioned/fastify/new-state-tracking.tap.js
garbados: I ended up placing them in their own test file because it seems the hook isn't affected by the differences between v2 and v3. LMK if this is sub-optimal!

garbados:
OK! Hook wrapper updated, tests added. The tests pass but only because there are some lifecycle events that don't get caught, which I then commented out. Specifically:

- preValidation
- preHandler
- preSerialization

I'll dive into this tomorrow but I thought getting this far was a nice milestone for the day.

garbados:
Sooooo as of this morning, those lifecycle events that mysteriously weren't firing yesterday... totally do fire this morning! 🤷‍♀️ 

I'll push what I have momentarily and then we'll let the CI make its call.

garbados:
Looks like we're good to go for review! Tests pass :D

test/versioned/fastify/add-hook.tap.js
```diff
@@ -0,0 +1,83 @@
+/*
```
bizob2828, line 1: i know this is not obvious, but you have to add this test to the `test/versioned/fastify/package.json` for each test stanza.  It looks like it does not work in v2 and am getting failures in v3

```
        not ok 2 - segment "WebTransaction/WebFrameworkUri/Fastify/GET//" should have child "Nodejs/Middleware/Fastify/preValidation/testHook" in position 3: expected 'Nodejs/Middleware/Fastify/onSend/testHook' to equal 'Nodejs/Middleware/Fastify/preValidation/testHook'
```

bizob2828: looking at the trace root there are only 3 of the 7 handlers firing

```
in lifecycle event onRequest
in lifecycle event preParsing
in lifecycle event onSend
```

bizob2828: looks like maybe because the signatures vary https://www.fastify.io/docs/latest/Hooks/.  Also I do see there are more hooks in there that I didn't previously call out like `onTimeout` and `onError` but I don't see them in the request lifecycle tree.

bizob2828 (changes requested):
instrumentation looking good, however the versioned tests aren't wired up and are failing for me in v3 and hanging in v2

test/versioned/fastify/add-hook.tap.js
```diff
@@ -0,0 +1,144 @@
...
+    })
+
+    agent.on('transactionFinished', (transaction) => {
+      t.equal('WebFrameworkUri/Fastify/GET//', transaction.getName(), `transaction name matched`)
+      // all the hooks are siblings of the route handler
+      metrics.assertSegments(transaction.trace.root, [
+        'WebTransaction/WebFrameworkUri/Fastify/GET//',
+        [
+          'Nodejs/Middleware/Fastify/routeHandler'
+          // `Nodejs/Middleware/Fastify/${hookName}/testHook`
```
garbados, line 131: I would expect the onError hook to be listed as a segment but... it isn't? Should I be worried?

lib/instrumentation/fastify/spec-builders.js
```diff
@@ -75,6 +75,12 @@ function buildMiddlewareSpecForRouteHandler(shim, path) {
      * The isFinal param determines whether or not a path is appended for
      * this particular piece of middleware. (i.e. if this is the final handler
      * that is actually handling the request, the path is actually left on)
+     *
+     * @param shim
```
bizob2828, line 79: I know this is unrelated but personally I'm not a fan of requiring specifying these in jsdoc without types

lib/instrumentation/undici.js
```diff
@@ -66,8 +66,8 @@ module.exports = function addUndiciChannels(agent, undici, modName, shim) {
    * We will check current segment for opaque and also attach
    * relevant headers to outgoing http request
    *
-   * @param {Object} params
-   * @param {Object} params.request undici request object
+   * @param {object} params
```
bizob2828, line 69: just curious why this file is touched in the PR?

test/versioned/fastify/add-hook.tap.js
bizob2828: it should def be there but the reason it is not is because in the instrumentation you didnt' add `onError` into the valid `REQUEST_HOOKS`. I added and i see the segment

screenshot 2021-10-07 at 10 42 18 AM

test/versioned/fastify/add-hook.tap.js
```diff
@@ -0,0 +1,144 @@
...
+    for (const [hookName, isOk] of Object.entries(ok)) {
+      t.equal(isOk, true, `${hookName} captured`)
+    }
+    t.end()
+  })
+
+  t.test('error hook', async function errorHookTest(t) {
+    // setup fastify route
+    fastify.get('/', function routeHandler() {
+      throw new Error('sup')
```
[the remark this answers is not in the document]
bizob2828: I think because in fastify 2 it didn't handle throwing errors in synchronous handlers.  So just turns this handler into an async function and it will work. It's mentioned in the 2 -> 3 upgrade guide https://github.com/fastify/fastify/blob/main/docs/Migration-Guide-V3.md#manage-uncaught-exception-2073

test/versioned/fastify/add-hook.tap.js
```diff
@@ -33,51 +48,97 @@ tap.test(function (t) {
...
+    })
+
+    const hookName = 'onError'
+    let ok = false
+
+    fastify.addHook(hookName, (req, reply, next) => {
```
bizob2828, line 119: it looks like the signature is diff https://github.com/fastify/fastify/blob/main/docs/Hooks.md#manage-errors-from-a-hook, it has 4 args and next is last

lib/instrumentation/undici.js
garbados: https://github.com/newrelic/node-newrelic/pull/936#discussion_r723759706

Issue #967, opened 2021-11-03
Process crash if unable to read /proc/cpuinfo and /proc/meminfo


jstaro:
`parseProcCPUInfo(data)` and `parseProcMemInfo(data)` cannot handle when `data` is `null`, which happens if the library was unable to read `/proc/cpuinfo` and `/proc/meminfo`

## Description

I am working on a node app capable of running both on multiple hardware devices and now in the cloud, so they now have restricted things to target a subset of functionality.

It turns out you now get an `ENOENT` when trying to cat `/proc/cpuinfo` and `/proc/meminfo`.

There is error handling in `readProc`, returning `null` if an error occurs, but the parsing functions cannot deal with that parameter being `null`. These parsing functions otherwise gracefully handle not being able to parse the string data, so my suggestion would be adding null-checking guard statements before trying to parse.

## Expected Behavior

The library should gracefully handle not being able to read `/proc/cpuinfo` and `/proc/meminfo`.

## Steps to Reproduce

N/A

## Your Environment

node-newrelic version: 8.5.0
Node version: 12.16.1
Operating System and version: unknown, custom Linux distro

michaelgoin:
Hi @jstaro,

Thank you for the issue. That is definitely not a use case we account for but looks like it is straightforward enough to handle.

michaelgoin:
@jstaro Got a quick fix out today with v8.5.1. Hopefully that resolves your issue.

PR #968, opened 2021-11-03
Adds graceful handling of null input to cpu and memory parse proc methods

michaelgoin:
## Proposed Release Notes

* Fixed bug where failure to retrieve CPU/Memory details for certain Linux distros could result in a crash.

  `parseProcCPUInfo` and `parseProcMeminfo` now check for `null` input prior to processing. 

## Links

* Fixes: https://github.com/newrelic/node-newrelic/issues/967

## Details

Also fixed a misspelling.


bizob2828 (approved):
nice defensive code + unit tests :shipit:

PR #985, opened 2021-11-18
Removes tracer.segment

michaelgoin:
## Proposed Release Notes

* Removed `tracer.segment` in place of direct usage of context manager.

## Details

Patch release as this is internal API (even though some usages of tracer have unfortunately bled outside the core agent).

Only test code was still using `tracer.segment`. I did not find any usages in external modules.

I cleaned up some usage of `tracer.getSegment()` in the test files touched but did not attempt a full cleanup. That can be a follow-up.

`getContextManager()` was added to the agent helper to try to avoid grabbing directly off the agent all over the place. It does currently soft-rely on the agent being loaded first which is brittle. I think ideally we'd have helper APIs that generally returned both but we can see how things evolve. I'll also need to add this to the external test helper. At some point, I'd love for those to not be separate. It would be ideal for the agent and external modules to use the same test helpers so we know everything is consistent in what is possible and what is executed for tests. Probably would be best if the agent either exported its own for the external modules to use or the helper was a separate module in the same repo so the updates can be made with the agent. Right now, it would be quite painful at times (chicken or egg problem) to have the agent use the helper directly from the test utils.

I went ahead and updated the defunct bench files so they stay in relatively up-to-date form as examples, should we tackle getting something similar in place to replace them.

### DIFF
--- a/lib/transaction/tracer/index.js
+++ b/lib/transaction/tracer/index.js
@@ -45,16 +45,6 @@ Tracer.prototype.wrapFunctionFirst = wrapFunctionFirst
 Tracer.prototype.wrapSyncFunction = wrapSyncFunction
 Tracer.prototype.wrapCallback = wrapCallback
 
-Object.defineProperty(Tracer.prototype, 'segment', {
-  get: function segmentGetter() {
-    return this._contextManager.getContext()
-  },
-  set: function segmentSetter(segment) {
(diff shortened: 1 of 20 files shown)


bizob2828 (approved):
verified no more usage of `tracer.segment`. Also searched all other external repos and they are not accessing `tracer.segment`.

PR #986, opened 2021-11-18
Updates module onResolved hooks to only fire the first time we've see…

michaelgoin:
## Proposed Release Notes

* Updated `onResolved` instrumentation hook to only be called the first time we see a specific module filepath resolved.


bizob2828 (approved):
I confirmed this fix addresses the issue we saw while working on aws-sdk v3 instrumentation. meaning the smithy client instrumentation only fires once even though the smithy client is getting resolved many times

PR #1001, opened 2021-12-15
fix canonicalization if-condition

garbados:
## Proposed Release Notes

* Fixed issue where expected status code ranges would not be parsed until ignored status codes were also defined.


michaelgoin:
Would change the release note on here as it is slightly misleading or implying a larger scope than it is.

Something like:

`* Fixed issue where expected status code ranges would not be parsed until ignored status codes were also defined.`

Individually listed expected status codes work fine.

michaelgoin (approved):
Good catch!

PR #1043, opened 2021-12-30
add external segment before host is known

garbados:
## Proposed Release Notes

- Add a transaction segment when a socket is first opened, before a request is fully formed, to account for possible socket errors.

## Links

- Closes https://github.com/newrelic/node-newrelic/issues/947

## Details

This PR introduces a segment when a request is first created, before the request host is known. This segment is inserted between the root segment and the segment created when the request's headers are sent, causing some changes to the test suite to reflect the new segment order.

Rather than refactor `recordExternal`, a fake host is used instead: `'<unknown>'`, so that these segments will appear to the user as `'External/<unknown>'` followed by a segment for the fully-formed request, such as `'External/httpbin.org/get'`. I used `'<unknown>'` in keeping with how we report on anonymous functions with `'<anonymous>'`.

Let me know if this is not an appropriate approach!

### DIFF
--- a/lib/instrumentation/undici.js
+++ b/lib/instrumentation/undici.js
@@ -103,13 +103,13 @@ module.exports = function addUndiciChannels(agent, undici, modName, shim) {
       request.addHeader(key, outboundHeaders[key])
     }
 
-    const name = NAMES.EXTERNAL.PREFIX + UNKNOWN_HOST
+    const name = NAMES.EXTERNAL.PREFIX + UNKNOWN_HOST + (request.method || 'get')
     const segment = shim.createSegment(name, recordExternal(UNKNOWN_HOST, 'undici'), parent)
     if (segment) {
       segment.start()
       shim.setActiveSegment(segment)
       segment.addAttribute('procedure', request.method || 'GET')
-      request[SYMBOLS.PARENT_SEGMENT] = segment
+      request[SYMBOLS.SEGMENT] = segment
     }
   })
 
@@ -144,17 +144,23 @@ module.exports = function addUndiciChannels(agent, undici, modName, shim) {
     const url = new URL(urlString)
 
     const name = NAMES.EXTERNAL.PREFIX + url.host + url.pathname
-    const segment = shim.createSegment(name, recordExternal(url.host, 'undici'), parentSegment)
-    if (segment) {
+    let segment
+    if (request[SYMBOLS.SEGMENT]) {
+      segment = request[SYMBOLS.SEGMENT]
+      segment.name = name
+    } else {
+      segment = shim.createSegment(name, recordExternal(url.host, 'undici'), parentSegment)
       segment.start()
       shim.setActiveSegment(segment)
+      segment.addAttribute('procedure', request.method || 'GET')
+      request[SYMBOLS.SEGMENT] = segment
+    }
+    if (segment) {
       segment.addAttribute('url', `${url.protocol}//${url.host}${url.pathname}`)
 
       url.searchParams.forEach((value, key) => {
         segment.addSpanAttribute(`request.parameters.${key}`, value)
       })
-      segment.addAttribute('procedure', request.method || 'GET')
-      request[SYMBOLS.SEGMENT] = segment
     }
   })
 
--- a/lib/instrumentation/undici.js
+++ b/lib/instrumentation/undici.js
@@ -16,6 +16,8 @@ const SYMBOLS = {
 }
 const { executionAsyncResource } = require('async_hooks')
 
+const UNKNOWN_HOST = '<unknown>'
+
 let diagnosticsChannel = null
 try {
   diagnosticsChannel = require('diagnostics_channel')
@@ -99,6 +101,15 @@ module.exports = function addUndiciChannels(agent, undici, modName, shim) {
     // eslint-disable-next-line guard-for-in
     for (const key in outboundHeaders) {
       request.addHeader(key, outboundHeaders[key])
+    }
+
+    const name = NAMES.EXTERNAL.PREFIX + UNKNOWN_HOST
+    const segment = shim.createSegment(name, recordExternal(UNKNOWN_HOST, 'undici'), parent)
+    if (segment) {
+      segment.start()
+      shim.setActiveSegment(segment)
+      segment.addAttribute('procedure', request.method || 'GET')
+      request[SYMBOLS.PARENT_SEGMENT] = segment
     }
   })
 
--- a/lib/instrumentation/undici.js
+++ b/lib/instrumentation/undici.js
@@ -150,10 +150,12 @@ module.exports = function addUndiciChannels(agent, undici, modName, shim) {
       segment.name = name
     } else {
       segment = shim.createSegment(name, recordExternal(url.host, 'undici'), parentSegment)
-      segment.start()
-      shim.setActiveSegment(segment)
-      segment.addAttribute('procedure', request.method || 'GET')
-      request[SYMBOLS.SEGMENT] = segment
+      if (segment) {
+        segment.start()
+        shim.setActiveSegment(segment)
+        segment.addAttribute('procedure', request.method || 'GET')
+        request[SYMBOLS.SEGMENT] = segment
+      }
     }
     if (segment) {
       segment.addAttribute('url', `${url.protocol}//${url.host}${url.pathname}`)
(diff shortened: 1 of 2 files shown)


garbados:
I've modified this patch so that the request segment is modified rather than another segment created for it. This allows tests to stay the same and users will not see a change in NRO, except that they will now be able to see issues from malformed requests.

My worry with *this* approach is that it may be possible for the segment to have been harvested between creation and modification, so that only the `'External/<unknown>'` segment appears in the interface. Is this something I should be worried about?

lib/instrumentation/undici.js
```diff
@@ -133,17 +144,25 @@ module.exports = function addUndiciChannels(agent, undici, modName, shim) {
     const url = new URL(urlString)
 
     const name = NAMES.EXTERNAL.PREFIX + url.host + url.pathname
-    const segment = shim.createSegment(name, recordExternal(url.host, 'undici'), parentSegment)
+    let segment
+    if (request[SYMBOLS.SEGMENT]) {
+      segment = request[SYMBOLS.SEGMENT]
+      segment.name = name
```
bizob2828, line 150: nice, glad this just works by reassigning name after it was created as unknown above

lib/instrumentation/undici.js
```diff
@@ -133,17 +144,25 @@ module.exports = function addUndiciChannels(agent, undici, modName, shim) {
     const url = new URL(urlString)
 
     const name = NAMES.EXTERNAL.PREFIX + url.host + url.pathname
-    const segment = shim.createSegment(name, recordExternal(url.host, 'undici'), parentSegment)
+    let segment
+    if (request[SYMBOLS.SEGMENT]) {
```
bizob2828, line 148: i'd follow the logic we have in other hook points where we do this and add it to the top of this hook `request:create`
```js
    const activeSegment = request[SYMBOLS.SEGMENT]
    if (!activeSegment) {
      return
    }
```

bizob2828 (changes requested):
A few suggestions but nice work on getting the segment rename 👏🏻 .  One big concern is you changed where segments are created and tests didn't change.  we have unit tests that should be updated to reflect new behavior and some will have to be updated if you address my changes I suggested

bizob2828:
I'll have to play around with it but it now may be easier to get this to feature parity like our http module where the connection time is a child segment of the external http request.

screenshot 2022-01-03 at 4 22 31 PM

I'm going to file a feature request for this

lib/instrumentation/undici.js
garbados: Yeah, though I am a little worried that an async harvest cycle might report this segment prior to this update -- what do you think?

lib/instrumentation/undici.js
```diff
@@ -133,17 +144,25 @@ module.exports = function addUndiciChannels(agent, undici, modName, shim) {
...
+      segment = request[SYMBOLS.SEGMENT]
+      segment.name = name
+    } else {
+      segment = shim.createSegment(name, recordExternal(url.host, 'undici'), parentSegment)
```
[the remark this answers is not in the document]
garbados: I found during tests that sometimes the segment hadn't been created yet (not sure why) so this catch was still necessary. Do you have any counsel on this?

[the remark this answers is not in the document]
bizob2828: it's because our async context access isn't perfect.  we shouldn't do this though if we're already creating the segment above. we can iterate on hardening accessing segment in a follow up

lib/instrumentation/undici.js
bizob2828: I'm not sure I understand what you mean.  we only remove segments when the response headers have been sent.

lib/instrumentation/undici.js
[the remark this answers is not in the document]
bizob2828: were you seeing the issues with segments not yet created with this feature flag on `agent.config.feature_flag.undici_async_tracking`?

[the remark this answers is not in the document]
bizob2828: was it that the segment wasn't created or when trying to get the active segment it wasn't defined? if the latter was this feature flag on `agent.config.feature_flag.undici_async_tracking`?

garbados:
> I'll have to play around with it but it now may be easier to get this to feature parity like our http module where the connection time is a child segment of the external http request.
> 
> screenshot 2022-01-03 at 4 22 31 PM
> 
> I filed this #1046

This certainly makes more sense than doing this segment-updating dance, as long as that's an appropriate approach.

bizob2828:
> > I'll have to play around with it but it now may be easier to get this to feature parity like our http module where the connection time is a child segment of the external http request.
> > screenshot 2022-01-03 at 4 22 31 PM
> > I filed this #1046
> 
> This certainly makes more sense than doing this segment-updating dance, as long as that's an appropriate approach.

It's a pattern we have to follow when we're implementing hook points.  It's the same thing we do in apollo server plugin.  As we discover more out about a given segment we decorate it accordingly.  https://github.com/newrelic/newrelic-node-apollo-server-plugin/blob/main/lib/create-plugin.js#L154 adds the name based on more metadata of the gql

garbados:
> > > I'll have to play around with it but it now may be easier to get this to feature parity like our http module where the connection time is a child segment of the external http request.
> > > screenshot 2022-01-03 at 4 22 31 PM
> > > I filed this #1046
> > 
> > 
> > This certainly makes more sense than doing this segment-updating dance, as long as that's an appropriate approach.
> 
> It's a pattern we have to follow when we're implementing hook points. It's the same thing we do in apollo server plugin. As we discover more out about a given segment we decorate it accordingly. https://github.com/newrelic/newrelic-node-apollo-server-plugin/blob/main/lib/create-plugin.js#L154 adds the name based on more metadata of the gql

This approach makes me think the approach I've got right now makes decent sense. FWIW I experimented with this child segment approach but it meant making big changes to the test suite, since there was suddenly a new segment involved. I can go back to that approach if you think it's appropriate.

lib/instrumentation/undici.js
garbados: Oh, well that's convenient. As for not understanding, it's just that I don't completely understand how harvest cycles work, so I was worried the segment would be removed before it was properly decorated.

michaelgoin: Segments are tied to the transaction lifecycle, so they won't generate their data or be cleaned up (at least from the transaction) until the transaction has ended. Although, these symbol usages are a bit different in that they propagate separate from that. If they happened to propagate after the transaction was ended then the segment updates would indeed be lost but then we also have other problems (which we actually might have, may or may not be related to this style of propagation tho).

lib/instrumentation/undici.js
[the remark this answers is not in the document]
michaelgoin: Hmm  yeah, we def only want to be creating and trusting that creation from a single place. We have a few problems we need to fix which may be causing problems here.

lib/instrumentation/undici.js
```diff
@@ -100,6 +102,15 @@ module.exports = function addUndiciChannels(agent, undici, modName, shim) {
     for (const key in outboundHeaders) {
       request.addHeader(key, outboundHeaders[key])
     }
+
+    const name = NAMES.EXTERNAL.PREFIX + UNKNOWN_HOST + (request.method || 'get')
```
michaelgoin, line 106: Given this style of doing things will be newer for externals, I'm thinking it would be useful to add trace level logging here for "creating external as unknown" and then later in the other event for "updating external to X". I'm somewhat curious what the existing logging shows, as well. I can imaging we may get some messages that are hard to tell why we can't add a segment because we don't have the context. Usually we'll have a name to go off of to find the external call but won't be able to here. That may just be a limitation we are stuck with based on how undici is written.

michaelgoin: Scanning the existing and http-outbound instrumentation, I don't think we typically include the verb in the name itself? As such, I don't think we should do so here unless I'm missing something.

If we were to include it, I'm guessing we'd want to use 'GET'.

Also, just want to double check `request.method` is available here? If it is but not specified means default to 'GET' (and below) then cool. If it always isn't available here, then defaulting to 'GET' would not actually be awesome cause we don't know. Prob is the case but just want to double-check.

michaelgoin:
Just noting this issue may impact things here or changes here may impact the issue: https://github.com/newrelic/node-newrelic/issues/972. Currently seems to be transaction conflation which is having the same transaction persist between requests and get reused.

garbados:
Closing for now as the PR has become somewhat stale. We'll return to it when we can.

PR #1111, opened 2022-03-02
instrument initializeUnorderedBulkOp, initializeOrderedBulkOp in mong…

dlants:
…odb v3

## Proposed Release Notes

 * Added support for `initializeUnorderedBulkOp`, and `initializeOrderedBulkOp` in mongodb v3 instrumentation.

## Details
I needed this monitoring for assessing load on our mongodb instance. Figured others could use it and maybe extend it to other versions of mongodb. I think it's a pretty essential feature that's currently missing from the driver.


dlants:
Hm. It seems like this implementation is not quite right. Even though the op is appearing correctly in the transaction breakdown and the database operations table, I do not see it in the histogram, and selecting it from the operations table shows "undefined undefined undefined" 
Screen Shot 2022-03-03 at 9 42 25 AM

lib/instrumentation/mongodb/common.js
```diff
@@ -84,6 +100,25 @@ common.makeQueryDescFunc = function makeQueryDescFunc(shim, methodName) {
   }
 }
 
+/**
+ * Sets up the desc for all instrumented query methods
```
bizob2828, line 104:
```suggestion
 * Sets up the desc for all instrumented bulk operations
```

lib/instrumentation/mongodb/common.js
```diff
@@ -84,6 +100,25 @@ common.makeQueryDescFunc = function makeQueryDescFunc(shim, methodName) {
...
+ *
+ * @param {Shim} shim
+ * @param {string} methodName
+ */
+common.makeBulkDescFunc = function makeBulkDescFunc(shim, methodName) {
```
bizob2828, line 109: the 2nd arg on this function is unused can be removed

bizob2828 (changes requested):
Hi @dlants, thanks for the contribution!  A few tiny comments but overall looks great!

Also, it looks like you have some linting errors.  There is a lot of noise with warnings that you can ignore but seems like you didn't have our dev deps installed as husky should've picked up staged changes and ran linting via lint-staged.  To fix this you can just run 

```sh
npx eslint --fix lib/instrumentation/mongodb/common.js
```

test/versioned/mongodb/bulk.tap.ts
```diff
@@ -0,0 +1,61 @@
+'use strict'
+
+const common = require('./collection-common')
+const mongoPackage = require('mongodb/package.json')
+const semver = require('semver')
+
+common.test('unorderedBulkOp', function unorderedBulkOpTest(t, collection, verify) {
```
bizob2828, line 7: this is one of the rough edges of our versioned tests but you need to add bulk.tap.js to the files array in `test/versioned/mongodb/package.json`. also these tests need to be semver gated to only run on v3

bizob2828: ```if(semver.satisfies(mongoPackage.version, '<4')) {```

bizob2828:
> Hm. It seems like this implementation is not quite right. Even though the op is appearing correctly in the transaction breakdown and the database operations table, I do not see it in the histogram, and selecting it from the operations table shows "undefined undefined undefined" Screen Shot 2022-03-03 at 9 42 25 AM

Ah ok.  Admittedly I just looked at the code.  But it also does look like the versioned tests you wrote do work.  A quick way to test just that file is to do this

```
cd test/versioned/mongodb
npm i mongodb@3 --no-save --no-package-lock
node bulk.tap.js
```

Also if you want to confirm the tests run against all v3 versions of mongo you can do this

```
npm run versioned mongodb
```

test/versioned/mongodb/bulk.tap.ts
```diff
@@ -0,0 +1,61 @@
+'use strict'
```
bizob2828, line 1: can you rename this to .js.  This doesn't contain any typescript

dlants:
> please log a ticket.
I think this is an issue with the one.newrelic.com web client, not the node agent. What's the best way to file an issue for that?

bizob2828:
> > please log a ticket.
> 
> I think this is an issue with the one.newrelic.com web client, not the node agent. What's the best way to file an issue for that?

If you can provide a repo case I can let the UI team know as I don't think that code is open source.

bizob2828 (changes requested):
It looks like this will fail on certain versions of mongodb 3.x.  Our CI will only run on the latest version for every major version line.  If you run 

```
npm run versioned mongodb
```

You can see the failure or you can just do this in `test/versioned/mongodb`

```
npm i mongodb@3.1.13 --no-save --no-package-lock
node bulk.tap.js
```

dlants:
>  If you can provide a repro case I can let the UI team know as I don't think that code is open source.

I wrote up the bug here: https://discuss.newrelic.com/t/bug-when-selecting-db-operations-from-db-operations-table/178575 If you could make it more visible to the UI team, that would be great.

bizob2828:
@dlants I triaged the failure and it appears to be broken up until 3.2.4.  Since those releases are almost 3 years old I'm going to update the semver guard. I think the [fix was here](https://github.com/mongodb/node-mongodb-native/commit/f548bd7).  I also plan on filing a follow up to get this support in 4.x. It looks like it works up until 4.1.4

bizob2828:
Actually I can't push to your remote.  Here's the diff that will fix the problem I was describing https://github.com/newrelic/node-newrelic/commit/a6c9e597372c37a370ad8ed1c90e43204139de3f

bizob2828:
Logged https://github.com/newrelic/node-newrelic/issues/1122 to get feature parity in v4 of the mongo instrumentation

bizob2828 (approved):
Looks great.  Again, thanks for your contribution @dlants !

PR #1121, opened 2022-03-11
Fix mysql test resource contention

michaelgoin:
## Proposed Release Notes

* Updated mysql and mysql2 versioned tests to run against their own databases on the MySQL instance.

## Details

Recently starting getting failures rotating between mysql and mysql2 tests which led me to assume contention on a table. Looking at the setup, these use the same user/database/table combination and part of the setup is to TRUNCATE (clear) the table. One of the failures was querying data from this table, so the theory seems to check out.

Often, I'll introduce generated GUID table names and such to prevent collisions but in this case we'll only ever have a single file from each running at a time. So to get a quick fix in, I instead have each suite run in their own database within the MySQL Server instance. I've wired these up to exposed variables exported from the `setup` files so this should be somewhat trivial to modify should we find our needs change here.

This **should** resolve the issue when a full --minor versioned test run occurs but there are never guarantees for race conditions that don't appear locally.

I also cleaned up some deprecated API warnings, misspellings and such in a separate commit.

Didn't take on consolidating the setup files for now.

### DIFF
--- a/test/versioned/mysql/setup.js
+++ b/test/versioned/mysql/setup.js
@@ -8,8 +8,15 @@
 const a = require('async')
 const params = require('../../lib/params')
 
+const USER = 'mysql_test_user'
+const DATABASE = 'mysql_agent_integration'
+const TABLE = 'test'
+
 module.exports = exports = setup
 exports.pool = setupPool
+exports.USER = USER
+exports.DATABASE = DATABASE
+exports.TABLE = TABLE
 
 function setup(mysql) {
   return a.series([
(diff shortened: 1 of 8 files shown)


test/versioned/mysql/setup.js
```diff
@@ -8,8 +8,15 @@
 const a = require('async')
 const params = require('../../lib/params')
 
+const USER = 'mysql_test_user'
+const DATABASE = 'mysql_agent_integration'
```
michaelgoin, line 12: These are the main difference. Although, user could have been kept the same.

test/versioned/mysql2/setup.js
```diff
@@ -8,8 +8,15 @@
 const a = require('async')
 const params = require('../../lib/params')
 
+const USER = 'mysql2_test_user'
+const DATABASE = 'mysql2_agent_integration'
```
michaelgoin, line 12: These are the main difference (as in mysql).

bizob2828 (approved):

PR #1171, opened 2022-05-04
Fix typo

eitoball:
## Proposed Release Notes

- Fixed a typo in lib/instrumentation/core/child_process.js


bizob2828 (approved):
yikes, thanks for fixing @eitoball.  I also searched code base and this seems to be the only offender of this issue.

PR #1173, opened 2022-05-04
Add magic comments to disable webpack dynamic require parsing

eitoball:
## Proposed Release Notes

- Add magic comments to where it requires node-builtin packages and configuration file in order to disable dynamic require parsing when bundling with webpack 5.17.0 or above.

## Details

- When bundling assets with Webpack, it parses `require` statements and rewrites to `__webpack_require__`. `__webpack_require__` requires from bundle file, not from external files. This behavior isn't desirable for some `require` statements such as requiring Node.js built-in packages or New Relic configuration file (`newrelic.js`). Adding `/* webpackIgnore: true */` (and set `module.parser.javascript.commonjsMagicComments` to `true`) disables this behavior.

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -1657,7 +1657,7 @@ function initialize(config) {
 
   let userConf
   try {
-    userConf = require(filepath).config
+    userConf = require(/* webpackIgnore: true */ filepath).config
   } catch (error) {
     logger.error(error)
 
--- a/lib/shimmer.js
+++ b/lib/shimmer.js
@@ -360,7 +360,7 @@ const shimmer = (module.exports = {
       let uninstrumented = null
 
       try {
-        uninstrumented = require(mojule)
+        uninstrumented = require(/* webpackIgnore: true */ mojule)
       } catch (err) {
         logger.trace('Could not load core module %s got error %s', mojule, err)
       }


bizob2828 (changes requested):
This might be a large ask but can you add some tests for this behavior?  I'm unsure why these are the only two places that you've added the ignore stanza.

eitoball:
> This might be a large ask but can you add some tests for this behavior? 

No, I cannot. I need to add webpack and check result by running webpack. This change is just adding comments, so wouldn't eslint be enough to check that syntax is correct?

> I'm unsure why these are the only two places that you've added the ignore stanza.

webpack re-writes `require` to `__webpack_require__` to require from bundled file when bundling assets. Where magic comments are added are places like:

1. requires configuration file (`newrelic.js`) at runtime. Requiring from bundled file is not desirable because file can be required from several places.
2. requires Node's standard library. webpack usually handles this case, but [this place](https://github.com/newrelic/node-newrelic/pull/1173/files#diff-3d0c4388e915f657046d7040af798b9376a45f953db048c457fe98a148d22b10L363) requires dynamically, so webpack cannot detect.

jordigh:
We have several other locations in the code that do dynamic requires, for example [here](https://github.com/newrelic/node-newrelic/blob/main/lib/shim/shim.js#L1727) and [here](https://github.com/newrelic/node-newrelic/blob/main/lib/proxy/grpc.js#L9). I assume that in your particular usage of webpack, only the two locations you changed were being collected by webpack. Should we be adding this webpack directive to all dynamic locations, not just the two you found?

bizob2828 (commented):
@eitoball after socializing this change internally on our team we are going to close this PR.  The agent doesn't officially support webpack and this change will not fully address that issue.
 
Have you tried using this [experimental package](https://github.com/newrelic-experimental/newrelic-webpack-plugin) with the agent and webpack?  Other customers have had success with it.  If not, you can also what you need here in webpack configuration.

Issue #1180, opened 2022-05-09
Logging: Collector - adhere to data usage metrics specification


coreyarnold:
Implement the [data usage metrics specification](https://source.datanerd.us/agents/agent-specs/blob/main/Data-Usage-Supportability-Metrics.md)

Generate top level metrics for the following endpoints
- [ ] Collector

Generate sub metrics for the collector endpoints where possible
- [ ] [Collector endpoints](https://source.datanerd.us/agents/agent-specs/blob/main/README.md)
- [ ] [logging endpoint](https://source.datanerd.us/agents/agent-specs/pull/570/files#diff-fdbbc261f836d83d3ffe5d685d33069c53f989eb54c57c2df547378acf8dd572)

Update the agent spec readme
- [ ] Update [readme](https://source.datanerd.us/agents/agent-specs/blob/main/README.md) to reflect adherence.

garbados:
In order to determine the byte size of a payload, we will need to log metrics from around https://github.com/newrelic/node-newrelic/blob/main/lib/collector/remote-method.js#L91 , after the payload is serialized but before it is sent. This will require passing the agent to the `RemoteMethod` class, which may not be desirable.

coreyarnold:
@garbados I updated the task list to callout the new log endpoint and point at the open PR.

PR #1183, opened 2022-05-10
Log collector data usage metrics

garbados:
## Proposed Release Notes

- Report metrics about the data usage of harvested data.

## Links

- Closes https://github.com/newrelic/node-newrelic/issues/1180

## Details

This PR logs metrics about harvested data according to the Data Usage Supportability Metrics specification, which is not a public document. Specifically, metrics are recorded *after* each operation -- preconnect, connect, shutdown, etc. -- in order to report the size of the payload sent and the payload received. Fields unused by the spec may contain garbage data that should be ignored.


lib/collector/api.js
```diff
@@ -68,21 +68,28 @@ function CollectorAPI(agent) {
...
+    'span_event_data',
+    'log_event_data'
+  ]
+    .map((methodName) => {
+      return new RemoteMethod(methodName, agent.config, initialEndpoint, agent.metrics)
+    })
+    .reduce((all, method) => {
+      all[method.name] = method
+      return all
+    }, {})
```
jordigh, line 92: I find the cognitive load of parsing the `.map(...).reduce(...)` to be slightly higher than the original copy-pastey code, but not significantly so to complain any louder than I already am.

jordigh (commented):
lgtm, kind of surprised that we can ensure that much numerical precision in the tests, but glad to see it works.

lib/collector/api.js
garbados: I just can't stand that kind of copypasta.

garbados: I've modified the way this is done to use a for-loop instead, which should be more readable than my sicko map-reduce shenanigans.

jordigh: It's not that important to me, so sorry to ask if you changed your mind and also sorry if I'm misunderstanding, but I am not seeing the change, did you push it or am I looking at the wrong version of the code?

lib/stats/index.js
```diff
@@ -63,9 +63,15 @@ function recordValueInMillis(totalTime, exclusiveTime) {
   this.recordValue(totalTime * FROM_MILLIS, exclusiveTime >= 0 ? exclusiveTime * FROM_MILLIS : null)
 }
 
-Stats.prototype.recordValueInBytes = function recordValueInBytes(bytes, exclusiveBytes) {
-  exclusiveBytes = exclusiveBytes || bytes
-  this.recordValue(bytes / BYTES_PER_MB, exclusiveBytes / BYTES_PER_MB)
+Stats.prototype.recordValueInBytes = function recordValueInBytes(bytes, exclusiveBytes, exact) {
```
bizob2828, line 66: I don't see any corresponding unit test updates for this.

lib/collector/api.js
```diff
@@ -68,20 +68,24 @@ function CollectorAPI(agent) {
...
-    traces: new RemoteMethod('transaction_sample_data', agent.config, initialEndpoint),
-    shutdown: new RemoteMethod('shutdown', agent.config, initialEndpoint),
-    events: new RemoteMethod('analytic_event_data', agent.config, initialEndpoint),
-    customEvents: new RemoteMethod('custom_event_data', agent.config, initialEndpoint),
-    queryData: new RemoteMethod('sql_trace_data', agent.config, initialEndpoint),
-    errorEvents: new RemoteMethod('error_event_data', agent.config, initialEndpoint),
-    spanEvents: new RemoteMethod('span_event_data', agent.config, initialEndpoint),
-    logEvents: new RemoteMethod('log_event_data', agent.config, initialEndpoint)
+  this._methods = {}
+  for (const [accessor, name] of [
```
bizob2828, line 72: nice refactor 👏🏻

lib/collector/remote-method.js
```diff
@@ -88,7 +90,32 @@ RemoteMethod.prototype.invoke = function invoke(payload, nrHeaders, callback) {
...
-      this._post(serialized, nrHeaders, callback)
+      this._post(
+        serialized,
+        nrHeaders,
+        function onResponse(err, res) {
+          if (err) {
+            return callback(err)
+          }
+          if (this._metrics) {
+            const payloadByteLength = byteLength(serialized)
```
bizob2828, line 101: it might be good to extract out this bit of code into its own function so it's clearer the intention when first reading it

lib/collector/remote-method.js
```diff
@@ -88,7 +90,32 @@ RemoteMethod.prototype.invoke = function invoke(payload, nrHeaders, callback) {
       if (err) {
         return callback(err)
       }
-      this._post(serialized, nrHeaders, callback)
+      this._post(
+        serialized,
+        nrHeaders,
+        function onResponse(err, res) {
+          if (err) {
```
bizob2828, line 97: I'm not sure you want to do this. if the request to collector endpoint fails we currently aren't recording those metrics, am I reading that correctly?

lib/collector/remote-method.js
```diff
@@ -88,7 +90,32 @@ RemoteMethod.prototype.invoke = function invoke(payload, nrHeaders, callback) {
...
+            return callback(err)
+          }
+          if (this._metrics) {
+            const payloadByteLength = byteLength(serialized)
+            const responseByteLength = res.payload ? byteLength(JSON.stringify(res.payload)) : 0
+            this._metrics.measureBytes(
+              `${NAMES.SUPPORTABILITY.NODEJS}/Collector/Output/Bytes`,
```
bizob2828, line 104: you should probably put all this into a new metric constant in `lib/metrics/names`

bizob2828: maybe 2 to account for the top level and then the sub destination

lib/collector/remote-method.js
```diff
@@ -88,7 +90,32 @@ RemoteMethod.prototype.invoke = function invoke(payload, nrHeaders, callback) {
...
+              true
+            )
+            this._metrics.measureBytes(
+              `${NAMES.SUPPORTABILITY.NODEJS}/Collector/${this.name}/Output/Bytes`,
+              payloadByteLength,
+              responseByteLength,
+              true
+            )
+          }
+          return callback(null, res)
```
bizob2828, line 116: i think here should just always return `callback(err, res)` based on my comment above

test/unit/collector/remote-method.test.js
```diff
@@ -725,3 +728,68 @@ tap.test('when generating the User-Agent string', (t) => {
...
+      host: agent.config.host,
+      port: agent.config.port
+    }
+  })
+
+  t.afterEach(() => {
+    agent && helper.unloadAgent(agent)
+  })
+
+  t.test('should aggregate bytes of uploaded payloads', async (t) => {
```
bizob2828, line 751: I'm not sure if we have to be exhaustive with every endpoint but a few tests I could see valuable are:
 * asserting response byte size(you'll have to probably stub the response payload)
 * having the remote method return an error and still verify metrics are recording(current bug in PR)

bizob2828 (changes requested):
for the most part this looks good.  I have not tested yet but I will once some comments are addressed.

lib/collector/remote-method.js
garbados: That's true. I can modify it to be more durable.

lib/collector/remote-method.js
garbados: Fair enough!

lib/collector/api.js
garbados: I think you're looking at the wrong version of the code. Have you refreshed the page?

jordigh: I swear I reloaded the page, so anyway, yeah, it looks good now.

garbados: Sometimes that's just how browsers are. Sorry for the trouble!

test/unit/collector/remote-method.test.js
garbados: I've added several tests, including the ones you outlined here. Thanks for suggesting them!

garbados:
CI passes! I've made a few style changes and added several tests. Looking forward to further feedback :)

lib/collector/remote-method.js
```diff
@@ -16,7 +16,11 @@ const Sink = require('../util/stream-sink')
 const agents = require('./http-agents')
 const certificates = require('./ssl/certificates')
 const isValidLength = require('../util/byte-limit').isValidLength
-const NAMES = require('../metrics/names')
+const { SUPPORTABILITY, DATA_USAGE } = require('../metrics/names')
+
+function getMetricName(name) {
+  return `${SUPPORTABILITY.NODEJS}/Collector/${name}/${DATA_USAGE.SUFFIX}`
```
bizob2828, line 22: you created a `DATA_USAGE.COLLECTOR` to take care of this prefix. I'd update it here

lib/collector/remote-method.js
```diff
@@ -74,6 +78,16 @@ RemoteMethod.prototype.serialize = function serialize(payload, callback) {
  * @param {Function} callback          What to do next. Gets passed any error.
  */
 RemoteMethod.prototype.invoke = function invoke(payload, nrHeaders, callback) {
+  const reportDataUsage = (sent, received) => {
```
bizob2828, line 81: nice consolidation of this logic. However, this is creating a function every time invoke is getting called. I'm sure it will be optimized by v8 but you could just put on the instance of RemoteMethod as `this_metrics` will be there right?

test/unit/stats.test.js
```diff
@@ -154,4 +154,26 @@ describe('Stats', function () {
     it('should store bytes as bytes, rescaling only at serialization')
     it('should store time as nanoseconds, rescaling only at serialization')
   })
+
+  describe('recordValueInBytes', function () {
```
bizob2828, line 158: 👏🏻

bizob2828 (changes requested):
few comments, but this looks very close

lib/collector/remote-method.js
garbados: `DATA_USAGE.COLLECTOR` is `${SUPPORTABILITY.NODEJS}/Collector/${DATA_USAGE.SUFFIX}`. Are you saying I should modify it to just `${SUPPORTABILITY.NODEJS}/Collector`?

lib/collector/api.js
```diff
@@ -68,20 +68,24 @@ function CollectorAPI(agent) {
...
+    ['traces', 'transaction_sample_data'],
+    ['shutdown', 'shutdown'],
+    ['events', 'analytic_event_data'],
+    ['customEvents', 'custom_event_data'],
+    ['queryData', 'sql_trace_data'],
+    ['errorEvents', 'error_event_data'],
+    ['spanEvents', 'span_event_data'],
+    ['logEvents', 'log_event_data']
+  ]) {
+    const method = new RemoteMethod(name, agent.config, initialEndpoint, agent.metrics)
```
bizob2828, line 87: this won't work. I noticed it while testing with a real app and not seeing these new output bytes metrics.  If you test for yourself and put a break point here `agent.metrics` is undefined.  This is because the collector instance is created before the metrics aggregator: https://github.com/newrelic/node-newrelic/blob/main/lib/agent.js#L77

garbados: Welp, that's frustrating. I guess we might need to pass the whole agent to the remote methods after all, to preserve that reference rather than passing undefined by value?

lib/collector/remote-method.js
bizob2828: oh i see.  I'm suggesting creating yet another constant so you don't have the raw string of `Collector` here

bizob2828: or you could pare the `DATA_USAGE.COLLECTOR` to exclude the suffix and just construct that below with 2 diff constants

garbados: all valid approaches. i'll follow up when i can

lib/collector/remote-method.js
garbados: good point. i'll refactor accordingly soon

lib/collector/api.js
jordigh: I think passing in the whole agent is the easiest thing to do. I started playing around trying some other more sophisticated things but I couldn't figure out how to do it simply.

lib/collector/remote-method.js
jordigh: Did that refactor, does it look good?

lib/collector/remote-method.js
jordigh: I've split it up into a prefix and a suffix.

lib/collector/remote-method.js
garbados: Sure does!

lib/collector/remote-method.js
```diff
@@ -68,6 +68,23 @@ RemoteMethod.prototype.serialize = function serialize(payload, callback) {
...
 
+/**
+ * This records the amount of data usage per function call and sends
+ * it to the metrics.
+ *
+ * @param {string} sent        The serialized object we sent
+ * @param {object} received    The raw object we got back as a response
+ */
+RemoteMethod.prototype._reportDataUsage = function reportDataUsage(sent, received) {
+  if (this._agent) {
```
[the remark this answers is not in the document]
garbados: In unit tests, RemoteMethod is usually instantiated without an agent parameter, such as:
```js
method = new RemoteMethod('test')
```
In this case, the guard is necessary.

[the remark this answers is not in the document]
bizob2828: not a fan of these kind of checks to appease sparse tests.  We have other params to remote method but it looks like those signatures are updated when it needs more than the 1st arg to the class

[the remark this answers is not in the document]
garbados: Great, I'll remove those guards and update the unit tests accordingly.

garbados:
Checking in here, I'm fixing the CI as-is before adding an integration test regarding this new feature.

garbados:
Alright! CI passes. Ready for review!

PR #1189, opened 2022-05-12
post-release: fetch the full repo, not a shallow clone

jordigh:
By default, github's actions/checkout action will do a shallow clone
of a single commit, without history. Since our workflow requires a
merge, we need the full history in order to be able to perform the
merge.


jordigh:
I just ran the workflow from this branch and it succeeded this time:

https://github.com/newrelic/node-newrelic/runs/6425238283?check_suite_focus=true

bizob2828 (approved):

Issue #1236, opened 2022-06-02
amqplib 0.10.0 fails test suite


garbados:
Like it says on the tin, the recently released v0.10.0 of amqplib fails our test suite. It's possible our instrumentation is itself broken by the new version. We should resolve this soon as it's blocking CI passing for other PRs.

PR #1237, opened 2022-06-02
ignore latest amqplib which breaks instrumentation

garbados:
## Proposed Release Notes

* Removed latest amqplib (v0.10.0) from versioned testing, as it fails to instrument properly.

## Links

* Schließt nicht https://github.com/newrelic/node-newrelic/issues/1236

## Details

This library is breaking CI and figuring out how to fix our instrumentation is not simple or obvious, so we're going to ignore it for now and come back to it after current blocked PRs are merged.

### DIFF
--- a/test/versioned/amqplib/package.json
+++ b/test/versioned/amqplib/package.json
@@ -8,7 +8,7 @@
         "node": ">=12"
       },
       "dependencies": {
-        "amqplib": ">=0.5.0"
+        "amqplib": ">=0.5.0 <0.10.0"
       },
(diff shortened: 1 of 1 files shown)


garbados:
That bit of German you see above means "Does not close" because if I wrote "Does not close" GitHub would interpret it as the same as "Closes" 🙃

jordigh (approved):
Okay, we've got other things to do for now, but looks like we'll have to figure out how to fix our instrumentation of this for later.

PR #1250, opened 2022-06-07
upgrades to patch vulnerability in protobufjs

ankoor7:
## Proposed Release Notes

Update @grpc/proto-loader to v0.6.13 to pickup security fixes in protobufjs.

## Links
Solves this security issue: https://github.com/advisories/GHSA-g954-5hwp-pp24

## Details
@grpc/proto-loader uses a fixed version of protobufjs.

The package protobufjs before 6.11.3 are vulnerable to Prototype Pollution which can allow an attacker to add/modify properties of the Object.prototype.

### DIFF
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -296,7 +296,7 @@ This product includes source derived from [@grpc/grpc-js](https://github.com/grp
 
 ### @grpc/proto-loader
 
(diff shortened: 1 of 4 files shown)


bizob2828 (commented):
@ankoor7 Is this causing you an issue?  The dependency you updated has protobuf as ` "protobufjs": "^6.10.0",` which will update accordingly to 6.11.3 the next time you re-install the agent(newrelic).  I'll happily merge but this change shouldn't be necessary to resolve your issue.  You can even see that by using the [semver calculator](https://semver.npmjs.com/)

screenshot 2022-06-07 at 11 16 22 AM

bizob2828 (approved):

PR #1274, opened 2022-07-08
gh-actions: add support statement to our automated release notes

jordigh:
* Added a support statement to our release notes (closes [NR-29755](https://issues.newrelic.com/browse/NR-29755))

### DIFF
--- a/bin/create-docs-pr.js
+++ b/bin/create-docs-pr.js
@@ -35,6 +35,11 @@ program.option(
 const RELEASE_NOTES_PATH =
   './src/content/docs/release-notes/agent-release-notes/nodejs-release-notes'
 
+const SUPPORT_STATEMENT = `
+### Support statement:
+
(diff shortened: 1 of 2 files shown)


bin/create-docs-pr.js
```diff
@@ -108,7 +113,7 @@ async function getReleaseNotes(version, releaseNotesFile) {
   const versionChangeLog = sections.find((section) => section.startsWith(version))
   // e.g. v7.1.2 (2021-02-24)\n\n
   const headingRegex = /^v\d+\.\d+\.\d+ \((\d{4}-\d{2}-\d{2})\)\s+/
-  const body = versionChangeLog.replace(headingRegex, '')
+  const body = versionChangeLog.replace(headingRegex, '') + SUPPORT_STATEMENT
```
michaelgoin, line 116: If we got to a point we could trigger these off of GitHub releases and use those notes, we could remove the need to duplicate adding this. Not a big deal currently.

michaelgoin (approved):

bin/create-docs-pr.js
jordigh: I couldn't figure out how to do it since the docs PR requires "owernship", to be assigned to a human on our team that will interface with the docs team. Maybe if we added our credentials to the github action that creates the release, that way github actions can generate a PR with the name of the person who triggered the release?

PR #1282, opened 2022-07-14
Bumps test self-signed cert to use sha256 and default size.

michaelgoin:
## Proposed Release Notes

* Updated self-signed test cert to use sha256 and default size.

## Links

* https://issues.newrelic.com/browse/NR-33629

## Details

Node 18 requires a hiring minimum cert algorithm than sha1. This caused tests to break with: `error:0A00018E:SSL routines::ca md too weak`.

Developers will need to clear out these files:
* `test/lib/self-signed-test-certificate.crt`
* `test/lib/test-key.key`
* test/lib/ca-*`

Then run `npm run prepare-test`.

### DIFF
--- a/bin/ssl.sh
+++ b/bin/ssl.sh
@@ -45,7 +45,7 @@ if [ -e $CERTIFICATE ]; then
 fi
 
 # generates an RSA key
-openssl genrsa -out $SSLKEY 1024
+openssl genrsa -out $SSLKEY
 
 # ca-index is the "certificate authority" database
 # and ca-serial is a file that openssl will read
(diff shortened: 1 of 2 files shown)


bin/ssl.sh
```diff
@@ -45,7 +45,7 @@ if [ -e $CERTIFICATE ]; then
 fi
 
 # generates an RSA key
-openssl genrsa -out $SSLKEY 1024
```
michaelgoin, line 48: This change wasn't necessary but didn't see why we needed to be forcing a smaller size than the default.

bizob2828 (approved):
verified this fixed my issues in node 18

PR #1287, opened 2022-07-15
Removes restify as a dev dependency.

michaelgoin:
## Proposed Release Notes

* Removed `restify` as a dev dependency.

## Details

Recent changes have removed usage of `restify` outside of the versioned tests.


bizob2828 (approved):
nice. we should probably do this for all instrumentation of libraries

PR #1311, opened 2022-08-01
Bumps minimum external instrumentation modules to new major versions.

michaelgoin:
## Proposed Release Notes

* Bumped external instrumentation to new major versions.

## Links

* Closes: https://issues.newrelic.com/browse/NR-35667

## Details

Test utils had already been updated with a separate PR.


bizob2828 (changes requested):
The native metrics optional dependency needs updated too, right?

bizob2828 (approved):

PR #1321, opened 2022-08-03
Revert "Dropped async lib from pricing meminfo and cpu tests"

mrickard:
Reverts newrelic/node-newrelic#1319


mrickard:
No need to revert.

PR #1348, opened 2022-09-09
Add ES Module loader initial implementation

jmartin4563:
## Proposed Release Notes
 * Added loader hook to support instrumentation of CommonJS dependencies in ES Module applications

## Links
Closes [NEWRELIC-3121](https://issues.newrelic.com/browse/NEWRELIC-3121)

## Details
* Created `esm-loader.mjs` with `resolve` hook, based on prototype by @michaelgoin (h/t)
* Added new npm script for running our ESM flavored unit tests (`npm run unit:esm`), and updated CI to run this script
* Introduced testdouble, a library that allows us to fake out `import` statements like proxyquire does for `require`
* Updated eslint configuration to allow for `import()`, and updated linting commands to run on `.mjs` files


esm-loader.mjs
```diff
@@ -0,0 +1,59 @@
...
+ *        The Node.js default resolve hook
+ * @returns {Promise} Promise object representing the resolution of a given specifier
+ */
+export async function resolve(specifier, context, nextResolve) {
+  if (!newrelic.shim) {
+    return nextResolve(specifier)
+  }
+
+  const instrumentationApi = newrelic.shim
+  const logger = instrumentationApi.logger.child({ component: 'esm-loader' })
```
michaelgoin, line 32: One thing to note is via this approach, we'll be instantiating a child logger every single invocation of `resolve`... so once per every import. This is probably pretty cheap / not noticeable but might recommend avoiding any extra work we can, that is obvious,  that can add to startup times for customer applications.

.eslintrc.js
```diff
@@ -8,6 +8,9 @@
 module.exports = {
   extends: ['@newrelic', 'plugin:jsdoc/recommended'],
   plugins: ['jsdoc'],
+  parserOptions: {
+    ecmaVersion: '2020'
```
michaelgoin, line 12: I think ideally when we bump allowed versions based on what's allowed (deprecating older node versions), we update the shared definition so all projects are consistent. Fine to land this sort of thing in the project but might follow-up with updating the shared eslint module and then removing here.

Also, I'm not sure what was needed to bump that version but calling out that this will probably allow "spread parameters after optional chaining" which is not available in Node 14. So there's a small risk of introducing something that won't work. Hopefully that would get caught by tests but noting we have been burned at least once since I've been involved with the Node team.

esm-loader.mjs
```diff
@@ -0,0 +1,59 @@
...
+
+      logger.debug(
+        `Registered CommonJS instrumentation for ${specifier} under ${instrumentationDefinitionCopy.moduleName}`
+      )
+    }
+
+    logger.debug(`${specifier} is not a CommonJS module, skipping for now`)
+  }
+
+  return { url, format }
```
michaelgoin, line 58: This would only matter for large projects, or may never matter, but the destructuring to later create an object to return the exact elements does result in an unnecessary allocation on a potentially hot startup path (application with many import statements). It may be negligible in this case but calling out as that is something to consider with agent development that often has less of an impact in CRUD/traditional development. Allocating too much on heavily invoked paths can start impacting GC, even when objects are short lived, which can have non-obvious impacts to perf sometimes (or obvious other times). A more obvious case where this might come up is instrumenting a database "read" method.

esm-loader.mjs
```diff
@@ -0,0 +1,59 @@
...
+      instrumentationDefinitionCopy.moduleName = url.replace('file://', '')
+
+      shimmer.registerInstrumentation(instrumentationDefinitionCopy)
+
+      logger.debug(
+        `Registered CommonJS instrumentation for ${specifier} under ${instrumentationDefinitionCopy.moduleName}`
+      )
+    }
+
+    logger.debug(`${specifier} is not a CommonJS module, skipping for now`)
```
michaelgoin, line 55: Since the 'commonjs' `if` block does not return, I believe this log message will be logged regardless. Seems like it may not happen anyways as we are unlikely to have anything registered unless one of these projects converts (or does hybrid support? maybe?) but probably want to clean up anyways.

esm-loader.mjs
```diff
@@ -0,0 +1,59 @@
...
+
+  if (instrumentationDefinition) {
+    logger.debug(`Instrumentation exists for ${specifier}`)
+
+    if (format === 'commonjs') {
+      // ES Modules translate import statements into fully qualified filepaths, so we create a copy of our instrumentation under this filepath
+      const instrumentationDefinitionCopy = Object.assign({}, instrumentationDefinition)
+
+      // Stripping the prefix is necessary because the code downstream gets this url without it
+      instrumentationDefinitionCopy.moduleName = url.replace('file://', '')
```
michaelgoin, line 46: To make sure it doesn't get lost, is there a specific story for tweaking/updating/fixing the automatically generated metrics for the instrumentation for these use cases? If not, you'll want a story (or to fix in this PR). 

Without updating that code, instead of generating metrics like: 
* `Supportability/Features/Instrumentation/OnRequire/express`
* `Supportability/Features/Instrumentation/OnRequire/express/Version/4`

You'll end up with metrics like:
* `Supportability/Features/Instrumentation/OnRequire//Users/theuser/some/path/to/app/node_modules/express/index.js`
* `Supportability/Features/Instrumentation/OnRequire//Users/theuser/some/path/to/app/node_modules/express/index.js/Version/4\`

michaelgoin: Ah nevermind, I found it mentioned in https://issues.newrelic.com/browse/NEWRELIC-3120. Leaving for note/reminder I checked this but I'll quickly resolve it.

THIRD_PARTY_NOTICES.md
```diff
@@ -8977,6 +8978,35 @@ CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 ```
 
+### testdouble
+
+This product includes source derived from [testdouble](https://github.com/testdouble/testdouble.js) ([v3.16.6](https://github.com/testdouble/testdouble.js/tree/v3.16.6)), distributed under the [MIT License](https://github.com/testdouble/testdouble.js/blob/v3.16.6/LICENSE.txt):
+
+```
+The MIT License (MIT)
```
michaelgoin, line 8986: MIT license ✔️

michaelgoin (commented):
Very clean implementation, nice work!

A few comments around doing unnecessary work. You/the team can decide if you want to care or not this early. My main thing is they are easily identifiable and I'm not sure how easy to track down if problems do come up with customers.

I'd normally look for an integration test with this sort of thing to confirm actual functionality but I believe that is intended to be tackled in a separate story. So given that, the unit testing looks legit.

.eslintrc.js
jmartin4563: The reason I introduced this is that doing `await import()` without it caused a parsing error on my unit test file, causing eslint to fail completely on the file (which would then fail CI). Fixing that is not worth the risk of potentially allowing introduction of bugs for v14 IMO, so what do you think about undoing this change and then adding my ESM unit test to `.eslintignore`?

michaelgoin: We have at times had to have a config specific to the tests folder to bump the ecmascript version earlier than the code base for testing newer features. For example, async-await while still supporting very old node versions.

It looks like we still have a file in there, so I might just bump that one to start: `test/.eslintrc.js`.

esm-loader.mjs
jmartin4563: whoops! that's my bad, will fix :)

esm-loader.mjs
jmartin4563: Updated to use the logging module directly, as that is a pattern applied by basically everything else in this repo

esm-loader.mjs
jmartin4563: Better to be cautious now than have to try and track down these GC thrashes in the wild

esm-loader.mjs
```diff
@@ -0,0 +1,59 @@
...
+ *        String identifier in an import statement or import() expression
+ * @param {object} context
+ *        Metadata about the specifier, including url of the parent module and any import assertions
+ *        Optional argument that only needs to be passed when changed
+ * @param {Function} nextResolve
+ *        The Node.js default resolve hook
+ * @returns {Promise} Promise object representing the resolution of a given specifier
+ */
+export async function resolve(specifier, context, nextResolve) {
+  if (!newrelic.shim) {
```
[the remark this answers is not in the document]
jmartin4563: Checking for that would only cover the case where the configuration explicitly disables it, it doesn't cover the case where the agent errors out on bootstrap. Checking for the `shim` or `agent` property covers both cases in one `if`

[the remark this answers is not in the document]
jordigh: Can we check for the agent itself, then? That seems a little more intentional than checking to see if the shim API is there.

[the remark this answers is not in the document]
jmartin4563: Sure! Will update to assert `newrelic.agent`

esm-loader.mjs
```diff
@@ -0,0 +1,59 @@
...
+ *        String identifier in an import statement or import() expression
+ * @param {object} context
+ *        Metadata about the specifier, including url of the parent module and any import assertions
+ *        Optional argument that only needs to be passed when changed
+ * @param {Function} nextResolve
+ *        The Node.js default resolve hook
+ * @returns {Promise} Promise object representing the resolution of a given specifier
+ */
+export async function resolve(specifier, context, nextResolve) {
+  if (!newrelic.agent) {
```
bizob2828, line 30: I think the check for `newrelic.shim` was done because it follows the patterns of our external modules.  I think checking for agent, shim or awsLambda is enough. but frankly everything we do in here I think is safe without checking anything agent properties but wasn't sure the intention when @michaelgoin wrote this into the prototype.

esm-loader.mjs
```diff
@@ -0,0 +1,59 @@
...
+
+      logger.debug(
+        `Registered CommonJS instrumentation for ${specifier} under ${instrumentationDefinitionCopy.moduleName}`
+      )
+    } else {
+      logger.debug(`${specifier} is not a CommonJS module, skipping for now`)
+    }
+  }
+
+  return resolvedModule
```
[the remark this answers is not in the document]
bizob2828: we need to await the parent(nextResolve) so we can get the resolved path.  In the case the agent doesn't exist, still question this, you can see above, it means continue the loader chain.  But we're essentially halting the loader chain.  But we are going to suggest to be last in that chain. so technically we could just return without calling the parent resolve

package.json
```diff
@@ -232,6 +234,7 @@
     "THIRD_PARTY_NOTICES.md",
     "lib/",
     "bin/tracetractor",
-    "bin/test-naming-rules.js"
+    "bin/test-naming-rules.js",
+    "esm-loader.mjs"
```
bizob2828, line 238: nice catch. this would've been a downer to get through the work and forget to include this here.

test/unit/esm-loader.test.mjs
```diff
@@ -0,0 +1,181 @@
...
+        }
+      }
+    }
+
+    await td.replaceEsm('../../index.js', {}, fakeNewrelic)
+    await td.replaceEsm('../../lib/shimmer.js', {}, fakeShimmer)
+    await td.replaceEsm('../../lib/logger.js', {}, fakeLogger)
+
+    // eslint-disable-next-line node/no-unsupported-features/es-syntax
+    loader = await import('../../esm-loader.mjs')
```
bizob2828, line 56: It's worth noting this is ignored because of https://github.com/mysticatea/eslint-plugin-node/issues/250.  I also logged a ticket to update our plugin so we can remove these: https://issues.newrelic.com/browse/NEWRELIC-3321

esm-loader.mjs
```diff
@@ -0,0 +1,59 @@
...
+ * @param {Function} nextResolve
+ *        The Node.js default resolve hook
+ * @returns {Promise} Promise object representing the resolution of a given specifier
+ */
+export async function resolve(specifier, context, nextResolve) {
+  if (!newrelic.agent) {
+    return nextResolve(specifier)
+  }
+
+  const resolvedModule = await nextResolve(specifier)
```
bizob2828, line 34: might be good to put a comment explaining why we are doing this here exactly.

bizob2828 (dismissed):
Good work on the paring down from prototype to actual solution.  Overall looks great.  I tested it with a branch where I'm porting versioned tests and all is well.

.eslintrc.js
bizob2828: for future posterity 16.3.0 is when es2020 spec is 100% implemented: see https://node.green/

bizob2828 (approved):

esm-loader.mjs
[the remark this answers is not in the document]
jordigh: My concern here was having a function that sometimes returns a promise and sometimes does not. I thought that might break API expectations, but if we tell our users, "put us last", then maybe this doesn't matter.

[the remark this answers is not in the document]
bizob2828: The API doesn't change. It's async we just do all the awaiting ourselves

mrickard (approved):

PR #1368, opened 2022-09-29
NEWRELIC-3769 add esm package test

jmartin4563:
## Proposed Release Notes
* Added test for asserting ESM loader functionality on ESM-only package

## Links
Closes [NEWRELIC-3769](https://issues.newrelic.com/browse/NEWRELIC-3769)

## Details
* `parse-json` is a small ESM package from the ecosystem that we can use as our guinea pig until something we actually want to implement is ESM-only
* Add "instrumentation" for `parse-json`, once we have API methods for doing this, we should update to use those instead of manually adding our instrumentation
* Enhanced test loader to support `load`


bizob2828 (approved):
LGTM, very simple. one small suggestion

test/versioned/esm-package/parse-json-instrumentation.mjs
```diff
@@ -0,0 +1,39 @@
...
+  shim.wrap(parseJson, 'default', function wrappedParseJsonLib(_shim, orig) {
+    return function wrappedParseJsonFunc() {
+      const result = orig.apply(this, arguments)
+      result.isInstrumented = true
+      return result
+    }
+  })
+
+  shim.wrap(parseJson, 'JSONError', function wrappedParseJsonLib(_shim, orig) {
+    // const WrappedError = function wrappedError() {
```
kibertoad, line 20: why is this commented out?

jmartin4563: @bizob2828 is helping me out with some debugging, so I pushed up what I was working on to make our debugging session faster/easier, will put this PR back into draft until said debugging is done

bizob2828 (approved):

PR #1384, opened 2022-10-21
NEWRELIC-4643 url encoded characters in ESM Loader filepaths 

jmartin4563:
## Proposed Release Notes
Updated ES Module loader to register instrumentation correctly when filepaths include url encoded characters

## Links
Closes NEWRELIC-4643

## Details
* Updated CJS instrumentation mechanism to make sure we store instrumentation copy under the decoded path
* Updated ESM instrumentation mechanism to make sure the code rewrite uses decoded path when wrapping
* `fileURLToPath` also handles the stripping of the file prefix off for us, and actually uncovered issue with how we were mocking URLs in the tests (we were missing a `/`)
* Updated our tests to include a copy of our ESM instrumentation in a folder with a space `test/lib/edge case/` to ensure correct behavior


bizob2828 (approved):
Nice tests to prove the fix 👏🏻 .

I tested this manually before and after fix with an app that contained a file path with a space.  All is well now 🎉

PR #1392, opened 2022-10-26
NR-57994 Requested changes for code coverage collection

jmartin4563:
## Links
Follow up from #1385 

## Details
There were a couple of late breaking requests for the original codecov implementation, so this PR addresses them
1. Move Async Local Context tests out into own GHA step, they're a bit flaky and having them lumped with the others means that re-runs take forever
2. Only run versioned tests on Async Local Context, as our integration/unit tests for Async Local Context are already dealt with by the original runs
3. Added a "cron" like job to run 4/5am eastern to gather coverage of versioned tests from the `main` branch, as we can only gather coverage for versioned tests with the `--major` flag due to c8 OOM'ing with the `--minor` flag

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -120,7 +120,7 @@ jobs:
         files: lcov.info
         flags: versioned-tests-${{ matrix.node-version }}
 
-  async-local-context-tests:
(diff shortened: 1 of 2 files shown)


bizob2828 (approved):
Very nice

PR #1397, opened 2022-10-27
Drop async lib from environment unit test

mrickard:
Replaced async lib with async functions
Replaced fs with fs/promises
Replaced rimraf with fs/promises

## Proposed Release Notes
In environment unit test: 
* Replaced async lib with async functions
* Replaced fs with fs/promises
* Replaced rimraf with fs/promises

## Links
https://issues.newrelic.com/browse/NR-39257

### DIFF
--- a/test/unit/environment.test.js
+++ b/test/unit/environment.test.js
@@ -257,8 +257,8 @@ describe('the environment scraper', function () {
 
   describe('with symlinks', function () {
     const nmod = path.resolve(__dirname, '../helpers/node_modules')
-    const makeDir = async (dirp, mkdirDb) => {
-      const code = await fs
+    const makeDir = (dirp) => {
+      return fs
         .mkdir(dirp)
         .then(() => null)
         .catch((err) => {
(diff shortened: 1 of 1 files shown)


test/unit/environment.test.js
```diff
@@ -259,11 +257,43 @@ describe('the environment scraper', function () {
...
+        .mkdir(dirp)
+        .then(() => null)
+        .catch((err) => {
+          if (err.code !== 'EEXIST') {
+            return err
+          }
+          return null
+        })
+      // vestigial--mkdirDb is the async callback, invisibly appended to arguments when apply is used
+      if (mkdirDb) {
```
mrickard, line 271: I could see removing this conditional block entirely, along with the `mkdirDb` property.

test/unit/environment.test.js
```diff
@@ -259,11 +257,38 @@ describe('the environment scraper', function () {
 
   describe('with symlinks', function () {
     const nmod = path.resolve(__dirname, '../helpers/node_modules')
+    const makeDir = (dirp) => {
```
jmartin4563, line 260: It seems weird to use both async/await and then/catch style, do you think we could update this function to be async/await for consistency?

mrickard: It should be something I can update--this already converted callback-style `fs` to `fs/promises`, which I could wrap in a try/catch block. Let's see.

mrickard: Should be updated!

jmartin4563 (approved):

PR #1398, opened 2022-10-28
NR-39257 dropping async from promise shim test

mrickard:
## Proposed Release Notes
Intended to drop async from promise shim test

## Links
https://issues.newrelic.com/browse/NR-39257

## Details

Currently in draft: this test still uses the async lib, as I'm running into failures in `tap` when I try to convert to async functions

### DIFF
--- a/test/unit/shim/promise-shim.test.js
+++ b/test/unit/shim/promise-shim.test.js
@@ -237,7 +237,8 @@ tap.test('PromiseShim', (t) => {
       })
     })
 
-    t.test('should reinstate lost context', (t) => {
+    t.test('should reinstate lost context', async (t) => {
+      t.autoend()
       helper.runInTransaction(agent, (tx) => {
         shim.setClass(TestPromise)
         const WrappedPromise = shim.wrapConstructor(TestPromise)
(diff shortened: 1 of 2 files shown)


test/unit/shim/promise-shim.test.js
```diff
@@ -237,44 +236,38 @@ tap.test('PromiseShim', (t) => {
...
-              new WrappedPromise((resolve) => {
-                t.sameTransaction(agent.getTransaction(), tx)
-                helper.runOutOfContext(resolve) // <-- Context loss before resolve.
-              })
-                .then(() => {
-                  t.sameTransaction(agent.getTransaction(), tx)
-                  cb()
-                })
-                .catch(cb)
+        const txTest = async (runOutOfContext, runNext) => {
```
jmartin4563, line 249: I tried really hard to rewrite this in a way that reads a bit easier, but I ran into the exact same problems that you described in standup 😭 

The problem is definitely that the second promise never resolves, so maybe something with that `runOutOfContext` helper is deeply tied to the `async.series` execution? Gonna do some more digging because this is really hard to grok and making sure that context restoration works is important

mrickard: I did dig into the `runOutOfContext` helper, and I've come around to thinking that it's not an interaction between this and `async`, particularly. The `async` library is used just to slow this down and avoid concurrent executions. We'd be able to solve that by making these functions truly async functions, but for me that would have required an extensive refactoring of this entire test. Which might be better--it'd at least be more clear. 

I'm not overly fond of how I solved this problem, recursing through. It removes the abstraction through `async`, at least, but the recursion isn't easy to follow.

jmartin4563: Yeah, getting rid of async once and for all is enough of a win that i'm willing to eat the strangeness of the test for now

jmartin4563 (approved):

test/unit/shim/promise-shim.test.js
mrickard: If nothing else, at least the full function code is available here, instead of depending on code deeply buried in that library.

PR #1404, opened 2022-11-07
drop async from benchmark

mrickard:
## Proposed Release Notes

Removed deprecated and failing suites from benchmark tests. 

## Links

https://issues.newrelic.com/browse/NR-40029

## Details

This doesn't yet remove async, so it's a draft PR. But it might be useful if you're looking to run the benchmark tests.

### DIFF
--- a/bin/compare-bench-results.js
+++ b/bin/compare-bench-results.js
@@ -133,7 +133,7 @@ const iterate = async () => {
   const results = files.map(async (file) => {
     return processFile(file)
   })
-  reportResults(null, results)
+  reportResults(results)
 }
 
 iterate()
--- a/bin/compare-bench-results.js
+++ b/bin/compare-bench-results.js
@@ -7,6 +7,7 @@
 /* eslint-disable sonarjs/no-duplicate-string, no-console */
 
 const fs = require('fs/promises')
+const { errorAndExit } = require('./utils')
 
 if (process.argv.length !== 4) {
   console.log('Usage: %s %s <baseline> <upstream>', process.argv[0], process.argv[1])
@@ -17,25 +18,12 @@ if (process.argv.length !== 4) {
   process.exit(1)
 }
 
-const errorAndExit = (err) => {
-  console.log('Failed to load files.')
-  console.log(err)
-  process.exit(2)
-}
-
 const processFile = async (file) => {
   try {
     const data = await fs.readFile(file, { encoding: 'utf8' })
-    let parsed = null
-    try {
-      parsed = JSON.parse(data)
-    } catch (parseError) {
-      return errorAndExit(parseError)
-    }
-
-    return parsed
+    return JSON.parse(data)
   } catch (err) {
-    return errorAndExit(err)
+    return errorAndExit(err, 'Failed to load files.', 2)
   }
 }
 
--- a/bin/compare-bench-results.js
+++ b/bin/compare-bench-results.js
@@ -6,8 +6,7 @@
 'use strict'
 /* eslint-disable sonarjs/no-duplicate-string, no-console */
 
-const async = require('async')
-const fs = require('fs')
+const fs = require('fs/promises')
 
 if (process.argv.length !== 4) {
   console.log('Usage: %s %s <baseline> <upstream>', process.argv[0], process.argv[1])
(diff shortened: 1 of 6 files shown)


test/benchmark/shimmer/instrumentation.bench.js
```diff
@@ -40,12 +40,13 @@ suite.add({
...
-  }
-})
+// This causes the cjs loader to fail, with benchmark not being found.
+// suite.add({
+//   name: 'shimmer.reinstrument()',
+//   agent: true,
+//   fn: function (agent) {
+//     return shimmer.reinstrument(agent, 'benchmark')
+//   }
+// })
```
mrickard, line 50: I was inclined to remove this block, but it may be possible to have this run with the right path, if we really wanted to run it for some reason.

mrickard: @jordigh The failure here is that `benchmark` can't be found from the context of the test, which is...weird. It's on my docket, along with the max-listeners-exceeded problem.

bin/run-bench.js
```diff
@@ -5,7 +5,7 @@
 
 'use strict'
 
-const a = require('async')
+// const a = require('async')
```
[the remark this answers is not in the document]
mrickard: Yep. WIP; I'll rebase before merge.

bin/run-bench.js
```diff
@@ -17,9 +17,17 @@ const tests = []
 const globs = []
 const opts = Object.create(null)
 
+const fakeCb = (err, payload) => {
+  if (err) {
```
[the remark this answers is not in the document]
mrickard: This is a kind of shim to keep the callback-heavy test style working as I convert over from `async`. Ultimately it'll be removed.

bin/run-bench.js
```diff
@@ -74,67 +82,77 @@ class JSONPrinter {
...
+  const spawnEachFile = (file, spawnCb) => {
+    const test = path.relative(benchpath, file)
+
+    const args = [file]
+    if (opts.inspect) {
+      args.unshift('--inspect-brk')
+    }
+
+    // / TODO: remove diagnostic --trace-warnings
+    args.push('--trace-warnings')
```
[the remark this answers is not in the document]
mrickard: Yes--I wanted to see if I could find the source of the max-listeners-exceeded error, in case this approach is what triggered it. It turns out that the benchmark tests exceed the max listeners without my refactoring.

test/lib/benchmark.js
```diff
@@ -35,104 +34,128 @@ class Benchmark {
...
     return (this.processedSamples = Object.keys(samples).reduce((acc, sampleName) => {
-      acc[sampleName] = new BenchmarkStats(samples[sampleName])
-      return acc
+      try {
+        acc[sampleName] = new BenchmarkStats(samples[sampleName])
+        return acc
+      } catch (e) {
+        // console.log(`********** ${sampleName} from ${this.name}`)
+        // console.log('this tests:', this.tests)
+        // console.log('err:', e)
```
mrickard, line 43: I'll remove these commented lines once I work out the db test errors.

bin/compare-bench-results.js
```diff
@@ -18,119 +17,126 @@ if (process.argv.length !== 4) {
...
-      console.log(err)
-      process.exit(2)
+const processFile = async (file) => {
+  try {
+    const data = await fs.readFile(file, { encoding: 'utf8' })
+    let parsed = null
+    try {
+      parsed = JSON.parse(data)
+    } catch (parseError) {
+      return errorAndExit(parseError)
```
jmartin4563, line 33: Since both these catches just `errorAndExit`, I don't think we need nested try's, we can just do one try/catch?

bin/run-bench.js
```diff
@@ -19,9 +18,18 @@ const tests = []
...
+    console.error(err)
+    return
+  }
+  return payload
+}
+
 process.argv.slice(2).forEach(function forEachFileArg(file) {
   if (/^--/.test(file)) {
-    opts[file.substr(2)] = true
+    opts[file.substring(2)] = true
```
jmartin4563, line 32: Curious why this changed, they're semantically doing the same thing?

bin/run-bench.js
```diff
@@ -76,67 +84,74 @@ class JSONPrinter {
 
 run()
 
-function run() {
+const errorAndExit = (err, message, code) => {
```
jmartin4563, line 87: We have two very similar implementations of this function, one here and one in compare-bench-results.js, can we just create one?

test/benchmark/shim/utilities.bench.js
```diff
@@ -41,13 +41,6 @@ suite.add({
   }
 })
 
-suite.add({
```
jmartin4563, line 44: Why was this removed?

test/benchmark/shimmer/instrumentation.bench.js
```diff
@@ -40,12 +40,13 @@ suite.add({
...
 })
 
-suite.add({
-  name: 'shimmer.reinstrument()',
-  agent: true,
-  fn: function (agent) {
-    return shimmer.reinstrument(agent, 'benchmark')
-  }
-})
+// This causes the cjs loader to fail, with benchmark not being found.
```
jmartin4563, line 43: Sorry to rehash a previous conversation, but did we log a ticket for fixing this?

test/lib/benchmark.js
```diff
@@ -6,9 +6,8 @@
 'use strict'
 
 const helper = require('./agent_helper')
-const async = require('async')
 
-exports.createBenchmark = function createBenchmark(opts) {
+exports.createBenchmark = (opts) => {
```
jmartin4563, line 10: was this change actually needed?

jmartin4563 (changes requested):
Couple of things, but `npm run bench` works for me now!

bin/run-bench.js
mrickard: `substr` is a [deprecated method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr), so I went with the alternative that's likely to have a longer support timeline.

bin/run-bench.js
mrickard: I was definitely leaning that way, but was biased toward not altering the expected behavior of the functions until I'd got them converted. Easy enough to do now!

test/benchmark/shim/utilities.bench.js
mrickard: @bizob2828 removed this method from the agent's shim interface with [this commit](https://github.com/newrelic/node-newrelic/commit/c470ff4fd223ba33a9c067463746b83918edd60d) in 2021; I documented that in the ticket comments. `events.EventEmitter.listenerCount` has apparently been deprecated from Node. Since this was removed from the agent code, we shouldn't be testing on it.

test/lib/benchmark.js
mrickard: This one...not in terms of functionality. I've been converting function definitions to arrow functions where I can, so this was stylistically in keeping with that.

jmartin4563 (approved):

PR #1411, opened 2022-11-11
NEWRELIC-4272 - Add eslint-plugin-sonar rules and address issues

jmartin4563:
## Proposed Release Notes
* Added [eslint-plugin-sonarjs](https://github.com/SonarSource/eslint-plugin-sonarjs) recommended ruleset to eslint configuration and fixed violations

## Links
Closes https://issues.newrelic.com/browse/NEWRELIC-4272

## Details
Added [sonarjs](https://github.com/SonarSource/eslint-plugin-sonarjs) eslint recommended ruleset to this repo
* Disabled two rules for tests (`no-duplicate-literals` and `cognitive-complexity`), as there are many non-trivial violations, created NEWRELIC-5257 to properly fix
* For actual code that has cognitive complexity issues we used eslint configuration comments to change the error threshold to reduce PR load, created NEWRELIC-5252 to properly fix
* Started some of the JSDoc warning fixes, but limited to try and keep this PR sane instead of blowing up the changeset, created NEWRELIC-5261 to properly fix


api.js
```diff
@@ -5,6 +5,8 @@
 
 'use strict'
 
+/* eslint sonarjs/cognitive-complexity: ["error", 21] -- TODO: https://issues.newrelic.com/browse/NEWRELIC-5252 */
```
bizob2828, line 8: shouldn't this be on the actual function that's in violation? it looks like there are actually 2 functions

api.js
```diff
@@ -1349,15 +1338,10 @@ API.prototype.instrumentWebframework = function instrumentWebframework(
  *  - `newrelic.instrumentMessages(moduleName, onRequire [,onError])`
  *  - `newrelic.instrumentMessages(options)`
  *
- * @param {object} options The options for this custom instrumentation.
- * @param {string} options.moduleName The module name given to require to load the module
- * @param {Function}  options.onResolved The function to call prior to module load after the filepath has been resolved
- * @param {Function}  options.onRequire The function to call when the module has been loaded
- * @param {Function} [options.onError] If provided, should `onRequire` throw an error, the error will be passed to
+ * @param {string} moduleName The module name given to require to load the module
```
bizob2828, line 1341: this jsdoc was correct before.  it could either be an object or string

api.js
```diff
@@ -1310,15 +1304,10 @@ API.prototype.instrumentDatastore = function instrumentDatastore(moduleName, onR
...
- * @param {string} options.moduleName The module name given to require to load the module
- * @param {Function}  options.onResolved The function to call prior to module load after the filepath has been resolved
- * @param {Function}  options.onRequire The function to call when the module has been loaded
- * @param {Function} [options.onError] If provided, should `onRequire` throw an error, the error will be passed to
- *  this function.
- * @param moduleName
- * @param onRequire
- * @param onError
+ 
+ * @param {string} moduleName The module name given to require to load the module
```
bizob2828, line 1308: this jsdoc was correct before. it could either be an object or string

api.js
```diff
@@ -1275,15 +1275,9 @@ API.prototype.instrumentConglomerate = function instrumentConglomerate(
...
- * @param {object} options The options for this custom instrumentation.
- * @param {string} options.moduleName The module name given to require to load the module
- * @param {Function}  options.onResolved The function to call prior to module load after the filepath has been resolved
- * @param {Function}  options.onRequire The function to call when the module has been loaded
- * @param {Function} [options.onError] If provided, should `onRequire` throw an error, the error will be passed to
- *  this function.
- * @param moduleName
- * @param onRequire
- * @param onError
+ * @param {string} moduleName The module name given to require to load the module
```
bizob2828, line 1278: this jsdoc was correct before. it could either be an object or string

api.js
```diff
@@ -1241,15 +1246,10 @@ API.prototype.instrument = function instrument(moduleName, onRequire, onError) {
...
- * @param {object} options The options for this custom instrumentation.
- * @param {string} options.moduleName The module name given to require to load the module
- * @param {Function}  options.onResolved The function to call prior to module load after the filepath has been resolved
- * @param {Function}  options.onRequire The function to call when the module has been loaded
- * @param {Function} [options.onError] If provided, should `onRequire` throw an error, the error will be passed to
- *  this function.
- * @param moduleName
- * @param onRequire
- * @param onError
+
```
bizob2828, line 1249: this jsdoc was correct before. it could either be an object or string

api.js
```diff
@@ -1208,15 +1219,9 @@ API.prototype.recordCustomEvent = function recordCustomEvent(eventType, attribut
...
- * @param {object} options The options for this custom instrumentation.
- * @param {string} options.moduleName The module name given to require to load the module
- * @param {Function}  options.onResolved The function to call prior to module load after the filepath has been resolved
- * @param {Function}  options.onRequire The function to call when the module has been loaded
- * @param {Function} [options.onError] If provided, should `onRequire` throw an error, the error will be passed to
- *  this function.
- * @param moduleName
- * @param onRequire
- * @param onError
+ * @param {string} moduleName The module name given to require to load the module
```
bizob2828, line 1222: this jsdoc was correct before. it could either be an object or string

lib/agent.js
```diff
@@ -5,6 +5,8 @@
 
 'use strict'
 
+/* eslint sonarjs/cognitive-complexity: ["error", 23] -- TODO: https://issues.newrelic.com/browse/NEWRELIC-5252 */
```
bizob2828, line 8: this should be on line 728 as _addIntrinsicAttrsFromTransaction is the function in violation. By doing this at the file level it sets cognitive complexity to error at 23 for this entire file

bizob2828 (changes requested):

lib/agent.js
bizob2828: ok we talked in slack. there's no nice way to override an eslint rule at a function level as it will apply for the remaining lines of the file. the only way to do this would be to override the value at start of function and restore after. since we plan on addressing these violations shortly after this PR lands, it is not worth the headache

test/unit/collector/api-connect.test.js
```diff
@@ -243,91 +243,6 @@ tap.test('succeeds when given a different port number for redirect', (t) => {
   })
 })
 
-tap.test('succeeds after one 503 on preconnect', (t) => {
```
bizob2828, line 246: did I do this or was it you @jmartin4563 ? this still seems legit

test/unit/instrumentation/core/promises.test.js
```diff
@@ -215,4 +215,5 @@ function checkTrace(t, tx) {
   t.equal(segment.children.length, 0)
   // verify current segment is same as trace root
   t.same(segment.name, helper.getContextManager().getContext().name)
+  return segment
```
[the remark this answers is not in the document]
bizob2828: I think because it's in a promise chain but it appears this is the end of the line so you could just add t.end to the end of checkTrace

test/unit/trace-aggregator.test.js
```diff
@@ -333,18 +333,16 @@ describe('TraceAggregator', function () {
       done()
     }
 
-    const testCallback = (err, props) => {
+    // Array iteration is too difficult to slow down, so this steps through recursively
+    txnCreator(0, maxTraces, function testCallback(err, props) {
```
bizob2828, line 337: so i think these refactors here and across the others could've been solved by reusing one callback function instead of putting it inline. personally this update appeases the sonar rule but makes less readable and still have the same copy pasta cb function

bizob2828 (changes requested):
overall looks good, just a few comments and suggestions. lmk if you need help fixing

bizob2828:
I think this can be moved to ready for review and provide proper context in the PR details

test/unit/collector/api-connect.test.js
jmartin4563: I removed this one, because the test right below it tests the exact same thing, just after five attempts. I can readd if you think it's still useful to have the one vs several attempts?

bizob2828: my only concern is the other test has this ```// TODO: 503 tests can likely be consolidated into single test func
// passed to t.test() while specifying different # of 503s.
```

we should probably just refactor to address this todo

File: lib/transaction/tracer/index.js [hunk 11]
```diff
@@ -8,6 +8,9 @@
 const Transaction = require('../index.js')
 const logger = require('../../logger').child({ component: 'tracer' })
 const symbols = require('../../symbols')
+const INACTIVE_TRANSACTION_MESSAGE = 'Not creating segment "%s" because no transaction was active'
```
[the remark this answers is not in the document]
jmartin4563: I have no problem doing a more invasive change to fix the underlying issue at the root, but that makes this PR even scarier than it already is, due to the amount of changes already here.

lib/shim/promise-shim.js
```diff
@@ -316,6 +297,9 @@ class PromiseShim extends Shim {
         })
 
         return __NR_wrappedPromisified
+        /**
```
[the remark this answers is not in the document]
jmartin4563: this is my VSCode trying to be helpful, reverting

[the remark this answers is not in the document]
jmartin4563: Actually, it's not VSCode, it's `npm run lint:fix`, which is run by lint-staged, so husky technically added this, which'll get fixed as part of the jsdocs ticket

lib/shim/promise-shim.js
```diff
@@ -566,6 +569,9 @@ class Contextualizer {
       }
     )
 
+    /**
```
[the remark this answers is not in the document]
jmartin4563: Same thing here, lint-staged and npm run lint:fix's fault

test/unit/trace-aggregator.test.js
bizob2828: this isn't worth the time to refactor the right way

lib/transaction/tracer/index.js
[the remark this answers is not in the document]
bizob2828: that's fair, we should log a ticket to address this

lib/config/index.js
```diff
@@ -5,6 +5,9 @@
 
 'use strict'
 
+/* eslint sonarjs/cognitive-complexity: ["error", 29] -- TODO: https://issues.newrelic.com/browse/NEWRELIC-5252 */
```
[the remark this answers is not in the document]
jmartin4563: These are two different rules being changed though, the second is for the massive switch statement in here

lib/transaction/tracer/index.js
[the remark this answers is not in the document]
jmartin4563: https://issues.newrelic.com/browse/NEWRELIC-5270

bizob2828 (approved):

PR #1482, opened 2023-01-20
Preserving GCP IDs that come in as numbers

mrickard:
## Proposed Release Notes

Added safety check to enquote GCP IDs the platform reports as numbers. These numbers can be long enough that JS converts them to scientific notation, losing precision. Updated cross-agent test fixtures to expose this, with a new test case for this issue. Closes NR-80407

## Links
https://issues.newrelic.com/browse/NR-80407


jmartin4563:
GCP Node.js SDK team's original bug where they had to fix the same issue (for historical context): https://github.com/googleapis/gcp-metadata/pull/74

jordigh (approved):
I do think the simplicity of the external dep was well-warranted!

jmartin4563 (approved):
🥇 :shipit:

PR #1485, opened 2023-01-25
Obfuscate url paths for distributed tracing and transactions

matewilk:
## Proposed Release Notes

* Added support for `url` obfuscation using regex:
  * for distributed tracing 
    * incoming and outgoing requests `path` obfuscation
    * transaction `request.url` attribute obfuscation
    * span `http.uri` path obfuscation
  * for transactions
    *  transaction trace details `url` obfuscation

An example config looks like this:

```js
url_obfuscation: {
   enabled: true,
   regex: {
     pattern: /(\/api\/v1\/users\/)([\d]+)(\/.*$)/,
     flags: "i",
     replacement: '$1**$3'
  } 
}
```

This would turn `/api/v1/users/12345456/edit` to `/api/v1/users/**/edit`

Or using env vars:
```yml
NEW_RELIC_URL_OBFUSCATION_ENABLED: "true",
NEW_RELIC_URL_OBFUSCATION_REGEX_PATTERN: '/(\/api\/v1\/users\/)([\d]+)(\/.*$)/',
NEW_RELIC_URL_OBFUSCATION_REGEX_FLAGS: 'i',
NEW_RELIC_URL_OBFUSCATION_REGEX_REPLACEMENT: '$1**$3'
```

## Details

Adds the ability to obfuscate `urls` using config either via env vars or `newrelic.js`:

A `regex` formatter has been added to validate the `*_PATTERN` env variable.

`pattern` attribute can be either a string or a regex expression, it's going to be put through `new RegExp()` internally in `urlutils.obfuscatePath` method which is the main part of the PR.

The `urlutils.obfuscatePath` method is being used in 3 different places:
- `lib/instrumentation/core/http-outbound.js` - when applying segment - which sends the obfuscated `url` `attribute` with the segment. (I think) this part also obfuscates the segment name.
span_name_url_obfuscation

- `lib/instrumentation/core/http.js` - when wrapping the transaction - here the `path` part of `transaction.url` is being obfuscated
transaction_uri_attribute_obfuscation

- `lib/transaction/trace/index.js` - when running `_getRequestUrl` prototype methods - here the `request.url` for transaction trace detail (`url` in the NR UI) is being obfuscated.
transaction_trace_url_obfuscation
transaction_trace_detail_obfuscation

I might have missed other places in the UI that this might (not) be obfuscated, so I'd love to know whether there are more places to look for/into.

### DIFF
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -1211,24 +1211,40 @@ defaultConfig.definition = () => ({
   },
 
   /**
-   * Default URL obfuscation regex - will never match anything
+   * Obfuscates URL parameters
+   * for outgoing and incoming requests
+   * for distrubuted tracing attributes - both transaction and span attributes
+   * for transaction trace transaction details
    */
   url_obfuscation: {
+    /**
+     * Toggles whether to obfuscate URL parameters
+     */
     enabled: {
       formatter: boolean,
       default: false
     },
 
     regex: {
+      /**
+       * A regular expression to match URL parameters to obfuscate
+       */
       pattern: {
         formatter: regex,
         default: null
       },
 
+      /**
+       * A string containing RegEx flags to use when matching URL parameters
+       */
       flags: {
         default: null
       },
 
+      /**
+       * A string containing a replacement value for URL parameters
+       * can contain refferences to capture groups in the pattern
+       */
       replacement: {
         default: null
       }
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -1238,7 +1238,7 @@ defaultConfig.definition = () => ({
        * A string containing RegEx flags to use when matching URL parameters
        */
       flags: {
-        default: null
+        default: ''
       },
 
       /**
@@ -1246,7 +1246,7 @@ defaultConfig.definition = () => ({
        * can contain refferences to capture groups in the pattern
        */
       replacement: {
-        default: null
+        default: ''
       }
     }
   }
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -1220,11 +1220,17 @@ defaultConfig.definition = () => ({
     },
 
     regex: {
-      formatter: regex,
-      default: {
-        pattern: null,
-        flags: null,
-        replacement: null
+      pattern: {
+        formatter: regex,
+        default: null
+      },
+
+      flags: {
+        default: null
+      },
+
+      replacement: {
+        default: null
       }
     }
   }
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -1223,8 +1223,8 @@ defaultConfig.definition = () => ({
       formatter: regex,
       default: {
         pattern: null,
-        flags: undefined,
-        replacement: undefined
+        flags: null,
+        replacement: null
       }
     }
   }
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -1208,7 +1208,12 @@ defaultConfig.definition = () => ({
       formatter: boolean,
       default: false
     }
-  }
+  },
+
+  /**
+   * Default URL obfuscation regex - will never match anything
+   */
+  url_obfuscation_regex: '~^'
 })
 
 /**
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -6,7 +6,7 @@
 'use strict'
 
 const defaultConfig = module.exports
-const { array, int, float, boolean, object, objectList, allowList } = require('./formatters')
+const { array, int, float, boolean, object, objectList, allowList, regex } = require('./formatters')
 
 /**
  * A function that returns the definition of the agent configuration
@@ -1213,7 +1213,21 @@ defaultConfig.definition = () => ({
   /**
    * Default URL obfuscation regex - will never match anything
    */
-  url_obfuscation_regex: '~^'
+  url_obfuscation: {
+    enabled: {
+      formatter: boolean,
+      default: false
+    },
+
+    regex: {
+      formatter: regex,
+      default: {
+        pattern: null,
+        flags: undefined,
+        replacement: undefined
+      }
+    }
+  }
 })
 
 /**
--- a/lib/util/urltils.js
+++ b/lib/util/urltils.js
@@ -171,12 +171,8 @@ module.exports = {
    * @returns {string} The obfuscated path or the original path
    */
   obfuscatePath: function obfuscatePath(config, path) {
-    if (typeof path !== 'string') {
-      return path
-    }
-
     const { enabled, regex } = config.url_obfuscation
-    if (!enabled || !regex) {
+    if (typeof path !== 'string' || !enabled || !regex) {
       return path
     }
 
--- a/lib/util/urltils.js
+++ b/lib/util/urltils.js
@@ -166,17 +166,19 @@ module.exports = {
    * Obfuscates path parameters with regex from config
    *
    * @param {Config} config The configuration containing the regex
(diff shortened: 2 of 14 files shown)


matewilk:
I'd like to mention, that I'm happy to modify this PR so that an array of different regex patterns can be used, for sizeable applications, this feature might be very useful when dealing with different paths.

So instead of just having the config file like this:
```
url_obfuscation: {
   regex: {
      pattern: '...',
      replacement: '...'
   }
}
```

it would be:
```
url_obfuscation: {
   regex: [{
      pattern: '...',
      replacement: '...'
   },
   {
      pattern: /pattern2/,
      replacement: 'replacement2'
   }]
}
```

lib/config/default.js
```diff
@@ -1208,6 +1208,47 @@ defaultConfig.definition = () => ({
...
     }
+  },
+
+  /**
+   * Obfuscates URL parameters
+   * for outgoing and incoming requests
+   * for distrubuted tracing attributes - both transaction and span attributes
+   * for transaction trace transaction details
+   */
+  url_obfuscation: {
```
bizob2828, line 1219: When this lands can you also open a PR in docs-website that describes these new configuration options?

lib/util/urltils.js
```diff
@@ -162,6 +162,31 @@ module.exports = {
...
+    const { enabled, regex } = config.url_obfuscation
+    if (!enabled || !regex) {
+      return path
+    }
+
+    const { pattern, flags, replacement } = regex
+    const f = !flags ? '' : flags
+    const r = !replacement ? '' : replacement
+    const regexPattern = new RegExp(pattern, f)
+    return path.replace(regexPattern, `/${r}`)
```
bizob2828, line 187:
```suggestion
    const { pattern, flags = '', replacement = '' } = regex
    const regexPattern = new RegExp(pattern, flags)
    return path.replace(regexPattern, `/${replacement}`)
```

bizob2828, line 187: I'm unclear what this `/$` is doing with the replacement, can you explain this please?

lib/util/urltils.js
```diff
@@ -162,6 +162,31 @@ module.exports = {
...
 
+  /**
+   * Obfuscates path parameters with regex from config
+   *
+   * @param {Config} config The configuration containing the regex
+   * @param {string} path The path to be obfuscated
+   * @returns {string} The obfuscated path or the original path
+   */
+  obfuscatePath: function obfuscatePath(config, path) {
+    // TODO: handle SyntaxError: Invalid regular expression
```
bizob2828, line 173: can you address this please?

test/unit/instrumentation/http/http.test.js
```diff
@@ -322,6 +322,35 @@ test('built-in http module instrumentation', (t) => {
       }
     )
 
+    t.test('when url_obfuscation regex pattern is set, obfuscate segment url attributes', (t) => {
```
bizob2828, line 325: can we add a few more tests. I get this is just compiling regex but something more realistic to the use cases of customers.  I'd love to see one that replaces both  `/api/v1/users/112324423/edit` and `/api/v1/tokens/adsfasf12312321` with `/api/v1/users/***/edit` and `/api/v1/tokens/***` respectively.

test/unit/trace.test.js
```diff
@@ -758,6 +773,38 @@ tap.test('should set URI to /Unknown when URL is not known/set on transaction',
   t.end()
 })
 
+tap.test('should obfuscate URI using regex when pattern is set', async (t) => {
+  const URL = '/abc/123/def'
```
bizob2828, line 777: since this is using the greedy flag, can you add an extra number path segment? right now the greedy isn't really doing anything.

test/unit/instrumentation/http/http.test.js
bizob2828: n/m i see something to this effect below

bizob2828 (changes requested):
overall this is looking great.  I don't know this problem as well as you but I'm wondering if 1 regex is enough but I don't know how to properly wire that up to support multiple based on some heuristics.

lib/util/urltils.js
matewilk: I was thinking about the same, this will require me to change default values from `null` to an empty string `''` in the `/lib/config/default.js` file. I think it's a better solution overall. I went with your initial suggestion to use `null` values in defaults and from what I'm seeing we can simply use `''` instead.

matewilk: Somewhere internally (not sure where) the `/` is being used by the UI, see the example below:
Screenshot 2023-02-03 at 10 58 56

If there was no `/` the text in the UI would look like this:
```
jsonplaceholder.typicode.comusers/***
```

I can comment on this in the code if you think it's an important piece of info.

lib/util/urltils.js
matewilk: I'm wondering whether this is the best place to address it.
I'm reckoning we want to `logger.error` if it happens, but I can't see any other `urltils` function using the `logger` so I'm wondering whether the logic should be handler somewhere.
Seems like the most reasonable place to handle it, otherwise I would be polluting other files with unrelated logic.

matewilk: I've added functionality but I can't figure out how to stub the `logger` so there is a failing test.

test/unit/urltils.test.js
```diff
@@ -273,4 +275,91 @@ describe('NR URL utilities', function () {
...
+        'child': function () {
+          return {
+            logger: {
+              warn: sinon.spy()
+            }
+          }
+        },
+        'warn': sinon.spy()
+      }
+      proxyquire('../../lib/util/urltils', { '../logger': loggerStub })
```
matewilk, line 298: @bizob2828 Are you able to look into it and see if you can stub the `logger`, I tried various combinations but can't make it work. Joys of working with `proxyquire` and alike 😿

lib/util/urltils.js
bizob2828: ah yea, we should set these to empty strings in `lib/config/default.js`, my bad

bizob2828: wouldn't that just be part of your regex to decide if you want to replace the current `/` or not?

test/unit/urltils.test.js
bizob2828: yea i can comment

lib/util/urltils.js
bizob2828: i don't think we should include an extra `/`. customers can add that in their regex capture groups, etc

bizob2828:
> I'd like to mention, that I'm happy to modify this PR so that an array of different regex patterns can be used, for sizeable applications, this feature might be very useful when dealing with different paths.
> 
> So instead of just having the config file like this:
> 
> ```
> url_obfuscation: {
>    regex: {
>       pattern: '...',
>       replacement: '...'
>    }
> }
> ```
> 
> it would be:
> 
> ```
> url_obfuscation: {
>    regex: [{
>       pattern: '...',
>       replacement: '...'
>    },
>    {
>       pattern: /pattern2/,
>       replacement: 'replacement2'
>    }]
> }
> ```
> 
> So different paths could be replaced with different replacement strings, since capture groups are also supported, this would give the ability to obfuscate sensitive information from urls but keep important information about urls to identify where the transaction/trace/span/request originated or is related to.

I think we should wait and get feedback. I don't want to over-engineer out the gate.  Another possibility is providing an API method that gets invoked when urls are resolved so customers can do whatever in the callback

bizob2828 (approved):

PR #1500, opened 2023-02-06
Fix Public API documentation generation process

jmartin4563:
## Proposed Release Notes
* Updated public api doc generation process to use new theme

## Links
Fixes https://github.com/newrelic/node-newrelic/actions/runs/4107219427/jobs/7086424484 (this has been failing for a while)

## Details
In #1490 we updated our version of `jsdoc` to remove a dependency of `TaffyDB`, which was being reported as a vulnerability. An unfortunate side effect of that is that our existing jsdoc theme (`minami`) requires `TaffyDB` _without a proper entry in their package.json_. `minami` also appears to be unmaintained.

So instead of adding `TaffyDB` to our list of dependencies as https://github.com/nijikokun/minami/issues/48 suggests, we'll just switch to a more maintained theme. The `api.js` change is purely JSDocs, as the `?` were causing parsing to fail, which caused our scripting to not copy over the images.

You can test what this will look like by running `npm run public-docs` then `open out/index.html`.


bizob2828 (approved):
Nice find.  This new theme feels more modern too, good work

PR #1515, opened 2023-02-22
Reduced cognitive complexity

mrickard:
## Proposed Release Notes
Reduced cognitive complexity in lib/metrics/normalizer.js

## Links
https://issues.newrelic.com/browse/NEWRELIC-5292

## Details

Closes NEWRELIC-5292

### DIFF
--- a/lib/metrics/normalizer.js
+++ b/lib/metrics/normalizer.js
@@ -116,7 +116,7 @@ MetricNormalizer.prototype.load = function load(json) {
  * @param rule {object} from this.config.rules
  * @param ctx {object} context from the calling function
  */
-const processNameRule = function processName(rule, ctx) {
+function processNameRule(rule, ctx) {
   if (!rule.pattern) {
     return logger.error({ rule: rule }, 'Simple naming rules require a pattern.')
   }
--- a/lib/metrics/normalizer.js
+++ b/lib/metrics/normalizer.js
@@ -5,8 +5,6 @@
 
 'use strict'
 
-/* eslint sonarjs/cognitive-complexity: ["error", 24] -- TODO: https://issues.newrelic.com/browse/NEWRELIC-5252 */
-
 const EventEmitter = require('events').EventEmitter
 const util = require('util')
 const logger = require('../logger').child({ component: 'metric_normalizer' })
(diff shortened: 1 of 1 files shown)


mrickard:
Draft PR because it makes sense to roll up several of these in one PR.

lib/metrics/normalizer.js
```diff
@@ -112,45 +110,53 @@ MetricNormalizer.prototype.load = function load(json) {
   }
 }
 
+/**
+ *  Helper for loadFromConfig
+ *
+ * @param rule {object} from this.config.rules
+ * @param ctx {object} context from the calling function
+ */
+const processNameRule = function processName(rule, ctx) {
```
jordigh, line 119: Is using an assignment to rename a function something that is less cognitively complex?

lib/metrics/normalizer.js
```diff
@@ -112,45 +110,53 @@ MetricNormalizer.prototype.load = function load(json) {
...
+    ctx.rules.splice(insert, 0, new NormalizerRule(json))
+  }
+}
 /**
  * Load any rules found in the configuration into a metric normalizer.
  *
  * Operates via side effects.
  */
 MetricNormalizer.prototype.loadFromConfig = function loadFromConfig() {
+  const ctx = this
```
jordigh, line 155: I'd just use `this` and invoke the function with `apply` in order to save ourselves one more variable to think about, but that's just me.

lib/metrics/normalizer.js
mrickard: Apparently so, oddly enough.

lib/metrics/normalizer.js
mrickard: Ordinarily, I'd reach for `.bind(this)`, and `.apply` is similar, but context is still getting lost in the `forEach` with those. I'm not super enthusiastic about passing it in as an argument, but it's proving to be more reliable than `bind` or `apply` here.

lib/metrics/normalizer.js
mrickard: Ah, wait--I see what you mean. I'd thought you were referring to pulling the processing block into a function. The cognitive complexity total is mostly affected by breaks in flow and nesting. Assignment doesn't seem to move the needle, but I can take it out if it's jarring.

jordigh: Thank you, yes, that's better!

jordigh (approved):

PR #1560, opened 2023-03-21
NEWRELIC-5252: simplify datastore-shim.js

jordigh:
## Links

* NEWRELIC-5252

## Details

Commit messages:

*NEWRELIC-5252: (datastore-shim) remove eslint rule for cognitive complexity

File now passes without this rule.

*NEWRELIC-5252: (datastore-shim) simplify _recordQuery

It's easier to use the spread operator and some default arguments than
to do all of those ternaries or logical operators to decide if
defaults should be used or not.

Another simplicity win is to move the remove the duplicate call to
`_normalizeParameters`.

*NEWRELIC-5252: (datastore-shim) simplify _normalizeParameters

Split up the function into two extra functions, because each side has
a bunch of bits irreducible fiddly logic.

*NEWRELIC-5295: (datastore-shim) simplify recordOperation

It's easier to use the spread operator and some default arguments than
to do all of those ternaries or logical operators to decide if
defaults should be used or not.

### DIFF
--- a/lib/shim/datastore-shim.js
+++ b/lib/shim/datastore-shim.js
@@ -357,19 +357,17 @@ function recordOperation(nodule, properties, opSpec) {
   return this.record(nodule, properties, function opRecorder(shim, fn, fnName, args) {
     shim.logger.trace('Recording datastore operation "%s"', fnName)
 
-    // Derive the segment information.
+    // Derive the segment information, starting from some defaults
     let segDesc = null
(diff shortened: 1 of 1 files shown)


jmartin4563 (approved):
You'll need to add back `Proposed Release Notes` header to the PR description so the release prep action doesn't hork, otherwise this is awesome!

PR #1569, opened 2023-03-24
Reduced cognitive complexity in transaction index

mrickard:
## Proposed Release Notes
* Reduced cognitive complexity in lib/transaction/index.js

## Links
https://issues.newrelic.com/browse/NEWRELIC-5301

## Details
Closes NEWRELIC-5301

### DIFF
--- a/lib/transaction/index.js
+++ b/lib/transaction/index.js
@@ -413,6 +413,7 @@ Transaction.prototype.isIgnored = function getIgnore() {
  * @param {number} statusCode - The response status code.
  */
 Transaction.prototype.finalizeNameFromUri = finalizeNameFromUri
+
 function finalizeNameFromUri(requestURL, statusCode) {
   if (logger.traceEnabled()) {
     logger.trace(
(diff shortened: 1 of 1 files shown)


lib/transaction/index.js
```diff
@@ -1025,32 +1040,33 @@ function acceptTraceContextPayload(traceparent, tracestate, transport) {
...
+      supportabilityMetric = 'DistributedTrace/AcceptPayload/Ignored/Multiple'
     }
-    return
+    this.agent.recordSupportability(supportabilityMetric)
+    return true
   }
-
+  return false
+}
+const _dtConfigTest = function _dtConfigTest() {
```
jordigh, line 1069: I kind of wonder, is doing these function assignments merely to avoid having to give these functions docstrings?

jordigh (approved):
Easy enough simplification, just breaking out a few functions into the top level.

lib/transaction/index.js
mrickard: The function assignment is stylistic--I just prefer function assignments (no hoisting, for good or for ill). The declaration doesn't trigger the docstring requirement, likely because these aren't exported in any way for anything else to use them.

jordigh: Ah, I see, interesting.

PR #1587, opened 2023-04-10
Reduced cognitive complexity in tracecontext

mrickard:
## Proposed Release Notes
Refactored lib/transaction/tracecontext.js to reduce cognitive complexity

## Links
https://issues.newrelic.com/browse/NEWRELIC-5302

## Details
Closes NEWRELIC-5302

### DIFF
--- a/lib/transaction/tracecontext.js
+++ b/lib/transaction/tracecontext.js
@@ -447,7 +447,6 @@ class TraceContext {
   _parseTraceState(params) {
     const { tracestate, hasTrustKey, expectedNrKey } = params
     let nrTraceStateValue = null
-    let traceStateValid = true
     const finalListMembers = []
     const vendors = []
     const incomingListMembers = tracestate.split(',')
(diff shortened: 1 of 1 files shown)


lib/transaction/tracecontext.js
```diff
@@ -419,13 +460,13 @@ class TraceContext {
...
         if (listMemberParts.length !== 2) {
-          tsd.traceStateValid = false
+          traceStateValid = false
           logger.debug('Unable to parse tracestate list members.')
           this.transaction.agent.recordSupportability(
             'TraceContext/TraceState/Parse/Exception/ListMember'
           )
 
-          return tsd
+          return { traceStateValid }
```
jmartin4563, line 469: can't we simplify this to just be `return { traceStateValid: false }`, and then down on 487 return true instead of keeping a var?

jmartin4563 (approved):
thank you!

PR #1588, opened 2023-04-11
Reduce cognitive complexity in lib/transaction/trace/index.js

mrickard:
## Proposed Release Notes
Refactored lib/transaction/trace/index.js to reduce cognitive complexity

## Links
https://issues.newrelic.com/browse/NEWRELIC-5303

## Details
Closes NEWRELIC-5303

### DIFF
--- a/lib/transaction/trace/index.js
+++ b/lib/transaction/trace/index.js
@@ -76,12 +76,7 @@ Trace.prototype.end = function end() {
 Trace.prototype.generateSpanEvents = function generateSpanEvents() {
   const config = this.transaction.agent.config
 
-  const infiniteTracingConfigured = Boolean(config.infinite_tracing.trace_observer.host)
-  const sampledOrInfinite = infiniteTracingConfigured || this.transaction.sampled
-  const shouldProcess =
-    sampledOrInfinite && config.span_events.enabled && config.distributed_tracing.enabled
-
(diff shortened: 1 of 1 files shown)


lib/transaction/trace/index.js
```diff
@@ -79,55 +77,55 @@ Trace.prototype.generateSpanEvents = function generateSpanEvents() {
   const config = this.transaction.agent.config
 
   const infiniteTracingConfigured = Boolean(config.infinite_tracing.trace_observer.host)
+  const sampledOrInfinite = infiniteTracingConfigured || this.transaction.sampled
+  const shouldProcess =
```
jmartin4563, line 81: I still find this hard to grok, since it's still essentially chained conditionals. My personal preference would be to abstract it into an function and we unit test the permutations.

I'd love for this to read like
```js
if (!shouldGenerateSpanEvents(config, this.transaction)) {
  return
}
```

Thoughts?

jmartin4563 (changes requested):
One suggestion about the configuration check, other than that, this was a great call, flipping it to return early.

lib/transaction/trace/index.js
mrickard: That makes sense. There are a *lot* of conditionals in here.

jmartin4563 (approved):

PR #1623, opened 2023-05-05
ci: add ability to generate release notes from conventional commits

jmartin4563:
## Proposed Release Notes
Created the `ConventionalChangelog` class, which has methods for parsing conventional commits from the git log, as well as methods for generating JSON and Markdown changelogs. JSON changelogs will be used by the docs-website PR script to generate the front-matter for Agent Version Metadata service. Markdown changelogs are for our own documentation and for customers to read.

## Details
The expectation is that this class will be used in NR-96236, where we'll conditionally (since the external repos use the same scripting) update `prepare-release.js` to use this class, and `create-docs-pr.js` to consume `changelog.json` for the front-matter fields.

To manually test, use the following script and observe the updates to NEWS.md and `changelog.json`
```
const ConventionalChangelog = require('./conventional-changelog')

const changelog = new ConventionalChangelog({
  newVersion: '10.1.0',
  previousVersion: '9.8.0',
  repo: 'node-newrelic',
  org: 'newrelic'
})

changelog
  .getFormattedCommits()
  .then((commits) => {
    return Promise.all([
      changelog.generateMarkdownChangelog(commits),
      changelog.generateJsonChangelog(commits)
    ])
  })
  .then(([markdownChangelog, jsonChangelog]) => {
    return Promise.all([
      changelog.writeMarkdownChangelog(markdownChangelog),
      changelog.writeJsonChangelog(jsonChangelog)
    ])
  })
  .then(() => {
    console.log("Job's finished")
  })
```


bin/templates/template.hbs
```diff
@@ -0,0 +1,21 @@
+{{> header}}
+
+{{#if noteGroups}}
+{{#each noteGroups}}
+#### ⚠ {{title}}
```
mrickard, line 5: Is that triangle-and-! character in the space intentionally?

mrickard: (Looks like it might be for breaking changes.)

mrickard (approved):

bin/templates/template.hbs
jmartin4563: It is for breaking changes, I'm not sure why they do this, but currently the only supported `note` type is a breaking change, so `noteGroups` only has one possible group, which is also breaking change (see https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-commits-parser/test/parser.spec.js for examples of what comes out of the parser)

PR #1635, opened 2023-05-11
docs: update Contribution Guide with Conventional Commit info

jmartin4563:
## Details
Would love to get more feedback on the PR template, not sure if we want to put an example in there, or have different headings.


CONTRIBUTING.md
```diff
@@ -62,6 +62,30 @@ In general, we try to limit adding third-party production dependencies. If one i
...
+  * `style` - changes that do not affect the meaning of the code (e.g. formatting, white-space, etc.)
+  * `test` - adding new tests or modifying existing tests
+* Use the imperative, present tense (e.g. "add feature" instead of "added feature")
+* Optionally, include a scope in parantheses after the type to indicate which part of the repository is affected (e.g. `feat(instrumentation): add support for Prisma Client`)
+
+Please note that we only require commits to the main branch to adhere to our Conventional Commit standards. Commits to other branches or in forks may follow a different convention as long as they are clear and descriptive. We use the [Squash and Merge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits) method when merging Pull Requests into the main branch. This means that all commits in a Pull Request will be combined into a single commit before merging. The commit message for this single commit must be Conventional, and will be used in our automated release note generation process.
```
bizob2828, line 87: after reading that how does this work? the last commit will be the initial line and then all subsequent commits will be the additional information which do not have to be conventional commit style?

bizob2828 (changes requested):

CONTRIBUTING.md
jmartin4563: Yeah, I think I just caused more confusion explaining how squash and merge works. For this, when you squash and merge, it'll use the PR title and description to fill out the commit message, it won't use the original commits. What do you think about this instead:

```
Please note that we use the Squash and Merge method when merging uppl requests into the main branch. We do not use the original commit messages from each individual commit. Instead, we use the Pull Request title and description as the commit message for the squashed commit, and require that the Pull Request title adheres to our Conventional Commit standards.
```

bizob2828: oh wait so the description is what's being used at the body of the conventional commit?  My previous comments about keeping links/description is going to be problematic.  If so, we need to tweak this. we need a place to provide context that doesn't go into the changelog but more for the reviewers and posterity

jmartin4563: Screenshot 2023-05-11 at 12 09 06 PM

These are the options for the default squash and merge commit message, we could just default to the PR title, and then the body will be blank, leaving the person merging it the ability to write whatever necessary information they want in the release notes there, and then the PR description is just the PR description, what're your thoughts about doing it that way?

jmartin4563: For future us: Bob and I talked on Zoom and switched to using `Default to pull request title`. The squash commit body will be empty by default, and will be the place to put extended information for the release notes on merge, the PR body will be for reviewers.

Issue #1645, opened 2023-05-19
Lack of instrumentation and accurate test for mysql poolCluster


mrickard:
## Description

A customer observed a loss of telemetry for mysql poolCluster. This gap in telemetry was missed in version tests for mysql pooling: these tests used createConnection (invoked in a setup script). Removing this showed that the agent was not creating segments for poolCluster queries.

## Expected Behavior

versioned/mysql/basic-pool.tap.js test should pass when using a poolCluster connection, and not rely on instrumentation via createConnection

## Steps to Reproduce

See @bizob2828 's changes in https://github.com/newrelic/node-newrelic/tree/mysql-fix

Running versioned tests in that branch will fail subtests in the poolCluster query test. For example, while transactions exist, no segment has been created.  

## Additional context

See https://issues.newrelic.com/browse/NR-116646

PR #1649, opened 2023-05-25
test: added unit tests for MySQL instrumentation

jmartin4563:
## Description
In the aftermath of #1645, we though it prudent to add additional automated tests to ensure that our use of shim symbols is correct and consistent, as the symbols are heavily used in our logic to determine if something is already instrumented or needs to be instrumented. In order to minimize the amount of code shuffling and room for error, I took the route of updating functions to be part of the export.

## How to Test
Tested via `npm run unit` and also with `npm run versioned:major mysql`

## Related Issues
Closes https://issues.newrelic.com/browse/NEWRELIC-4868

### DIFF
--- a/lib/instrumentation/mysql/mysql.js
+++ b/lib/instrumentation/mysql/mysql.js
@@ -19,7 +19,7 @@ exports.callbackInitialize = function callbackInitialize(shim, mysql) {
       return
     }
     shim.logger.debug('Wrapping Connection#query')
-    if (wrapQueryable(shim, connection, false)) {
+    if (exports.wrapQueryable(shim, connection, false)) {
       const connProto = Object.getPrototypeOf(connection)
       connProto[symbols.storeDatabase] = true
       shim[symbols.unwrapConnection] = true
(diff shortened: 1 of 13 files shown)


lib/instrumentation/mysql/mysql.js
```diff
@@ -108,11 +111,11 @@ function wrapGetConnection(shim, connectable) {
   return true
 }
 
-function wrapGetConnectionCallback(shim, cb) {
+exports.wrapGetConnectionCallback = function wrapGetConnectionCallback(shim, cb) {
```
mrickard, line 114: Would it make sense to define the module exports together elsewhere in the file, so that these usages wouldn't need the explicit export every time they were used?

jmartin4563: Sure! I can update to do that instead

mrickard: My thought there was that we could avoid having to refer to each function as `exports.functionName`, but still be able to use the exported functions for testing.

mrickard (approved):

PR #1659, opened 2023-06-02
chore(tests): unmochaify transaction unit tests

jordigh:
## Description

* Closes NEWRELIC-8839.

### DIFF
--- a/test/unit/transaction.test.js
+++ b/test/unit/transaction.test.js
@@ -6,13 +6,6 @@
 'use strict'
 
 const tap = require('tap')
-// TODO: convert to normal tap style.
-// Below allows use of mocha DSL with tap runner.
-tap.mochaGlobals()
-
-const chai = require('chai')
-const should = chai.should()
-const expect = chai.expect
 const helper = require('../lib/agent_helper')
 const API = require('../../api')
 const AttributeFilter = require('../../lib/config/attribute-filter')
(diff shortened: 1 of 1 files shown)


test/unit/transaction.test.js
```diff
@@ -1285,7 +1367,7 @@ describe('Transaction', function () {
           ty: 'Mobile',
           ac: trustedAccountKey,
           ap: '51424',
-          id: '5f474d64b9cc9b2a',
+          id: '5f4n74d64b9cc9b2a',
```
mrickard, line 1370: It looks like this id value has changed.

jordigh: Whoa! I know how that happened. Emacs-induced typo. :( `C-a` means "beginning of line" so sometimes I typo and leave a stray "a" somewhere.

mrickard: Though this one looks like an inserted `n`, I'd think.

jordigh: `C-n` is next line. :P (down arrow)

mrickard: Ah, that'd explain it, yeah.

mrickard, line 1370:
```suggestion
          id: '5f474d64b9cc9b2a',
```

test/unit/transaction.test.js
```diff
@@ -23,528 +16,534 @@ const Segment = require('../../lib/transaction/trace/segment')
...
   })
+})
 
-  it('should know when it is not a web transaction', function () {
-    const tx = new Transaction(agent)
-    tx.type = Transaction.TYPES.BG
-    expect(tx.isWeb()).to.be.false
+tap.test('Transaction naming tests', (t) => {
+  t.autoend()
+  const agent = helper.loadMockedAgent({
```
mrickard, line 211: Previously, this test group used a new agent for each subtest. Is it important to preserve that?

test/unit/transaction.test.js
```diff
@@ -23,528 +16,534 @@ const Segment = require('../../lib/transaction/trace/segment')
...
+      txn.finalizeNameFromUri(url, 200)
+      t.equal(txn.name, 'WebTransaction/NormalizedUri/test/*/string/*')
+      t.end()
+    })
 
-      it('produces the custom name even when status is 501', function () {
-        trans.finalizeNameFromUri('/test/string?do=thing&another=thing', 501)
-        expect(trans.name).equal('WebTransaction/Custom/test')
-      })
+    t.test('should not scope web transactions to their URL', (t) => {
```
mrickard, line 440: Is this subtest new for this group?

jordigh: Yes, but I think it belongs with this group. Originally it was alone at the "top level". We are checking that its name isn't scoped. In particular, when a partial name isn't set.

test/unit/transaction.test.js
jordigh: Okay, I've moved the tear-down/setup logic closer to each subtest to ensure the agent gets thrown out every time.

test/unit/transaction.test.js
```diff
@@ -23,1251 +16,1326 @@ const Segment = require('../../lib/transaction/trace/segment')
...
-        txn.end()
-      })
-    })
+    t.end()
   })
 
-  it('should hand itself off to the agent upon finalization', function (done) {
-    agent.on('transactionFinished', function (inner) {
-      expect(inner).equal(trans)
+  t.test('with distributed tracing enabled', (t) => {
```
mrickard, line 74: Is this the same as the block on line 59?

jordigh: It's not, on line 59 we are not ignoring the transaction. I've relabelled the tests to make this clearer.

test/unit/transaction.test.js
```diff
@@ -71,11 +71,11 @@ tap.test('Transaction unit tests', (t) => {
     t.end()
   })
 
-  t.test('with distributed tracing enabled', (t) => {
+  t.test('with DT enabled, should not produce span events when ignored', (t) => {
```
mrickard, line 74: Ah, this is much clearer--thank you!

mrickard (approved):

PR #1684, opened 2023-06-21
fix(prisma): check for query args in string property

### LINKED ISSUE #1679: Prisma 4.16.0 breaks our instrumentation
## Description

Our versioned tests are no longer passing on Prisma 4.16.0, which was released a few hours ago:

https://github.com/newrelic/node-newrelic/actions/runs/5323833971/jobs/9642251497?pr=1671#step:6:33119

We need to investigate what broke and how to fix it.

jordigh:
Prisma 4.16.0 changed where the query comes from, so let's just make sure we get it from the right place.

Closes #1679


jmartin4563 (approved):
Nice fix!

PR #1723, opened 2023-07-13
feat: enable code_level_metrics by default

bcanzanella:
## Description

Enables code level metrics by default

## How to Test

## Related Issues

### DIFF
--- a/test/unit/shim/webframework-shim.test.js
+++ b/test/unit/shim/webframework-shim.test.js
@@ -599,10 +599,14 @@ test('WebFrameworkShim', function (t) {
         const segment = wrapped(req)
 
         t.ok(segment.attributes)
-        t.same(segment.getAttributes(), {
-          'request.parameters.route.foo': 'bar',
-          'request.parameters.route.biz': 'bang'
-        })
+        const attrs = segment.getAttributes()
+        t.equal(attrs['request.parameters.route.foo'], 'bar')
+        t.equal(attrs['request.parameters.route.biz'], 'bang')
+        const filePathSplit = attrs['code.filepath'].split('/')
+        t.equal(filePathSplit[filePathSplit.length - 1], 'webframework-shim.test.js')
+        t.equal(attrs['code.function'], 'getActiveSegment')
+        t.equal(attrs['code.lineno'], 37)
+        t.equal(attrs['code.column'], 50)
         t.end()
       })
     })
@@ -618,7 +622,10 @@ test('WebFrameworkShim', function (t) {
         const segment = wrapped(req)
 
         t.ok(segment.attributes)
-        t.same(segment.getAttributes(), {})
+        const attrs = Object.keys(segment.getAttributes())
+        const requestParameters = /request\.parameters.*/
+
+        t.notOk(attrs.some((attr) => requestParameters.test(attr)))
         t.end()
       })
     })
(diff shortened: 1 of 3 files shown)


bizob2828 (approved):
Can you also PR the docs website please? it's says defaulting to [false](https://docs.newrelic.com/docs/apm/agents/nodejs-agent/installation-configuration/nodejs-agent-configuration/#code-level-metrics-enabled) currently

bizob2828 (changes requested):
tests are broken. just run `npm run unit` and you will see

planteater:
@bizob2828 I'll take care of the docs.

test/unit/shim/webframework-shim.test.js
```diff
@@ -601,7 +601,12 @@ test('WebFrameworkShim', function (t) {
         t.ok(segment.attributes)
         t.same(segment.getAttributes(), {
           'request.parameters.route.foo': 'bar',
-          'request.parameters.route.biz': 'bang'
+          'request.parameters.route.biz': 'bang',
+          'code.filepath':
+            '/Users/bcanzanella/code/node-newrelic/test/unit/shim/webframework-shim.test.js',
```
bcanzanella, line 606: @bizob2828 any suggestion for how to make this generic? checking the entire object won't work unless the value for code.filepath is modified. could check the length of keys and check each key/value pair with the code.filepath doing an endsWith check? thoughts?

test/unit/shim/webframework-shim.test.js
```diff
@@ -618,7 +623,13 @@ test('WebFrameworkShim', function (t) {
         const segment = wrapped(req)
 
         t.ok(segment.attributes)
-        t.same(segment.getAttributes(), {})
+        t.same(segment.getAttributes(), {
+          'code.filepath':
```
bcanzanella, line 627: @bizob2828 same comment as the other

test/unit/shim/webframework-shim.test.js
bizob2828: instead of `t.same` you could just assert the individual attrs.

```js
const attrs = segment.getAttributes()
t.equal(attrs['request.parameters.route.foo'], 'bar')
t.equal(attrs['request.parameters.route.biz'], 'bang')
```

I'm ok with that. we have other places asserting CLM data

test/unit/shim/webframework-shim.test.js
bizob2828: i would just update this assertion to make sure there are no request parameters.

```js
const attrs = Object.keys(segment.getAttributes())
const requestParams = /request\.parameters.*/
t.notOk(attrs.some((attr) => requestParameters.test(attr)))
```

bcanzanella:
ready for another look @bizob2828

bizob2828 (approved):

menocomp:
I am a little bit confused here.
Docs says this feature is off by default but the changes in the code says it is On by default.
Screenshot from 2023-11-09 11-41-19

bizob2828:
> @bizob2828 I'll take care of the docs.

@planteater looks like the docs were never updated. https://github.com/newrelic/node-newrelic/pull/1723#issuecomment-1803007745

bizob2828:
> I am a little bit confused here.
> 
> Docs says this feature is off by default but the changes in the code says it is On by default.
> 
> Screenshot from 2023-11-09 11-41-19
> 
> 

@menocomp this is a docs issues. It looks it was missed 4 months ago. We will get it fixed.

planteater:
Oops! Looks like I updated the following page, but didn't catch that reference. I'll take care of it.
 
https://docs.newrelic.com/docs/apm/agents/nodejs-agent/installation-configuration/codestream-integration/

PR #1880, opened 2023-11-17
feat: Ensure API errors are tracked

### LINKED ISSUE #1845: Create error traces on failures of both embedding and chat creation

jsumners-nr:
This PR is to resolve #1845.


lib/llm-events/openai/llm-error.js
```diff
@@ -0,0 +1,35 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+/**
+ * Represents an error object, to be tracked via `agent.errors`, that is the
+ * result of some error returned from the OpenAI API.
+ */
+class OpenAiLlmError extends Error {
```
bizob2828, line 12: i'll review more at length later but wanted to call out we already had openai/error-message.js.  That class should be pared down but you can see there are no complex objects. when dealing with nested keys they are just `.` delimited like `error.code` or `http.statusCode`

jsumners-nr: What is the underlying reasoning behind `obj['error.code']` style? My guess is that it is something around data transmission and storage? Just wanting to understand.



> we already had openai/error-message.js

Should I update that object with the extra properties required by the spec? Is the spec defining the minimum properties for the error, or is it defining the whole set of properties?

bizob2828: we spoke in slack about this

bizob2828: This should just be a class, not inheriting from Error.  When you call `agent.errors.add` the signature is `transaction, err, customAttrs`. This class should be providing all the custom attrs. Since the NR exceptions do not handle error.code and error.param those will be properties on this class.

lib/instrumentation/openai.js
```diff
@@ -173,19 +184,25 @@ module.exports = function initialize(agent, openai, moduleName, shim) {
...
-          recordChatCompletionMessages({
+          const summary = recordChatCompletionMessages({
             segment,
             request,
             response
           })
+
+          if (err) {
+            const llmError = new LlmErrorMessage({ cause: err, summary, request, response })
+            shim.agent.errors.add(segment.transaction, llmError)
```
bizob2828, line 203: the 2nd arg should be `err`. `llmError` are the custom attributes

lib/instrumentation/openai.js
```diff
@@ -218,6 +240,12 @@ module.exports = function initialize(agent, openai, moduleName, shim) {
           })
 
           recordEvent('LlmEmbedding', embedding)
+
+          if (err) {
+            const llmError = new LlmErrorMessage({ cause: err, embedding, request, response })
+            shim.agent.errors.add(segment.transaction, llmError)
```
bizob2828, line 246: see comment above regarding signature

lib/llm-events/openai/error-message.js
```diff
@@ -4,18 +4,36 @@
...
+   * @param {object} [params.response] Instance of an incoming message.
+   * @param {object} [params.cause] An instance of the OpenAI error object.
+   * @param {LlmChatCompletionSummary} [params.summary] Details about the
+   * conversation if it was a chat completion conversation.
+   * @param {LlmEmbedding} [params.embedding] Details about the conversation
+   * if it was an embedding conversation.
+   */
+  constructor({ request, response, cause, summary, embedding } = {}) {
     this.api_key_last_four_digits = response?.api_key && `sk-${response.api_key.slice(-4)}`
     this['request.model'] = request.model || request.engine
```
bizob2828, line 25: this can be pared down since this includes attributes from before we agreed to pare this down

lib/instrumentation/openai.js
```diff
@@ -173,19 +184,25 @@ module.exports = function initialize(agent, openai, moduleName, shim) {
...
-          // if (err) {}
-
-          recordChatCompletionMessages({
+          const summary = recordChatCompletionMessages({
             segment,
             request,
             response
           })
+
+          if (err) {
```
bizob2828, line 201: also if errored we need to add `error: true` to the embedding and summary events

bizob2828 (changes requested):
I may have more comments after another round but I think I called out everything that needs to be addressed

bizob2828 (approved):
nice work 👏🏻

lib/instrumentation/openai.js
```diff
@@ -127,14 +135,18 @@ module.exports = function initialize(agent, openai, moduleName, shim) {
    * @param {TraceSegment} params.segment active segment from chat completion
    * @param {object} params.request chat completion params
    * @param {object} params.response chat completion response
+   * @param {boolean} [params.hasError=false] indicates that the completion
+   * resulted in some sort of error case
+   * @returns {LlmChatCompletionSummary} A summary object.
    */
-  function recordChatCompletionMessages({ segment, request, response }) {
+  function recordChatCompletionMessages({ segment, request, response, hasError = false }) {
```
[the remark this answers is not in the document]
jsumners-nr: What is your suggestion?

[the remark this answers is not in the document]
bizob2828: just calling out the arg coming in is diff than it going into the event creation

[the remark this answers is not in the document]
jsumners-nr: In my mind there's a difference between the two, but I'm totally open to changing it one way or the other. Which would you prefer?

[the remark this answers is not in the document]
bizob2828: Ah I guess so. Just keep it.

PR #1949, opened 2024-01-16
chore: Updated some JSDoc declarations in shim

mrickard:
## Description

Updated JSDoc declarations in shim.js.

## Related Issues

NR-134511

### DIFF
--- a/lib/shim/shim.js
+++ b/lib/shim/shim.js
@@ -474,7 +474,7 @@ Shim.prototype[symbols.unwrap] = unwrapAll
  * Entry point for executing a spec.
  *
  * @param {object|Function} nodule Class or module containing the function to wrap.
- * @param {Spec} spec @see Spec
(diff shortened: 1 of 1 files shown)


lib/shim/shim.js
```diff
@@ -471,8 +473,8 @@ Shim.prototype[symbols.unwrap] = unwrapAll
 /**
  * Entry point for executing a spec.
  *
- * @param nodule
- * @param spec
+ * @param {object|Function} nodule Class or module containing the function to wrap.
+ * @param {Spec} spec @see Spec
```
jsumners-nr, line 477: `@see` is a top-level key like `@param`. To inline it, you'd do:
```suggestion
 * @param {Spec} spec {@link Spec}
```

But that is redundant because `@param {Spec}` _should_ link to "Spec".

PR #2062, opened 2024-03-04
chore: Added some delays to docs publishing

jsumners-nr:
After adding verbosity to docs gen (https://github.com/newrelic/node-newrelic/actions/runs/8145456792/job/22261755414) it looks like some operations are happening out-of-order (it could just be the way GitHub Actions records the logs, though). This PR adds some delays to try and solve that problem.


bizob2828 (commented):
I'm not sure this is necessary. I think it's just how GHA buffers and outputs stdout and stderr

bizob2828 (approved):

PR #2099, opened 2024-03-26
feat: Allow for different base branch for prepare-release CI

### LINKED ISSUE #2098: Release notes CI should use branch parameter
The prep-release job in CI takes a `branch` parameter (e.g. `refs/heads/main` or `refs/heads/master`), but the `prepare-release.js` script specifies `main` as the base branch for the PR. This reusable workflow should allow for other base branches. 

## Feature Description
`prepare-release.js` should make use of the value of the branch parameter for setting the base of the PR. 

## Priority
Blocker for completing automation for the serverless plugin releases.

mrickard:
## Description

Our prepare-release job takes `branch` as an input parameter, but it's ignored when creating the release notes PR, and is overridden with the value `main`. For repos that have a different main branch name, we could use the submitted `branch` option. 

Closes #2098 
Closes NR-252885

### DIFF
--- a/bin/prepare-release.js
+++ b/bin/prepare-release.js
@@ -72,6 +72,7 @@ async function prepareReleaseNotes() {
   const options = program.opts()
   console.log('Script running with following options: ', JSON.stringify(options))
   const [owner, repo] = options.repo.split('/')
+  const [, , baseBranch] = options.branch.split('/')
 
   logStep('Validation')
 
@@ -175,7 +176,7 @@ async function prepareReleaseNotes() {
 
     const prOptions = {
       head: newBranchName,
-      base: 'main',
+      base: baseBranch,
       title,
       body,
       draft: true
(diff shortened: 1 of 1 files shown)


bin/prepare-release.js
```diff
@@ -72,6 +72,7 @@ async function prepareReleaseNotes() {
   const options = program.opts()
   console.log('Script running with following options: ', JSON.stringify(options))
   const [owner, repo] = options.repo.split('/')
+  const [, , baseBranch] = options.branch.split('/')
```
jsumners-nr, line 75: This isn't going to work for simple names or for differently styled refs. For example:

```js
const [, , foo] = 'main'.split('/')
// foo => undefined
```

My understanding of `--branch` is that it is the branch to generate release notes from. It is not necessarily the branch to target with the resulting PR.

mrickard: True--the intention was to collect the release notes from some branch, most frequently `main`. The plugin repo, though, uses `master`, which in GHA was expressed as `refs/heads/master`.  In any case, I'm closing this because I've renamed the main branch.

PR #2100, opened 2024-03-27
chore: Updated versioned security agent workflow

### LINKED ISSUE #2088: Implement security agent testing on PRs for security agent updates
### Description
When the security agent team issues a new release the submit a pull request (PR) to use to bump the version, e.g. https://github.com/newrelic/node-newrelic/pull/2080. However, our continuous integration (CI) setup only runs the standard tests suites, which do not enable the security agent, for all PRs. So it is possible for us to see breakages when the nightly tests run that _do_ enable the security agent, e.g. https://github.com/newrelic/node-newrelic/pull/2087

### Acceptance Criteria
1. Update the CI configuration to enable the security agent when the PR contains a change to the security agent dependency version such that CI enables the security agent for the runs against the PR
2. ~Change the security agent dependency to a pinned version instead of a semver qualifier range of `^`~ (solved in https://github.com/newrelic/node-newrelic/pull/2089)

jsumners-nr:
This PR resolves #2088. It adds a new job to the `versioned-security-agent` workflow that detects if the `package.json` has been updated, and if so, whether or not the `@newrelic/security-agent` version number has been changed. The results of this job are utilized by the already existing job to determine if it should run. After this PR there are three cases where the versioned tests will run with the security agent enabled:

1. When the workflow is directly run through the GitHub interface.
2. When nightly scheduled job is executed.
3. When a pull request includes a change in the version for the `@newrelic/security-agent` dependency.

-----

Note: the CI run https://github.com/newrelic/node-newrelic/actions/runs/8455455203/job/23162959574?pr=2100 as a result of opening this PR shows that the security agent enabled tests do not run for every PR (this PR does not change the dependency version).

### DIFF
--- a/.github/workflows/versioned-security-agent.yml
+++ b/.github/workflows/versioned-security-agent.yml
@@ -21,9 +21,49 @@ on:
         required: false
   schedule:
     - cron:  '0 9 * * 1-5'
+  pull_request:
+    branches:
+      - main
 
 jobs:
+  should_run:
+    # Used to determine if the `@newrelic/security-agent` dependency has
+    # been updated in any new pull requests. This job _must_ always run because
+    # the `security-agent-tests` job depends on the outputs of this job. Thus,
+    # we cannot put a conditional on this job to only run if the event name
+    # is "pull_request".
+    name: Should Run
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


mrickard (approved):
LGTM!

Issue #2112, opened 2024-04-02
Remove `USE_NEW_RELEASE` flag from prep release workflow


svetlanabrennan:
We no longer need this `USE_NEW_RELEASE` flag when all the repos are updated to use the new release style. 

These 4 repos need to be updated first to use the new release process:

- https://github.com/newrelic/node-newrelic-superagent/issues/123
- https://github.com/newrelic/node-native-metrics/issues/241
- https://github.com/newrelic/eslint-config-newrelic/issues/25
- https://github.com/newrelic/newrelic-node-opentelemetry-integration/issues/29

Code to remove:
- https://github.com/newrelic/node-newrelic/blob/158c29566f6a5993ec77f84aeaa266ceab0bf4f0/.github/workflows/prepare-release.yml#L17
- https://github.com/newrelic/node-newrelic/blob/158c29566f6a5993ec77f84aeaa266ceab0bf4f0/.github/workflows/prep-release.yml#L54 (just the part that passes in the input for using new release)
- dead code in `bin/prepare-release.js` that uses the new release option since that should be the only option now

PR #2124, opened 2024-04-04
ci: Removed usage of `use_new_release` from prepare release workflow

svetlanabrennan:
## Description

We no longer need to use `use_new_release` since all repos workflows are (or will be) updated to use the new release workflow. 

## Related Issues

Fixes issue https://github.com/newrelic/node-newrelic/issues/2112

## ToDo

- [ ] Remove `USE_NEW_RELEASE` from github action variables


jsumners-nr (approved):
Looks good to me.

PR #2191, opened 2024-05-13
fix: Added support for AWS API Gateway v2 payloads

### LINKED ISSUE #2189: Fix parsing of v2 API Gateway headers for Lambda
## Description

The Node agent automatically parses incoming HTTP headers to continue a distributed trace. In a recent test with API Gateway V2 in Lambda,  the incoming trace header was not extracted from the request, and a new trace ID was generated. 

This process is invoked in the wrapped handler with `shim.tracer.getTransaction()` ([here](https://github.com/newrelic/node-newrelic/blob/main/lib/serverless/aws-lambda.js#L137)), which leverages the transaction tracer's `getTransaction` method. The agent iterates over headers in the [api-gateway.js](https://github.com/newrelic/node-newrelic/blob/main/lib/serverless/api-gateway.js) file, which iterates over headers, looking for headers defined in the [event-sources.json](https://github.com/newrelic/node-newrelic/blob/main/lib/serverless/event-sources.json#L18-L25) file. 

## Expected Behavior

The Node agent's method of detecting the incoming trace header should work seamlessly with either V2 or V1 API Gateway event payloads.

## Steps to Reproduce

We'll be supplied with the test functions that were used to detect this, and we'll deploy them in the dev account. If deploying with the Serverless framework, we can specify a V2 gateway by including an event definition like this: 

    events:
      - httpApi:
          method: GET
          path: /

By comparison, the equivalent V1 gateway is defined with 

    events:
      - http: GET /

## Additional context

https://new-relic.atlassian.net/browse/NR-265371

jsumners-nr:
This PR resolves #2189.


lib/serverless/event-sources.json
```diff
@@ -24,13 +25,32 @@
             "requestContext.stage"
         ]
     },
+
+    "apiGatewayV2": {
+        "attributes": {
+            "aws.lambda.eventSource.accountId": "requestContext.accountId",
+            "aws.lambda.eventSource.apiId": "requestContext.apiId",
+            "aws.lambda.eventSource.stage": "requestContext.stage"
+        },
```
jsumners-nr, line 34: "resourceId" and "resourcePath", found in the v1 payload, are not present in v2 payloads.

mrickard (approved):

PR #2208, opened 2024-05-22
chore: Removed package-lock.json

jsumners-nr:
As we discussed prior to stand-up on 2024-05-22, this PR removes the `package-lock.json` from this repo. Doing so provides:

1. No more useless Dependabot updates that change items in the lock file: when people `npm install newrelic`, the lock file does not get consulted and thus none of the Dependabot changes have any affect on our customers.
2. Easier detection of dependency updates in workflows. This will allows #2205 to be updated in such a fashion that we can detect changes to dependency blocks in `package.json` so that test suites can be triggered or skipped accordingly. With the lock file in place, this sort of workflow detection will be very difficult.


.github/workflows/release-creation.yml
```diff
@@ -45,9 +45,9 @@ jobs:
     - name: Install Dependencies
       run: |
         # Install deps in caller repo
-        npm ci
+        npm install
```
jsumners-nr, line 48: The only concern I have in this PR is this line. I do not expect any problem as `npm install` should still install all dependencies. The difference would be if a repo that is reusing this workflow has updates to `package.json` and _not_ `package-lock.json`. In that case, `npm ci` would notice the discrepancy and generate an error. Whereas `npm install` will install the dependencies as noted by `package.json`.

bizob2828 (changes requested):
I think we should add package-lock.json to .gitignore to be safe. I know you set it in npmignore but wouldn't hurt.

jsumners-nr:
> I think we should add package-lock.json to .gitignore to be safe. I know you set it in npmignore but wouldn't hurt.

Good call. Done.

bizob2828 (approved):

PR #2323, opened 2024-07-01
refactor: reduce duplication in the error-collector

amychisholm03:
- Refactored _processUserErrors, _processTransactionExceptions, and _processTransactionErrors into a single function: _processErrors

Original Ticket:
- There's a lot of duplication here--for example, in the error-collector, _processUserErrors and _processTransactionExceptions are nearly identical, apart from the property they inspect. _processTransactionErrors behaves only slightly differently. Those three could be collapsed into one processor.
- First half of this Jira ticket (https://new-relic.atlassian.net/browse/NR-96722)

### DIFF
--- a/lib/errors/error-collector.js
+++ b/lib/errors/error-collector.js
@@ -88,6 +88,51 @@ class ErrorCollector {
     return false
   }
 
+  _getIterableProperty(transaction, errorType) {
+    let iterableProperty = null
+    if (errorType === 'user' && transaction.userErrors.length) {
+      iterableProperty = transaction.userErrors
+    }
+    if (errorType === 'transactionException') {
+      iterableProperty = transaction.exceptions
+    }
+    return iterableProperty
+  }
+
+  /**
+   * Helper method for processing errors that are created with .noticeError(), exceptions
+   * on the transaction (transaction.exceptions array), and inferred errors based on Transaction metadata.
+   *
+   * @param {Transaction} transaction the collected exception's transaction
+   * @param {number} collectedErrors the number of errors we've succesfully .collect()-ed
+   * @param {number} expectedErrors the number of errors marked as expected in noticeError
+   * @param {string} errorType the type of error to be processed; "user", "transactionException", "transaction"
+   * @returns {Array.<number>} the updated [collectedErrors, expectedErrors] numbers post processing
+   */
+  _processErrors(transaction, collectedErrors, expectedErrors, errorType) {
+    const iterableProperty = this._getIterableProperty(transaction, errorType)
+    if (iterableProperty) {
+      iterableProperty.forEach((exception) => {
+        if (this.collect(transaction, exception)) {
+          ++collectedErrors
+          if (
+            urltils.isExpectedError(this.config, transaction.statusCode) ||
+            errorHelper.isExpectedException(transaction, exception, this.config, urltils)
+          ) {
+            ++expectedErrors
+          }
+        }
+      })
+    } else if (errorType === 'transaction' && this.collect(transaction)) {
+      ++collectedErrors
+      if (urltils.isExpectedError(this.config, transaction.statusCode)) {
+        ++expectedErrors
+      }
+    }
+
+    return [collectedErrors, expectedErrors]
+  }
+
   /**
    * Helper method for processing errors that are created with .noticeError()
    *
@@ -183,10 +228,11 @@ class ErrorCollector {
 
     // errors from noticeError are currently exempt from
     // ignore and exclude rules
-    ;[collectedErrors, expectedErrors] = this._processUserErrors(
+    ;[collectedErrors, expectedErrors] = this._processErrors(
       transaction,
       collectedErrors,
-      expectedErrors
+      expectedErrors,
+      'user'
     )
 
     const isErroredTransaction = urltils.isError(this.config, transaction.statusCode)
@@ -194,16 +240,18 @@ class ErrorCollector {
 
     // collect other exceptions only if status code is not ignored
     if (transaction.exceptions.length && !isIgnoredErrorStatusCode) {
-      ;[collectedErrors, expectedErrors] = this._processTransactionExceptions(
+      ;[collectedErrors, expectedErrors] = this._processErrors(
         transaction,
         collectedErrors,
-        expectedErrors
+        expectedErrors,
+        'transactionException'
       )
     } else if (isErroredTransaction) {
-      ;[collectedErrors, expectedErrors] = this._processTransactionErrors(
+      ;[collectedErrors, expectedErrors] = this._processErrors(
         transaction,
         collectedErrors,
-        expectedErrors
+        expectedErrors,
+        'transaction'
       )
     }
 
--- a/lib/errors/error-collector.js
+++ b/lib/errors/error-collector.js
@@ -88,6 +88,12 @@ class ErrorCollector {
     return false
   }
 
+  /**
+   * Gets the iterable property from the transaction based on the error type
+   * @param {Transaction} transaction the collected exception's transaction
+   * @param {string} errorType the type of error to be processed; "user", "transactionException", "transaction"
+   * @returns
+   */
   _getIterableProperty(transaction, errorType) {
     let iterableProperty = null
     if (errorType === 'user' && transaction.userErrors.length) {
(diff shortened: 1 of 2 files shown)


lib/errors/error-collector.js
```diff
@@ -89,69 +89,47 @@ class ErrorCollector {
...
+    return iterableProperty
+  }
+
+  /**
+   * Helper method for processing errors that are created with .noticeError(), exceptions
+   * on the transaction (transaction.exceptions array), and inferred errors based on Transaction metadata.
    *
    * @param {Transaction} transaction the collected exception's transaction
-   * @param {number} collectedErrors the number of errors we've successfully .collect()-ed
+   * @param {number} collectedErrors the number of errors we've succesfully .collect()-ed
```
mrickard, line 113:
```suggestion
   * @param {number} collectedErrors the number of errors we've successfully .collect()-ed
```

lib/errors/error-collector.js
```diff
@@ -89,69 +89,47 @@ class ErrorCollector {
...
    * @param {number} expectedErrors the number of errors marked as expected in noticeError
-   * @returns {Array.<number>} the updated [collectedErrors, expectedErrors] numbers post-processing
+   * @param {string} errorType the type of error to be processed; "user", "transactionException", "transaction"
```
mrickard, line 115:
```suggestion
   * @param {string} errorType the type of error to be processed: "user", "transactionException", "transaction"
```

mrickard: I'm suggesting a colon here, since we're introducing a list of potential error types in the JSDoc line here, rather than joining two independent clauses.

lib/errors/error-collector.js
```diff
@@ -89,69 +89,47 @@ class ErrorCollector {
   }
 
   /**
-   * Helper method for processing errors that are created with .noticeError()
+   * Gets the iterable property from the transaction based on the error type
+   * @param {Transaction} transaction the collected exception's transaction
+   * @param {string} errorType the type of error to be processed: "user", "transactionException", "transaction"
+   * @returns
```
jsumners-nr, line 95: What is being returned by this function? Is it a string?

lib/errors/error-collector.js
```diff
@@ -89,69 +89,47 @@ class ErrorCollector {
...
+   */
```
jsumners-nr, line 96: In my opinion, it is easier to read, and maintain, a JSDoc block when a little extra formatting is applied. As an example, I prefer to write them like:

```js
/**
 * A helpful introduction, hard wrapped at 80 characters.
 *
 * @param {object} params Input parameters.
 * @param {boolean} params.hasSomething Indicates if something is present.
 * @param {string} params.type The type of something. Possible values
 * are: "foo", "bar", or "baz".
 * @param {object} params.something The something to work with.
 *
 * @throws {Error} When something bad happens.
 * @returns {object} A new something!
 */
function changeSomething(params) {
  // do stuff
}
```

Notice the breaks between the introduction, the parameter definitions, and the results sections.

lib/errors/error-collector.js
```diff
@@ -89,69 +89,47 @@ class ErrorCollector {
...
             errorHelper.isExpectedException(transaction, exception, this.config, urltils)
           ) {
             ++expectedErrors
           }
         }
-      }
-    }
-
-    return [collectedErrors, expectedErrors]
-  }
```
jsumners-nr, line 117: I find it helpful if the main logic of a function is at the leftmost indentation possible. Consider the following functions:

```js
function foo (bar) {
  let result
	if (bar != null) { // coercive check to cover null and undefined
		result = ''
		for (const ele of bar) {
			result += ele.baz
		}
	}

	return result
```

vs

```js
function foo (bar) {
	if (bar == null) { // coercive check to cover null and undefined
		return
	}

	let result = ''
	for (const ele of bar) {
		result += ele.baz
	}

	return result
}
```

In my opinion, it's easier in the second version to see and review the purpose of the function. Versus the first version where we have to reason through nesting to find the main purpose of the function.

lib/errors/error-collector.js
```diff
@@ -89,69 +89,47 @@ class ErrorCollector {
...
+   * @returns {Array.<number>} the updated [collectedErrors, expectedErrors] numbers post processing
    */
-  _processUserErrors(transaction, collectedErrors, expectedErrors) {
-    if (transaction.userErrors.length) {
-      for (let i = 0; i < transaction.userErrors.length; i++) {
-        const exception = transaction.userErrors[i]
+  _processErrors(transaction, collectedErrors, expectedErrors, errorType) {
+    const iterableProperty = this._getIterableProperty(transaction, errorType)
+    if (iterableProperty) {
+      iterableProperty.forEach((exception) => {
```
jsumners-nr, line 121: I'm not sure if we are in a "hot path" here, so it may not be worth a refactor, but I'd like to call attention to the performance difference between the options for looping here.

Utilizing this benchmark script:

```js
'use strict'

const { Bench } = require('tinybench')
const bench = new Bench({ time: 100 })

const input = new Array(1_000_000)
bench
	.add('traditional for', function () {
		let x
		for (let i = 0; i < input.length; i += 1) {
			x = input[i]
		}
	})
	.add('for...of', function () {
		let x
		for (const i of input) {
			x = i
		}
	})
	.add('forEach', function () {
		let x
		input.forEach((i) => {
			x = i
		})
	})

bench
	.warmup()
	.then(() => bench.run())
	.then(() => {
		console.table(bench.table())
	})
```

We can see a result set like:
```
┌─────────┬───────────────────┬─────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name         │ ops/sec │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼───────────────────┼─────────┼────────────────────┼──────────┼─────────┤
│ 0       │ 'traditional for' │ '2,042' │ 489546.9463414662  │ '±0.74%' │ 205     │
│ 1       │ 'for...of'        │ '1,576' │ 634174.5822784823  │ '±0.27%' │ 158     │
│ 2       │ 'forEach'         │ '773'   │ 1293309.2948717843 │ '±0.64%' │ 78      │
└─────────┴───────────────────┴─────────┴────────────────────┴──────────┴─────────┘
```

I ran this test with Node.js 20.15.0 on my M1 Max based system. The `forEach` is the slowest of the methods due to the additional cost of invoking a function on each iteration. Consider how `forEach` might be implemented:

```js
Array.prototype.forEach = function (iterator) {
	for (let i = 0; i < this.length; i += 1) {
		iterator(this[i])
	}
}
```

In conclusion, `forEach` is a nice convenience tool but _may_ not always be appropriate.

jsumners-nr (commented):
Looks good. I have some feedback on JSDoc formatting, and a couple items for consideration, but this is very close to ready.

lib/errors/error-collector.js
amychisholm03: An array of exceptions or errors; I'll post my changes shortly

lib/errors/error-collector.js
```diff
@@ -89,75 +89,62 @@ class ErrorCollector {
...
   /**
-   * Helper method for processing errors that are created with .noticeError()
+   * Gets the iterable property from the transaction based on the error type
    *
    * @param {Transaction} transaction the collected exception's transaction
-   * @param {number} collectedErrors the number of errors we've successfully .collect()-ed
-   * @param {number} expectedErrors the number of errors marked as expected in noticeError
-   * @returns {Array.<number>} the updated [collectedErrors, expectedErrors] numbers post-processing
+   * @param {string} errorType the type of error: "user", "transactionException", "transaction"
+   * @returns {Array} the iterable property from the transaction based on the error type
```
jsumners-nr, line 96: It's a good idea to give a stronger indication of what the value is. In this case, an array of objects is returned, so we would use the `object[]` syntax to denote it:

```suggestion
   * @returns {object[]} the iterable property from the transaction based on the error type
```

https://jsdoc.app/tags-type explains the various syntaxes available for describing parameter types.

lib/errors/error-collector.js
```diff
@@ -89,75 +89,62 @@ class ErrorCollector {
...
-      const exception = transaction.exceptions[i]
-      if (this.collect(transaction, exception)) {
-        ++collectedErrors
-        // if we could collect it, then check if expected
-        if (
-          urltils.isExpectedError(this.config, transaction.statusCode) ||
-          errorHelper.isExpectedException(transaction, exception, this.config, urltils)
-        ) {
-          ++expectedErrors
+  _processErrors(transaction, collectedErrors, expectedErrors, errorType) {
```
jsumners-nr, line 119: Let's add a new sub-test suite to https://github.com/amychisholm03/node-newrelic/blob/error-collector/test/unit/errors/error-collector.test.js to cover the various cases of this function, e.g.:

```js
tap.test('_proccessErrors', t => {
	t.test('no iterableProperty and type is transaction', async t => {
		t.pass()
	})

	t.test('interableProperty is null', async t => {
		t.pass()
	})

	t.end()
})
```

test/unit/errors/error-collector.test.js
```diff
@@ -2370,3 +2370,80 @@ test('When using the async listener', (t) => {
...
+      }
+    })
+    transaction = new Transaction(agent)
+    transaction.url = '/'
+    errorCollector = agent.errors
+  })
+
+  t.afterEach(() => {
+    helper.unloadAgent(agent)
+  })
```
jsumners-nr, line 2396: I'm aware that the other sub-tests in this suite follow the pattern used here, but I'd like us to avoid it. First, a little information:

1. The test runner, `tap`, runs individual test files (suites) in parallel with their own environments. That is `tap suite1.js suite2.js` will run both `suite1.js` and `suite2.js` at the same time. No variables or state defined _wholly_ within each file will interact with the other's (external resources, e.g. a database, does not have that guarantee).
2. It is _possible_ to define `tap.jobs = 2 // any integer, really` to trigger running at least 2 subtests at a time in parallel (https://tapjs.github.io/tapjs/classes/_tapjs_test.index.Test.html#jobs).

If we were to ever do 2, then the setup in this subtest will introduce race conditions on the value for `agent`, `transaction`, and `errorCollector`. Thus making it indeterminate as to what values they have at any given instant.

We can avoid such a scenario by utilizing the `t.context` object as shown in https://github.com/newrelic/node-newrelic/blob/3a910ef29c76cfd05903f01fb84d6775f8669578/test/unit/utilization/docker-info.test.js#L16-L56

Even if we never introduce parallel execution of subtests, the pattern I am suggesting gives each subtest a guaranteed (as much as can be with the mock agent, anyway) clean slate for doing its work. In effect, we will know for a fact that the result of one subtest is not affect the state of another subtest.

amychisholm03: Okay, that makes sense. Do you think in the future we should try to remove this pattern in the other tests in favor of the one you provided?

jsumners-nr: If there isn't much work involved, I think it is a good idea. It's not something we need to go out of our way to do, though.

test/unit/errors/error-collector.test.js
```diff
@@ -2370,3 +2370,76 @@ test('When using the async listener', (t) => {
...
+  t.beforeEach(() => {
+    t.context.agent = helper.loadMockedAgent({
+      attributes: {
+        enabled: true
+      }
+    })
+
+    transaction = new Transaction(t.context.agent)
+    transaction.url = '/'
+    errorCollector = t.context.agent.errors
```
jsumners-nr, line 2387: We want them all scoped the same way as the agent:
```suggestion
    const transaction = new Transaction(t.context.agent)
    transaction.url = '/'
    t.context.transaction = transaction
    
    t.context.errorCollector = t.context.agent.errors
```

test/unit/errors/error-collector.test.js
```diff
@@ -2370,3 +2370,76 @@ test('When using the async listener', (t) => {
...
+    helper.unloadAgent(t.context.agent)
+  })
+
+  t.test('invalid errorType should return no iterableProperty', (t) => {
+    const errorType = 'invalid'
+    const result = errorCollector._getIterableProperty(transaction, errorType)
+
+    t.equal(result, null)
+    t.end()
+  })
```
jsumners-nr, line 2400: Given the previous suggestion, this test would look like:

```suggestion
  t.test('invalid errorType should return no iterableProperty', (t) => {
    const { errorCollector, transaction } = t.context
    const errorType = 'invalid'
    const result = errorCollector._getIterableProperty(transaction, errorType)

    t.equal(result, null)
    t.end()
  })
```

test/unit/errors/error-collector.test.js
```diff
@@ -2370,3 +2370,76 @@ test('When using the async listener', (t) => {
     })
   }
 })
+
+tap.test('_processErrors', (t) => {
+  let transaction = null
+  let errorCollector = null
+
+  t.beforeEach(() => {
```
jsumners-nr, line 2378:
```suggestion
  t.beforeEach((t) => {
```

It gets a little muddy with the multiple shadows of `t`, but:

```js
tap.test('a test', t => { // this t is the test object for this outer test
	// t.{beforeEach, test, end} all reference the above t
	t.beforeEach(t => { // this t is a reference to the subtest being executed
		t.context.foo = 'foo'
	})

	t.test('subtest', async t => { // this t is local to the subtest (and was augmented by beforeEach)
		const { foo } = t.context
		t.equal(foo, 'foo')
	})

	t.end() // marks that the outer test has finished
})
```

test/unit/errors/error-collector.test.js
```diff
@@ -2370,3 +2370,76 @@ test('When using the async listener', (t) => {
...
+    })
+
+    transaction = new Transaction(t.context.agent)
+    transaction.url = '/'
+    errorCollector = t.context.agent.errors
+  })
+
+  t.afterEach(() => {
+    helper.unloadAgent(t.context.agent)
+  })
```
jsumners-nr, line 2392:
```suggestion
  t.afterEach((t) => {
    helper.unloadAgent(t.context.agent)
  })
```

jsumners-nr (commented):
We're really close. I believe that addressing these comments will finish it off.

jsumners-nr (approved):
Fantastic! Looks good.

PR #2328, opened 2024-07-03
refactor: got rid of redundant isExpected in the Exception class

amychisholm03:
- Got rid of the `isExpected` function in the Exception class to avoid duplication with the `isExpected` function in the error helper.


jsumners-nr (approved):
Looks good to me.

PR #2334, opened 2024-07-08
fix: Refactored benchmark tests to complete async functions

mrickard:
## Description

Benchmark tests for datastore-shim and promises were throwing an error: when run, and benchmark stats were generated, the stat constructor failed, as there were no values in samples. The tests used both callbacks and async/await, so the stats were invoked before the async promises were resolved. This PR removes the async-handling callback code path, so async tests aren't expected to fire a callback on completion. 

## How to Test

`npm run bench datastore-shim`
`npm run bench promises`

## Related Issues

Closes NR-67805
Closes NR-281838

### DIFF
--- a/test/benchmark/datastore-shim/test-datastore.js
+++ b/test/benchmark/datastore-shim/test-datastore.js
@@ -7,15 +7,24 @@
 
 class TestDatastore {
   testOp(cb) {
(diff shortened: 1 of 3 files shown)


mrickard:
I'll wait to merge until the node-gyp PR is merged.

PR #2383, opened 2024-07-22
chore: Reverted "ci: Pin Node 22 to 22.4.1"

jsumners-nr:
Reverts newrelic/node-newrelic#2375

GitHub fixed the Actions infra this morning. We should be able to revert this now.


jsumners-nr:
> did you mean to say that Node.js fixed this in 22.5.1 or github pinned 22.x to 22.4.1?

No, I meant that we had to wait literal days after Node.js fixed it because GitHub had to let a human press a merge button https://github.com/actions/node-versions/pull/182

PR #2411, opened 2024-07-31
chore: Converted agent unit tests to node:test

### LINKED ISSUE #2380: Convert agent unit tests to `node:test`
As outlined in #2373, we want to migrate our tests to utilize `node:test`. Mostly arbitrarily picked, we should start with https://github.com/newrelic/node-newrelic/tree/c6a822b0d80e2a9380c46bc4a605a7d3553ec531/test/unit/agent as an example and testbed for this work.

jsumners-nr:
This PR resolved #2380.

There are couple of main differences between `tap` and `node:test` patterns:

1. In `node:test` every `test()` is actually a promise.
2. In `node:test` every test function is wrapped in a promise.
3. In `node:test` there is not a subtest `.end` or `.autoEnd` concept.

So if we have a `tap` test like:

```js
tap.test('some stuff', t => {
	t.test('one', t => {
		t.equal(1, 1)
	})

	t.test('two', t => {
		foo('bar', (error, val) => {
			t.error(error, 'no error should occur')
			t.equal(val, 'bar')
			t.end()
		})

		function foo(param, cb) {
			cb(param)
		}
	})
})
```

The equivalent in `node:test` would be:

```js
test('some stuff', async t => {
	await t.test('one', t => {
		assert.equal(1, 1)
	})

	await t.test('two', async t => {
		const { promise, resolve } = Promise.withResolvers()

		foo('bar', (error, val) => {
			assert.equal(error, undefined, 'no error should occur')
			assert.equal(val, 'bar')
			resolve()
		})
		await promise

		function foo(param, cb) {
			cb(param)
		}
	})
})
```

### DIFF
--- a/test/lib/test-collector.js
+++ b/test/lib/test-collector.js
@@ -1,0 +1,194 @@
+/*
+ * Copyright 2024 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+// This provides an in-process http server to use in place of
+// collector.newrelic.com. It allows for custom handlers so that test specific
+// assertions can be made.
+
+const https = require('node:https')
+const querystring = require('node:querystring')
+const helper = require('./agent_helper')
+const fakeCert = require('./fake-cert')
+
+class Collector {
+  #handlers = new Map()
+  #server
+  #address
+
+  constructor() {
+    this.#server = https.createServer({
+      key: fakeCert.privateKey,
+      cert: fakeCert.certificate
+    })
+    this.#server.on('request', (req, res) => {
+      const qs = querystring.decode(req.url.slice(req.url.indexOf('?') + 1))
+      const handler = this.#handlers.get(qs.method)
+      if (typeof handler !== 'function') {
+        res.writeHead(500)
+        return res.end('handler not found: ' + req.url)
+      }
+
+      res.json = function ({ payload, code = 200 }) {
+        this.writeHead(code, { 'content-type': 'application/json' })
+        this.end(JSON.stringify(payload))
+      }
+
+      handler.isDone = true
+      handler(req, res)
+    })
+
+    // We don't need this server keeping the process alive.
+    this.#server.unref()
+  }
+
+  /**
+   * A configuration object that can be passed to an "agent" instance so that
+   * the agent will communicate with this test server instead of the real
+   * server.
+   *
+   * Important: the `.listen` method must be invoked first in order to have
+   * the `host` and `port` defined.
+   *
+   * @returns {object}
+   */
+  get agentConfig() {
+    return {
+      host: this.host,
+      port: this.port,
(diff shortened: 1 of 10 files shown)


test/lib/test-collector.js
jsumners-nr, line 1: This was devised to replace the usage of `nock` in `test/unit/agent/agent.test.js`. I encountered an issue where one of the subtests was unable to be resolved due to a timing issue between promise resolution and the event loop finishing. Utilizing "real" network requests helped me resolve the problem. If I remember correctly, it was this test causing grief: https://github.com/newrelic/node-newrelic/blob/09636a4ce90969e7aea229ef008bd35f57e09217/test/unit/agent/agent.test.js#L494-L504

Maybe the test could have been reworked with `nock` still being used, but I am less convinced of `nock`'s utility nowadays than I used to be. While the pattern is not used by any tests in this PR, by using the approach in this PR we can do things like:

```js
test('something', async t => {
	const { collector } = t.ctx
	const { promise, resolve = Promise.withResolvers()

	collector.addHandler('/some/endpoint', (req, res) => {
		// Verify that the client generated the correct request structure:
		assert.equal(req.url, '/some/endpoint?with=query_params')
		res.writeHead(200)
		res.end()
		resolve()
	})

	await promise
})
```

Yes, it is possible with `nock`, but, despite using it for many years, I find myself having to go reference the documentation _every time_ I want to do something like that.

test/lib/test-collector.js
```diff
@@ -0,0 +1,194 @@
...
+// This provides an in-process http server to use in place of
+// collector.newrelic.com. It allows for custom handlers so that test specific
+// assertions can be made.
+
+const https = require('node:https')
+const querystring = require('node:querystring')
+const helper = require('./agent_helper')
+const fakeCert = require('./fake-cert')
+
+class Collector {
```
jsumners-nr, line 17: This should be generic enough to utilize in all tests that need to talk to the remote collector. I haven't actually looked, but I bet we can replace the `fake-collector` references in `test/integration/collector-remote-method.tap.js` with this.

test/unit/agent/agent.test.js
```diff
@@ -16,643 +19,562 @@ const Transaction = require('../../../lib/transaction')
...
+    await promise
   })
 
-  t.test('should start aggregators after initial harvest', (t) => {
+  await t.test('should start aggregators after initial harvest', async (t) => {
+    const { promise, resolve } = promiseResolvers()
+    const { agent, collector } = t.nr
+
     sinon.stub(agent.harvester, 'start')
+    t.after(() => sinon.restore())
```
jsumners-nr, line 433: This `sinon.restore()` is the sole reason for updating to `sinon@5` (and evidently `sinon@18` is the current version). Basically, in `sinon@5` the default import is itself a "sandbox" and thus obviates the need to `const sandbox = sinon.sandbox()` and do a dance around that for this simple test.

PR #2450, opened 2024-08-12
chore: Remove promise resolvers from callback based agent unit tests

jsumners-nr:
This PR removes the promise resolvers workaround from the agent unit tests. I discovered that the test function supports a second `done`/`end` parameter that is a function to be invoked when the test is complete.

Note:

```js
// This test will not work as the promise from the `async` takes precedence:
test('something', async (t, end) => {
	someCallback('does something', (error) => {
		end(error)
	})
})

// This test will work as expected because no promises are involved:
test('something', (t, end) => {
	someCallback('does something', (error) => {
		end(error)
	})
})

// If we need async operations, we should still use promise resolvers:
test('something', async (t) => {
	const { promise, resolve, reject } = Promise.withResolvers()
	const aThing = await getThing()
	someCallback('does something', { with: aThing }, (error) => {
		if (error) { return reject(error) }
		resolve()
	})
	await promise
})
```

### DIFF
--- a/test/unit/agent/agent.test.js
+++ b/test/unit/agent/agent.test.js
@@ -7,7 +7,6 @@
 
 const test = require('node:test')
 const assert = require('node:assert')
-const promiseResolvers = require('../../lib/promise-resolvers')
 const Collector = require('../../lib/test-collector')
 
 const sinon = require('sinon')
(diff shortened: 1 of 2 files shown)


jsumners-nr:
> when should we use the done callback vs writing promise based tests when converting?

I think that would be a case-by-case basis, but I suspect the majority of our tests will be callback based due to needing to wrap them with various agent callbacks.

PR #2473, opened 2024-08-15
chore: Reverted to upstream require-in-the-middle

jsumners-nr:
This PR gets us back to using the upstream version of `require-in-the-middle` instead of our fork.


bizob2828 (approved):
LGTM

PR #2496, opened 2024-08-20
fix: Improved AWS Lambda event detection

### LINKED ISSUE #2291: Undefined header errors with LambdaProxyWebRequest and destination lambdas

## Description

We use the [serverless-newrelic-lambda-layers](https://www.npmjs.com/package/serverless-newrelic-lambda-layers) plugin for our lambdas. This includes both http lambdas and destination lambdas. We have experienced an issue whereby the `x-forwarded-port` header is attempted to be retrieved from non-existent headers when our destination lambda is invoked. This is preventing the lambda from running.

## Expected Behavior

This issue started occurring on May 16th. It appears that the [node-newrelic v11.17.0](https://github.com/newrelic/node-newrelic/releases/tag/v11.17.0) release introduced a change that is causing this problem, which is being used by the `serverless-newrelic-lambda-layers`.

```
"TypeError: Cannot read properties of undefined (reading 'x-forwarded-port')",
       "    at new LambdaProxyWebRequest (/opt/nodejs/node_modules/newrelic/lib/serverless/api-gateway.js:20:29)",
       "    at Runtime.wrappedHandler (/opt/nodejs/node_modules/newrelic/lib/serverless/aws-lambda.js:152:28)",
       "    at runInContextCb (/opt/nodejs/node_modules/newrelic/lib/shim/shim.js:1188:22)",
       "    at AsyncLocalStorage.run (node:async_hooks:338:14)",
       "    at AsyncLocalContextManager.runInContext (/opt/nodejs/node_modules/newrelic/lib/context-manager/async-local-context-manager.js:65:36)",
       "    at TransactionShim.applySegment (/opt/nodejs/node_modules/newrelic/lib/shim/shim.js:1178:25)",
       "    at Runtime.transactionWrapper (/opt/nodejs/node_modules/newrelic/lib/shim/transaction-shim.js:414:17)",
       "    at Runtime.patchHandlerSync [as handler] (/opt/nodejs/node_modules/newrelic-lambda-wrapper/index.js:140:25)",
```

Looking at this release it appears there may be an issue in https://github.com/newrelic/node-newrelic/pull/2191 around the APIGateway check to see if this is a lambda proxy event. This previously used to check for the presence of headers, but was [updated to just check the version number, e.g. returns `true` if the version is 1.0.](https://github.com/newrelic/node-newrelic/blob/main/lib/serverless/api-gateway.js#L105-L113). If this is true then it attempts to normalise the headers and pull the `x-forwarded-port` header from it, which fails as the [`lowerCaseHeaders` are undefined](https://github.com/newrelic/node-newrelic/blob/main/lib/serverless/api-gateway.js#L15-L20).

For the lambda experiencing this issue it is an AWS destination lambda vs. http so does not have those headers. The lambda proxy check in this change is however [incorrectly identifying it as needing the web proxy wrapped around it](https://github.com/newrelic/node-newrelic/blob/main/lib/serverless/aws-lambda.js#L151-L152).

## Steps to Reproduce

This can be replicated by using two lambdas.
* One lambda deployed as a [destination lambda](https://aws.amazon.com/blogs/compute/introducing-aws-lambda-destinations/)
* The second lambda, performing any given operation, and configured to invoke the destination lambda on success or failure

After successfully executing the normal lambda operation the destination lambda will not be invoked and the logs will contain the information as per the description re: the undefined object when trying to get the headers.

## Your Environment

* ex: Node version:

## Additional context

As per the server-newrelic-lambda-layers documentation we have added the custom `layerArn` value to pin to an earlier version of the lambda layers (`68` as `69` introduces the problem) that contains the previous version of the node agent that does not contain this breaking change. This is working as expected.

```
newRelic:
    layerArn: arn:aws:lambda:ap-southeast-2:451483290750:layer:NewRelicNodeJS18X:68
```

jsumners-nr:
This PR resolves #2291.

### DIFF
--- a/lib/serverless/api-gateway.js
+++ b/lib/serverless/api-gateway.js
@@ -106,38 +106,45 @@ function isLambdaProxyEvent(event) {
   return isGatewayV1Event(event) || isGatewayV2Event(event)
 }
 
+const v1Keys = [
+  'body',
+  'headers',
+  'httpMethod',
+  'isBase64Encoded',
+  'multiValueHeaders',
+  'multiValueQueryStringParameters',
+  'path',
+  'pathParameters',
+  'queryStringParameters',
+  'requestContext',
+  'resource',
+  'stageVariables',
+  'version'
+].join(',')
+
 function isGatewayV1Event(event) {
-  let result = false
-
-  if (event?.version === '1.0') {
-    result = true
-  } else if (
-    typeof event?.path === 'string' &&
-    (event.headers ?? event.multiValueHeaders) &&
-    typeof event?.httpMethod === 'string'
-    // eslint-disable-next-line sonarjs/no-duplicated-branches
-  ) {
-    result = true
-  }
-
-  return result
+  const keys = Object.keys(event).sort().join(',')
+  return keys === v1Keys && event?.version === '1.0'
 }
 
+const v2Keys = [
+  'body',
+  'cookies',
+  'headers',
+  'isBase64Encoded',
+  'pathParameters',
+  'queryStringParameters',
+  'rawPath',
+  'rawQueryString',
+  'requestContext',
+  'routeKey',
+  'stageVariables',
+  'version'
+].join(',')
+
 function isGatewayV2Event(event) {
-  let result = false
-
-  if (event?.version === '2.0') {
-    result = true
-  } else if (
-    typeof event?.requestContext?.http?.path === 'string' &&
-    Object.prototype.toString.call(event.headers) === '[object Object]' &&
-    typeof event?.requestContext?.http?.method === 'string'
-    // eslint-disable-next-line sonarjs/no-duplicated-branches
-  ) {
-    result = true
-  }
-
-  return result
+  const keys = Object.keys(event).sort().join(',')
+  return keys === v2Keys && event?.version === '2.0'
 }
 
 /**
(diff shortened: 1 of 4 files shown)


lib/serverless/api-gateway.js
```diff
@@ -106,38 +106,45 @@ function isLambdaProxyEvent(event) {
...
-    result = true
-  } else if (
-    typeof event?.path === 'string' &&
-    (event.headers ?? event.multiValueHeaders) &&
-    typeof event?.httpMethod === 'string'
-    // eslint-disable-next-line sonarjs/no-duplicated-branches
-  ) {
-    result = true
-  }
+const v1Keys = [
```
jsumners-nr, line 109: I can't find any information on what AWS does when some field from this list is not included in the payload. Does it send with `field: null`? Or does it completely omit the field. We are assuming that it sends `field: null` with this implementation, because the v1 example shows some such fields.

test/unit/serverless/fixtures.js
```diff
@@ -0,0 +1,181 @@
...
+  isBase64Encoded: false,
+  stageVariables: {
+    stageVariable1: 'value1',
+    stageVariable2: 'value2'
+  }
+}
+
+// Event used when one Lambda directly invokes another Lambda.
+// https://docs.aws.amazon.com/lambda/latest/dg/invocation-async-retain-records.html#invocation-async-destinations
+const lambaV1InvocationEvent = {
```
jsumners-nr, line 150: I'm not completely convinced this is the only payload shape that Lambda sends with direct invocations. But I can't find any other documentation to prove that.

PR #2517, opened 2024-08-26
chore: Convert `config` to `node:test`

amychisholm03:


test/unit/config/attribute-filter.test.js
```diff
@@ -139,29 +126,35 @@ tap.test('#filter', (t) => {
...
+    assert.equal(filter.filterTransaction(DESTS.TRANS_COMMON, 'abc'), DESTS.NONE)
 
-    t.end()
+    assert.equal(filter.filterTransaction(DESTS.NONE, 'a.c'), DESTS.TRANS_COMMON)
+    assert.equal(filter.filterTransaction(DESTS.NONE, 'abc'), DESTS.NONE)
   })
 })
 
-function makeFilterAssertions(t, filter) {
+function makeFilterAssertions(filter) {
```
jsumners-nr, line 137: When I read `makeFilterAssertions` I expect it will return some validator function that would then need to be invoked. But it's really doing the validation itself. I think we should rename this function:

```suggestion
function validateFilter(filter) {
```

test/unit/config/config-formatters.test.js
```diff
@@ -5,175 +5,146 @@
...
+      assert.deepStrictEqual(result, [{ key: 'value' }])
     })
 
-    t.test('should log error and return null if it cannot parse option as json', (t) => {
+    await t.test('should log error and return null if it cannot parse option as json', () => {
       const loggerMock = { error: sinon.stub() }
       const val = 'invalid'
-      t.notOk(formatters.objectList(val, loggerMock))
-      t.equal(
+      assert.ok(!formatters.objectList(val, loggerMock))
```
jsumners-nr, line 107: `assert.notOk` exists. But I would write:
```suggestion
      assert.equal(formatters.objectList(val, loggerMock), null)
```

test/unit/config/config-location.test.js
```diff
@@ -61,48 +60,41 @@ tap.test('when overriding the config file location via NEW_RELIC_HOME', (t) => {
...
-      delete process.env.NEW_RELIC_HOME
-    })
-
-    let configuration
-
-    t.doesNotThrow(() => {
-      configuration = Config.initialize()
-    })
+  await t.test('should ignore the configuration file completely when so directed', (t, end) => {
+    try {
```
jsumners-nr, line 76: I believe you've opted for a `try/finally` because `t.teardown` doesn't exist. But it actually does:

```js
t.after(() => {
	delete process.env.thing
})
``

jsumners-nr (approved):

Issue #2526, opened 2024-08-28
Convert the `custom-events` and `distributed_tracing` unit tests to `node:test`


amychisholm03:
Converts the tests in `test/unit/custom-events` (only 1 test in here, so I'm lumping it in w/ `distributed_tracing`) and `test/unit/distributed_tracing` to use `node:test` instead of `tap`

PR #2527, opened 2024-08-29
test: Replace distributed tracing tests with `node:test`

amychisholm03:
This PR resolves https://github.com/newrelic/node-newrelic/issues/2526.

### DIFF
--- a/test/unit/custom-events/custom-event-aggregator.test.js
+++ b/test/unit/custom-events/custom-event-aggregator.test.js
@@ -4,7 +4,8 @@
  */
 
 'use strict'
-const tap = require('tap')
+const test = require('node:test')
+const assert = require('node:assert')
 const CustomEventAggregator = require('../../../lib/custom-events/custom-event-aggregator')
 const Metrics = require('../../../lib/metrics')
 const NAMES = require('../../../lib/metrics/names')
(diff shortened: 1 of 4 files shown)


test/unit/custom-events/custom-event-aggregator.test.js
```diff
@@ -33,36 +32,34 @@ tap.test('Custom Event Aggregator', (t) => {
     )
   })
 
-  t.afterEach(() => {
-    eventAggregator = null
+  t.afterEach((ctx) => {
+    ctx.nr.eventAggregator = null
   })
```
jsumners-nr, line 37: Technically this is not necessary because each test will have a new scope. Thus, the `ctx.nr.eventAggregator` from test A will not be the same one in test B. But having the `afterEach` clear it is not harming anything. So this comment is informational, not blocking.

test/unit/distributed_tracing/dt-cats.test.js
```diff
@@ -126,13 +127,13 @@ tap.test('distributed tracing', function (t) {
 
             const attributes = event[0]
             arbitrary.forEach((key) => {
-              t.ok(attributes[`${key}`], `${type} should have ${key}`)
+              assert.ok(attributes[`${key}`], `${type} should have ${key}`)
             })
             unexpected.forEach((key) => {
-              t.notOk(attributes[`${key}`], `${type} should not have ${key}`)
+              assert.ok(!attributes[`${key}`], `${type} should not have ${key}`)
```
jsumners-nr, line 133: `assert.notOk` is also available. I would find that easier to read than `!something`.

jsumners-nr (approved):
Looks good. None of my comments are blockers.

test/unit/distributed_tracing/dt-cats.test.js
amychisholm03: I don't see it in this list (https://nodejs.org/api/assert.html) and whenever I use it it says that assert.notOk is not a function. Maybe I have a different version?

jsumners-nr: Sorry. I must be thinking of `notEqual`.

Issue #2548, opened 2024-09-09
Convert `serverless` tests to `node:test`


jsumners-nr:
As titled.

PR #2561, opened 2024-09-11
chore: Document emitted events

jsumners-nr:
While working on #2548 I needed to understand when some events are firing and learned that we do not have any documentation around events. Ideally, we'd have a document that outlines the common event sequences, but I don't have that picture right now. What this PR includes is jsdoc for almost every event the codebase generates. There is a whole class events in the `Config` namespace that I could not document. These events are emitting events named after configuration keys. There are two problems here:

1. It's not clear what the event names will actually be. As an example, what would be fired if `enabled` from `{ feature_name: { enabled: true } }` is changed to `false`? Would the event name be `enabled`, `feature_name`, or something like `feature_name.enabled`?
2. There's no way to document dynamic event names with jsdoc. At least no way that I can think of that would really clarify anything.

Also, _this PR changes_ the events around data sending. The current event names are like `starting error_data data send.`. This PR changes them to `starting_data_send-error_data`. We can't document event names with spaces, and other non-alphanum characters, in them. Indeed, it was a bit of a challenge to find a separator character to separate the prefix and actual event name that jsdoc would handle correctly.

So, while not everything is fully documented, I think the PR puts us in a better position than before.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -58,6 +58,94 @@ const INITIAL_HARVEST_DELAY_MS = 1000
 const DEFAULT_HARVEST_INTERVAL_MS = 60000
 
 /**
+ * Indicates that the agent has finished connecting. It is preceded by
+ * {@link Agent#event:connecting}.
+ *
+ * @event Agent#connected
+ */
+
+/**
+ * Indicates that the agent is in the connecting state. That is, it is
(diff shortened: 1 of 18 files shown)


bizob2828 (approved):
LGTM

PR #2583, opened 2024-09-17
chore: Removed noisy test log

jsumners-nr:
Refactors the `test/unit/util/obfuscate-sql.test.js` script to generate less noise.


bizob2828 (approved):
I know this one annoyed you before. Personally I don't see a diff, it's still "noisy" just in a diff way. But it's fine with me

jsumners-nr:
> I know this one annoyed you before. Personally I don't see a diff, it's still "noisy" just in a diff way. But it's fine with me

See https://github.com/newrelic/node-newrelic/pull/2582/files

image

PR #2721, opened 2024-11-11
feat: Added support for Bedrock invoke model user request

cmcadams-newrelic:
## Description

This Pr will improve the support for using the invoke model with bedrock LLM. the Invoke API expects the body to be of type `{[content: "user or assistant message"]}`.

We will now check for the type of the input if the type is a string we will return the value as is. If the param is an object we will attempt to reduce the object to just the message value.

## How to Test

run test suite. 

## Related Issues

https://new-relic.atlassian.net/browse/NR-326020

### DIFF
--- a/lib/llm-events/aws-bedrock/bedrock-command.js
+++ b/lib/llm-events/aws-bedrock/bedrock-command.js
@@ -85,7 +85,11 @@ class BedrockCommand {
       result = this.#body.prompt
     } else if (this.isClaude3() === true) {
       result = this.#body?.messages?.reduce((acc, curr) => {
-        acc += curr?.content ?? ''
+        if (Object.keys(curr?.content).length) {
+          acc += curr?.content.text ?? ''
+        } else {
+          acc += curr?.content ?? ''
+        }
         return acc
       }, '')
     }
--- a/lib/llm-events/aws-bedrock/bedrock-command.js
+++ b/lib/llm-events/aws-bedrock/bedrock-command.js
@@ -77,7 +77,7 @@ class BedrockCommand {
     } else if (this.isCohereEmbed() === true) {
       result = this.#body.texts.join(' ')
     } else if (
-      (this.isClaude() === true && this.isClaude3() === false) ||
+      this.isClaude() === true ||
       this.isAi21() === true ||
       this.isCohere() === true ||
       this.isLlama() === true
--- a/lib/llm-events/aws-bedrock/bedrock-command.js
+++ b/lib/llm-events/aws-bedrock/bedrock-command.js
@@ -77,7 +77,7 @@ class BedrockCommand {
     } else if (this.isCohereEmbed() === true) {
       result = this.#body.texts.join(' ')
     } else if (
-      this.isClaude() === true ||
+      (this.isClaude() === true && this.isClaude3() === false) ||
       this.isAi21() === true ||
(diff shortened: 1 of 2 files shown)


lib/llm-events/aws-bedrock/bedrock-command.js
```diff
@@ -77,15 +77,19 @@ class BedrockCommand {
     } else if (this.isCohereEmbed() === true) {
       result = this.#body.texts.join(' ')
     } else if (
-      this.isClaude() === true ||
+      (this.isClaude() === true && this.isClaude3() === false) ||
```
jsumners-nr, line 80: Why is this condition change necessary? The `isClaude` method does not match for v3, or it shouldn't.

jsumners-nr (changes requested):
I do not see any tests for this change.

lib/llm-events/aws-bedrock/bedrock-command.js
cmcadams-newrelic: Thank you for your review, and yes, you are correct! `this.isClaude3() === false` is unneeded at this time, and is pushing the solution into redundancy. Im happy to remove this change from the PR.

____
For context: 

I want to add this change because I think it improves readability by explicitly calling out this will not match `claude 3.x`. 

Currently, Claude is the only llm that we treat as a separate llm based on its version. I get that is because of the API. But thinking about the future would `isClaude` support `anthropic.claude-v4` or would we need to continue the pattern of creating a new possible llm `isClaude4`?

Furthermore, I think this will help developers of all exp. levels to quickly consume the code and be able to contribute.

jsumners-nr: I think it depends on how they provide these things via the AWS API. We initially had one "Claude". So we implemented an `isClaude` method. Then they put out a "Claude v3" with a completely different identifier. So we added an `isClaude3` method.

I think this discussion is valid, but probably one to have separate from this PR.

cmcadams-newrelic:
> I do not see any tests for this change.

@jsumners-nr. Thank you for the review, I have included tests in my latest commit. the tests are to cover the changes in Claude 3.5 and the invoke model. In addition I have updated this PR's description calling out the scope of this change.

https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html#API_runtime_InvokeModel_RequestBody

test/unit/llm-events/aws-bedrock/bedrock-command.test.js
```diff
@@ -182,6 +189,30 @@ test('claude3 complete command works', async (t) => {
...
+  t.nr.updatePayload(payload)
+  const cmd = new BedrockCommand(t.nr.input)
+  assert.equal(cmd.isClaude3(), true)
+  assert.equal(cmd.maxTokens, 25)
+  assert.equal(cmd.modelId, payload.modelId)
+  assert.equal(cmd.modelType, 'completion')
+  assert.equal(cmd.prompt, payload.body.messages[0].content.text)
+  assert.equal(cmd.temperature, payload.body.temperature)
+})
 test('cohere minimal command works', async (t) => {
```
jsumners-nr, line 216:
```suggestion
})

test('cohere minimal command works', async (t) => {
```

jsumners-nr (approved):

PR #2741, opened 2024-11-14
fix: Corrected expected shape of user input in get Prompt()

cmcadams-newrelic:
## Description

The shape of the users prompt message was incorrect for claude 3.5 
```
 // Prepare the payload for the model.
  const payload = {
    anthropic_version: "bedrock-2023-05-31",
    max_tokens: 1000,
    messages: [
      {
        role: "user",
        content: [{ type: "text", text: prompt }],
      },
    ],
  };
```
https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/javascript_bedrock-runtime_code_examples.html#anthropic_claude

## How to Test

run unit test suite.

### DIFF
--- a/lib/llm-events/aws-bedrock/bedrock-command.js
+++ b/lib/llm-events/aws-bedrock/bedrock-command.js
@@ -86,7 +86,7 @@ class BedrockCommand {
     } else if (this.isClaude3() === true) {
       result = this.#body?.messages
         ?.reduce((acc, curr) => {
-          if (curr?.role === 'user') {
+          if (curr?.role !== 'assistant') {
             if (typeof curr?.content === 'string') {
               acc.push(curr?.content)
             } else {
--- a/lib/llm-events/aws-bedrock/bedrock-command.js
+++ b/lib/llm-events/aws-bedrock/bedrock-command.js
@@ -84,18 +84,19 @@ class BedrockCommand {
     ) {
       result = this.#body.prompt
     } else if (this.isClaude3() === true) {
-      result = this.#body?.messages?.reduce((acc, curr) => {
-        if (curr?.role === 'user') {
-          if (typeof curr?.content === 'string') {
-            acc += curr?.content
-          } else {
-            acc += curr?.content.map((msgObj) => msgObj.text)
+      result = this.#body?.messages
+        ?.reduce((acc, curr) => {
+          if (curr?.role === 'user') {
+            if (typeof curr?.content === 'string') {
+              acc.push(curr?.content)
+            } else {
+              const mappedMsgObj = curr?.content.map((msgObj) => msgObj.text)
+              acc.push(mappedMsgObj)
+            }
           }
-        } else {
-          acc += ' '
-        }
-        return acc
-      }, '')
+          return acc
+        }, [])
+        .join(' ')
     }
     return result
   }
--- a/lib/llm-events/aws-bedrock/bedrock-command.js
+++ b/lib/llm-events/aws-bedrock/bedrock-command.js
@@ -87,8 +87,8 @@ class BedrockCommand {
       result = this.#body?.messages?.reduce((acc, curr) => {
         if (typeof curr?.content === 'string') {
           acc += curr?.content
-        } else if (Object.keys(curr?.content).length) {
-          acc += curr?.content.text ?? ''
+        } else if (Array.isArray(curr?.content) && Object.keys(curr?.content[0]).length) {
+          acc += curr?.content[0].text ?? ''
         }
         return acc
       }, '')
--- a/lib/llm-events/aws-bedrock/bedrock-command.js
+++ b/lib/llm-events/aws-bedrock/bedrock-command.js
@@ -85,10 +85,14 @@ class BedrockCommand {
       result = this.#body.prompt
     } else if (this.isClaude3() === true) {
       result = this.#body?.messages?.reduce((acc, curr) => {
-        if (typeof curr?.content === 'string') {
-          return curr?.content
-        } else if (Array.isArray(curr?.content) && Object.keys(curr?.content[0]).length) {
-          return curr?.content[curr?.content.length - 1].text ?? ''
+        if (curr?.role === 'user') {
+          if (typeof curr?.content === 'string') {
+            acc += curr?.content
+          } else {
+            acc += curr?.content.map((msgObj) => msgObj.text)
+          }
+        } else {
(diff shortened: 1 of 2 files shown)


lib/llm-events/aws-bedrock/bedrock-command.js
```diff
@@ -86,9 +86,9 @@ class BedrockCommand {
     } else if (this.isClaude3() === true) {
       result = this.#body?.messages?.reduce((acc, curr) => {
         if (typeof curr?.content === 'string') {
-          acc += curr?.content
-        } else if (Object.keys(curr?.content).length) {
-          acc += curr?.content.text ?? ''
+          return curr?.content
+        } else if (Array.isArray(curr?.content) && Object.keys(curr?.content[0]).length) {
+          return curr?.content[curr?.content.length - 1].text ?? ''
```
cmcadams-newrelic, line 91: we only care about the last message, which would be the most recent user prompt.

jsumners-nr: If that is the case, do we need to retain the `.reduce`?

cmcadams-newrelic: Im thinking in the future we might want to use the reduce to filter out messages from `{type: text}` and `{type: image}` at this current point in time. I don't see a benefit with using the `reduce` function.

test/unit/llm-events/aws-bedrock/bedrock-command.test.js
```diff
@@ -27,7 +27,7 @@ const claude = {
 const claude35 = {
   modelId: 'anthropic.claude-3-5-sonnet-20240620-v1:0',
   body: {
-    messages: [{ role: 'user', content: { type: 'text', text: 'who are you' } }]
+    messages: [{ role: 'user', content: [{ type: 'text', text: 'who are you' }] }]
```
cmcadams-newrelic, line 30: addresses the bug where the shape of the object was incorrect. content is an array of objects

jsumners-nr:
@cmcadams-newrelic in Slack you mentioned wanting to get this PR ready for review by the end of today. I see that you have marked it ready, but I don't see any new commits. It sounded like you would be making some changes. Is this ready?

lib/llm-events/aws-bedrock/bedrock-command.js
jsumners-nr: How are other agent teams handling these cases? I haven't seen a spec that specifically calls out this model's operations.

RyanKadri: I think for 1, it's just a relatively new API that got introduced after we added Bedrock support. Python also seems to only handle the string case and not the multi-chunk case. I'm not positive but I'd guess that's the same in other agents (who support Bedrock) as well.

For 2, I think Claude is actually getting closer to how we handle Chat Completions in OpenAI [here](https://github.com/newrelic/node-newrelic/blob/57e6be9f4717fde3caada0e3ca3680959180f928/lib/instrumentation/openai.js#L139). For one completion call, we already split out several message events based on the context messages that were passed in

jsumners-nr (changes requested):
Tests are failing.

cmcadams-newrelic:
@jsumners-nr Sorry for the delay. I have the fixed the tests and pushed up all required changes

jsumners-nr:
@cmcadams-newrelic the unit tests are failing. Please rebase against `main` and run `npm run unit`.

cmcadams-newrelic:
@jsumners-nr I have rebased this branch with main as well as corrected the failing test.

lib/llm-events/aws-bedrock/bedrock-command.js
```diff
@@ -84,14 +84,19 @@ class BedrockCommand {
...
+      result = this.#body?.messages
+        ?.reduce((acc, curr) => {
+          if (curr?.role !== 'assistant') {
+            if (typeof curr?.content === 'string') {
+              acc.push(curr?.content)
+            } else {
+              const mappedMsgObj = curr?.content.map((msgObj) => msgObj.text)
+              acc.push(mappedMsgObj)
+            }
+          }
```
jsumners-nr, line 96:
```suggestion
          if (curr?.role !== 'assistant') {
            return acc
          }
          if (typeof curr?.content === 'string') {
            acc.push(curr?.content)
          } else {
            const mappedMsgObj = curr?.content.map((msgObj) => msgObj.text)
            acc.push(mappedMsgObj)
          }
```

test/unit/llm-events/aws-bedrock/bedrock-command.test.js
jsumners-nr, line 30:
```suggestion
    messages: [
      { role: 'user', content: [{ type: 'text', text: 'who are' }] },
      { role: 'assistant', content: [{ type: 'text', text: 'researching' }] },
      { role: 'user', content: [{ type: 'text', text: 'you' }] }
    ]
```

test/unit/llm-events/aws-bedrock/bedrock-command.test.js
```diff
@@ -207,7 +207,7 @@ test('claude35 minimal command works', async (t) => {
   assert.equal(cmd.maxTokens, undefined)
   assert.equal(cmd.modelId, claude35.modelId)
   assert.equal(cmd.modelType, 'completion')
-  assert.equal(cmd.prompt, claude35.body.messages[0].content.text)
+  assert.equal(cmd.prompt, claude35.body.messages[0].content[0].text)
```
jsumners-nr, line 210:
```suggestion
  assert.equal(cmd.prompt, 'who are you')
```

test/unit/llm-events/aws-bedrock/bedrock-command.test.js
```diff
@@ -221,7 +221,7 @@ test('claude35 complete command works', async (t) => {
   assert.equal(cmd.maxTokens, 25)
   assert.equal(cmd.modelId, payload.modelId)
   assert.equal(cmd.modelType, 'completion')
-  assert.equal(cmd.prompt, payload.body.messages[0].content.text)
+  assert.equal(cmd.prompt, payload.body.messages[0].content[0].text)
```
jsumners-nr, line 224:
```suggestion
  assert.equal(cmd.prompt, 'who are you')
```

jsumners-nr (approved):

PR #2773, opened 2024-11-20
chore: Updated q, superagent, and when tests to node:test

### LINKED ISSUE #2734: Migrate `q`, `when` and `superagent` versioned tests to `node:test`

jsumners-nr:
This PR resolves #2734.


test/versioned/when/segments.test.js
```diff
@@ -0,0 +1,448 @@
...
+        process.nextTick(transaction.end.bind(transaction))
+      })
+    })
+
+    await plan.completed
+  })
+
+  // Note: this test was originally a `promise_segments: false` test listed
+  // under the segments enabled suite.
+  await t.test('then handler that returns a new promise', async (t) => {
```
jsumners-nr, line 86: This test was originally https://github.com/newrelic/node-newrelic/blob/dffdac7f450a23c6e23a173c91bda583d40e573d/test/versioned/when/legacy-promise-segments.js#L67-L78

I don't think it should have been under the "promise_segments: true" group to begin with.

test/versioned/when/segments.test.js
```diff
@@ -0,0 +1,448 @@
...
+          process.nextTick(transaction.end.bind(transaction))
+        })
+    })
+
+    await plan.completed
+  })
+
+  // Note: this test was originally a `promise_segments: false` test listed
+  // under the segments enabled suite.
+  await t.test('catch handler with error from subsequent promise', async (t) => {
```
jsumners-nr, line 189: This one was originally https://github.com/newrelic/node-newrelic/blob/dffdac7f450a23c6e23a173c91bda583d40e573d/test/versioned/when/legacy-promise-segments.js#L144-L156

It also should not have been under `promise_segments: true`.

bizob2828 (approved):
LGTM, thanks for cleaning up the when tests

jsumners-nr:
> LGTM, thanks for cleaning up the when tests

You're welcome. Really, I just couldn't even do the conversion without doing it 😭

PR #2780, opened 2024-11-21
fix: Fixed matching for ALB and API Gateway (v1 & v2) events for Lambda

### LINKED ISSUE #2753: Amazon API Gateway HTTP API (payload v2) event not detected correctly
## Description

The Node.js agent incorrectly assumes that all fields are present in an API Gateway v2 proxy event object. While this appears to be true for API Gateway v1 proxy event, it is not true for API Gateway v2. For example, in v1, if there are no stage variables configured on the API Gateway stage, the `stageVariables` field in the proxy event is null. However, in v2, the field is absent. This applies to other fields - for example, if there are not path parameters or query string parameters in the request, the `pathParameters` and `queryStringParameters` fields are absent from the event object.

The following logic assumes that all fields are present in API Gateway v2 proxy event.

https://github.com/newrelic/node-newrelic/blob/96400304e451502ce2f09b569b46eaff271fd9ab/lib/serverless/api-gateway.js#L142-L160

The implication is that `isGatewayV2Event` is returning false for valid API Gateway v2 proxy events. In turn, this means that [`isApiGatewayLambdaProxy`](https://github.com/newrelic/node-newrelic/blob/96400304e451502ce2f09b569b46eaff271fd9ab/lib/serverless/aws-lambda.js#L144) is false. For my use case, this meant that tracing headers were not being respected in the proxy event and a new trace was created, rather than adding segments to the existing trace.

The workaround is to set the payload version to 1.0 for API Gateway routes.

## Expected Behavior

API Gateway v2 proxy events with absent fields are recognised as being valid. In turn, trace headers are pulled out of the event and used for distributed tracing.

## Steps to Reproduce

1. Deploy attached CloudFormation template (drop .txt extension, if required). This template creates an Lambda function, API Gateway HTTP API, and sets up an associated integration and route.
2. Perform a HTTP GET to the / route of the API. Observe in the Lambda function's logs that the `stageVariables` key among others (for example, `queryStringParameters` and `pathParameters`) is absent in the event object.

## Your Environment

AWS Lambda, nodejs18.x
newrelic 12.7.0
[template.yaml.txt](https://github.com/user-attachments/files/17808781/template.yaml.txt)

### LINKED ISSUE #2779: Missing attributes for Lambda functions triggered by ALB
A change to Lambda event detection in 12.3.0 resulted in the agent not tracking some attributes of ALB-triggered Lambda requests: https://github.com/newrelic/node-newrelic/commit/5e8b2608d9914e2a4282f7c9c42ff17dfa9f793e

## Expected Behavior

Ensure that web transactions triggered via ALB record 
- http.statusCode
- request.headers.accept
- request.headers.contentLength
- request.headers.contentType
- request.headers.host
- request.method
- request.uri
- response.headers.contentType

mrickard:
## Description
Filtering of API Gateway v1 vs v2 did not allow for Lambda functions triggered by ALB/ELB. Prior to our most recent change to that filtering, ALB events were traced without explicit filtering. This adds explicit checking for an ALB event type, and allows ALB-triggered web requests to be treated as web transactions. 

## How to Test
This adds a unit test at `test/unit/serverless/alb-event.test.js` checking for the expected request properties. Existing tests at `test/unit/serverless/api-gateway-v2.test.js`, `test/unit/serverless/aws-lambda.test.js`, and `test/unit/serverless/utils.test.js` also test the new comparison method.

## Related Issues

Closes NR-341636
Closes #2779 
Closes NR-342521
Closes #2753 
Closes NR-341279

### DIFF
--- a/lib/serverless/api-gateway.js
+++ b/lib/serverless/api-gateway.js
@@ -22,7 +22,7 @@ class LambdaProxyWebRequest {
     }
     this.method = ''
 
-    if (isGatewayV1Event(event) === true) {
+    if (isGatewayV1Event(event) === true || isAlbEvent(event) === true) {
       this.url.path = event.path
       this.method = event.httpMethod
     } else if (isGatewayV2Event(event) === true) {
@@ -54,7 +54,10 @@ class LambdaProxyWebResponse {
  * @returns {Object<string, string>} The normalized query string map
  */
 function normalizeQueryStringParameters(event) {
-  if (!event.multiValueQueryStringParameters) {
+  if (
+    !event.multiValueQueryStringParameters ||
+    Object.keys(event.multiValueQueryStringParameters).length === 0
+  ) {
     return event.queryStringParameters
   }
   return Object.fromEntries(
@@ -107,7 +110,7 @@ function normalizeHeaders(event, lowerCaseKey = false) {
  *                    to create a web transaction.
  */
 function isLambdaProxyEvent(event) {
-  return isGatewayV1Event(event) || isGatewayV2Event(event)
+  return isGatewayV1Event(event) || isGatewayV2Event(event) || isAlbEvent(event)
 }
 
 // See https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format
@@ -158,6 +161,25 @@ function isGatewayV2Event(event) {
   const keys = Object.keys(event).sort().join(',')
   return keys === httpApiV2Keys && event?.version === '2.0'
 }
+const albKeys = [
+  'requestContext',
+  'httpMethod',
+  'path',
+  'queryStringParameters',
+  'headers',
+  'body',
+  'isBase64Encoded',
+  'rawHeaders',
+  'multiValueQueryStringParameters',
+  'pathParameters'
+]
+  .sort()
+  .join(',')
+
+function isAlbEvent(event) {
+  const keys = Object.keys(event).sort().join(',')
+  return keys === albKeys && event?.requestContext?.elb !== undefined
+}
 
 /**
  * Determines if Lambda event appears to be a valid Lambda Proxy response.
@@ -176,5 +198,6 @@ module.exports = {
   isLambdaProxyEvent,
   isValidLambdaProxyResponse,
   isGatewayV1Event,
-  isGatewayV2Event
+  isGatewayV2Event,
+  isAlbEvent
 }
--- a/lib/serverless/api-gateway.js
+++ b/lib/serverless/api-gateway.js
@@ -114,14 +114,30 @@ function isLambdaProxyEvent(event) {
 }
 
 /**
- * Iterates over the minimum signature properties of an event triggered by API Gateway
- * to determine the version. V1 and V2 have a lot of overlap, but some signature differences.
+ * Iterates over the minimum signature properties of an event received by this Lambda function
+ * to determine if this request is triggered from a proxy (API Gateway V1, V2, ALB) or some other service.
+ * If API Gateway, we need to determine which version: V1 and V2 have a lot of overlap, but some signature
+ * differences for which we can test.
+ *
+ * The test is designed to look only at signature properties used by each version, and returns true with the first
+ * top-level match. Each test array has only four elements, and each invocation should incur
+ * only five comparisons: one match for its matching type, and four comparisons for the non-matching type.
+ *
+ * API Gateway v2 HTTP: top-level 'rawPath', 'rawQueryString', 'routeKey'. Possibly 'cookies'
+ * API Gateway v1 HTTP: top-level 'httpMethod', 'resource'. Possibly 'multiValueHeaders', 'multiValueQueryStringParameters'
+ * API Gateway v1 REST same as HTTP, but without top-level `version`
+ * ALB: very similar to API Gateway v1, but requestContext contains an elb property
+ *
+ * In tests, this set of required API Gateway v1 properties has consistently been delivered in event payloads.
+ * Similar tests with API Gateway V2 shows that the cookies property is *only* defined if cookies are present.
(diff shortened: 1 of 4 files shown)


lib/serverless/api-gateway.js
```diff
@@ -158,6 +161,25 @@ function isGatewayV2Event(event) {
   const keys = Object.keys(event).sort().join(',')
   return keys === httpApiV2Keys && event?.version === '2.0'
 }
+const albKeys = [
```
jsumners-nr, line 164:
```suggestion
}

const albKeys = [
```

jsumners-nr, line 164: Please document this as our other event keys are documented.

lib/serverless/api-gateway.js
```diff
@@ -158,6 +161,25 @@ function isGatewayV2Event(event) {
...
+  'path',
+  'queryStringParameters',
+  'headers',
+  'body',
+  'isBase64Encoded',
+  'rawHeaders',
+  'multiValueQueryStringParameters',
+  'pathParameters'
+]
+  .sort()
```
jsumners-nr, line 176: I recommend listing them in sorted order to avoid the unnecessary sorting operation cost.

lib/serverless/api-gateway.js
```diff
@@ -54,7 +54,10 @@ class LambdaProxyWebResponse {
  * @returns {Object<string, string>} The normalized query string map
  */
 function normalizeQueryStringParameters(event) {
-  if (!event.multiValueQueryStringParameters) {
+  if (
+    !event.multiValueQueryStringParameters ||
+    Object.keys(event.multiValueQueryStringParameters).length === 0
```
jsumners-nr, line 59: I don't understand. If we have `multiValueQueryStringParameters` defined at all, we should not have `queryStringParameters` defined. So why would we return `queryStringParameters` if the multi valued property is defined with zero elements?

jsumners-nr (changes requested):

lib/serverless/api-gateway.js
mrickard: In the sample event captured by the customer, both `queryStringParameters` and `multiValueQueryStringParameters` were defined, event though each had zero properties. I could see changing this to check for properties in each. The definition of one, though, doesn't exclude the definition of the other, based on the input they've captured.

lib/serverless/api-gateway.js
mrickard: I'm dispensing with the sort and string conversion, as we're no longer looking for an exact match of all keys.

lib/serverless/api-gateway.js
```diff
@@ -107,56 +110,82 @@ function normalizeHeaders(event, lowerCaseKey = false) {
...
+/**
+ * We need to determine whether this invocation event was triggered from a proxy: API Gateway V1, V2, ALB, or some other service.
+ * API Gateway V1/V2 and ALB events aren't guaranteed always to have the same properties in all cases. There are, however, properties
+ * that are vary consistently depending on the service triggering the event.
+ *
+ * API Gateway v2 HTTP
+ * API Gateway v1 HTTP: top-level httpMethod, resource, multiValueHeaders (possibly), multiValueQueryStringParameters (possibly)
+ * API Gateway v1 REST same as HTTP, but without version
+ * ALB: very similar to API Gateway v1, but requestContext contains an elb property
+ */
```
jsumners-nr, line 143: What is this docblock style comment meant to be attached to?

lib/serverless/api-gateway.js
```diff
@@ -107,56 +110,82 @@ function normalizeHeaders(event, lowerCaseKey = false) {
...
+ * @returns {boolean} Whether this event has matches for the keys we're checking
+ */
+function findProperties(targetEvent, searchFor) {
+  const keys = Object.keys(targetEvent)
+  for (const el of searchFor) {
+    if (keys.indexOf(el) > -1) {
+      return true
+    }
+  }
+  return false
```
jsumners-nr, line 131: Isn't this returning true if _any_ of the elements of `searchFor` are present in `targetEvent`? Shouldn't we replicate the functionality of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/isSubsetOf here?

lib/serverless/api-gateway.js
```diff
@@ -107,56 +110,82 @@ function normalizeHeaders(event, lowerCaseKey = false) {
...
 // See https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html
-const httpApiV1Keys = [...restApiV1Keys.split(','), 'version'].join(',')
 
 function isGatewayV1Event(event) {
-  const keys = Object.keys(event).sort().join(',')
-  if (keys === httpApiV1Keys && event?.version === '1.0') {
+  if (event?.requestContext?.elb !== undefined) {
+    return false
+  }
+  const hasKeys = findProperties(event, uniqueHttpApiV1Keys)
```
jsumners-nr, line 160: Would `eventHasRequriedKeys` convey a more accurate description of the function?

```suggestion
  const hasKeys = eventHasRequiredKeys(event, uniqueHttpApiV1Keys)
```

lib/serverless/api-gateway.js
mrickard: This content best fits with the function that tests for these properties, so I'll consolidate that content.

lib/serverless/api-gateway.js
mrickard: It is indeed a test for *any*, which is intentional. With API Gateway v1, properties like `multiValueHeaders` aren't guaranteed to be defined on the event. If we do see that, we can be certain it's a v1 payload, but its absence doesn't mean that a payload isn't v1. If there's a more efficient test pattern than "test for any signature property and bail on a positive," we should use it...but I'm wary of false negatives in this case.

jsumners-nr: Please document these details in the code.

jsumners-nr (approved):

PR #2825, opened 2024-12-09
fix: Fixed matching for Lambda events containing resource but which aren't API Gateway

### LINKED ISSUE #2824: Setting lambda handler is broken for certain lambda payloads since 12.8.1
## Description

When you have an lambda which receives an event payload which is no api gateways payload but contains one of the key words which are used to check, it is triggered as lambda proxy event and then fails for url parsing. 
image

Also the payload invoking our lambda comes from commercetools so that is nothing we can change, and it could potentially affect more people

Issue was introduced with https://github.com/newrelic/node-newrelic/pull/2780 as here the check is made the second the key resource is used it is an api gateways v1 event

## Expected Behavior

* setting the lambda handler does not throw errors

## Steps to Reproduce

use the following event payload
```
{
  "action": "Update",
  "resource": {  }
}
```

setting the lambda layer should not try to set a transaction for the web request

Engerim:
Fixes #2824

### DIFF
--- a/lib/serverless/api-gateway.js
+++ b/lib/serverless/api-gateway.js
@@ -173,7 +173,7 @@ function isGatewayV1Event(event) {
 const requiredHttpApiV2Keys = ['rawPath', 'rawQueryString', 'routeKey', 'cookies']
 
 function isGatewayV2Event(event) {
-  if (event?.requestContext?.elb !== undefined) {
+  if (event?.requestContext?.elb !== undefined || event?.version === undefined) {
     return false
   }
   return eventHasRequiredKeys(event, requiredHttpApiV2Keys) && event?.version === '2.0'
--- a/lib/serverless/api-gateway.js
+++ b/lib/serverless/api-gateway.js
@@ -158,7 +158,7 @@ const requiredHttpApiV1Keys = [
 
 // See https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html
 function isGatewayV1Event(event) {
-  if (event?.requestContext?.elb !== undefined) {
+  if (event?.requestContext?.elb !== undefined || event?.version === undefined) {
     return false
   }
   const hasKeys = eventHasRequiredKeys(event, requiredHttpApiV1Keys)
--- a/lib/serverless/api-gateway.js
+++ b/lib/serverless/api-gateway.js
@@ -158,7 +158,7 @@ const requiredHttpApiV1Keys = [
 
 // See https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html
 function isGatewayV1Event(event) {
-  if (event?.requestContext?.elb !== undefined || event?.version === undefined) {
+  if (event?.requestContext === undefined || event?.requestContext?.elb !== undefined) {
     return false
   }
   const hasKeys = eventHasRequiredKeys(event, requiredHttpApiV1Keys)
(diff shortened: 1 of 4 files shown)


test/unit/serverless/aws-lambda.test.js
```diff
@@ -108,6 +108,45 @@ test('AwsLambda.patchLambdaHandler', async (t) => {
       body: 'worked'
     }
 
+    await t.test('should not create web transaction', (t, end) => {
```
jsumners-nr, line 111:
```suggestion
    await t.test('should not create web transaction for custom direct invocation payload', (t, end) => {
```

lib/serverless/api-gateway.js
```diff
@@ -158,7 +158,7 @@ const requiredHttpApiV1Keys = [
 
 // See https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html
 function isGatewayV1Event(event) {
-  if (event?.requestContext?.elb !== undefined) {
+  if (event?.requestContext?.elb !== undefined || event?.version === undefined) {
```
jsumners-nr, line 161: Do we not need the same check on line 176?

jsumners-nr (changes requested):
Looking great. I have a suggestion for the test name, and a question about the check. Also, please run `npm run lint`

lib/serverless/api-gateway.js
Engerim: hmm we can do it also here for the sake of it, I works atm because in L179 we check the keys and version 2.0 so even if the check would return true, as long as the custom payload has not by accident also version with 2.0 it will fail

jsumners-nr (dismissed):
Looks good to me.

mrickard (dismissed):

Engerim:
hmmm given that https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format that the request event for the rest api gateway does not contain a version, my fix would break rest api gateway but, as fare as I see, rest api gateway and also http api gateway contain a `requestContext` key

jsumners-nr (dismissed):

mrickard:
@Engerim Thank you! If you want to test the matching with a specific payload, you could add one to the fixtures in `test/unit/serverless/fixtures.js`, which have explicit matching tests in `utils.test.js`. (There's already an explicit test for REST API, which does pass for your PR.)

mrickard (dismissed):

mrickard:
> > @Engerim Thank you! If you want to test the matching with a specific payload, you could add one to the fixtures in `test/unit/serverless/fixtures.js`, which have explicit matching tests in `utils.test.js`. (There's already an explicit test for REST API, which does pass for your PR.)
> 
> you mean something like, I can add that if you guys would like to have it
> 
> ```
>   assert.equal(isAlbEvent(webEventLookALike), false)
>   assert.equal(isGatewayV1Event(webEventLookALike), false)
>   assert.equal(isGatewayV2Event(webEventLookALike), false)
> ```
I think that'd be worth adding, just to be sure that this is behaving the way we'd want. Thank you!

Engerim:
> > > @Engerim Thank you! If you want to test the matching with a specific payload, you could add one to the fixtures in `test/unit/serverless/fixtures.js`, which have explicit matching tests in `utils.test.js`. (There's already an explicit test for REST API, which does pass for your PR.)
> > 
> > 
> > you mean something like, I can add that if you guys would like to have it
> > ```
> >   assert.equal(isAlbEvent(webEventLookALike), false)
> >   assert.equal(isGatewayV1Event(webEventLookALike), false)
> >   assert.equal(isGatewayV2Event(webEventLookALike), false)
> > ```
> 
> I think that'd be worth adding, just to be sure that this is behaving the way we'd want. Thank you!

have added it

jsumners-nr (dismissed):

Engerim:
sorry had a copy and paste issue 🫤

mrickard:
OK, this suggests that the comparison isn't working completely the way we'd expect. We may want to decouple API Gateway v1 HTTP from API Gateway v1 REST, depending on what's going on here.

Engerim:
Now it is correct as ofc the event should be always false, before I had the wrong assertion because of copy and paste

mrickard (approved):

jsumners-nr (approved):

PR #2851, opened 2024-12-23
chore: Updated eslint configuration

### LINKED ISSUE #2852: Update eslint configuration
Our eslint configuration has not kept up with the changes in the eslint ecosystem. We should keep this up-to-date so that our code base is able to keep linting errors to a minimum during updates.

jsumners-nr:
This PR resolve #2852.

This PR updates the eslint configuration, and plugins, to the latest. I discovered that our shared configuration has not kept up-to-date on the packages that it uses. In particular:

1. `eslint` released v9 that radically changes (and simplifies) how configuration is written and parsed.
2. `eslint-plugin-node` has been out-of-maintenance for a long time, replaced by `eslint-plugin-n`

The result was a difficult to update shared configuration package. Now that this PR is ready, I _could_ go back and update the shared package to eslint@9 standards, but I'm not convinced the shared package is worth the maintenance cost. Most of the configuration is specific to the repository. I am quite certain that most of the packages we maintain can get by with the simple `neostandard` configuration.

Having been out-of-date for so long, our code base has developed many "errors" in regard to linting. The sonarjs plugin saw a lot of additions that now trigger errors. So there are _many_ files touched in this PR. I took the approach of overriding the majority of rules for the tests, but using inline comments in the actual library code. I think our tests can be more lenient, and that we can investigate on a case-by-case basis all of the manual overrides.

I think we should wait for the `next` branch to be merged to `main`, at which point I will rebase and update this branch, before we merge. But I'm open to merging this first if the team thinks it will not cause too much strife in keeping `next` synchronized.


test/unit/spans/map-to-streaming-type.test.js
```diff
@@ -46,7 +46,7 @@ test('should correctly convert bools when false', async () => {
 })
 
 test('should correctly convert integers', async () => {
-  const intValue = 9999999999999999
+  const intValue = 999999999999999
```
jsumners-nr, line 49: This one generated an error about the value overflowing. I don't think we were trying to test against an overflowed value, so I just shortened it to fit.

eslint-plugin-newrelic-header.js
jsumners-nr, line 1: The `eslint-plugin-header` module has not been maintained in several years (last real commit in 2021 wasn't even a code commit). I found it simpler to add our own extension that covers our use cases.

test/unit/api/api-instrument-loaded-module.test.js
```diff
@@ -97,9 +97,8 @@ test('Agent API - instrumentLoadedModule', async (t) => {
     const EMPTY_MODULE = {}
     let mod = EMPTY_MODULE
     try {
-      // eslint-disable-next-line node/no-missing-require
```
bizob2828, line 100: i know this switched to `eslint-plugin-n` but this still should fail, not sure what's going on here

eslint.config.js
```diff
@@ -0,0 +1,184 @@
...
+    // There are times we'd rather have import statements close to the
+    // thing that needed them:
+    'import-x/first': 'off',
+
+    // Prefer single quotes, but utilize others to avoid escaping:
+    '@stylistic/quotes': ['error', 'single', { avoidEscape: true }],
+
+    // These rules would be too disruptive for an initial migration to
+    // neostandard:
+    '@stylistic/space-before-function-paren': 'off'
```
bizob2828, line 53: are we going to cut a ticket to address this separately?

eslint.config.js
```diff
@@ -0,0 +1,184 @@
...
+  },
+
+  linterOptions: {
+    reportUnusedDisableDirectives: 'error'
```
bizob2828, line 57: very nice

bizob2828 (changes requested):
this is a very large PR, i went through most. you removed all the ignoring of the no console and it won't show this as a warning because lint is quiet but not sure if we want to restore those or not.  i also noticed now that lint is quiet there are a few tests that are using `t.diagnostic` and they should prob be removed

test/unit/api/api-instrument-loaded-module.test.js
jsumners-nr: The current configuration applies the `eslint-plugin-node` "recommended" rules. The configuration in this PR omits them. I experimented with adding them back in yesterday afternoon, but it flagged a whole bunch of things as not supported prior to Node 18/20. It was doing so because it thinks we are trying to target `>=16` (the finally fallback as described in https://github.com/eslint-community/eslint-plugin-n?tab=readme-ov-file#configured-nodejs-version-range).

I have implemented some tweaking of the rules for this code base and included it in my latest commit.

jsumners-nr:
> you removed all the ignoring of the no console and it won't show this as a warning because lint is quiet but not sure if we want to restore those or not.

Supposedly it was set as "warn" in the original configuration https://github.com/newrelic/eslint-config-newrelic/blob/c5780025fff0c92d6d9012e6132e84fb42ad2459/index.js#L54C1-L54C26

> i also noticed now that lint is quiet there are a few tests that are using `t.diagnostic` and they should prob be removed

Can you clarify this for me? How does the quiet flag affect the `t.diagnostic` invocations?

I set the main lint script to quiet because we have been ignoring warnings anyway. Do we want the noise to return and just keep overlooking it?

test/unit/api/api-instrument-loaded-module.test.js
bizob2828: OIC, it doesn't define the `no-missing-require`: https://github.com/neostandard/neostandard/blob/main/lib/configs/base.js#L160

lib/instrumentation/core/http.js
```diff
@@ -74,6 +74,7 @@ function wrapEmitWithTransaction(agent, emit, isHTTPS) {
     tracer.bindEmitter(response, segment)
 
     // the error tracer needs a URL for tracing, even though naming overwrites
+    // eslint-disable-next-line n/no-deprecated-api
```
bizob2828, line 77: this is fine but we had this in our rules before to omit this

```sh
    'node/no-deprecated-api': [
      'error',
      {
        ignoreModuleItems: [
          'url.parse'
        ]
      }
    ],
```

lib/custom-events/custom-event-aggregator.js
```diff
@@ -12,6 +12,7 @@ class CustomEventAggregator extends EventAggregator {
   constructor(opts, agent) {
     opts = opts || {}
     opts.method = opts.method || 'custom_event_data'
+    // eslint-disable-next-line no-self-assign
     opts.metricNames = opts.metricNames
```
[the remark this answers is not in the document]
bizob2828: yea it just looks like it's assigning itself

lib/shim/promise-shim.js
```diff
@@ -419,6 +421,7 @@ function wrapHandler({ handler, index, argsLength, useAllParams, ctx, shim }) {
       ret = shim.applySegment(handler, promSegment, true, this, arguments)
     } finally {
       if (ret && typeof ret.then === 'function') {
+        // eslint-disable-next-line sonarjs/no-dead-store
```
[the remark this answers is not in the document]
bizob2828: yea i don't get this finally block

lib/shim/message-shim/subscribe-consume.js
```diff
@@ -225,6 +225,7 @@ function createConsumerWrapper({ shim, spec, consumer }) {
     } finally {
       if (shim.isPromise(ret)) {
         shim.logger.trace('Got a promise, attaching tx %s ending to promise', tx.id)
+        // eslint-disable-next-line sonarjs/no-dead-store
```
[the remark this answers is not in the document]
bizob2828: i don't really get the reason for the try/finally

bizob2828:
> > you removed all the ignoring of the no console and it won't show this as a warning because lint is quiet but not sure if we want to restore those or not.
> 
> Supposedly it was set as "warn" in the original configuration https://github.com/newrelic/eslint-config-newrelic/blob/c5780025fff0c92d6d9012e6132e84fb42ad2459/index.js#L54C1-L54C26
> 
> > i also noticed now that lint is quiet there are a few tests that are using `t.diagnostic` and they should prob be removed
> 
> Can you clarify this for me? How does the quiet flag affect the `t.diagnostic` invocations?
> 
> I set the main lint script to quiet because we have been ignoring warnings anyway. Do we want the noise to return and just keep overlooking it?

We can address the `t.diagnostic` in a different PR. but i'm referring to the noise in the PRs now:

screenshot 2025-01-03 at 11 27 57 AM

lib/instrumentation/core/http.js
jsumners-nr: I missed that.

jsumners-nr:
> but i'm referring to the noise in the PRs now:

Ah, okay. Those have been there since we migrated to `node:test`, but they were hidden by all of the eslint warnings.

bizob2828:
> > but i'm referring to the noise in the PRs now:
> 
> Ah, okay. Those have been there since we migrated to `node:test`, but they were hidden by all of the eslint warnings.

Yep, fixed in #2858

bizob2828 (approved):
LGTM, we can always tweak this as we go

lib/shim/message-shim/subscribe-consume.js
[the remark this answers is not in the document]
jsumners-nr: If only we could ask the author of the code https://github.com/newrelic/node-newrelic/commit/3d1caaf7d7f7de1c2e139eb7943f557b896fc1e2#diff-c3b4dadcc3b24d7d31b5192193cdb89f33e7c2e9cb2f4b9ddbf7e50067c9240aR192

🤣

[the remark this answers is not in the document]
bizob2828: nah, i just moved this

bizob2828 (dismissed):

bizob2828 (approved):

PR #2887, opened 2025-01-16
chore: Replaced backtracking regex with new algorithm

### LINKED ISSUE #2860: Improve header renaming regular expression
https://github.com/newrelic/node-newrelic/blob/d2fba9da8b9ba28d94e50c24fa494a35c442f0a5/lib/header-attributes.js#L97-L101

This regular expression is triggering sonar-js's "slow regex" rule. We need to investigate what the regular expression does, and see if we can improve it.

jsumners-nr:
This PR resolves #2860. If we were to insist of using a simple regular expression to match the characters we want to strip, that regular expression would _always_ include backtracking. This is because we want to replace _all_ occurrences of the offending characters. The only solution I can determine that will satisfy Sonar is to split and rejoin.

### DIFF
--- a/lib/header-attributes.js
+++ b/lib/header-attributes.js
@@ -95,10 +95,10 @@ function _headerToCamelCase(header) {
   const newHeader = header.charAt(0).toLowerCase() + header.slice(1)
 
   // Converts headers in the form 'header-name' to be in the form 'headerName'
-  // eslint-disable-next-line sonarjs/slow-regex
-  return newHeader.replace(/[\W_]+(\w)/g, function capitalize(m, $1) {
-    return $1.toUpperCase()
-  })
+  return newHeader.split(/[\W_]/).map((ele, i) => {
+    if (i === 0) return ele
+    return ele.slice(0, 1).toUpperCase() + ele.slice(1)
+  }).join('')
 }
 
 function _collectHeaders(headers, nameMap, prefix, transaction) {
--- a/test/unit/header-attributes.test.js
+++ b/test/unit/header-attributes.test.js
@@ -77,6 +77,23 @@ test('#collectRequestHeaders', async (t) => {
     })
   })
 
+  await t.test('should replace repeating non-word characters', (t, end) => {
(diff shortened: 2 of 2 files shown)


jsumners-nr:
I wrote the below benchmark to evaluate the two algorithms. In my tests, the split method is consistently faster than the regex replace method. Given that, I am going to merge this PR.

```
/tmp/16 took 8m 15.4s
❯ node index.js
with regex duration: 2797038958
with split duration: 1204851250
regex is faster: false

/tmp/16 took 4.1s
❯ node index.js
with regex duration: 2838906166
with split duration: 1201614500
regex is faster: false

/tmp/16 took 4.1s
❯ node index.js
with regex duration: 2735240000
with split duration: 1210803583
regex is faster: false
```

```js
'use strict'

const iterations = 1_000_000
const longHeaderKey = 'a-b-c-d-e-f-g-h-i-j-k-l-m-n-o-p-q-r-s-t-u-v-w-x-y-z'
const expected = 'aBCDEFGHIJKLMNOPQRSTUVWXYZ'

const aStart = process.hrtime.bigint()
  const found = replaceWithRegex(longHeaderKey)
  assert.equal(found, expected)
}
const aEnd = process.hrtime.bigint()

const bStart = process.hrtime.bigint()
  const found = replaceWithSplit(longHeaderKey)
  assert.equal(found, expected)
}
const bEnd = process.hrtime.bigint()

const regexDuration = aEnd - aStart
const splitDuration = bEnd - bStart
console.log(`with regex duration: ${regexDuration}`)
console.log(`with split duration: ${splitDuration}`)
console.log(`regex is faster: ${regexDuration < splitDuration}`)

function replaceWithRegex(input) {
  return input.replace(/[\W_]+(\w)/g, (m, $1) => $1.toUpperCase())
}

function replaceWithSplit(input) {
  return input.split(/[\W_]/).map((ele, i) => {
    if (i === 0) return ele
    return ele.slice(0,1).toUpperCase() + ele.slice(1)
  }).join('')
}
```

PR #2911, opened 2025-01-29
chore: Added regex note to sql obfuscator

### LINKED ISSUE #2861: Improve sql obfuscation regular expressions
https://github.com/newrelic/node-newrelic/blob/d2fba9da8b9ba28d94e50c24fa494a35c442f0a5/lib/util/sql/obfuscate.js#L10-L17

https://github.com/newrelic/node-newrelic/blob/d2fba9da8b9ba28d94e50c24fa494a35c442f0a5/lib/db/statement-matcher.js#L8-L10

These regular expressions are triggering sonar-js's "slow regex" rule. We need to investigate what the regular expressions do, and see if we can improve them.

jsumners-nr:
This PR resolves #2861.

### DIFF
--- a/lib/util/sql/obfuscate.js
+++ b/lib/util/sql/obfuscate.js
@@ -7,13 +7,30 @@
 
 module.exports = obfuscate
 
-// eslint-disable-next-line sonarjs/slow-regex
+// All eslint rules in this file that have a comment description of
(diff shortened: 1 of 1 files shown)


bizob2828 (approved):
LGTM

PR #2964, opened 2025-02-26
chore: Refactored otel attribute reconciling

### LINKED ISSUE #2961: Refactor span attribute reconciling
We duplicated a lot of code while devising how to reconcile span attributes in https://github.com/newrelic/node-newrelic/blob/258ad7de7751df59cf6d68334e342f751bb3a934/lib/otel/span-processor.js. We would like to explorer refactoring this repetitive code into something that utilizes a rule set. It is possible that we just leave the code as it is if the rule set is too complicated.

jsumners-nr:
This PR resolves #2961.


bizob2828 (approved):
very nice. LGTM

PR #2981, opened 2025-03-04
feat: Added support for response streaming Lambda functions

### LINKED ISSUE #2901: Response Streaming Lambda support
### Description
AWS introduced response streaming for Node Lambda functions in 2023. Our agent's Lambda instrumentation throws errors when customers wrap these, because the argument order (event, responseStream, context) is different from the traditional functions we instrument (event, context, callback). 

The `.streamifyResponse` method is defined for the Lambda execution environment, on the global `awslambda` object. When invoked, it returns the original handler function, with an additional symbol added as a property. See https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/main/src/UserFunction.js#L204-L213 ; the symbol is `aws.lambda.runtime.handler.streaming`.

We can use the presence of this property to detect automatically if the handler is intended to stream a response, and wrap it accordingly, with no additional input needed from the customer. 

### Acceptance Criteria
This spike should either add response streaming support, or at least determine the scale of any blockers.

mrickard:
## Description

AWS introduced response streaming for Node Lambda functions in 2023, allowing Lambda functions to respond asynchronously from other resources (like LLMs). The argument order of streamified Lambda handlers `(event, responseStream, context)` is different from the traditional arguments `(event, context, callback).

Support required a new code path in `lib/serverless/aws-lambda.js` for detecting streamified functions, and wrapping the new responseStream argument. 

## How to Test

A new unit test customized for response streaming is in `test/unit/serverless/aws-lambda-stream-response.test.js`. Tests are mostly equivalent to the `aws-lambda.test.js` tests, though tests and handlers have been updated to be async. Test handlers now have the correct arguments, and they write to a responseStream for testing. Callback-based tests have been removed, as streamified Lambda handlers take no callback argument.

## Related Issues

Closes #2901 
Closes NR-362185


test/unit/serverless/aws-lambda-stream-response.test.js
```diff
@@ -0,0 +1,1882 @@
...
+        plan.equal(transaction.getFullName(), expectedBgTransactionName)
+        plan.equal(transaction.isActive(), true)
+        responseStream.end()
+        return validResponse
+      })
+
+      const wrappedHandler = awsLambda.patchLambdaHandler(handler)
+      await wrappedHandler(nonApiGatewayProxyEvent, responseStream, context)
+
+      await plan.completed
```
jsumners-nr, line 246: 🤦‍♂️ the missing `.completed` was all me. I think I'll see if I can alter `tspl` to support both `await plan` and `await plan.completed`.

lib/serverless/aws-lambda.js
```diff
@@ -114,6 +114,9 @@ class AwsLambda {
       return handler
     }
 
+    const isStreamHandler = handler[Symbol.for('aws.lambda.runtime.handler.streaming')] === 'response'
+    // TODO: add supportability metric for streaming?
```
bizob2828, line 118: are you going to create a metric for this?

bizob2828 (commented):
I fixed tests and we seem to have parity between streaming/non-streaming. Have you tested this PR with a real lambda using response streaming?

lib/serverless/aws-lambda.js
```diff
@@ -122,13 +126,26 @@ class AwsLambda {
...
+    for (const symbol of Object.getOwnPropertySymbols(handler)) {
+      logger.trace(`Setting symbol ${symbol.toString()} on handler`)
+      nrHandler[symbol] = handler[symbol]
+    }
+
+    return nrHandler
 
     function wrappedHandler() {
       const args = shim.argsToArray.apply(shim, arguments)
+      logger.trace('handler args', args)
```
bizob2828, line 143: do we want this to ship?

lib/serverless/aws-lambda.js
```diff
@@ -122,13 +126,26 @@ class AwsLambda {
...
 
       const event = args[0]
-      const context = args[1]
+      logger.trace('Is this a response streaming Lambda function?', isStreamHandler)
```
bizob2828, line 146: do we want this to ship?

lib/serverless/aws-lambda.js
```diff
@@ -122,13 +126,26 @@ class AwsLambda {
...
+      const context = isStreamHandler === true ? args[2] : args[1]
+      logger.trace('Context', context)
```
bizob2828, line 148: do we want this to ship?

bizob2828 (changes requested):
just a few questions around the logging we're doing and if we want to keep it

lib/serverless/aws-lambda.js
mrickard: I'm adding one in the next commit, unless we don't want to record this at all.

lib/serverless/aws-lambda.js
mrickard: Removing this.

lib/serverless/aws-lambda.js
mrickard: This could be useful as we roll out this feature. We're doing essentially the same thing for API Gateway, so it fits the pattern of trace logging for this module.

lib/serverless/aws-lambda.js
mrickard: We found it useful in debugging, but I'm not 100% in favor of logging out this context. I could see the value in logging out the context type

lib/serverless/aws-lambda.js
```diff
@@ -114,6 +115,9 @@ class AwsLambda {
       return handler
     }
 
+    const isStreamHandler = handler[HANDLER_STREAMING] === 'response'
+    this.agent.recordSupportability('Nodejs/Serverless/Lambda/ResponseStreaming')
```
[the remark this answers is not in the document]
mrickard: Ah, right--fixed.

test/unit/serverless/aws-lambda-stream-response.test.js
```diff
@@ -0,0 +1,1932 @@
...
+ */
+class HttpResponseStream {
+  static from(originalStream, prelude) {
+    originalStream.setContentType('application/vnd.awslambda.http-integration-response')
+    const streamMeta = JSON.stringify(prelude)
+    originalStream._onBeforeFirstWrite = (write) => {
+      // If we finish writing all of the required unit tests, and this assert
+      // never gets triggered, then there is no reason to have this
+      // `HttpResponseStream` thing.
+      assert.fail('_onBeforeFirstWrite')
```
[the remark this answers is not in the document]
jsumners-nr: If the `assert.fail` line were ever hit by these tests, then an uncaught error would be thrown. Therefore, this code path is never utilized.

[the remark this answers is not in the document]
mrickard: That seems like an argument to delete this class declaration and remove it from the handlers. There don't seem to be any adverse effects when I remove it, so I'll take them all out.

lib/serverless/aws-lambda.js
```diff
@@ -251,6 +278,17 @@ class AwsLambda {
     }
   }
 
+  wrapStreamAndCaptureError(transaction, txnEnder, stream) {
+    const shim = this.shim
+    stream.on('error', (error) => {
+      shim.agent.errors.add(transaction, error)
+    })
+    stream.on('end', () => {
```
bizob2828, line 286: i also think this event should be `finish` based on the docs. 

```
Event: 'finish'[#](https://nodejs.org/api/stream.html#event-finish)
Added in: v0.9.4
The 'finish' event is emitted after the [stream.end()](https://nodejs.org/api/stream.html#writableendchunk-encoding-callback) method has been called, and all data has been flushed to the underlying system.
```

test/unit/serverless/aws-lambda-stream-response.test.js
```diff
@@ -0,0 +1,1855 @@
...
+    }
+  })
+
+  stream.setContentType = function () {}
+
+  stream.on('error', (error) => {
+    if (responseDoneReject) responseDoneReject(error)
+  })
+
+  stream.on('end', () => {
```
bizob2828, line 117: this should be `finish`

test/unit/serverless/aws-lambda-stream-response.test.js
```diff
@@ -0,0 +1,1855 @@
...
+      plan.equal(message, errorMessage)
+      plan.equal(type, 'SyntaxError')
+    })
+
+    let transaction
+    const handler = decorateHandler(async (event, responseStream) => {
+      transaction = agent.tracer.getTransaction()
+      const chunks = ['step 1', 'step 2', 'step 3']
+      await writeToResponseStream(chunks, responseStream, 100)
+      responseStream.end()
```
bizob2828, line 1705: if you update based on my suggestions, ending the stream halts this test so you can't end the stream and then throw a promise rejection

test/unit/serverless/aws-lambda-stream-response.test.js
```diff
@@ -0,0 +1,1855 @@
...
+test('should record standard background metrics', async (t) => {
+  const plan = tspl(t, { plan: 9 })
+  const { agent, awsLambda, event, responseStream, context } = t.nr
+  agent.on('harvestStarted', confirmMetrics)
+
+  const handler = decorateHandler(async (event, responseStream) => {
+    const chunks = ['step 1', 'step 2', 'step 3']
+    await writeToResponseStream(chunks, responseStream, 100)
+    responseStream.end()
+    return validResponse
```
bizob2828, line 1825: I'm not sure this matters but we're not checking that the response stream returns the right data.  You could add `responseDone` to the test context and after you end the stream 

````
const data = await responseDone
plan.equal(data, 'step1step2step3')
```

bizob2828 (changes requested):
We're not handling the `finish` event of stream, just `end` which doesn't exist. The response stream tests are also ending the stream in the case of errors which we shouldn't be ending the streaming in those cases. Lastly, we don't have a test that emits an error over the stream so the testing of the instrumented `error` event on stream isn't getting hit.

lib/serverless/aws-lambda.js
```diff
@@ -114,6 +115,11 @@ class AwsLambda {
       return handler
     }
 
+    const isStreamHandler = handler[HANDLER_STREAMING] === 'response'
+    if (isStreamHandler) {
+      this.agent.recordSupportability('Nodejs/Serverless/Lambda/ResponseStreaming')
```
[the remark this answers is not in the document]
mrickard: Looks like the agent helper in tests stubs out the recording of supportability metrics. I'm open to finding a way to test this here.

lib/serverless/aws-lambda.js
```diff
@@ -122,13 +128,25 @@ class AwsLambda {
...
+    return nrHandler
 
     function wrappedHandler() {
       const args = shim.argsToArray.apply(shim, arguments)
 
       const event = args[0]
-      const context = args[1]
+      logger.trace({ isStreamHandler }, 'Lambda handler type')
+      const context = isStreamHandler === true ? args[2] : args[1]
+      logger.trace('Handler context type', typeof context)
```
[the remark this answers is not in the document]
mrickard: If context is a function at this point, that's a sign that the agent detected the handler as response streaming, but the runtime environment parses the handler as non-streaming. I could make that more explicit, or just dispense with this log line.

lib/serverless/aws-lambda.js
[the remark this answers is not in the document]
bizob2828: It stubs it out but you can spy on it and assert it's getting called with proper arg and not called when not streaming

lib/serverless/aws-lambda.js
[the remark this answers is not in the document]
bizob2828: Maybe log out the `context?.functionName`. That was the root of the original issue

lib/serverless/aws-lambda.js
mrickard: Toggling to `finish` works, and reveals four test failures: 
* should record error event when func is async an UnhandledPromiseRejection is thrown
* should record error event when error is thrown
* should not end transactions twice
* should record standard background metrics

bizob2828: That's because they're ending the streaming before an error exists. If you remove ending the stream it works.

test/unit/serverless/aws-lambda-stream-response.test.js
mrickard: That makes sense to check against what the final stream value is, though with this usage I'm seeing responseDone as undefined.

bizob2828 (approved):

mrickard (commented):
👍

bizob2828:
This does not handle streaming from API gateway.  Based on their [announcment](https://aws.amazon.com/blogs/compute/introducing-aws-lambda-response-streaming/) it is not supported:

> You can not use [Amazon API Gateway](https://aws.amazon.com/api-gateway/) and [Application Load Balancer](https://aws.amazon.com/elasticloadbalancing/application-load-balancer/) to progressively stream response payloads, but you can use the functionality to return larger payloads with API Gateway.

If they add support later, we can log a separate story to intercept the stream and handle the transaction as web and assign the appropriate status code and response headers

PR #2987, opened 2025-03-11
chore: Clarified system metrics sampler naming

jsumners-nr:
While researching what it might take to support honoring the w3c sampling flag, I discovered that we have multiple "samplers." In order to reduce confusion when working with one sampler or the other, this PR renames the system metrics sampler from just "sampler" to a very clear "systemMetricsSampler."


bizob2828 (commented):
should you rename the class as well? Right now it's `Sampler`

PR #3017, opened 2025-03-31
feat: Azure Function utilization

### LINKED ISSUE #2996: Add utilization for Azure Functions
### Description
The Node agent detects if it's running in an Azure environment. The latest version of our Utilization spec adds a `vendor` key for `azurefunction`, consisting of `faas.app_name` and `cloud.region`. See the Utilization spec (agent-specs/Utilization.md in datanerd) for the structure of `faas.app_name`.

### Acceptance Criteria
`/lib/utilization/azure-info.js` should be updated to detect if the current application is an Azure Function (see the Azure-Function spec for details on the environment; we're targeting the v4 programming model). This should be verified with a unit test.

amychisholm03:
This PR resolves #2996.


lib/utilization/azurefunction-info.js
```diff
@@ -0,0 +1,46 @@
...
+  if (!agent.config.utilization || !agent.config.utilization.detect_azurefunction) {
+    return setImmediate(callback, null, null)
+  }
+
+  if (results) {
+    return setImmediate(callback, null, results)
+  }
+
+  // Detect if Azure Function.
+  // TODO: How do we access Azure Function invocation context?
```
mrickard, line 27: The agent would be attached to the Azure Function app code in order to execute, so we wouldn't have to worry about the environment except to simulate it during testing. To get the utilization info, these environment variables would be made available as properties of `process.env`.

lib/utilization/azurefunction-info.js
```diff
@@ -0,0 +1,46 @@
...
+  if (process.env.REGION_NAME && process.env.WEBSITE_OWNER_NAME && process.env.WEBSITE_SITE_NAME) {
+    // Derive vendor metadata from environment variables.
+    const subscriptionId = process.env.WEBSITE_OWNER_NAME.split('+')[0]
+    // eslint-disable-next-line no-useless-escape
+    const resourceGroupName = process.env.WEBSITE_RESOURCE_GROUP ?? process.env.WEBSITE_OWNER_NAME.match(/\+([a-zA-Z0-9\-]+)-[a-zA-Z0-9]+(?:-Linux)?/)?.[0]?.slice(1)
```
mrickard, line 32: You can avoid the useless-escape warning by changing the regex to `([a-zA-Z0-9-]+)-[a-zA-Z0-9]+(?:-Linux)?`

jsumners-nr, line 32: Let's skip all this regex silliness. I don't know why the spec author decided to encumber us with that when all we really need to do, based upon the information included in the spec, is:

```js
const input = 'b999997b-cb91-49e0-b922-c9188372bdba+testing-rg-EastUS2webspace-Linux'
const groupName = input.split('+').pop().split('-Linux').shift()
console.log(groupName) // testing-rg-EastUS2webspace
```

lib/utilization/azurefunction-info.js
jsumners-nr, line 27: As I am reading https://source.datanerd.us/agents/agent-specs/blob/main/Utilization.md#azure-functions, we do not need access to the invocation context. The app name should be present in the `WEBSITE_SITE_NAME` environment variable. There's really no way for us to satisfy the utilization spec for this platform if we have to rely on the invocation context. Utilization reporting happens during agent bootstrapping, long before any application code is invoked.

I think the spec should be updated to remove the reference to "the invocation context."

lib/utilization/azurefunction-info.js
mrickard: I'd second this. The more complicated the pattern is, the more brittle it's going to be.

lib/utilization/azurefunction-info.js
```diff
@@ -0,0 +1,46 @@
...
+
+  if (results) {
+    return setImmediate(callback, null, results)
+  }
+
+  // Detect if Azure Function.
+  if (process.env.REGION_NAME && process.env.WEBSITE_OWNER_NAME && process.env.WEBSITE_SITE_NAME) {
+    // Derive vendor metadata from environment variables.
+    const subscriptionId = process.env.WEBSITE_OWNER_NAME.split('+')[0]
+    const resourceGroupName = process.env.WEBSITE_RESOURCE_GROUP ?? process.env.WEBSITE_OWNER_NAME.split('+').pop().split('-Linux').shift()
```
jsumners-nr, line 30:
```suggestion
    const resourceGroupName = process.env.WEBSITE_RESOURCE_GROUP ?? process.env.WEBSITE_OWNER_NAME?.split('+').pop().split('-Linux').shift() ?? 'unknown'
```

lib/utilization/azurefunction-info.js
```diff
@@ -0,0 +1,46 @@
...
+    const azureFunctionAppName = process.env.WEBSITE_SITE_NAME
+    if (subscriptionId && resourceGroupName) {
+      results = {
+        'faas.app_name': `/subscriptions/${subscriptionId}/resourceGroups/${resourceGroupName}/providers/Microsoft.Web/sites/${azureFunctionAppName}`,
+        'cloud.region': process.env.REGION_NAME
+      }
+    }
+  } else {
+    logger.debug('Azure Function metadata was invalid.')
+    // TODO: Create NAMES.UTILIZATION.AZUREFUNCTION_ERROR ?
```
jsumners-nr, line 40: Do we need this TODO?

lib/utilization/azurefunction-info.js
```diff
@@ -0,0 +1,46 @@
...
+  if (!agent.config.utilization || !agent.config.utilization.detect_azurefunction) {
+    return setImmediate(callback, null, null)
+  }
+
+  if (results) {
+    return setImmediate(callback, null, results)
+  }
+
+  // Detect if Azure Function.
+  if (process.env.REGION_NAME && process.env.WEBSITE_OWNER_NAME && process.env.WEBSITE_SITE_NAME) {
```
jsumners-nr, line 27: We can make this code a bit easier to read if we do:

```js
const { REGION_NAME, WEBSITE_OWNER_NAME, WEBSITE_SITE_NAME } = process.env
if (!REGION_NAME || !WEBSITE_OWNER_NAME || !WEBSITE_SITE_NAME) {
	logger.debug('boop boop')
	return setImmediate(callback, null, null)
}
```

lib/utilization/azurefunction-info.js
```diff
@@ -0,0 +1,46 @@
...
+      }
+    }
+  } else {
+    logger.debug('Azure Function metadata was invalid.')
+    // TODO: Create NAMES.UTILIZATION.AZUREFUNCTION_ERROR ?
+    agent.metrics.getOrCreateMetric(NAMES.UTILIZATION.AZURE_ERROR).incrementCallCount()
+  }
+
+  // Call back!
+  callback(null, results)
```
jsumners-nr, line 45:
```suggestion
  callback(null, results)
```

jsumners-nr (commented):
Looking good. I have a few suggestions.

lib/utilization/azurefunction-info.js
jsumners-nr: I think it's okay to do so. Otherwise we have to PR an addition to the supportability service to recognize a new string.

lib/utilization/azurefunction-info.js
amychisholm03: I was thinking that if `resourceGroupName` is still undefined after the two possibilities, then it is likely not an Azure Function. That's why I added the `if (subscriptionId && resourceGroupName)` check before setting `result`. Because isn't it possible that another service could have the same env vars set but isn't an Azure Function?

mrickard: @amychisholm03 As far as I can tell, in Azure *everything* needs to belong to a resource group, so is going to have a resource group name. `process.env. FUNCTIONS_WORKER_RUNTIME` is defined for Functions, so if that's defined, this is definitely an Azure Function. ([docs](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=javascript%2Cwindows%2Cazure-cli&pivots=nodejs-model-v4#environment-variables))

test/integration/utilization/azurefunction-info.test.js
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright 2025 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const test = require('node:test')
+const assert = require('assert')
```
jsumners-nr, line 9: For consistency:

```suggestion
const assert = require('node:assert')
```

test/integration/utilization/azurefunction-info.test.js
```diff
@@ -0,0 +1,84 @@
...
+const helper = require('../../lib/agent_helper')
+const fetchAzureFunctionInfo = require('../../../lib/utilization/azurefunction-info')
```
jsumners-nr, line 11:
```suggestion
const helper = require('#testlib/agent_helper.js')
const fetchAzureFunctionInfo = require('#agentlib/utilization/azurefunction-info.js')
```

test/integration/utilization/azurefunction-info.test.js
```diff
@@ -0,0 +1,84 @@
...
+
+test('should return null if required environment variables are missing', (ctx, end) => {
+  const agent = ctx.nr.agent
+
+  fetchAzureFunctionInfo(agent, (err, result) => {
+    assert.strictEqual(err, null)
+    assert.strictEqual(result, null)
+    end()
+  })
+})
```
jsumners-nr, line 49: This test should include and assertion that the error metric was incremented correctly.

jsumners-nr (changes requested):

jsumners-nr (approved):
Looks good to me.

PR #3021, opened 2025-04-02
feat: Support Azure functions HTTP trigger type

### LINKED ISSUE #2995: Instrumentation of @azure/functions library http methods
### Description

Supporting Azure functions requires instrumenting the [@azure/functions Node.js library](https://github.com/Azure/azure-functions-nodejs-library). Azure Functions applications rely on this library for access to the Functions runtime environment. Function applications also rely on it for code and file structure in place of other application frameworks. We should support the current V4 version of this library.

Azure Functions applications can be triggered by different events, which our spec abstracts into these categories: 
* timer
* datasource
* http
* pubsub
* other 

For the purposes of this ticket, we'll focus on http methods, which will be instrumented specifically as web transactions. (All methods are exposed in `/src/app.ts`.)

A typical function will import `app` from @azure/functions, and invoke one of its trigger methods: 
```
const { app } = require('@azure/functions');

app.http('HttpExample', {
    methods: ['GET'],
    authLevel: 'anonymous',
    handler: async (request, context) => {
        const name = request.query.get('name') || await request.text() || 'world';
        return { body: JSON.stringify(r.body) };
    }
});
```

We'll want to instrument the handler, request, and aspects of the context. The context argument includes properties we can use for instrumentation: functionName, invocationId, traceContext, and others. 

Context is present with each handler, while the properties of the first argument differ with each trigger type. The HTTP Request argument [is documented here](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=javascript%2Cwindows%2Cazure-cli&pivots=nodejs-model-v4#http-request).

We have the option of instrumenting this library similarly to the way we instrument other libraries, or we can now use the [Node diagnostics channel](https://nodejs.org/api/diagnostics_channel.html). (We considered leveraging the otel instrumentation for Azure Functions via the otel bridge, but only emits logs and doesn't produce spans.) Based on our schedule, this new instrumentation will follow the historical pattern we've followed for instrumenting. 

#### Function Invocation Lifecycle
Function apps offer app-level invocation hooks `preInvocation` and `postInvocation`.  These have a `context` argument, so developers can filter pre/post behavior based on any of the properties included in the context object. [These are documented here. ](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=javascript%2Cwindows%2Cazure-cli&pivots=nodejs-model-v4#invocation-hooks)

More generally, the `@azure/functions`-supplied `app` object offers hooks for `appStart` and `appTerminate`. [These are documented here](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=javascript%2Cwindows%2Cazure-cli&pivots=nodejs-model-v4#app-hooks).

### What's not included: 
* http trigger functions for which HTTP streaming is enabled
* background transactions
* easy install of the agent
* V3 @azure/functions syntax
* ECMAScript support. Functions supports ES Modules, but this is officially "in preview." ES Module operation is enabled by ensuring a filename has the `.mjs` extension. 

### Acceptance Criteria
Requests made to @azure/functions http methods should result in complete transactions and traces in New Relic. Transactions and traces should have the appropriate names. Traces and spans should have the attributes described in the agent spec for Azure Functions. There should be a test suite added to verify instrumentation.

jsumners-nr:
This PR resolves #2995.


mrickard:
LGTM! Did we want to address the case of handling `http` requests with the `generic` API? I'm not sure how common that is, but it's documented: https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=javascript%2Cwindows%2Cazure-cli&pivots=nodejs-model-v4#generic-inputs-and-outputs

jsumners-nr:
> LGTM! Did we want to address the case of handling `http` requests with the `generic` API? I'm not sure how common that is, but it's documented: [learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=javascript%2Cwindows%2Cazure-cli&pivots=nodejs-model-v4#generic-inputs-and-outputs](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=javascript%2Cwindows%2Cazure-cli&pivots=nodejs-model-v4#generic-inputs-and-outputs)

I think we can address that later. It's an edge case, and it will require enough inspection work as to unnecessarily complicate this PR.

mrickard (approved):

PR #3028, opened 2025-04-08
feat: Support Azure functions background trigger types

### LINKED ISSUE #2997: Instrument background transactions for Azure Functions
### Description
Supporting Azure functions requires instrumenting the [@azure/functions Node.js library](https://github.com/Azure/azure-functions-nodejs-library). Azure Functions applications rely on this library for access to the Functions runtime environment. Function applications also rely on it for code and file structure in place of other application frameworks. We should support the current V4 version of this library.

For the purpose of this ticket, we'll focus on background transactions, which are triggered by the following Azure events:
* timer
* storageBlob
* storageQueue
* serviceBusQueue
* serviceBusTopic
* eventHub
* eventGrid
* cosmosDB
* warmup
* sql
* mySql
* webPubSub

These will be grouped into these spec categories: 
* timer
* datasource (cosmosDB, sql, mySql)
* pubsub (storageQueue, serviceBusQueue, serviceBusTopic, webPubSub, eventHub)
* other (storageBlob, eventGrid, warmup)

### Acceptance Criteria
Requests made to these @azure/functions methods should result in complete transactions and traces in New Relic. Transactions and traces should have the appropriate names. Traces and spans should have the attributes described in the agent spec for Azure Functions. There should be a test suite added to verify instrumentation.

jsumners-nr:
This PR resolves #2997.

-----

I believe that our other instrumentations will fill in any other required attributes. For example, if the trigger is a Redis based trigger, the work in the PR should start the transaction and then the Redis instrumentation should add spans that include required datastore attributes.

I wanted to write a test that proves this. But as of right now, the only such trigger I think we could test with is Redis, and the current version of `@azure/functions` doesn't actually support it:

https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-cache-trigger-redispubsub?tabs=isolated-process%2Cnode-v3%2Cpython-v1&pivots=programming-language-javascript
image

I looked at their RabbitMQ docs as well, https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-rabbitmq-trigger?tabs=isolated-process&pivots=programming-language-javascript, but there's no indication there of what an incoming payload looks like, nor any code examples. I think the "background" triggers are not well supported or documented. So we will just have to fix these up as we go.

----

I have left out support for distributed tracing in these methods. It's not clear to me that there would ever be any invocations with such headers, and I can't find any documentation that shows how headers might be delivered to such functions. So I am going with the "MAY" portion of our internal spec:

> Support for distributed tracing for other invocation triggers MAY be added at the discrection of the language agent.

### DIFF
--- a/lib/instrumentation/@azure/functions.js
+++ b/lib/instrumentation/@azure/functions.js
@@ -50,7 +50,20 @@ module.exports = function initialize(agent, azureFunctions, _moduleName, shim, {
   const httpMethods = ['http', 'get', 'put', 'post', 'patch', 'deleteRequest']
   shim.wrap(azureFunctions.app, httpMethods, wrapAzureHttpMethods)
 
-  const backgroundMethods = ['timer']
+  const backgroundMethods = [
+    'cosmosDB',
+    'eventGrid',
+    'eventHub',
+    'mySql',
+    'serviceBusQueue',
+    'serviceBusTopic',
+    'sql',
+    'storageBlob',
+    'storageQueue',
+    'timer',
+    'warmup',
+    'webPubSub'
+  ]
   shim.wrap(azureFunctions.app, backgroundMethods, wrapAzureBackgroundMethods)
 }
 
@@ -165,7 +178,7 @@ function wrapAzureBackgroundMethods(shim, appMethod) {
       tx.setPartialName(`AzureFunction/${context.functionName}`)
 
       const segment = tracer.createSegment({
-        name: 'timer-trigger',
+        name: `${appMethod}-trigger`,
         recorder: backgroundRecorder,
         parent: ctx.segment,
         transaction: tx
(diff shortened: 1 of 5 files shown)


jsumners-nr:
> MS briefly describes the `traceContext` property of the Function handler's `context` argument [in their SDK doc](https://learn.microsoft.com/en-us/javascript/api/@azure/functions/tracecontext?view=azure-node-latest). It does look like standard w3c traceContext, when I inspect on sample requests.

As I noted, it's not clear when nor how such things might be supplied to one of the "background" functions. In fact, from https://learn.microsoft.com/en-us/azure/azure-functions/opentelemetry-howto?tabs=app-insights&pivots=programming-language-javascript:

> At this time, only HTTP trigger and Azure SDK-based triggers are supported with OpenTelemetry outputs.

So their official OTEL support is also likely skipping DT stuff for these specific methods. And their generic tracing docs focus on HTTP handlers for DT stuff https://learn.microsoft.com/en-us/azure/azure-monitor/app/distributed-trace-data.

I am convinced that we do not need to implement DT parsing for these methods.

mrickard (approved):

Issue #3029, opened 2025-04-08
Fix failing prisma versioned test


svetlanabrennan:
CI failing because of prisma issue. 

![Image](https://github.com/user-attachments/assets/da9a3c59-0ee4-4a27-a26b-25f291da099f)

PR #3030, opened 2025-04-08
test: Pinned prisma version below 7

svetlanabrennan:
We're getting a CI failure with prisma 7.0.0 so we need to pin temporarily while we troubleshoot the issue. 

Relates to issue #3029


mrickard:
Closing in favor of #3031

PR #3110, opened 2025-05-20
chore: Added event source type to  txn name for Lambda APM Mode

### LINKED ISSUE #3104: chore: Add transaction naming for Lambda APM
### Description
If Lambda APM mode is enabled, the event type should be inserted into the function name segment of the transaction name. 

### Acceptance criteria

In `lib/serverless/aws-lambda.js`:
* detect the `NEW_RELIC_APM_LAMBDA_MODE` environment variable
* if set, and if the event source is available, reset the transaction name to prepend the event source type (in all uppercase) to the Function Name segment of the transaction name. 

For more information, see [Transactions Traces <-> Serverless Invocations document](https://newrelic.atlassian.net/wiki/spaces/FDN/pages/4270031787/APM+Transactions+Traces+-+Serverless+Invocations). 

### Additional context
https://github.com/newrelic/node-newrelic/issues/3041
https://new-relic.atlassian.net/browse/NR-405918
https://github.com/newrelic/node-newrelic/issues/3090
https://new-relic.atlassian.net/browse/NR-411864

mrickard:
Lambda APM Mode will support Lambda telemetry in the same UI as the one used by APM. To support this unified view, agent teams will need to inject an all-uppercase event source type in the function name segment of the transaction name. 

If `NEW_RELIC_APM_LAMBDA_MODE` is not enabled, or if there is no event source, the transaction name is unchanged. 

## How to Test

One test has been added to each of `aws-lambda.test.js` and `aws-lambda-stream-response.test.js`. In the latter case, the test uses a `cloudwatchScheduled` test event, as response streaming Lambda functions don't support API Gateway. Presumably any response streaming Lambda function using `NEW_RELIC_APM_LAMBDA_MODE` will be triggered by some background event, if any.

## Related Issues

Closes #3104 
Closes NR-416510

### DIFF
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -1231,6 +1231,9 @@ defaultConfig.definition = () => ({
       default: true
     }
   },
+  /**
+   * Specifies whether AWS Lambda telemetry is going to be compatible with the APM UI.
+   */
   lambda: {
     apm_mode: {
       default: false,
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -1231,13 +1231,16 @@ defaultConfig.definition = () => ({
       default: true
     }
   },
+
   /**
-   * Specifies whether AWS Lambda telemetry is going to be compatible with the APM UI.
+   * When `true`, the AWS Lambda instrumentation will add the necessary
+   * data to support the new (as of 2025) unified APM UI.
    */
   apm_lambda_mode: {
     default: false,
     formatter: boolean
   },
+
   /**
    * Specifies whether the agent will be used to monitor serverless functions.
    * For example: AWS Lambda
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -1234,11 +1234,9 @@ defaultConfig.definition = () => ({
   /**
    * Specifies whether AWS Lambda telemetry is going to be compatible with the APM UI.
    */
-  lambda: {
-    apm_mode: {
-      default: false,
-      formatter: boolean
-    }
+  apm_lambda_mode: {
+    default: false,
+    formatter: boolean
   },
   /**
    * Specifies whether the agent will be used to monitor serverless functions.
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -1231,6 +1231,12 @@ defaultConfig.definition = () => ({
       default: true
     }
   },
+  lambda: {
+    apm_mode: {
+      default: false,
+      formatter: boolean
+    }
+  },
   /**
    * Specifies whether the agent will be used to monitor serverless functions.
    * For example: AWS Lambda
--- a/lib/serverless/aws-lambda.js
+++ b/lib/serverless/aws-lambda.js
@@ -125,7 +125,7 @@ class AwsLambda {
     const awsAttributes = this._getAwsAgentAttributes(event, context)
     let trigger = ''
     // Lambda environment variables are passed as strings, so a simple truthiness check could be misleading
-    if (this.agent.config.lambda?.apm_mode === true &&
+    if (this.agent.config.apm_lambda_mode === true &&
         awsAttributes &&
         typeof awsAttributes['aws.lambda.eventSource.eventType'] === 'string') {
       // In Lambda APM mode, we prepend the event type (if present) in upper case to
--- a/lib/serverless/aws-lambda.js
+++ b/lib/serverless/aws-lambda.js
@@ -466,7 +466,7 @@ function lowercaseObjectKeys(original) {
   }, {})
 }
 
-function endTransaction(transaction, enderIndex, eventType) {
+function endTransaction(transaction, enderIndex) {
   if (transactionEnders.length === 0 || transactionEnders[enderIndex] === cleanClosure) {
     // In the case where we have already been called, we return early. There may be a
     // case where this is called more than once, given the lambda is left in a dirty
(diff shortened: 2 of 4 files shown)


test/unit/serverless/aws-lambda-stream-response.test.js
```diff
@@ -268,6 +271,35 @@ test('should create a transaction for handler', async (t) => {
   await plan.completed
 })
 
+test('should name transaction correctly in Lambda APM Mode', async (t) => {
+  const plan = tspl(t, { plan: 2 })
+  const { agent, awsLambda, responseStream, context, responseDone } = t.nr
+  const event = lambdaSampleEvents.cloudwatchScheduled
+  process.env.NEW_RELIC_APM_LAMBDA_MODE = 'true'
```
jsumners-nr, line 278: Note, if we rely on `agent.config` instead of a direct environment variable check, then we'll need to provide that configuration prior to constructing the agent. Which means we would need to refactor how the tests are bootstrapped to something like:

```js
async function beforeEach(ctx) { /* stuff */ }

test('does a thing', async t => {
	process.env.THING = 'value'
	await beforeEach(t)

	// test code
})
```

jsumners-nr (changes requested):

lib/serverless/aws-lambda.js
```diff
@@ -121,7 +121,21 @@ class AwsLambda {
 
   createSegment({ event, context, transaction, recorder }) {
     const shim = this.shim
-    const functionName = context.functionName
+
+    const awsAttributes = this._getAwsAgentAttributes(event, context)
+    let trigger = ''
+    // Lambda environment variables are passed as strings, so a simple truthiness check could be misleading
+    if (process.env.NEW_RELIC_APM_LAMBDA_MODE === 'true' &&
```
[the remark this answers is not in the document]
mrickard: @jsumners-nr Typically in Lambda deployments, customers don't use the config file, as there's much easier access to environment variables in the AWS UI. Some config (like the path to the intended Lambda handler) has to be in env vars, since it configures the Lambda wrapper which loads the agent. While we *could* add this to agent config, most customers expect to use the environment variables only.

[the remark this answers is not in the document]
mrickard: So whether or not the customer defines it that way, at this point we can expect it as `agent.config`. Got it.

mrickard:
Looks like I have to fiddle with the commit signature.

test/unit/serverless/aws-lambda-stream-response.test.js
```diff
@@ -392,3 +401,34 @@ test('should record standard background metrics', async (t) => {
...
   }
 })
+
+test('should name transaction correctly in Lambda APM Mode', async (t) => {
+  helper.unloadAgent(t.nr.agent)
+  await beforeEach(t, lambdaApmConfig)
+  const plan = tspl(t, { plan: 2 })
+  const { agent, awsLambda, responseStream, context, responseDone } = t.nr
+  const event = lambdaSampleEvents.cloudwatchScheduled
+  process.env.NEW_RELIC_APM_LAMBDA_MODE = 'true'
```
jsumners-nr, line 411: Setting this here is not doing anything. You've already defined the value in `lambdaApmConfig`. We can validate that the configuration parses the correct environment variable by:

```js
helper.unloadAgent(t.nr.agent)
process.env.NEW_RELIC_APM_LAMBDA_MODE = 'true'
await beforeEach(t)

// rest of the test
```

test/unit/serverless/aws-lambda.test.js
```diff
@@ -1557,3 +1566,28 @@ test('AwsLambda.patchLambdaHandler', async (t) => {
...
+  const agentAttributes = await transaction.trace.attributes.get(ATTR_DEST.TRANS_EVENT)
+  const trigger = agentAttributes[EVENTSOURCE_TYPE].toUpperCase()
+  const expectedApmTxnName = `WebTransaction/Function/${trigger} ${functionName}`
+
+  assert.equal(agentAttributes[EVENTSOURCE_TYPE], 'apiGateway')
+  assert.ok(transaction)
+  assert.equal(transaction.type, 'web')
+  assert.equal(transaction.getFullName(), expectedApmTxnName)
+
+  await afterTests(t)
```
jsumners-nr, line 1592: This should be registered directly after the `beforeTests` function is invoked by doing:

```js
t.after(async () => { await afterTests(t) })
```

In the current implementation, if the test fails before it reaches the `afterTests` invocation, the clean-up will never happen.

lib/config/default.js
```diff
@@ -1231,6 +1231,13 @@ defaultConfig.definition = () => ({
...
     }
   },
+  /**
+   * Specifies whether AWS Lambda telemetry is going to be compatible with the APM UI.
+   */
+  apm_lambda_mode: {
+    default: false,
+    formatter: boolean
+  },
   /**
```
jsumners-nr, line 1241:
```suggestion
  },

  /**
   * When `true`, the AWS Lambda instrumentation will add the necessary
   * data to support the new (as of 2025) unified APM UI.
   */
  apm_lambda_mode: {
    default: false,
    formatter: boolean
  },

  /**
```

jsumners-nr (approved):
Looks good. One suggestion on the configuration description.

jsumners-nr (approved):

jsumners-nr (approved):

PR #3113, opened 2025-05-21
fix: common.getKeys logic

amychisholm03:
Addendum to https://github.com/newrelic/node-newrelic/pull/3111


lib/utilization/common.js
```diff
@@ -38,8 +38,12 @@ exports.getKeys = function getKeys(data, keys, allValuesMustBeValid = true) {
...
     if (!properties.hasOwn(data, key) || !data[key]) {
       logger.debug('Key %s missing from metadata', key)
-      return null
+      if (allValuesMustBeValid) {
+        return null
+      }
+      continue
     }
+
     let value = data[key]
```
mrickard, line 47: Would it be possible to add tests for these uncovered lines? (I'd think asserting the intended behavior for `allValuesMustBeValid = true` and for `false`.)

PR #3119, opened 2025-05-27
feat: `@google/gen-ai` Gemini AI LLM instrumentation

### LINKED ISSUE #2890: Gemini LLM support - node.js
Have support for gemini LLM, see [CDD]https://newrelic.atlassian.net/wiki/spaces/INST/pages/4027023361/Python+Agent+Gemini+Instrumentation 

(https://newrelic.atlassian.net/wiki/spaces/AIO11y/pages/3227944456/Agents+supports)

Similar to OpenAI, Customers are using gemini LLM heavily and would like new relic to support it as well.

Basically looking for gemini support similar to: https://docs.newrelic.com/docs/ai-monitoring/intro-to-ai-monitoring/

[Feature request here](https://new-relic.atlassian.net/browse/NR-354216)

Python agent added support in - https://docs.newrelic.com/docs/release-notes/agent-release-notes/python-release-notes/python-agent-101100/

amychisholm03:
This PR closes #2890. This instruments the `@google/gen-ai` library which provides an SDK for Google's Gemini AI.

For reviewers, refer to:
* Python agent CDD for Gemini AI instrumentation
* LLM Agent Spec
* Run the example app on [this branch](https://github.com/newrelic/newrelic-node-examples/tree/gemini-ai)
* [`@google/gen-ai` repo](https://github.com/googleapis/js-genai)

Dev TODO:
- [ ] Confirm no relevant response headers can be extracted


lib/instrumentation/@google/genai.js
```diff
@@ -0,0 +1,298 @@
...
+  // initially declaring the variable.
+  TRACKING_METRIC = `${TRACKING_METRIC}/${shim.pkgVersion}`
+
+  const models = googleGenAi.Models
+
+  /**
+   * Instruments chat completion creation
+   * and creates the LLM events
+   */
+  shim.record(models.prototype, 'generateContentInternal',
```
jsumners-nr, line 204: I think the comment format on `TRACKING_METRIC` is appropriate. The comment format on the `shim.record` lines is not the one I would use. My recommendation is to reserve multi-line comments with a leading `/**` for things that would make sense to have included in generated documentation. That's things like functions, classes, and fields.

Basically, the fact that jsdoc requires the multi-line comment format with a leading `/**` makes using them ambiguous in code. For example, do we think this `shim.record` comment should be in generated documentation? Or is it merely an explanation for a future developer working on this code. I'm pretty confident it is the second scenario, but it would be clear if the block were written with single line comments.

Additionally, single line comments make it easier to comment out large chunks of code if one needs to do that. Consider:

```js
/*
 * something something
 */
shim.record()
```

If we try to comment out that whole section with a multi-line comment, it will not work:

```js
/*
/*
 * something something
 */
shim.record()
*/
```

But if we use single line comments, we can easily do it:

```js
/*
//
// something something
//
shim.record
*/
```

jsumners-nr (commented):
It'll take some time to go through the actual code. But I wanted to get this style feedback over sooner rather than later.

lib/instrumentation/@google/genai.js
```diff
@@ -0,0 +1,292 @@
...
+  if (err) {
+    const llmError = new LlmErrorMessage({ cause: err, summary: completionSummary, response })
+    agent.errors.add(transaction, err, llmError)
+  }
+}
+
+function instrumentStream ({ agent, shim, request, response, segment, transaction }) {
+  if (!agent.config.ai_monitoring.streaming.enabled) {
+    shim.logger.warn(
+      '`ai_monitoring.streaming.enabled` is set to `false`, stream will not be instrumented.'
```
mrickard, line 126:
```suggestion
      '`ai_monitoring.streaming.enabled` is set to `false`. Stream will not be instrumented.'
```

lib/instrumentation/@google/genai.js
```diff
@@ -0,0 +1,292 @@
...
+    segment,
+    transaction,
+    request,
+    response,
+    withError: err != null
+  })
+
+  // Only take the first response message and append to input messages
+  // request.contents can be a string or an array of strings
+  // response.candidates is an array of candidates (choices), we only take the first one
```
mrickard, line 96:
```suggestion
  // response.candidates is an array of candidates (choices); we only take the first one
```

lib/instrumentation/@google/genai.js
amychisholm03: This is the same warning message that other LLM libraries use and what the tests expect. I _could_ change it everywhere but seems unnecessary.

lib/instrumentation/@google/genai.js
```diff
@@ -0,0 +1,292 @@
...
+    }
+  )
+
+  // Instruments chat completion streaming
+  // and creates the LLM events
+  shim.record(models.prototype, 'generateContentStreamInternal',
+    function wrapGenerateContentStream(shim, func, name, args) {
+      if (!agent.config.ai_monitoring.streaming.enabled) {
+        shim.logger.warn(
+          '`ai_monitoring.streaming.enabled` is set to `false`, stream will not be instrumented.'
```
mrickard, line 232:
```suggestion
          '`ai_monitoring.streaming.enabled` is set to `false`; stream will not be instrumented.'
```

lib/instrumentation/@google/genai.js
```diff
@@ -0,0 +1,292 @@
...
+
+          if (err) {
+            const llmError = new LlmErrorMessage({ cause: err, embedding, response })
+            shim.agent.errors.add(transaction, err, llmError)
+          }
+        }
+      })
+    }
+  )
+}
```
mrickard, line 292: This file seems to be missing test coverage for several functions: 

- recordChatCompletionMessages (67-121)
- instrumentStream (123-185)
- recordEvent (30-37)
- addLlmMeta (46-49)
- shim.record for generateContentInternal (204-223)
- returning the recorderSpec for generateContentStreamInternal (237-246)
- shim.record for wrapEmbedContent (255-290)

amychisholm03: Yeah I was trying to figure where to add these tests and I finally found it in test/versioned/google-genai which I'm working on right now

lib/llm-events/google-genai/chat-completion-message.js
```diff
@@ -0,0 +1,42 @@
...
+    request = {},
+    response = {},
+    index = 0,
+    message,
+    completionId,
+    transaction
+  }) {
+    super({ agent, segment, request, response, transaction })
+    this.id = makeId(36)
+    this.role = message?.role ?? 'user' // Role is only defined if it is 'model'
```
mrickard, line 23: Could it also be defined as `assistant` or `system` as well? In either case, the code makes sense.

lib/llm-events/google-genai/chat-completion-message.js
```diff
@@ -0,0 +1,43 @@
...
```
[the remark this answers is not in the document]
amychisholm03: I can rewrite the comment; message.role is only defined if it equals 'model'.

test/unit/instrumentation/@google/genai.test.js
```diff
@@ -0,0 +1,96 @@
...
+    const agent = helper.loadMockedAgent()
+    agent.config.ai_monitoring = { enabled: true, streaming: { enabled: true } }
+    const shim = new GenericShim(agent, '@google/genai')
+    sandbox.stub(shim.logger, 'debug')
+    sandbox.stub(shim.logger, 'warn')
+
+    ctx.nr.agent = agent
+    ctx.nr.shim = shim
+    ctx.nr.sandbox = sandbox
+    ctx.nr.initialize = require('../../../../lib/instrumentation/@google/genai.js')
```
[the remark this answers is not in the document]
jsumners-nr: Yes, there is a reason: the local `package.json` in versioned test suites interferes with the resolution of our root `package.json`. Therefore, the aliases will not work in the versioned test suites. I may research possible fixes for this in the future.

test/versioned/google-genai/chat-completions.test.js
```diff
@@ -0,0 +1,470 @@
...
+    },
+    streaming: {
+      enabled: true
+    }
+  })
+  const { GoogleGenAI } = require('@google/genai')
+  ctx.nr.client = new GoogleGenAI({
+    apiKey: 'fake-versioned-test-key',
+    vertexai: false
+  })
```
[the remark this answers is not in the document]
amychisholm03: Yup that worked. Thank you

test/unit/instrumentation/@google/genai.test.js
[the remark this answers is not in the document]
amychisholm03: This test is in `unit` though?

test/versioned/google-genai/chat-completions.test.js
```diff
@@ -0,0 +1,474 @@
...
+test.afterEach((ctx) => {
+  helper.unloadAgent(ctx.nr.agent)
+  ctx.nr.server?.close()
+  removeModules('@google/genai')
+})
+
+test('should create span on successful models generateContent', (t, end) => {
+  const { client, agent, host, port } = t.nr
+  helper.runInTransaction(agent, async (tx) => {
+    const result = await client.models.generateContent({
```
mrickard, line 64: When executing this in versioned tests, a typeError occurs when `shim`'s `_applyRecorderSegment` function executes `shim.applyContext`.

lib/instrumentation/@google/genai.js
```diff
@@ -0,0 +1,289 @@
...
+const {
+  LlmChatCompletionMessage,
+  LlmChatCompletionSummary,
+  LlmEmbedding,
+  LlmErrorMessage
+} = require('../../../lib/llm-events/google-genai')
+const { RecorderSpec } = require('../../../lib/shim/specs')
+const { extractLlmContext } = require('../../util/llm-utils')
+
+const { AI } = require('../../../lib/metrics/names')
```
mrickard, line 17: These imports could make use of our import aliases.

test/versioned/google-genai/mock-server.js
```diff
@@ -0,0 +1,107 @@
...
+ * Build a mock server that listens on a 127.0.0.1 and a random port that
+ * responds with pre-defined responses based on the "prompt" sent by the
+ * Google GenAI client library.
+ *
+ * @example
+ * const { server, port } = await GoogleGenAIMockServer()
+ * const client = new GoogleGenAI({
+ *   vertexai: false
+ *   apiKey: 'some gemini api key'
+ *  }
```
jsumners-nr: This example is missing the injection of the destination server details.

lib/instrumentation/@google/genai.js
```diff
@@ -0,0 +1,289 @@
...
+  err,
+  transaction
+}) {
+  if (!response) {
+    // If we get an error, it is possible that `response = null`.
+    // In that case, we define it to be an empty object.
+    response = {}
+  }
+
+  // Explicitly end segment to consistent duration
```
mrickard:
```suggestion
  // Explicitly end segment to provide consistent duration
```
Is there a missing word in the comment here?

lib/instrumentation/@google/genai.js
mrickard: Running the versioned tests does generate a coverage report for files in `lib/llm-events/google-genai`, but doesn't generate an entry for `lib/instrumentation/@google/genai.js`, which is covered by unit tests. I'd think it would make sense to alter the test at `test/unit/instrumentation/@google/genai.test.js` so that it directly tested the methods in `lib/instrumentation/@google/genai.js`. As an experiment, I've placed breakpoints in this instrumentation file, and not seen those code paths executed when running the genai unit test.

amychisholm03: I just don't know if we should deviate how we do unit tests for the LLM libraries, considering this is how `openai` and `bedrock` did their unit tests. This code _does_ get tested.

lib/llm-events/error-message.js
```diff
@@ -23,9 +23,21 @@ module.exports = class LlmErrorMessage {
...
    */
   constructor({ response, cause, summary, embedding, vectorsearch, tool } = {}) {
-    this['http.statusCode'] = response?.status ?? cause?.status
+    // For @google/genai only, cause does not have the `error` or `status` fields,
+    // but it does have `message` with the info we need. So, we need to parse
+    // the relevant fields from cause.message to get `status` and `error`.
+    let parsedError
+    if ((embedding?.vendor === 'gemini' || summary?.vendor === 'gemini') && cause?.message) {
+      // Extract the JSON portion of the cause.message
+      const jsonStartIndex = cause.message.indexOf('{')
```
mrickard: This approach seems fragile. Do we have an example of this kind of error message? It'd be helpful to include a test for this case in `test/unit/llm-events/error.test.js`.

test/versioned/google-genai/chat-completions.test.js
```diff
@@ -0,0 +1,464 @@
...
+  const api = helper.getAgentApi()
+  function cb(model, content) {
+    assert.equal(model, expectedModel)
+    if (content === promptContent || content === promptContent2) {
+      return 53
+    } else if (content === res) {
+      return 11
+    }
+  }
+  api.setLlmTokenCountCallback(cb)
```
mrickard, line 234: While the token count callback is set in this test, this doesn't appear to assert that the callback is called. That may be leading to a lack of coverage for the token count callback in`lib/llm-events/google-genai/chat-completion-message.js`

lib/llm-events/error-message.js
amychisholm03: I agree; I'll add a test now. There's an example error in test/versioned/google-genai/mock-responses.js

lib/llm-events/error-message.js
```diff
@@ -23,9 +23,21 @@ module.exports = class LlmErrorMessage {
...
+      if (jsonStartIndex !== -1) {
+        const jsonString = cause.message.substring(jsonStartIndex)
+        parsedError = JSON.parse(jsonString)?.error
```
jsumners-nr, line 36: `JSON.parse` can throw an error. It should be handled. Particularly since we are extracting the source from something else. It is quite possible that we do not extract a full JSON string to parse.

lib/llm-events/google-genai/embedding.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * @typedef {object} LlmEmbeddingParams
+ * @property {string} input - The input message for the embedding call
+ */
+/**
+ * @type {LlmEmbeddingParams}
+ */
+
+class LlmEmbedding extends LlmEvent {
```
jsumners-nr: These doc blocks seem incomplete. Also, as written, `class LlmEmbedding` will get documented as "type" `LlmEmbeddingParams`.

amychisholm03: Yeah, this might be some lovely code-complete I missed. The other LlmEvent classes don't have doc blocks any how.

lib/llm-events/error-message.js
mrickard: Thank you! I can see that the code will parse the example errors in that test. I think, though, that @jsumners-nr point remains: JSON.parse itself can error if the string it's parsing is not valid JSON. That error case still needs to be handled. One possibility is a try/catch block around the parse, falling back to the `cause.message` string if it can't be parsed. Or simply supplying the `cause.message` without attempting to parse it.

lib/llm-events/error-message.js
```diff
@@ -23,9 +23,24 @@ module.exports = class LlmErrorMessage {
...
+    const isGeminiVendor = embedding?.vendor === 'gemini' || summary?.vendor === 'gemini'
+    if (isGeminiVendor && cause?.message) {
+      try {
+        // Extract the JSON portion of the cause.message
+        const jsonStartIndex = cause.message.indexOf('{')
+        const jsonString = cause.message.substring(jsonStartIndex)
+        parsedError = JSON.parse(jsonString)?.error
+      } catch {
+        parsedError = undefined
+      }
```
jsumners-nr, line 39: We do need to add a test to cover this. Simply setting `jsonString` to something like `{bad:"json"` will trigger the path.

lib/llm-events/error-message.js
```diff
@@ -23,9 +23,24 @@ module.exports = class LlmErrorMessage {
...
+    let parsedError
+    const isGeminiVendor = embedding?.vendor === 'gemini' || summary?.vendor === 'gemini'
+    if (isGeminiVendor && cause?.message) {
+      try {
+        // Extract the JSON portion of the cause.message
+        const jsonStartIndex = cause.message.indexOf('{')
+        const jsonString = cause.message.substring(jsonStartIndex)
+        parsedError = JSON.parse(jsonString)?.error
+      } catch {
+        parsedError = undefined
```
mrickard, line 38: Would it be better to pass the `cause.message` as a string, since we know it's defined at this point?
```suggestion
        parsedError = cause.message
```

lib/llm-events/error-message.js
```diff
@@ -23,9 +23,24 @@ module.exports = class LlmErrorMessage {
...
+      }
+    }
+    this['http.statusCode'] = response?.status ?? cause?.status ?? parsedError?.code
     this['error.message'] = cause?.message
-    this['error.code'] = response?.code ?? cause?.error?.code
+    this['error.code'] = response?.code ?? cause?.error?.code ?? parsedError?.code
```
mrickard, line 43: ...though if we define `parsedError` from `cause.message`, we'll need to change this reference to `parsedError.code`

lib/instrumentation/@google/genai.js
mrickard: @amychisholm03 That's true--running the versioned tests, I can see the execution pass through `lib/instrumentation/@google/genai.js`. And it's also true that this is consistent with how we do unit tests for other LLM libraries.

lib/llm-events/error-message.js
amychisholm03: I think it's better to set parsedError to undefined that way we don't attempt to set http.statusCode or error.code if that value doesn't actually exist.

mrickard: And it's true that we're capturing `cause.message` in `error.message`. 

I'd recommend we add one test of a malformed `cause.message` string in `test/versioned/google-genai/mock-responses.js`, just to make sure that this error handling does what we think it does.

test/unit/llm-events/google-genai/chat-completion-message.test.js
```diff
@@ -0,0 +1,225 @@
...
+      segment: {},
+      request: {},
+      response: {}
+    })
+    assert.equal(chatMessageEvent['llm.conversation_id'], conversationId)
+    end()
+  })
+})
+
+test('respects record_content', (t, end) => {
```
mrickard: Perhaps this test description could explain a bit more--this test verifies that content will not be recorded if record_content is not enabled, but `respects record_content` can be interpreted in other ways.

test/unit/llm-events/google-genai/chat-completion-message.test.js
```diff
@@ -0,0 +1,225 @@
...
+        message: res.candidates[0].content,
+        index: 2
+      })
+      assert.equal(chatMessageEvent.token_count, expectedCount)
+      end()
+    })
+  })
+})
+
+test('should not set token_count if not set in usage nor a callback registered', (t, end) => {
```
mrickard: Perhaps 
```suggestion
test('should not set token_count if it is not set in usage, or if no callback is registered', (t, end) => {
```

test/unit/llm-events/google-genai/embedding.test.js
```diff
@@ -0,0 +1,178 @@
...
+      segment,
+      request: req,
+      response: res
+    })
+    assert.equal(embeddingEvent.token_count, 21)
+    end()
+  })
+})
+
+test('should not set token count when not present in usage nor tokenCountCallback', (t, end) => {
```
mrickard:
```suggestion
test('should not set token count when not present in usage or tokenCountCallback', (t, end) => {
```

test/unit/llm-events/google-genai/chat-completion-message.test.js
```diff
@@ -0,0 +1,225 @@
...
+        message: res.candidates[0].content,
+        index: 2
+      })
+      assert.equal(chatMessageEvent.token_count, undefined)
+      end()
+    })
+  })
+})
+
+test('should not set token_count if not set in usage nor a callback registered returns count', (t, end) => {
```
mrickard:
```suggestion
test('should not set token_count if not set in usage or if a callback registered does not return count', (t, end) => {
```

lib/llm-events/error-message.js
amychisholm03: Yup I added that test in `test/unit/llm-events/google-genai/error.test.js` per James' suggestion

PR #3121, opened 2025-05-29
chore: Added preliminary otel configuration

### LINKED ISSUE #3120: Add OTEL compatibility feature configuration block
Currently, our OTEL compatibility feature is gated with a singular feature flag. We still need the feature flag while we are developing this feature, but we have reached the point that we need some well defined configuration. In particular, we currently want to be able to enable/disable the metrics API support individually. So we need to add support for the following configuration:

```js
{
  otel: {
    bridge: {
      enabled: { default: false, formatter: boolean }
    },
    metrics: {
      enabled: { default: false, formatter: boolean }
    }
  }
}
```

We also need to update the implementation for the metrics API to recognize this configuration:

https://github.com/newrelic/node-newrelic/blob/9db49707ffbfe7e5608b0ac539da76cd6f41016e/lib/otel/setup.js#L42

jsumners-nr:
This PR resolves #3120.


mrickard (approved):
Considering that this is such a large feature, it does make sense to have two layers of enablement (`opentelemetry_bridge` and `traces`/`metrics`).

PR #3127, opened 2025-06-02
test: Added JSDoc annotation for benchmark interface

mrickard:
Annotated arguments for benchmark tests. This goes along with additional documentation on the team drive. 

## Related Issues

Closes 2707
Closes NR-337788

### DIFF
--- a/test/lib/benchmark.js
+++ b/test/lib/benchmark.js
@@ -11,6 +11,13 @@ exports.createBenchmark = (opts) => {
   return new Benchmark(opts)
 }
 
+/**
+ * Represents the benchmark test
+ * @class
+ * @param {object} opts test suite configuration options
+ * @param {string} opts.name name of test suite
+ * @param {number} opts.runs number of iterations to run. 1000 by default
+ */
 class Benchmark {
   constructor(opts) {
     this.name = opts.name || 'Anonymous Suite'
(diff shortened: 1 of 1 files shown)


test/lib/benchmark.js
```diff
@@ -11,6 +11,13 @@ exports.createBenchmark = (opts) => {
   return new Benchmark(opts)
 }
 
+/**
+ * Represents the benchmark test
+ * @class
+ * @param {object} opts test suite configuration options
+ * @param {string} opts.name name of test suite
+ * @param {number} opts.runs number of iterations to run. 1000 by default
```
jsumners-nr: You can use the dsl to indicate the default:

```suggestion
 * @param {number} [opts.runs=1000] number of iterations to run.
```

Looks like the `name` is optional as well.

https://jsdoc.app/tags-param#optional-parameters-and-default-values

test/lib/benchmark.js
```diff
@@ -26,6 +33,18 @@ class Benchmark {
...
   }
 
+  /**
+   * @param {object} opts benchmark test configuration options
+   * @param {string} opts.name name of benchmark test in the suite
+   * @param {function} opts.fn function invoking the agent method we're testing
+   * @param {function} [opts.initialize] function executed before the tests run
+   * @param {function} [opts.teardown] function executed after the tests run
+   * @param {function} [opts.before] function executed before each test run; accepts agent as its param
+   * @param {function} [opts.after] function executed after each test run
```
jsumners-nr: I think the signature of these functions should be documented. Are they expected to be synchronous only? Should they return a promise?

mrickard: The suite was initially written to be synchronous with some callbacks. We refactored it to replace many of those with async/await, though synchronous code remains.

jsumners-nr: Okay. I am saying these are the details that should be in the documentation.

mrickard: I've updated the documentation throughout this file.

jsumners-nr: Thank you for doing that. However, the documentation for the `function` parameters on this method are still missing the details I outlined above. This method is rather important to understand because it is the primary method.

jsumners-nr (approved):

PR #3139, opened 2025-06-06
feat: Support `openai.responses.create` api

### LINKED ISSUE #3135: Support `responses.create` API in openai >= 5.0.0
### Description
OpenAI released version 5 of their [SDK](https://github.com/openai/openai-node), and now has a new API (`client.responses.create`) that they are strongly recommending to users. This API was not present in other versions of OpenAI, so we will not instrument these methods if a user were to take their advice.

Thankfully, the old API, `client.chat.completions.create` will be supported indefinitely, so we can keep that in. We will need to instrument at least the `client.responses.create` method.  

We will **_not_** instrument streaming support in this story; that will be a different issue.

### Example of New API
```
import OpenAI from 'openai';

const client = new OpenAI({
  apiKey: process.env['OPENAI_API_KEY'], // This is the default and can be omitted
});

const response = await client.responses.create({
  model: 'gpt-4o',
  instructions: 'You are a coding assistant that talks like a pirate',
  input: 'Are semicolons optional in JavaScript?',
});

console.log(response.output_text);
```

### Acceptance Criteria
* After instrumenting `client.responses.create`, we can see the same AI monitoring data in NR UI as we would if we called `client.chat.completions.create`.

amychisholm03:
This PR resolves #3135. This only supports the non-streaming `openai.responses.create` function; streaming will come later.


lib/instrumentation/openai.js
```diff
@@ -23,6 +23,64 @@ const { DESTINATIONS } = require('../config/attribute-filter')
...
+  const message = { content: undefined, role: undefined }
+  if (response?.object === 'response') {
+    message.content = response?.output?.[0]?.content?.[0]?.text
+    message.role = response?.output?.[0]?.role
+  } else {
+    message.content = response?.choices?.[0]?.message?.content
+    message.role = response?.choices?.[0]?.message?.role
+  }
+
+  return message
```
mrickard: I'd think this could be a bit more compact. What about something like this?
```suggestion
  let content
  let role 
  if (response?.object === 'response') {
    content = response?.output?.[0]?.content?.[0]?.text
    role = response?.output?.[0]?.role
  } else {
    content = response?.choices?.[0]?.message?.content
    role = response?.choices?.[0]?.message?.role
  }

  return { content, role }
```

lib/llm-events/openai/chat-completion-message.js
```diff
@@ -22,21 +22,31 @@ module.exports = class LlmChatCompletionMessage extends LlmEvent {
     this.role = message?.role
     this.sequence = index
     this.completion_id = completionId
-    this.is_response = response?.choices?.[0]?.message?.content === message?.content
 
+    // The chat completion API and responses API have different structures for the response.
+    const isResponseAPI = response?.object === 'response'
+    if (isResponseAPI) {
+      this.is_response = message.content === response?.output?.[0]?.content?.[0]?.text
```
mrickard, line 31: Under what conditions would we expect `message.content` and `response?.output?.[0]?.content?.[0]?.text` ever to be equal? If they're not, would that imply that this is not a response?

amychisholm03: This is a check to determine if `message` is from the `request` or `response`. `response?.output?.[0]?.content?.[0]?.text === message.content` will be true if the given `message` is within the `response` object. If the `message` is not present in the `response`, then it must be part of the `request`. 

The `isResponseAPI` boolean determines what part of `response` to look at for the `message.content`. `openai.responses.create` has `response.output`, while `openai.chat.completions.create` has `response.choices`.

mrickard: That makes sense. I'd recommend adding a brief explanatory comment here.

test/unit/llm-events/openai/common-responses-api.js
```diff
@@ -0,0 +1,108 @@
...
+
+  return expected
+}
+
+module.exports = {
+  req,
+  res,
+  chatRes,
+  getExpectedResult
+}
```
mrickard, line 108: It's great to bundle these standard fixtures in commonly shared files this way.

test/versioned/openai/chat-completions-v5.test.js
```diff
@@ -0,0 +1,299 @@
+/*
```
mrickard, line 1: While there's a lot of overlap with the pre-version-5 chat completions test, I agree with splitting this out in its own file for clarity.

lib/instrumentation/openai.js
```diff
@@ -23,6 +23,69 @@ const { DESTINATIONS } = require('../config/attribute-filter')
...
+  } else if (Array.isArray(request?.messages)) {
+    return request.messages
+      .filter(msg => msg?.content && msg?.role)
+      .map(msg => ({
+        content: msg.content,
+        role: msg.role
+      }))
+  } else {
+    shim.logger.warn('No valid messages found in OpenAI request object.')
+    return []
```
mrickard: If every condition in a conditional statement results in a `return`, and the last part of the statement is an `else`, that could be refactored to have the last `return` in the outer scope. (See [this post](https://medium.com/@rachelchervin/if-else-vs-if-return-using-implicit-returns-in-javascript-and-react-bbac63656e1b) for an example.) Alternatively, you could use the conditional to assign values to a a particular variable that's returned by the function.

PR #3145, opened 2025-06-11
chore: Added logging of used New Relic environment variables

### LINKED ISSUE #3144: Add logging of which environment variables are read
GTS has asked if we log which environment variables that we read. We do not. Our .NET agent does do this. We should implement this feature as well. Values should be redacted.

https://github.com/newrelic/node-newrelic/blob/b144d712f6e75265dc4be6f82a2fc3977ac1c640/lib/config/index.js#L1158-L1196

jsumners-nr:
This PR resolves #3144.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,10 +21,10 @@ services:
       interval: 30s
       timeout: 10s
       retries: 5
-  
+
   opensearch:
     container_name: nr_node_opensearch
-    image: opensearchproject/opensearch:2.1.0 
+    image: opensearchproject/opensearch:2.1.0
     environment:
       - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
       # Disable password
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -48,15 +48,15 @@ services:
   # `zookeeper` and `kafka` services.
   zookeeper:
     container_name: nr_node_kafka_zookeeper
-    image: confluentinc/cp-zookeeper
+    image: confluentinc/cp-zookeeper:7.9.1
```
jsumners-nr, line 51: The released an `8.0.0` tag earlier today (2025-06-11) that fails to start.

docker-compose.yml
```diff
@@ -48,15 +48,15 @@ services:
...
     ports:
       - '2181:2181'
     environment:
       ZOOKEEPER_CLIENT_PORT: 2181
       ZOOKEEPER_TICK_TIME: 2000
   kafka:
     container_name: nr_node_kafka
-    image: confluentinc/cp-kafka
+    image: confluentinc/cp-kafka:7.9.1
```
jsumners-nr, line 59: The released an 8.0.0 tag earlier today (2025-06-11) that fails to start.

PR #3156, opened 2025-06-13
test: Added mock HTTP2 server for Bedrock tests

mrickard:
This avoids the protocol errors in versioned tests for Bedrock@3.798.0 and higher, but shows that full instrumentation of requests will require full HTTP2 instrumentation. External segments aren't being created as we don't instrument HTTP2. 

Relates to #3060


mrickard:
Superseded by work in https://github.com/newrelic/node-newrelic/pull/3284

PR #3172, opened 2025-06-24
feat: Added support for knex

jsumners-nr:
This PR adds support for `knex` as the initial implementation of automatic OTEL bridge configuration.

### DIFF
--- a/test/versioned/knex/knex.test.js
+++ b/test/versioned/knex/knex.test.js
@@ -60,6 +60,19 @@ test('records queries', (t, end) => {
     assert.equal(agent.queries.samples.size === 1, true, 'should have recorded query')
     const sample = agent.queries.samples.values().next().value
     assert.equal(sample.trace.query, 'insert into `users` (`email`, `username`) values (?, ?)')
+    assert.equal(sample.total > 0, true, 'sample should have positive duration')
+
+    const metrics = agent.metrics._metrics.unscoped
+    const expectedMetrics = [
+      'Datastore/better-sqlite3/all',
+      'Datastore/better-sqlite3/allWeb',
+      'Datastore/instance/better-sqlite3/localhost/0',
+      'Datastore/operation/better-sqlite3/insert',
+      'Datastore/statement/better-sqlite3/users/insert'
+    ]
+    for (const expectedMetric of expectedMetrics) {
+      assert.equal(metrics[expectedMetric].callCount, 1)
+    }
 
     end()
   })
--- a/lib/otel/span-processor.js
+++ b/lib/otel/span-processor.js
@@ -139,6 +139,12 @@ module.exports = class NrSpanProcessor {
         continue
       }
 
+      if (Object.prototype.hasOwnProperty.call(attribute, 'value') === true) {
+        // The transformation rule is specifying a static value for the
+        // attribute. So we need to use that instead of whatever was found.
+        value = attribute.value
+      }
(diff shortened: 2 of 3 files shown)


lib/otel/bootstrap-traces.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }))
+
+  interceptSpanKey(agent)
+  opentelemetry.context.setGlobalContextManager(new ContextManager(agent))
+  opentelemetry.propagation.setGlobalPropagator(new TracePropagator(agent))
+
+  registerInstrumentations({
+    instrumentations: [
+      new KnexInstrumentation({ maxQueryLength: -1, requireParentSpan: true })
```
bizob2828, line 39: this instrumentation will conflict with our agent depending on the underlying db engine.  I found during our initial POC that if you rely on db engines that we instrument you get duplicate segments.  This needs to be solved before this is merged

lib/otel/transformation-rules.json
```diff
@@ -483,7 +483,40 @@
...
+  },
+  {
+    "name": "OtelDbClientNoServer1_24",
+    "type": "db",
+    "matcher": {
+      "required_span_kinds": [
+        "client"
+      ],
+      "required_attribute_keys": [
+        "db.system"
```
bizob2828, line 497: this rule is too generic. it'll never hit any of the rules that < 1.24.0 which is why you have failing tests.  Which db engine lacks server.address and server.port? You'll have to find other span attrs that are present in this case. based on the spec maybe i'ts `db.instance.id`?

test/versioned/knex/knex.test.js
```diff
@@ -0,0 +1,66 @@
...
+})
+
+test.afterEach(ctx => {
+  helper.unloadAgent(ctx.nr.agent)
+  ctx.nr.knex.destroy()
+  // TODO: remove otel modules?
+  removeModules(['knex', 'better-sqlite3'])
+})
+
+test('records queries', (t, end) => {
```
bizob2828, line 53: if we're writing versioned tests they should align with other db libraries we instrument to ensure parity.  you can check mysql, pg, etc for reference

lib/otel/transformation-rules.json
bizob2828: also without a server/port NR cannot link the db in the system map. There has to be some attribute that represents it

jsumners-nr: > this rule is too generic. it'll never hit any of the rules that < 1.24.0 which is why you have failing tests. Which db engine lacks server.address and server.port? You'll have to find other span attrs that are present in this case. based on the spec maybe i'ts `db.instance.id`?

I haven't given this any thought yet. This WIP is just a first pass to sketch everything out. But we really don't have many attributes to go on in this case:

image


> also without a server/port NR cannot link the db in the system map. There has to be some attribute that represents it

There really isn't a server or port. This is an in-process database: sqlite.

test/versioned/knex/knex.test.js
jsumners-nr: I believe this has been addressed.

jsumners-nr:
Closing this because I don't think it's going to go anywhere. It uncovered some issues, e.g. how do we handle traces that are duplicated due to the knex instrumentation in combination with our native database instrumentations?, and likely isn't going to go anywhere.

PR #3187, opened 2025-06-27
fix: Fixed logic around static values in OTEL rules

jsumners-nr:
I discovered that we don't actually support the static value feature of an OTEL transformation rule. This PR resolves that problem.

### DIFF
--- a/lib/otel/rules.js
+++ b/lib/otel/rules.js
@@ -116,7 +116,7 @@ class Rule {
   }
 
   /**
-   * Determines if the given span satifies this rule's requirements.
+   * Determines if the given span satisfies this rule's requirements.
    *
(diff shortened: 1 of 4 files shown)


bizob2828 (commented):
I don't think this PR is needed.  The solution is to define an attribute mapping like this:

```js
 { 
  value: 'hard-coded',
  name: 'test-attr',
  target: 'segment'
}
```

This will add a segment attribute of `test-attr` with the value of `hard-coded`. We already have rules with attribute mappings with value and they work just fine, like this [one](https://github.com/newrelic/node-newrelic/blob/5498c15bb6cf3329fe50595da46ab4f2b1385170/lib/otel/transformation-rules.json#L613)

PR #3191, opened 2025-06-30
feat: Instrument AWS Bedrock Converse API

### LINKED ISSUE #3153: [spike] Research viability of instrumenting aws-sdk-v3 Converse API
### Description
AWS Bedrock drops support for different AI models and adds support for new models somewhat frequently. We should not have to re-do our instrumentation/tests every time a new model is added or dropped. Thankfully, AWS has released their [Converse API](https://community.aws/content/2dtauBCeDa703x7fDS9Q30MJoBA/amazon-bedrock-converse-api-developer-guide), which aims to standardize requests to Bedrock, regardless of AI model. 

We should research if instrumenting Converse API will reduce our work for keeping our `@aws-sdk/client-bedrock-runtime` instrumentation up-to-date.

### Acceptance Criteria
- [ ] Decide whether or not to instrument Converse API.
- [ ] Can we use Converse API to determine what models are currently supported by Bedrock? If so, can we make a service that informs us when models are added/dropped?

amychisholm03:
Verified that Ryan (from AIM)'s previous work on Converse API is correct and up-to-date (with some tweaks); supports `ConverseCommand` and `ConverseStreamCommand`. Resolves #3153.


lib/instrumentation/aws-sdk/v3/bedrock.js
```diff
@@ -222,7 +242,10 @@ function createBedrockResponse({ bedrockCommand, response, err }) {
   if (err) {
     bedrockResponse = new BedrockResponse({ bedrockCommand, response: err, isError: err !== null })
   } else {
-    bedrockResponse = new BedrockResponse({ bedrockCommand, response })
+    bedrockResponse =
+      bedrockCommand instanceof BedrockConverseCommand
+        ? new BedrockConverseResponse({ bedrockCommand, response })
+        : new BedrockResponse({ bedrockCommand, response })
```
jsumners-nr: According to the docblock, `bedrockCommand` shouldn't ever be anything other than an instance of `BedrockCommand`.

lib/llm-events/aws-bedrock/bedrock-converse-command.js
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * The question posed to the LLM.
+   *
+   * @returns {string|string[]|undefined}
+   */
+  get prompt() {
+    const result = []
+    for (const message of this.#messages) {
+      if (typeof message?.content === 'string') {
+        result.push({ role: message.role, content: message?.content })
```
jsumners-nr: Is this condition covered by tests?

lib/llm-events/aws-bedrock/bedrock-converse-command.js
```diff
@@ -0,0 +1,86 @@
...
+      } else if (message?.content != null && Array.isArray(message.content)) {
```
jsumners-nr: `Array.isArray` is going to return `false` if it is given a nullish value:

```suggestion
      } else if (Array.isArray(message.content) === true) {
```

lib/llm-events/aws-bedrock/bedrock-converse-response.js
```diff
@@ -0,0 +1,115 @@
...
+   * @param params.isError
+   */
+  constructor({ response, bedrockCommand, isError = false }) {
+    this.#innerResponse = isError ? response.$response : response.response
+    this.#originalResponse = response
+    this.isError = isError
+
+    if (this.isError) {
+      return
+    }
```
jsumners-nr: Do we have a test that covers this error case?

lib/llm-events/aws-bedrock/bedrock-converse-command.js
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+'use strict'
+
+const { stringifyConverseChunkedMessage } = require('./utils')
+
+/**
+ * Parses an AWS invoke command instance into a re-usable entity.
+ */
+class BedrockConverseCommand {
```
jsumners-nr: Why doesn't this extend `BedrockCommand`?

lib/llm-events/aws-bedrock/chat-completion-message.js
```diff
@@ -57,7 +58,7 @@ class LlmChatCompletionMessage extends LlmEvent {
 
   #setId(index) {
     const cmd = this.bedrockCommand
-    if (cmd.isTitan() === true || cmd.isClaude() === true) {
+    if (cmd instanceof BedrockConverseCommand || cmd.isTitan() === true || cmd.isClaude() === true) {
```
jsumners-nr: If the new object extended `BedrockCommand` we wouldn't need to rely on an `instanceof` check. We could:

```js
if (cmd?.isConverse()) { /* do stuff */ }
```

The `?.` is important because the method would only be defined if it were the extended object. I don't know which would be more performant, but I think the `is` check would be easier to read here since the rest of the conditionals are in that format.

lib/instrumentation/aws-sdk/v3/bedrock.js
amychisholm03: We could do `BedrockConverseResponse extends BedrockResponse`. That would fix this and provide continuity with `BedrockConverseCommand extends BedrockCommand`.

lib/llm-events/aws-bedrock/bedrock-converse-command.js
jsumners-nr, line 1: Reviewing this file side-by-side with `BedrockCommand`, I'm not seeing much reason to create this new file. The only significant difference is using `#messages` over `#body`. I think the original constructor could be updated to something like this:

```js
constructor(input) {
  this.#input = input
  
  if (Object.hasOwn(input, 'body') === true) {
    this.#body = JSON.parse(this.#input.body)
    this.#isConverse = false
  } else if (Object.hasOwn(input, 'messages') === true) {
    this.#messages = input.messages
    this.#isConvers = true
  }
  
  this.#modelId = this.#input.modelId?.toLowerCase() ?? ''
}
```

Then we'd update the other methods appropriately.

lib/instrumentation/aws-sdk/v3/bedrock.js
jsumners-nr: See my latest comment on `BedrockConverseCommand`. I think it will help with these checks.

lib/llm-events/aws-bedrock/bedrock-converse-command.js
amychisholm03: Yeah, that's what I'm working through now. Originally, I was going to have a parent class `BedrockCommand` with children `BedrockInvokeCommand` and `BedrockConverseCommand`, but it's a bit over-engineered. Making it just one class seems like a better approach.

lib/llm-events/aws-bedrock/bedrock-command.js
```diff
@@ -32,17 +50,23 @@ class BedrockCommand {
...
+        result = this.#body.max_tokens_to_sample
+      } else if (this.isClaude3() === true || this.isCohere() === true) {
+        result = this.#body.max_tokens
+      } else if (this.isLlama() === true) {
+        result = this.#body.max_gen_length
+      } else if (this.isTitan() === true) {
+        result = this.#body.textGenerationConfig?.maxTokenCount
+      }
+      return result
     }
```
jsumners-nr: Given that the `if` block returns from the function, the `else` block is unnecessary. We can write the code like:

```js
function foo() {
  if (something) {
    return 'a'
  }

  /* do some more stuff */
  return 'b'
}
```

lib/llm-events/aws-bedrock/bedrock-command.js
```diff
@@ -60,61 +84,84 @@ class BedrockCommand {
...
-      this.isClaudeTextCompletionApi() === true ||
-      this.isCohere() === true ||
-      this.isLlama() === true
-    ) {
-      return [{ role: 'user', content: this.#body.prompt }]
-    } else if (this.isClaudeMessagesApi() === true) {
-      return normalizeClaude3Messages(this.#body?.messages)
+      }
+      return result
+    } else {
```
jsumners-nr: Same note about the `else` block as above.

lib/llm-events/aws-bedrock/bedrock-command.js
```diff
@@ -60,61 +84,84 @@ class BedrockCommand {
...
-      this.isClaude() === true ||
-      this.isClaude3() === true ||
-      this.isCohere() === true ||
-      this.isLlama() === true
-    ) {
-      result = this.#body.temperature
+    if (this.#isConverseCommand) {
+      // Logic for Converse
+      return this.#input?.inferenceConfig?.temperature
+    } else {
```
jsumners-nr: Unwrap `else` block.

lib/llm-events/aws-bedrock/bedrock-command.js
```diff
@@ -143,21 +190,24 @@ class BedrockCommand {
     return this.#modelId.startsWith('amazon.titan-embed')
   }
 
-  isClaudeMessagesApi() {
-    return (this.isClaude3() === true || this.isClaude() === true) && 'messages' in this.#body
+  // These methods are specific to the InvokeModelCommand's body structure
+  // and will only be called when isConverse is false.
```
jsumners-nr: I don't think this note is going to age well. It's likely that more methods will be added, or these will be shifted around. We don't know. Instead, if we feel we need to highlight that a method is specific to a certain scenario, we should call that out in the method's docblock.

lib/llm-events/aws-bedrock/bedrock-response.js
```diff
@@ -18,60 +18,51 @@ const { stringifyClaudeChunkedMessage } = require('./utils')
...
+   * @param {boolean} params.isError
    */
   constructor({ response, bedrockCommand, isError = false }) {
+    this.#originalResponse = response
     this.#innerResponse = isError ? response.$response : response.response
     this.#command = bedrockCommand
     this.isError = isError
 
+    // Determine if it's a Bedrock Converse API response based on the command
+    this.#isConverse = this.#command.isConverse
```
jsumners-nr: The comment isn't providing us any context we can't see from the code line:
```suggestion
    this.#isConverse = this.#command.isConverse
```

lib/llm-events/aws-bedrock/bedrock-response.js
```diff
@@ -82,32 +73,40 @@ class BedrockResponse {
...
    */
   get completions() {
-    return this.#completions
+    if (this.#isConverse) {
+      const content = this.#output?.content
+      if (!content) return []
+      return [typeof content === 'string' ? content : stringifyConverseChunkedMessage(content)]
+    } else {
+      return this.#completions
+    }
```
jsumners-nr: We shift the unnecessary `else` block out like described above here as well. But I would flip this one a bit so that the "easy" logic is handled first and the "complicated" logic forms the bulk of the function:

```suggestion
    if (this.#isConverse === false) {
      return this.#completions
    }

    const content = this.#output?.content
    if (!content) return []
    return [typeof content === 'string' ? content : stringifyConverseChunkedMessage(content)]
```

lib/llm-events/aws-bedrock/bedrock-response.js
```diff
@@ -82,32 +73,40 @@ class BedrockResponse {
...
-      result = this.#parsedBody.stop_reason
-    } else if (cmd.isCohere() === true) {
-      result = this.#parsedBody.generations?.find((r) => r.finish_reason !== null)?.finish_reason
-    } else if (cmd.isLlama() === true) {
-      result = this.#parsedBody.stop_reason
-    } else if (cmd.isTitan() === true) {
-      result = this.#parsedBody.results?.find((r) => r.completionReason !== null)?.completionReason
+    if (this.#isConverse) {
+      return this.#originalResponse.output.stopReason
+    } else {
```
jsumners-nr: Unwrap `else` block.

lib/llm-events/aws-bedrock/bedrock-command.js
```diff
@@ -55,17 +55,19 @@ class BedrockCommand {
...
+      if (this.isClaude3() === true || this.isCohere() === true) {
+        return this.#body.max_tokens
+      }
+      if (this.isLlama() === true) {
+        return this.#body.max_gen_length
+      }
+      if (this.isTitan() === true) {
+        return this.#body.textGenerationConfig?.maxTokenCount
+      }
+      return undefined
```
jsumners-nr: This isn't technically necessary. In JavaScript, every function always returns `undefined` if the end of the function is reached and there isn't an explicit `return` statement.

lib/llm-events/aws-bedrock/bedrock-command.js
```diff
@@ -100,41 +102,39 @@ class BedrockCommand {
       for (const message of this.#messages) {
         if (typeof message?.content === 'string') {
           result.push({ role: message.role, content: message?.content })
-        } else if (Array.isArray(message.content) === true) {
+        }
+        if (Array.isArray(message.content) === true) {
```
jsumners-nr: This code should be equivalent to the previous code, but I'd argue that the `else` provides clearer intent. That intention being "it's either this thing or that thing."

This is not a blocker. I just want to clarify my reasoning behind unwrapping the other `else` blocks versus this one.

lib/llm-events/aws-bedrock/bedrock-command.js
```diff
@@ -100,41 +102,39 @@ class BedrockCommand {
...
+      this.isLlama()
+    ) {
+      return [{ role: 'user', content: this.#body.prompt }]
+    }
+
+    if (this.isClaudeMessagesApi(this.#body)) {
+      return normalizeClaude3Messages(this.#body?.messages)
+    }
+
+    return []
```
jsumners-nr, line 135: With this change, the docblock should be updated as the function is no longer returning `undefined`.

lib/llm-events/aws-bedrock/utils.js
```diff
@@ -31,6 +31,43 @@ function stringifyClaudeChunkedMessage(chunks) {
...
 
+/**
+ *
+ * @param {object[]} chunks - The "chunks" that make up a single conceptual message. In a multi-modal scenario, a single message
+ * might have a number of different-typed chunks interspersed
+ * @returns {string} - A stringified version of the message. We make a best-effort effort attempt to represent non-text chunks. In the future
+ * we may want to extend the agent to support these non-text chunks in a richer way. Placeholders are represented in an XML-like format but
+ * are NOT intended to be parsed as valid XML
+ */
+function stringifyConverseChunkedMessage(chunks) {
```
jsumners-nr, line 42: We should have unit tests to cover the cases in this function.

amychisholm03: I added some more in `utils.test.js`.

jsumners-nr: Running `c8 --reporter=html -o ./coverage/unit borp 'test/unit/llm-events/**/*.test.js'` and then looking at `coverage/unit/index.html`, I can see that this is totally covered now. There's some other stuff not covered by unit tests, but I assume they are covered by the versioned tests since CodeCov is now happy. 👍

lib/llm-events/aws-bedrock/converse-stream-handler.js
jsumners-nr, line 1: The CodeCov annotations in this file are valid according to `c8 --reporter=html -o ./coverage/unit borp 'test/unit/llm-events/**/*.test.js'`. That is, as of this post, the following lines are missing coverage:

+ 108
+ 130
+ 134

These look to be core logic lines in the handler, so they should be tested.

lib/llm-events/aws-bedrock/bedrock-command.js
```diff
@@ -60,61 +85,80 @@ class BedrockCommand {
...
-      return [
-        {
-          role: 'user',
-          content: this.#body.inputText
+    if (this.#isConverseCommand) {
+      // Logic for Converse
+      const result = []
+      for (const message of this.#messages) {
+        if (typeof message?.content === 'string') {
+          result.push({ role: message.role, content: message?.content })
```
jsumners-nr: This line is not covered by any of our tests, presumably. The local unit test coverage (`c8 --reporter=html -o ./coverage/unit borp 'test/unit/llm-events/**/*.test.js'`) shows the whole loop not being covered. Since CodeCov is only complaining about this one line, I assume the others are handled in the versioned tests.

jsumners-nr (commented):
We are close!

lib/llm-events/aws-bedrock/bedrock-command.js
amychisholm03: Looking at [AWS's docs](https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference-call.html), it doesn't seem that `typeof message?.content === 'string'` would ever be true because it doesn't match the expected `message` format. I will remove this.

PR #3254, opened 2025-07-28
docs: update distributed tracing example target to use the correct repo

throck:
## Description

I was looking through the examples linked here, and noticed that the distributed tracing link 404'ed, because it points to an external+personal repo rather than newrelic-node-examples, therefore this simple change simply repoints this target to the correct repo so that the DT example can be easily found


mrickard:
Thank you for your PR, @throck ! Would you be able to sign the CLA?

throck:
@mrickard done!

amychisholm03 (approved):

amychisholm03:
Thank you for catching this! @throck

PR #3256, opened 2025-07-28
ci: Fail benchmark ci if tests fail

amychisholm03:
Bob and I were reviewing our benchmark test code and noticed that the coordinating GH action was not reporting a failure if any of the benchmark tests fail. This PR aims to "propagate" the failure, so that GH actions properly reports failure for a test that errors or exits abnormally.

Review [this run on my fork](https://github.com/amychisholm03/node-newrelic/actions/runs/16581118949) to see it properly fail.

### DIFF
--- a/bin/run-bench.js
+++ b/bin/run-bench.js
@@ -19,6 +19,7 @@ const tests = []
 const testPromises = []
 const globs = []
 const opts = Object.create(null)
+let hasFailures = false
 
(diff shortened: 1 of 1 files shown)


jsumners-nr (commented):
By the way, you could set `process.exitCode = 1` in the locations where failures should trigger a different exit code and not have to track that failures have occurred. See https://nodejs.org/api/process.html#processexitcode_1

PR #3280, opened 2025-08-05
Optimize url scrubbing

svetlanabrennan:
## Description

## How to Test

## Related Issues

### DIFF
--- a/lib/transaction/index.js
+++ b/lib/transaction/index.js
@@ -409,7 +409,6 @@ function finalizeNameFromUri(requestURL, statusCode) {
     )
   }
 
-  this.url = urltils.scrub(requestURL)
   this.statusCode = statusCode
   this.name = this.getFullName()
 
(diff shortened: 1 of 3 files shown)


lib/transaction/index.js
amychisholm03, line 1: Does `if (!this.agent.config.url_obfuscation.enabled)` need to be added on line 321 (in `function _partialNameFromUri`)? There's another `urltils.scrub` there. Or do we always scrub then no matter what?

jsumners-nr (approved):

lib/transaction/index.js
svetlanabrennan: So it looks like in this case (line 321), we don't need to add this if statement because the `requestUrl` parameter in `_partialNameFromUri` function is `this.url` (which is transaction.url) so it's already obfuscated by the time this function is called so calling `scrub` on it again doesn't undue obfuscation. 

Whereas the passed in `requestUrl` in `finalizeNameFromUri` function is `transaction.parsedUrl` is some cases and `transaction.url` in other cases. So `transaction.url` was already parsed before `finalizeNameFromUri` was called and when the passed in `transaction.parsedUrl` was scrubbed here, it replaced the already obfuscated `transaction.url` with this one. 

So in this pr, we won't do anything with line 321 but it should be addressed later when we optimize url scrubbing.

PR #3285, opened 2025-08-07
ci: Send benchmark metrics to NR One

### LINKED ISSUE #2116: Benchmark tests should send results to New Relic
Benchmark tests currently send results to stdout. We should add the ability to ship results to the connect service.

If these tests are going to deliver (metric) telemetry to New Relic, we'll want to add a collector or aggregator for them, which can deliver the metrics after tests are run. 

Acceptance criteria
* A collector or aggregator to receive benchmark metrics, replacing or augmenting the current `ConsolePrinter`
* A transport mechanism to send collected benchmark metrics to NR
* Test runs should be tagged with Node version and agent version
* Unit tests for the new collector/aggregator

Relates to #2114

### LINKED ISSUE #2117: feat: Create dashboard for benchmark tests
Create a dashboard for presenting benchmark test results. We should be able to see the agent performance over time, tagged with agent version and node runtime (assuming these are run in a matrix of currently-supported versions).

Acceptance criteria: 
* A dashboard displaying benchmark metrics tagged with agent version and Node version, updated by running the tests.

Relates to #2114

amychisholm03:
This PR resolves #2116 and resolves #2117. See dashboard here: https://onenr.io/0VjY5Bz2qj0

## How to Test
Locally:
```
cd bin
# Supply NEW_RELIC_LICENSE_KEY to your environment
node run-bench.js
```
GH workflow:
* Run the Benchmark Tests workflow on this branch (`NR-2116/benchmark-metrics-to-nr`): https://github.com/amychisholm03/node-newrelic/actions/workflows/benchmark-tests.yml


bin/otel-metrics-sender.js
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright 2025 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const { MeterProvider, PeriodicExportingMetricReader } = require('@opentelemetry/sdk-metrics')
+const { OTLPMetricExporter } = require('@opentelemetry/exporter-metrics-otlp-http')
```
jsumners-nr: We should be able to use the exporter we already leverage to provide the metrics API. See https://github.com/newrelic/node-newrelic/blob/c71185f7730fc12e5cb8860517394b53880a91b7/lib/otel/metrics/bootstrap-metrics.js#L94-L100

amychisholm03: How would this work if there's no agent in the main process (i.e. `run-bench.js` which calls `otel-metrics-exporter.js`)?

jsumners-nr: I'm saying that instead adding another dependency, we should be able to utilize the existing dependency. So:

1. `npm remove newly-added-dep`
2. Replace `require('newly-added-dep')` in your script

bin/env.sample
jsumners-nr, line 1: I'm conflicted on whether or not we should add an env file to the bin directory. I understand why it is desired, but it seems out-of-place and I am confident mistakes will happen that end up with some variant of the file being included in the repository (and leaking secrets). I have another suggestion to consider below.

bin/otel-metrics-sender.js
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright 2025 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const { MeterProvider, PeriodicExportingMetricReader } = require('@opentelemetry/sdk-metrics')
```
jsumners-nr: Given that this script relies on certain environment variables to be set, I'd assure they are present before continuing. For example:

```js
'use strict'

const NR_KEY = process.env.NEW_RELIC_LICENSE_KEY
if (NR_KEY == null) {
  console.error('Missing required environment variable: NEW_RELIC_LICENSE_KEY')
  process.exit(1)
}
```

Declaring all of them at the start of the script like this gives future readers an easy place to reference for the required variables, keeps everything defined close to the usage, and eliminates the necessity of including an environment file sample.

amychisholm03: Ok, sounds good. I will implement.

bin/env.sample
amychisholm03: The .env file is only really required if we wanted to run locally. For initial development and testing it is required, but once we are more confident with it, we can remove the .env and just test by running through the GH workflow.

jsumners-nr: I do not think we need it for local runs or development. The script verifying the required environment is present and refusing to run if not solves the issue. It tells the user what is required, and they can solve it in a variety of ways, using a file passed via `node --env-file` being one of them. We can see in the commit history of this PR how quickly such a file can get out-of-date and how difficult it is to link it to the right code.

amychisholm03: Ok, I'll remove.

jsumners-nr (approved):

PR #3311, opened 2025-08-14
chore: Updated eslint configuration on next

jsumners-nr:
Updates the `next` branch with the latest eslint configuration and applies fixes.


jsumners-nr:
I cherry picked the commit from `main`, but it looks like the GH UI is not recognizing that. It should all shake out when we merge `next`.

PR #3337, opened 2025-08-21
feat: MCP `StreamableHTTPClientTransport` support

### LINKED ISSUE #3328: Instrument MCP HTTP Streaming
### Description
We have already instrumented MCP's `stdio` transport; now we need to instrument MCP's [**streamable HTTP transport**](https://modelcontextprotocol.io/docs/learn/architecture#transport-layer).

### Acceptance Criteria
The following spans/segments get created according to this [spec](https://source.datanerd.us/agents/agent-specs/compare/main...mcp-spec):
* `readResource`
* `callTool`
* `getPrompt`

Appropriate versioned tests are created.

amychisholm03:
This PR resolves #3328. Because we only need to instrument the MCP SDK's `Client` class and not the `Transport` classes (`StdioClientTransport` and `StreamableHTTPClientTransport`), we already have HTTP streaming support from our [initial instrumentation](https://github.com/newrelic/node-newrelic/pull/3247). This is confirmed by the following tests. The appropriate spans for HTTP streaming are created according to the [MCP spec](https://source.datanerd.us/agents/agent-specs/pull/743):
* `Llm/tool/MCP/callTool/{primitive_name}`
* `Llm/tool/MCP/readResource/{primitive_name}`
* `Llm/tool/MCP/getPrompt/{primitive_name}`

Also, we have discovered that `client.callTool`, `client.getPrompt`, and `client.readResource` are just wrappers around `client.request`, so we will just instrument that instead.

## How to Test
```
npm run versioned mcp-sdk
```


bizob2828 (commented):
I don't think this is truly testing streaming to client, looking at their [example](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/client/simpleStreamableHttp.ts) they're using `client.setRequestHandler` and `client.setNotificationHandler` to receive data then call `client.request`

amychisholm03:
> I don't think this is truly testing streaming to client, looking at their [example](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/examples/client/simpleStreamableHttp.ts) they're using `client.setRequestHandler` and `client.setNotificationHandler` to receive data then call `client.request`

`client.callTool`, `client.readResource`, and `client.getPrompt` are just wrappers around `client.request` (see [here](https://github.com/modelcontextprotocol/typescript-sdk/blob/3bc2235d747c320dfa0b6227cc84414c6d0add89/src/client/index.ts#L396)). We could instrument it alongside the other methods as seen in this [commit](https://github.com/newrelic/node-newrelic/pull/3337/commits/701b9cb125f99e6e0699a33e23e3f18b4439d1c4), but it's essentially a duplicate span. Also, `client.setNotificationHandler` and `client.setRequestHandler` (in the example you linked) just handle the tool's result; they're not awaiting an actual stream.

To my understanding, `StreamableHTTPClientTransport` abstracts away from the user dealing with the actual stream: https://github.com/modelcontextprotocol/typescript-sdk/blob/3bc2235d747c320dfa0b6227cc84414c6d0add89/src/client/streamableHttp.ts#L314. They're no need to await the result like a traditional streaming method.

lib/subscribers/mcp-sdk/client-request.js
```diff
@@ -0,0 +1,58 @@
...
+  constructor({ agent, logger }) {
+    super({ agent, logger, channelName: 'nr_request' })
+    // We only want to record a `request` segment if it is
+    // not a child of another MCP span e.g. `callTool`.
+    // This is for use cases, like streaming, where the user
+    // calls `client.request` directly.
+    this.internal = true
+  }
+
+  determineFunction(args0) {
```
amychisholm03, line 26: If we do want to instrument `client.request`, we could just remove the other three subscribers and rely on this entirely.

lib/subscribers/mcp-sdk/client-request.js
```diff
@@ -0,0 +1,58 @@
...
+      functionName = 'getPrompt'
+      primitiveName = args0?.params?.name ?? 'prompt'
+    } else if (methodName === 'resources/read') {
+      mcpPrefix = MCP.RESOURCE
+      functionName = 'readResource'
+      const uri = args0?.params?.uri
+      primitiveName = typeof uri === 'string' ? uri.split('://')[0] : 'resource'
+    } else {
+      this.logger.debug(`@modelcontextprotocol/sdk: Will not create '${methodName}' segment.`)
+    }
```
mrickard, line 47: This conditional works fine, and it's understandable, though I wonder if there might not be a more compact way of doing this.

mrickard (dismissed):
This looks good to me! I agree with what you're seeing in the MCP client code.

amychisholm03:
> This looks good to me! I agree with what you're seeing in the MCP client code.

Do you think we should get rid of the three subscribers for `callTool`, `readResource`, and `getPrompt` in favor of just instrumenting the underlying `request`?

mrickard:
> Do you think we should get rid of the three subscribers for `callTool`, `readResource`, and `getPrompt` in favor of just instrumenting the underlying `request`?

Would we still be able to derive the functionName of the function that was originally called? I'd think in terms of a customer trying to track down a performance issue, it's helpful that we record the functionName in the segment name.

amychisholm03:
> > Do you think we should get rid of the three subscribers for `callTool`, `readResource`, and `getPrompt` in favor of just instrumenting the underlying `request`?
> 
> Would we still be able to derive the functionName of the function that was originally called? I'd think in terms of a customer trying to track down a performance issue, it's helpful that we record the functionName in the segment name.

Yes, `determineFunction` in `client-request.js` ([link](https://github.com/newrelic/node-newrelic/pull/3337/files#diff-b14a6351666db9fbf77e7ca55e5869cfe5a4aab3498e08fdf13d81b868c5fa03R21)) handles this.

mrickard:
> > > Do you think we should get rid of the three subscribers for `callTool`, `readResource`, and `getPrompt` in favor of just instrumenting the underlying `request`?
> > 
> > 
> > Would we still be able to derive the functionName of the function that was originally called? I'd think in terms of a customer trying to track down a performance issue, it's helpful that we record the functionName in the segment name.
> 
> Yes, `determineFunction` in `client-request.js` ([link](https://github.com/newrelic/node-newrelic/pull/3337/files#diff-b14a6351666db9fbf77e7ca55e5869cfe5a4aab3498e08fdf13d81b868c5fa03R21)) handles this.

In that case, I'd be in favor of instrumenting the underlying `request`, since it's used by the other methods.

test/versioned/mcp-sdk/streaming-server.js
```diff
@@ -0,0 +1,113 @@
+/*
+ * Copyright 2025 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+const express = require('express')
+const { McpServer, ResourceTemplate } = require('@modelcontextprotocol/sdk/server/mcp.js')
+const { StreamableHTTPServerTransport } = require('@modelcontextprotocol/sdk/server/streamableHttp.js')
+const { z } = require('zod')
```
jsumners-nr, line 10: This dependency isn't listed in any `package.json`. The `express` package is listed in the root `package.json` as a dev dependency, so that's good.

test/versioned/mcp-sdk/client-streaming.test.js
```diff
@@ -0,0 +1,191 @@
...
+
+const {
+  MCP
+} = require('../../../lib/metrics/names')
+
+async function startServer() {
+  const { spawn } = require('child_process')
+
+  // Spin up streaming server in background
+  const server = spawn('node', ['streaming-server.js'])
```
jsumners-nr: Why do we need to run the server in a separate process? If we look at other tests where we need a local server to communicate with, we can see that we usually keep them in the same process. For example https://github.com/newrelic/node-newrelic/blob/1fc1e1a4086200bd3fd76081913498d7ca344a3a/test/versioned/aws-sdk-v3/api-gateway.test.js#L17-L20

If we are going to keep the separate process, though, we need to make this update:
```suggestion
  const server = spawn('node', [path.join(__dirname, 'streaming-server.js')])
```

Without that change, we cannot run `node test/versioned/mcp-sdk/client-streaming.test.js` because the target script will not be found and the subprocess will not start.

I'd further change it to:
```suggestion
  const server = spawn(process.argv0, [path.join(__dirname, 'streaming-server.js')])
```

test/versioned/mcp-sdk/client-streaming.test.js
```diff
@@ -0,0 +1,191 @@
...
+    })
+
+    const agentMetrics = agent.metrics
+    const expectedPkgMetrics = [
+      [{ name: 'Supportability/Features/Instrumentation/OnRequire/@modelcontextprotocol/sdk' }],
+      [{ name: `Supportability/Features/Instrumentation/OnRequire/@modelcontextprotocol/sdk/Version/${semver.major(pkgVersion)}` }]
+    ]
+    assertMetrics(agentMetrics, expectedPkgMetrics, false, false)
+  })
+})
```
jsumners-nr: It's not clear to me that `helper.runInTransaction` will return a promise to await. So I'd change these tests to look like:
```suggestion
test('should create span for callTool', (t, end) => {
  const { agent, client, pkgVersion } = t.nr
  helper.runInTransaction(agent, async (tx) => {
    const result = await client.callTool({
      name: 'echo',
      arguments: {
        message: 'example message'
      }
    })

    assert.ok(result, 'should return a result from the tool call')

    const name = `${MCP.TOOL}/callTool/echo`
    assertSegments(tx.trace, tx.trace.root, [name], { exact: false })

    tx.end()
    assertSpanKind({
      agent,
      segments: [
        { name, kind: 'internal' }
      ]
    })

    const agentMetrics = agent.metrics
    const expectedPkgMetrics = [
      [{ name: 'Supportability/Features/Instrumentation/OnRequire/@modelcontextprotocol/sdk' }],
      [{ name: `Supportability/Features/Instrumentation/OnRequire/@modelcontextprotocol/sdk/Version/${semver.major(pkgVersion)}` }]
    ]
    assertMetrics(agentMetrics, expectedPkgMetrics, false, false)
    
    end()
  })
})
```

test/versioned/mcp-sdk/streaming-server.js
mrickard: I don't know that we need to use a validation dependency on mock servers for versioned tests. We'd control any data sent to the test servers, so we shouldn't need validation.

amychisholm03: `zod` is required to define the `inputSchema` for a given tool. I'll add it to the `package.json`.

test/versioned/mcp-sdk/client-streaming.test.js
amychisholm03: Interestingly enough when I implement this change, I get the hanging issue @mrickard was describing when running `npm run versioned mcp-sdk`. I'll work on implementing the local server instead.

jsumners-nr (dismissed):

mrickard (approved):

PR #3357, opened 2025-09-08
refactor: Benchmark test names

### LINKED ISSUE #3355: Refactor benchmark tests to follow a consistent naming scheme
### Description
Our benchmark tests (`test/benchmark`) don't follow a standard naming scheme (e.g. `PriorityQueue` tests live in `test/benchmark/events` instead of the expected `test/benchmark/lib/priority-queue`). We should follow the naming convention with our other tests, like `test/benchmark/lib/db/query-parsers/sql.bench.js`.

### Acceptance Criteria

- [x] All tests in `test/benchmark` are in a folder/path that matches the class/object/function that is being tested.
- [ ] The individual benchmark tests are named consistently (e.g. is this a comparison test or just testing a specific method?).

amychisholm03:
This PR resolves #3355.

In most cases, benchmark test files have been renamed to match the `require` of the class/method they are benchmarking. For example, if we are benchmarking the `PriorityQueue` class (`lib/priority-queue`), then its benchmark tests would live in `test/benchmark/lib/priority-queue`. If only 1 test file is needed, then the test file would be named `test/benchmark/lib/priority-queue.bench.js`.

Some benchmark tests test internal things that don't have a direct `require`, e.g. the `tracer` tests, in which the tracer is accessed with `agent_helper`'s `getTracer()` method. These tests still live inside `test/benchmark/lib` because they are testing internal functionality.

Finally, there are benchmark tests that use external classes/packages to showcase "no agent" vs "with agent" i.e. so far, `http`, `promise`, and `async-hook`. These do not live in `lib`, just `test/benchmark` because they are not benchmarking any specific internal classes/methods.


test/benchmark/lib/config/attribute-filter.bench.js
```diff
@@ -5,12 +5,12 @@
...
-const AttributeFilter = require('../../../lib/config/attribute-filter')
-const benchmark = require('../../lib/benchmark')
-const { makeAttributeFilterConfig } = require('../../lib/agent_helper')
+const AttributeFilter = require('#agentlib/config/attribute-filter.js')
+const benchmark = require('#testlib/benchmark.js')
+const { makeAttributeFilterConfig } = require('#testlib/agent_helper.js')
 
 const suite = benchmark.createBenchmark({
-  name: 'config.filter'
+  name: 'filterTransaction'
```
mrickard: Including the test directory name in the test name seems like something we'd want to keep--it'd make it easier to find a specific test file when tracing back from output. (I agree that `filterTransaction` is better than just `filter`.)

test/benchmark/lib/prioritized-attributes.bench.js
```diff
@@ -23,7 +23,7 @@ const halfLowHalfHighPriorityAttributes = new PrioritizedAttributes(SEGMENT_SCOP
 batchAddAttributes(halfLowHalfHighPriorityAttributes, 32, ATTRIBUTE_PRIORITY.LOW)
 batchAddAttributes(halfLowHalfHighPriorityAttributes, 32, ATTRIBUTE_PRIORITY.HIGH)
 
-const suite = benchmark.createBenchmark({ name: 'priority attributes', runs: 100000 })
+const suite = benchmark.createBenchmark({ name: 'addAttribute', runs: 100000 })
```
mrickard: Considering that these tests are setting priority attributes, does it make sense to drop "priority" from the test name?

mrickard:
Most of our test directories generally follow the pattern that `lib/` contains files used by tests, but which aren't tests themselves. The benchmark tests have been an outlier, though, in the past having one test (`db/query-parsers/sql-bench.js`) in `lib`. What's the benefit of moving benchmark tests down into the `benchmark/lib` subdirectory? Is that to indicate that it's testing agent lib functionality?

amychisholm03:
> Most of our test directories generally follow the pattern that `lib/` contains files used by tests, but which aren't tests themselves. The benchmark tests have been an outlier, though, in the past having one test (`db/query-parsers/sql-bench.js`) in `lib`. What's the benefit of moving benchmark tests down into the `benchmark/lib` subdirectory?

That's a good point. I also mentioned this at the top of the PR, but there are some benchmark tests that don't specifically test internal methods/classes, so introducing `lib` denotes internal vs. not. I'm ok with changing it though.

mrickard:
I'd vote for not having tests inside `lib`, and reserving that for non-test files used by benchmark tests (if there are any).

test/benchmark/lib/prioritized-attributes.bench.js
amychisholm03: With this benchmark suite (and the above comment), I wanted the suite name to reflect the function name being tested. When the results get sent up to NR, they are displayed like `path.to.file.fileName, benchmarkSuiteName`, so it is descriptive enough as a whole. In this scenario, it would be `lib.prioritized_attributes, addAttribute`.

PR #3368, opened 2025-09-10
style: Require `jsdoc/valid-types` and `jsdoc/check-param-names`

### LINKED ISSUE #2503: Address JSDoc eslint issues in Node.js agent codebase
### Description
There are currently a large amount of linting warnings with regards to JSDocs in node-newrelic. We should fix these warnings, and update the configuration to error instead of warn, so that we can programatically enforce good code documentation practices

### Additional context
see this ticket for more details - https://new-relic.atlassian.net/browse/NR-134511

amychisholm03:
This PR relates to #2503. `jsdoc/valid-types` and `jsdoc/check-param-names` are now errors in our eslint config. I will do the rest of the `jsdoc/*` rules in separate PRs.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -66,7 +66,9 @@ const jsdocConfig = {
           'EventEmitter'
         ]
       }
-    ]
+    ],
+    'jsdoc/valid-types': 'error',
+    'jsdoc/check-param-names': 'error'
   }
 }
 const jsdocOverrides = {
--- a/lib/metrics/recorders/database-operation.js
+++ b/lib/metrics/recorders/database-operation.js
@@ -13,7 +13,7 @@ const metrics = require('../names')
  * - `recordOperationMetrics(segment [, scope])`
  *
  * @private
- * @this DatastoreShim
+ * @memberof DatastoreShim.prototype
  * @implements {MetricFunction}
  * @param {TraceSegment}  segment - The segment being recorded.
  * @param {string}        [scope] - The scope of the segment.
--- a/lib/shim/datastore-shim.js
+++ b/lib/shim/datastore-shim.js
@@ -521,7 +521,7 @@ function getDatabaseNameFromUseQuery(query) {
  * `STATEMENT` metric.
  *
  * @private
- * @this DatastoreShim
+ * @memberof DatastoreShim.prototype
  * @param {string} suffix
  *  Suffix to be added to the segment name.
  * @param {object | Function} nodule
@@ -606,7 +606,7 @@ function _getSpec({ spec, shim, fn, fnName, args }) {
  * - `_extractQueryStr(fn, fnName, spec, ctx, args)`
  *
  * @private
- * @this DatastoreShim
+ * @memberof DatastoreShim.prototype
  * @param {Function}  fn      - The query function to be executed.
  * @param {string}    fnName  - The name of the query function.
  * @param {QuerySpec} spec    - The query spec.
@@ -639,7 +639,7 @@ function _extractQueryStr(fn, fnName, spec, ctx, args) {
  * from `localhost` to the actual host name.
  *
  * @private
- * @this DatastoreShim
+ * @memberof DatastoreShim.prototype
  * @param {object} [parameters] - The segment parameters to clean up.
  * @returns {object} - The normalized segment parameters.
  */
--- a/lib/collector/parse-response.js
+++ b/lib/collector/parse-response.js
@@ -16,18 +16,15 @@ const logger = require('../logger').child({ component: 'new_relic_response' })
  * about errors as possible, and to parse out the return value as often
  * as possible.
  *
- * @param string         name     Remote method name that was invoked.
- * @param ServerResponse response HTTP response stream
- * @param Function       callback Function that will be called with any
- *                                error, the value returned by the server
- *                                (if any), and the raw JSON of the
- *                                server's response.
- * @param name
- * @param response
- * @param callback
- * @returns Function Another callback that is meant to be invoked with
- *                   any errors from reading the response stream, as
- *                   well as a string containing the full response.
+ * @param {string}         name     Remote method name that was invoked.
+ * @param {ServerResponse} response HTTP response stream
+ * @param {Function}       callback Function that will be called with any
+ *                                  error, the value returned by the server
+ *                                  (if any), and the raw JSON of the
+ *                                  server's response.
+ * @returns {Function} Another callback that is meant to be invoked with
+ *                     any errors from reading the response stream, as
+ *                     well as a string containing the full response.
  */
 module.exports = function parse(name, response, callback) {
   if (!callback) {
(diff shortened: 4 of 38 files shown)


eslint.config.js
```diff
@@ -47,7 +47,7 @@ const newrelicConfigOverrides = {
 const jsdocConfig = {
   plugins: { jsdoc },
   rules: {
-    'jsdoc/require-jsdoc': 'off',
+    'jsdoc/require-jsdoc': 'error',
```
jsumners-nr: If the way we satisfy this is as follows, then I don't think we should change the value.

```js
function foo () {} // errors with new config
```

```js
/**
 *
 */
function foo () {}
```

Satisfying the rule requirement with an empty docblock just litters the code.

eslint.config.js
```diff
@@ -66,17 +66,12 @@ const jsdocConfig = {
...
-  files: [
-    './lib/shim/*.js',
-    'lib/transaction/handle.js',
-    'api.js'
-  ],
-  rules: {
-    'jsdoc/require-jsdoc': 'warn'
+    ],
+    'jsdoc/require-param-description': 'warn',
+    'jsdoc/require-param-type': 'warn',
```
jsumners-nr: I think this should be set to `error` if the rule is going to be added. `@param foo`, which I have just learned is [actually valid](https://jsdoc.app/tags-param), doesn't convey anything more than the function signature already does.

eslint.config.js
```diff
@@ -66,17 +66,12 @@ const jsdocConfig = {
...
+    'jsdoc/require-returns': 'warn',
```
jsumners-nr: Personally, I find this rule annoying. I have found that in the majority of cases `@returns {some_type}` is sufficient due to proper naming of the function.

eslint.config.js
```diff
@@ -66,17 +66,12 @@ const jsdocConfig = {
...
+    'jsdoc/valid-types': 'error',
```
jsumners-nr, line 76: I am interested to see how this one works out.

eslint.config.js
```diff
@@ -66,17 +66,12 @@ const jsdocConfig = {
...
+    'jsdoc/check-param-names': 'error'
```
jsumners-nr, line 77: This one will also be interesting to observe. The [current docs](https://github.com/gajus/eslint-plugin-jsdoc/blob/2644c6e9ef73e22ae75c7ab3df82984b210b9379/.README/rules/check-param-names.md) don't show any example of good vs bad.

eslint.config.js
```diff
@@ -66,17 +66,12 @@ const jsdocConfig = {
...
-  }
-}
-const jsdocOverrides = {
-  files: [
-    './lib/shim/*.js',
-    'lib/transaction/handle.js',
-    'api.js'
-  ],
-  rules: {
-    'jsdoc/require-jsdoc': 'warn'
```
jsumners-nr, line 79: These overrides were specifically added to cover the files we parse in order to generate https://newrelic.github.io/node-newrelic/. I get that they shouldn't be necessary with the proposed changes, but I would be remiss if I didn't point this out.

eslint.config.js
amychisholm03: Agreed. I saw all of the empty docblocks before vacation and didn't want to deal with them then 😂 I'll make the change today.

eslint.config.js
amychisholm03: I agree; however, I will make this change in a separate PR because it will also be a big change. I think I'll remove it for now.

lib/metrics/recorders/database-operation.js
```diff
@@ -13,7 +13,7 @@ const metrics = require('../names')
  * - `recordOperationMetrics(segment [, scope])`
  *
  * @private
- * @this DatastoreShim
+ * @memberof DatastoreShim.prototype
```
jsumners-nr: I see a few of these `@this` => `@memberof` changes. Why are they being made?

amychisholm03: apparently this is a rule in `jsdoc/valid-types`: `error  @this should not have a name in "typescript" mode`

jsumners-nr: I don't know what "typescript mode" is, but let's turn that off. `@this` is a much more accurate tag for what is being conveyed, and it most certainly should have a name -- https://jsdoc.app/tags-this.

amychisholm03: Yeah, I'm not sure what "typescript mode" is either; I'll figure out how to turn it off

lib/shim/datastore-shim.js
```diff
@@ -521,7 +521,7 @@ function getDatabaseNameFromUseQuery(query) {
  * `STATEMENT` metric.
  *
  * @private
- * @this DatastoreShim
+ * @memberof DatastoreShim.prototype
```
amychisholm03: @jsumners-nr It's also worth noting that most of our codebase uses `@memberof Class`; there are only a few instances of `@this Class`. This file had 3 `this`'s but most of the file uses `memberof` (see [line 503](https://github.com/newrelic/node-newrelic/pull/3368/files/9ea683d3bbd2338b9cd19c14cd21d930985ea694#diff-2c668028f828a40287750884468d77ef8e5c8fc0374533f6b5cc548077036481L503) for example)

jsumners-nr: They were documented this way on purpose. Look at how the functions are actually used. They are not attached as members of `DatastoreShim`. They are invoked with their `this` binding set to it.

amychisholm03: Ah ok, thanks for pointing that out. It seems the issue is that there is no `{ }` around the class. Pushing up a fix now

jsumners-nr: The docs for jsdoc don't indicate curly brackets are valid, much less required. 🤔

amychisholm03: Yeah, it's odd. The documentation for this specific rule is very limited; I found [this](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#this) in TS docs, but I'm curious on why it's being enforced on JS projects.

jsumners-nr: People nowadays keep claiming TypeScript-specific annotations are "jsdoc" when they very clearly are not.

jsumners-nr: If I write the following script (index.js):

```
/**
 * @this Array
 */
function foo() {
}

/**
 * @this {Array}
 */
function bar() {}
```

And then run `npx jsdoc index.js`, the resulting `out/index.html` will look like:

image

I don't think `@this {thing}` is correct. I think it should be `@this thing`.

eslint.config.js
```diff
@@ -45,6 +45,12 @@ const newrelicConfigOverrides = {
 }
 
 const jsdocConfig = {
+  settings: {
+    jsdoc: {
+      // This is explicity set, so we don't get TypeScript-specific linting errors.
+      mode: 'jsdoc'
```
amychisholm03, line 51: @jsumners-nr This resolves the eslint error for `@this MyObject`

PR #3381, opened 2025-09-19
chore: Updated metrics for OTel Bridge enablement/disablement

### LINKED ISSUE #3378: Update supportability metrics in tracing
As with #3375, the supportability metric names have been standardized to a slightly different format. We need to update the following to `Supportability/Tracing/Nodejs/OpenTelemetryBridge/enabled`:

https://github.com/newrelic/node-newrelic/blob/66a9fd543d1a7854b7405c7b5b884d8a71773b85/lib/otel/setup.js#L60-L62

Additionally, we need to update the following to report the `Supportability/Tracing/Nodejs/OpenTelemetryBridge/disabled` metric:

https://github.com/newrelic/node-newrelic/blob/66a9fd543d1a7854b7405c7b5b884d8a71773b85/lib/otel/setup.js#L22-L27

Finally, Angler must be updated accordingly.

mrickard:
## Description

The standard for supportability metric names has changed to a different format. See #3378 

## Related Issues

Closes #3378

### DIFF
--- a/lib/otel/setup.js
+++ b/lib/otel/setup.js
@@ -23,6 +23,9 @@ function setupOtel(agent, logger = defaultLogger) {
     logger.warn(
       '`opentelemetry_bridge` is not enabled, skipping setup of opentelemetry-bridge'
     )
+    agent.metrics
+      .getOrCreateMetric('Supportability/Tracing/Nodejs/OpenTelemetryBridge/disabled')
+      .incrementCallCount()
     return
   }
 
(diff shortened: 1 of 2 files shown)


jsumners-nr (approved):
Excellent. Please remember to update angler.

mrickard:
> Excellent. Please remember to update angler.

Do we want to replace existing metrics in Angler with the new format? I'd think during the changeover, we'd want to add the new ones and deprecate the old ones in the future.

jsumners-nr:
> > Excellent. Please remember to update angler.
> 
> Do we want to replace existing metrics in Angler with the new format? I'd think during the changeover, we'd want to add the new ones and deprecate the old ones in the future.

I'm genuinely not clear how we handle this scenario. I think that if we add both, then we have to remember to go back and remove one and that is unlikely to happen. I think we can get away with simply replacing the old ones with the new ones since this feature is not advertised, still experimental, and not used by anyone that we know of. But I'd check with product for their opinion.

PR #3391, opened 2025-09-23
style: Require jsdoc `require-param-description` and `require-param-type`

amychisholm03:
This PR relates to #2503. This makes `jsdoc/require-param-description` and `jsdoc/require-param-type` errors in our eslint config.

### DIFF
--- a/lib/subscribers/amqplib/consume.js
+++ b/lib/subscribers/amqplib/consume.js
@@ -15,7 +15,7 @@ class ConsumeSubscriber extends MessageConsumer {
   /**
    * Handles the consumption of a message(s),
    * this will create a new transaction every time this function is called.
-   * @param data
+   * @param {object} data the data associated with the `asyncStart` event
    */
   asyncStart(data) {
     const [queue] = data?.arguments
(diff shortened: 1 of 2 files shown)


eslint.config.js
```diff
@@ -69,12 +69,15 @@ const jsdocConfig = {
...
           'MetricAggregator',
+          'Metrics',
           'EventEmitter'
         ]
       }
     ],
     'jsdoc/valid-types': 'error',
-    'jsdoc/check-param-names': 'error'
+    'jsdoc/check-param-names': 'error',
+    'jsdoc/require-param-description': 'error',
```
jsumners-nr, line 80: I'm torn on this one. A lot of times it just isn't necessary to add a description for an appropriately named parameter. In my view, the most important part of `@param {string} foo Is a foo.` is `{string}`. But we can see from this diff that the majority of changes are a direct result of this rule, and it would be clearer if everything had a description.

I think it would be best to poll the team to get their thoughts on this rule.

amychisholm03: I am leaning towards making it a requirement -- having redundant descriptions of some parameters with the gain of having some more complex params being described is worth it in my opinion. Also, if we want to do `jsdoc/require-description` as well, we'll run into a similar issue -- Redundant descriptions of functions (especially if it's properly named or small or has a descriptive return) at the gain of more complicated functions being described.

jsumners-nr (dismissed):

jsumners-nr (approved):

PR #3400, opened 2025-09-29
style: Require `check-types` and `require-returns-type` jsdoc

amychisholm03:
This PR relates to #2503. This changes the following jsdoc eslint rule:

- `check-types`: `off` -> `error`
- `require-returns-type`: `warn` -> `error`


lib/instrumentation/amqplib/utils.js
jsumners-nr, line 1: I think your branch is out-of-date. You should `git fetch upstream && git rebase upstream/main`.

PR #3404, opened 2025-09-29
style: Require jsdoc `check*`, `implements-on-classes`, and some `no-undefined-types`

### LINKED ISSUE #2503: Address JSDoc eslint issues in Node.js agent codebase

amychisholm03:
This PR closes #2503 

- `jsdoc/check-tag-names`: warn -> error
- `jsdoc/check-access`: warn -> error
- `jsdoc/implements-on-classes`: warn -> error
- `jsdoc/no-undefined-types`: warn -> warn (fix egregious violations e.g. `strng` instead of `string`)

### DIFF
--- a/lib/metrics/recorders/database-operation.js
+++ b/lib/metrics/recorders/database-operation.js
@@ -12,13 +12,12 @@ const metrics = require('../names')
  *
  * - `recordOperationMetrics(segment [, scope])`
  *
+ * This function is an implementation of {@link MetricFunction} (see #agentlib/shim/shim.js)
  * @private
- * @this DatastoreShim
- * @type {MetricFunction}
  * @param {TraceSegment}  segment - The segment being recorded.
  * @param {string}        [scope] - The scope of the segment.
+ * @param {Transaction}   transaction - The transaction associated with the segment.
  * @see DatastoreShim#recordOperation
- * @see MetricFunction
  */
 function recordOperationMetrics(segment, scope, transaction) {
   const duration = segment.getDurationInMillis()
--- a/lib/shim/shim.js
+++ b/lib/shim/shim.js
@@ -247,7 +247,7 @@ Shim.prototype[symbols.unwrap] = unwrapAll
  * their own recorders.
  *
  * @private
- * @typedef {Function} MetricFunction
+ * @callback MetricFunction
  * @param {TraceSegment}  segment - The segment to record.
  * @param {string}        [scope] - The scope of the recording.
  */


lib/metrics/recorders/database-operation.js
```diff
@@ -14,7 +14,7 @@ const metrics = require('../names')
  *
  * @private
  * @this DatastoreShim
- * @implements {MetricFunction}
+ * @type {MetricFunction}
```
jsumners-nr: We seem to be missing the definition for `MetricFunction`. I cannot find it anywhere. I believe we should define it at the top of `segment.js` because I think the top-most reference to it is at:

https://github.com/newrelic/node-newrelic/blob/100807342b9c9283f43042024f15bb575de219ee/lib/shim/specs/segment.js#L90-L97

The definition should look very similar to:

https://github.com/newrelic/node-newrelic/blob/100807342b9c9283f43042024f15bb575de219ee/lib/shim/specs/message.js#L23-L32

lib/shim/promise-shim.js
```diff
@@ -382,7 +382,6 @@ function _wrapExecutorContext(shim, args) {
  * @param {object} context context object
  * @param {Function} fn function that is wrapped
  * @returns {Function} wrapped function
- * @private
```
jsumners-nr, line 385: Why has this been removed? This function is a private function.

lib/shim/webframework-shim/index.js
```diff
@@ -344,9 +344,9 @@ function savePossibleTransactionName(req) {
  * responded.
  *
  * @memberof WebFrameworkShim.prototype
- * @param {function(args, object): boolean} pred
+ * @param {function(Array<object>, object): boolean} pred
```
jsumners-nr:
```suggestion
 * @param {function(object[], object): boolean} pred
```

Personally, I'd add a `typedef` and name it. The current syntax is not JavaScript and difficult to read.

lib/shim/webframework-shim/index.js
```diff
@@ -344,9 +344,9 @@ function savePossibleTransactionName(req) {
...
  *  Function which should return true if the object passed to it is considered
- *  a response.
+ *  a response. First param is args.
```
jsumners-nr: The additional sentence is not providing any clarity to me. I think the suggestion above to add a `typedef` will solve this.

lib/w3c/traceparent.js
```diff
@@ -11,7 +11,7 @@ const invalidParentId = '0'.repeat(16)
 /**
  * Parses a W3C traceparent header into an object representation.
  *
- * @link https://www.w3.org/TR/trace-context/#traceparent-header
+ * @see https://www.w3.org/TR/trace-context/#traceparent-header
```
jsumners-nr: The correct form is as follows (it seems):

```suggestion
 * @see {@link https://www.w3.org/TR/trace-context/#traceparent-header}
```

https://jsdoc.app/tags-see

lib/w3c/traceparent.js
```diff
@@ -88,7 +88,7 @@ class Traceparent {
   /**
    * Construct a new instance from an OTEL span context.
    *
-   * @link https://opentelemetry.io/docs/concepts/signals/traces/#span-context
+   * @see https://opentelemetry.io/docs/concepts/signals/traces/#span-context
```
jsumners-nr:
```suggestion
   * @see {@link https://opentelemetry.io/docs/concepts/signals/traces/#span-context}
```

lib/w3c/tracestate.js
```diff
@@ -11,7 +11,7 @@ const TracestateIntrinsics = require('./tracestate-intrinsics')
 /**
  * Parses a W3C traceparent header into an object representation.
  *
- * @link https://www.w3.org/TR/trace-context/#tracestate-header
+ * @see https://www.w3.org/TR/trace-context/#tracestate-header
```
jsumners-nr:
```suggestion
 * @see {@link https://www.w3.org/TR/trace-context/#tracestate-header}
```

test/lib/custom-assertions/assert-span-kind.js
```diff
@@ -17,7 +17,7 @@
  *
  * @param {object} params to function
  * @param {Agent} params.agent instance of agent
- * @param {segments} params.segments collection of span names and their respective span kind
+ * @param {Array<object>} params.segments collection of span names and their respective span kind
```
jsumners-nr:
```suggestion
 * @param {object[]} params.segments collection of span names and their respective span kind
```

eslint.config.js
```diff
@@ -98,7 +107,8 @@ const jsdocOverrides = {
     'api.js'
   ],
   rules: {
-    'jsdoc/require-jsdoc': 'warn'
+    // Public-facing functions should always have jsdoc
+    'jsdoc/require-jsdoc': 'error'
```
jsumners-nr, line 111: This is the one that introduced all of the empty docblock if I'm not mistaken. We should probably set the `enableFixer` option to `false`:

https://github.com/gajus/eslint-plugin-jsdoc/blob/main/docs/rules/require-jsdoc.md#enablefixer

lib/shim/promise-shim.js
amychisholm03: It had two `@private`s; see line 381.

eslint.config.js
amychisholm03: This for the `jsdcoOverrides` object which only affects `api.js`, `lib/transaction/handle.js`, and `lib/shim/*.js`, which all already have jsdoc blocks.

jsumners-nr: Ah, makes sense.

lib/metrics/recorders/database-operation.js
amychisholm03: It looks like it's defined as a `callback` in `lib/shim/shim.js`: https://github.com/newrelic/node-newrelic/blob/100807342b9c9283f43042024f15bb575de219ee/lib/shim/shim.js#L244-#L253. Changing it to `typedef` doesn't get rid of the `implements` error in `database-operation.js`.

jsumners-nr: Not sure how I missed that 🤦‍♂️. Maybe it should just be in the description as:

```
This function is an implementation of {@link MetricFunction}.
```

https://jsdoc.app/tags-callback

amychisholm03: I committed something similar.

PR #3409, opened 2025-10-03
feat: `adaptive_sampling_target` logic

### LINKED ISSUE #3353: Add support for `adaptive_sampling_target` value
Issue #3352 describes the work needed to add the configuration parameter. This ticket is for implementing the feature support.

We need to:

1. Update the adaptive sampler to honor the value: https://github.com/newrelic/node-newrelic/blob/fa27046d734f875e4e2d16e83fcdc1b9a02553eb/lib/agent.js#L290-L295 https://github.com/newrelic/node-newrelic/blob/fa27046d734f875e4e2d16e83fcdc1b9a02553eb/lib/adaptive-sampler.js#L9-L26
2. Verify that we continue to override the local configuration when a server value is received: https://github.com/newrelic/node-newrelic/blob/15253c08f6d8381f516040e339a0aedaf04e365e/test/unit/agent/agent.test.js#L1011-L1034

amychisholm03:
This PR resolves #3353. Instead of replacing every instance of `Config.sampling_target` with `Config.distributed_tracing.sampler.adaptive_sampling_target`, we can just turn `sampling_target` into a getter/setter property that will update `adaptive_sampling_target`. This needs to be released silently.


lib/config/index.js
```diff
@@ -199,8 +200,20 @@ function Config(config) {
...
 util.inherits(Config, EventEmitter)
 
+// Expose sampling_target as an accessor that proxies to distributed_tracing.sampler.adaptive_sampling_target.
+Object.defineProperty(Config.prototype, 'sampling_target', {
+  get: function () {
+    return this.distributed_tracing.sampler.adaptive_sampling_target
+  },
+  set: function (value) {
+    this.distributed_tracing.sampler.adaptive_sampling_target = value
+    this.emit('sampling_target', value)
```
jsumners-nr, line 213: I'm not seeing where currently emit this event. Can you point it out to me?

amychisholm03: In config/index.js, emits for attribute changes are typically like `this.emit(key, ...)` where key is the name of the attribute. Example: https://github.com/amychisholm03/node-newrelic/blob/626fc4e16249d7f0812e069ca880a0467accf5d6/lib/config/index.js#L663. I made an explicit emit here so that if `dt.sampler.adaptive_sampling_target` changes, the agent is notified.

Here is where the agent relies on this event: https://github.com/amychisholm03/node-newrelic/blob/626fc4e16249d7f0812e069ca880a0467accf5d6/lib/agent.js#L850-L852

jsumners-nr: I see. Thank you. That being the case, we should continue using the same method to update the value:

https://github.com/newrelic/node-newrelic/blob/626fc4e16249d7f0812e069ca880a0467accf5d6/lib/config/index.js#L405-L410

amychisholm03: Ah, are you suggesting adding `case 'distributed_tracing.sampler.adaptive_sampling_target'`? I'm not sure that this value is going to come from the server.

jsumners-nr: I am saying:

```js
Object.defineProperty(Config.prototype, 'sampling_target', {
	set (value) {
		this._alwaysUpdateIfChanged('target.path', value)
	}
})
```

amychisholm03: I don't think this works because that function specifies that `json` ( the first argument) is a blob sent from the collector, which is not what we are doing here. We are just trying to corroborate that when `distributed_tracing.sampler.adaptive_sampling_target` changes from the config that the agent should reassign `sampling_target` in its active sampler.

jsumners-nr: I think the idea of using the getter/setter is a good one, but we need to account for the setting coming from the server. The `_alwaysUpdateIfChanged` function is used to process the configuration coming from the server and, uh, update the local value if it was changed on the server. So I think we are going to have to go the more arduous route and update all of the references to point to the new configuration path.

amychisholm03: However, we need to keep `sampling_target` support with the server, according to the spec: "Upon agent connect, the connect response **MUST** provide the value of `sampling_target` based on this configuration setting's value. The `sampling_target` value from the connect response **SHOULD** be used as the sampling target value for adaptive sampling in the agent."

jsumners-nr: I suggest asking in the agent dev channel if this should be updated in any way.

amychisholm03: This, as in the PR? I looked at Java's implementation (their PR 2471), and it seems that they're doing something similar by sending and receiving the value as `sampling_target` from the server.

jsumners-nr: The spec. Should the spec be updated now that the parameter is under a different path?

Issue #3042, opened 2025-04-21
[AIM] Token calculation support in agent


kmudduluru:
Refer to this notes for more details. Waiting on agent specs.

kmudduluru:
setup time with josh

mrickard:
Still awaiting a spec. There's a draft PR for Gemini-specific token count tracking [here](https://github.com/newrelic/newrelic-python-agent/pull/1377). 

@bizob2828 has pointed out that the Node agent previously tracked token counts for OpenAI and Langchain, though they were removed as per [this issue](https://github.com/newrelic/node-newrelic/issues/2057).

kmudduluru:
some of this logic is earlier and removed it.

PR #3430, opened 2025-10-14
feat: Updated token calculation for `openai` LLM Events

svetlanabrennan:
## Description

### Token Calculation

- See "`response.usage.*` attributes: " section below for details on the token usage data we look for. 

These will be set by obtaining the token values from the tokenCallback or from the token usage metadata in the response (in this order).

   - If prompt tokens and completion tokens are not obtained from the tokenCallback or the response object, don’t set them at all.

  - use the tokenCallback to get the input tokens and set `response.usage.total_tokens`
  - For the embedding event, if total tokens is not obtained from the tokenCallback or the response object, don’t set it at all.

## How to Test

Run `npm run test`.

## Related Issues

Partially fixes issue #3042


lib/llm-events/openai/chat-completion-message.js
```diff
@@ -38,17 +40,46 @@ module.exports = class LlmChatCompletionMessage extends LlmEvent {
...
-    if (typeof tokenCB !== 'function') {
+
+    if (tokenCB) {
+      const messages = request?.input || request?.messages
+      let promptContent
+
+      if (typeof messages === 'string') {
+        promptContent = messages
+      } else {
+        promptContent = messages.map((msg) => msg.content).join(' ')
```
mrickard: This assumes that `messages` is going to be an array if it's not a string. Would it be possible for messages to be undefined or null here?

lib/llm-events/openai/chat-completion-message.js
```diff
@@ -38,17 +40,46 @@ module.exports = class LlmChatCompletionMessage extends LlmEvent {
...
+      }
+
+      let completionContent
+      if (response?.output) {
+        completionContent = response.output.map((resContent) => resContent.content[0].text).join(' ')
+      } else {
```
mrickard: Is it certain that `response.choices` will exist and be iterable if `response.output` isn't defined?

lib/llm-events/openai/chat-completion-summary.js
```diff
@@ -24,5 +26,43 @@ module.exports = class LlmChatCompletionSummary extends LlmEvent {
...
+
+    this.setTokens(agent, request, response)
+  }
+
+  setTokens(agent, request, response) {
+    const tokenCB = agent.llm?.tokenCountCallback
+
+    // Prefer callback for prompt and completion tokens; if unavailable, fall back to response data.
+    if (tokenCB) {
+      const messages = request?.input || request?.messages
```
mrickard, line 38: Similar to the chat completion message, how will `setTokens` behave if `input` and `messages` aren't in the expected form, or if `response.choices` isn't defined or iterable?

lib/llm-events/openai/embedding.js
```diff
@@ -14,9 +15,33 @@ module.exports = class LlmEmbedding extends LlmEvent {
...
+  }
+
+  setTotalTokens(agent, request, response) {
+    const tokenCB = agent.llm?.tokenCountCallback
+
+    // For embedding events, only total token count is relevant.
+    // Prefer callback for total tokens; if unavailable, fall back to response data.
+    if (tokenCB) {
+      const content = request.contents?.toString()
+      const totalTokenCount = calculateCallbackTokens(tokenCB, this['request.model'], content)
```
mrickard, line 34: How is this likely to handle content being `undefined`?  (Does that matter at all for calculating the tokens?)

lib/llm-events/openai/chat-completion-message.js
svetlanabrennan: According to their docs, for chat.completions.create, messages is required which is an array: https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages

And for response.create, the input is optional which can be a string or array. https://platform.openai.com/docs/api-reference/responses/create

lib/llm-events/openai/chat-completion-message.js
svetlanabrennan: The docs say it's returned and it's an array. https://platform.openai.com/docs/api-reference/chat/object

lib/llm-events/openai/chat-completion-summary.js
svetlanabrennan: Well according to the docs, those should exist but I can add an additional check to make sure promptContent and completionContent is not undefined before calling `setTokenUsageFromCallback`. 

How does that sound?

lib/llm-events/openai/embedding.js
svetlanabrennan: Looks like this was suppose to be `request.input` instead of `request.contents`. It's `google-genai` that has `request.contents`.

I just pushed up the fix for it. 

Request.input is required according to docs: https://platform.openai.com/docs/api-reference/embeddings/create

lib/llm-events/openai/chat-completion-summary.js
mrickard: It might be worth starting with one test case, just to verify how the agent code handles it.

test/unit/llm-events/openai/chat-completion-message.test.js
mrickard, line 158: Am I reading this right that `chatMessageEvent.token_count` would be expected to be `0` here, when the required callback is returning `4`?

svetlanabrennan: Correct. `token_count` will always be 0. This is done for backwards compatibility. And in the summary event, we look at the callback and actually use the value from the callback to set the token usage attributes on the summary event.

mrickard (approved):
👍 LGTM! (The changes look good. See below for one last thing.)

mrickard:
@svetlanabrennan Ah, this will need to be rebased against `main` to pick up the `langchain` fix.

PR #3431, opened 2025-10-14
feat: Updated token calculation for `google-genai` LLM Events

svetlanabrennan:
## Description

### Token Calculation

- See "response.usage.*` attributes" section below for details on the token usage data we look for. 

These will be set by obtaining the token values from the tokenCallback or from the token usage metadata in the response (in this order).

   - If prompt tokens and completion tokens are not obtained from the tokenCallback or the response object, don’t set them at all.

  - use the tokenCallback to get the input tokens and set as `response.usage.total_tokens`
  - For the embedding event, if total tokens is not obtained from the tokenCallback or the response object, don’t set it at all.

## How to Test

Run `npm run test`.

## Related Issues

Partially fixes issue #3042

### DIFF
--- a/lib/llm-events/google-genai/chat-completion-message.js
+++ b/lib/llm-events/google-genai/chat-completion-message.js
@@ -31,9 +31,8 @@ module.exports = class LlmChatCompletionMessage extends LlmEvent {
     const responseText = response?.text ?? response?.candidates?.[0]?.content?.parts?.[0]?.text
     this.is_response = responseText === message?.parts?.[0]?.text
 
-    const content = this.is_response ? message?.parts?.[0]?.text : message
     if (agent.config.ai_monitoring.record_content.enabled === true) {
-      this.content = content
+      this.content = this.is_response ? message?.parts?.[0]?.text : message
     }
 
     this.setTokenCount(agent, request, response)
--- a/test/unit/llm-events/google-genai/common.js
+++ b/test/unit/llm-events/google-genai/common.js
@@ -18,9 +18,9 @@ const res = {
     }
   ],
   usageMetadata: {
-    promptTokenCount: '10',
-    candidatesTokenCount: '20',
-    totalTokenCount: '30'
+    promptTokenCount: 10,
+    candidatesTokenCount: 20,
+    totalTokenCount: 30
   }
 }
 


lib/llm-events/utils.js
```diff
@@ -0,0 +1,110 @@
+/*
```
svetlanabrennan, line 1: Note: This is the same file as this in PR:https://github.com/newrelic/node-newrelic/pull/3430/files#diff-d2cde635cbccfcbf23c4e9a87a19ee85e1e80ce57198a7284fc837cfe86975f9 so I'll rebase this PR once that other PR is merged.

test/unit/llm-events/google-genai/common.js
```diff
@@ -18,9 +18,9 @@ const res = {
     }
   ],
   usageMetadata: {
-    promptTokenCount: 10,
-    candidatesTokenCount: 20,
-    totalTokenCount: 30
+    promptTokenCount: '10',
```
mrickard: These are coming in as string values? Tricky.

lib/llm-events/google-genai/chat-completion-message.js
```diff
@@ -29,18 +31,47 @@ module.exports = class LlmChatCompletionMessage extends LlmEvent {
     const responseText = response?.text ?? response?.candidates?.[0]?.content?.parts?.[0]?.text
     this.is_response = responseText === message?.parts?.[0]?.text
 
+    const content = this.is_response ? message?.parts?.[0]?.text : message
```
mrickard: Considering that we're parsing these ungainly `message?.parts?.[0]?.text` in a couple places, would it make sense to do it once to define a constant and reference that constant?

lib/llm-events/utils.js
mrickard: This should be ready to be rebased.

lib/llm-events/google-genai/chat-completion-message.js
svetlanabrennan: I was previously passing content to setTokenCount but that's not relevant anymore so I can undue this change and leave it how it was before.

test/unit/llm-events/google-genai/common.js
svetlanabrennan: These should have been integers. It's fixed now.

mrickard (approved):

PR #3432, opened 2025-10-15
docs: documented how to use the subscriber based instrumentation

### LINKED ISSUE #3295: Subscriber base classes need documentation
https://github.com/newrelic/node-newrelic/blob/436c3c9b0e4263d5c5add1882a3b92166710cdca/lib/subscribers/db.js#L10

^ as an example.

-----

I am looking in to what it will take for #3289, and I am having difficulty understanding which parts I will need to use and how to use those parts. I think it would be easier to understand if the base classes were extensively documented with jsdoc.

For further illustration of this issue, what is the difference between a `DbSubscriber` and a `DbQuerySubscriber`? Which one, if either, should be used for instrumenting a traditional DBMS engine?

I would expect to see documentation like the following on all of the baseline subscriber classes:

```js
/**
 * DbQuerySubscriber provides a base class for instrumenting
 * database engines like `MySQL` and `PostgreSQL`. See
 * {@link DbOperationSubscriber} for key-value database engines
 * and {@link Whatever} for document store engines.
 *
 * Probably some more text here that provides pointers for
 * implementing new instrumentations based on this class.
 * Maybe link to a particularly easy to understand specific
 * implementation as an example to follow.
 */
class DbQuerySubscriber extends DbSubscriber {}
```

amychisholm03:
This PR closes #3295. Adds a README.md in `lib/subscribers`, briefly describing how subscribers work and how to implement them. Adds further comments to most Subscriber classes.

### DIFF
--- a/lib/subscribers/README.md
+++ b/lib/subscribers/README.md
@@ -34,7 +34,7 @@ const config = {
   instrumentations: [
     {
       /**
-       * The channelName must be prefixed with `nr_` and must contain at least the expressionName or methodName.
+       * By convention, we prefix channelNames with `nr_` and include at least the expressionName or methodName.
        * It could also contain the moduleName or className to further differentiate between subscribers.
        */
       channelName: 'nr_functionName',


lib/subscribers/README.md
```diff
@@ -0,0 +1,101 @@
...
+```javascript
+// in lib/subscribers/<package_name>/config.js
+
+const config = {
+  path: './<package_name>/<subscriber_name>.js',
+  instrumentations: [
+    {
+      /**
+       * The channelName must be prefixed with `nr_` and must contain at least the expressionName or methodName.
+       * It could also contain the moduleName or className to further differentiate between subscribers.
```
jsumners-nr: I'd say "by convention, we prefix channel names with `nr_` and ....". The "must" here implies some magic that will happen due to the names which is not fully explained. Noting that the naming is merely convention allows the reader to accept it without making them wonder what other docs they are missing.

jsumners-nr:
I'm merging this. We can alter/improve it as we go.

PR #3433, opened 2025-10-16
chore: Fixed misspellings in code comments, method and variable names

brnhensley:
## Description

I saw some typos in the `default.js` that I was going to correct and then I realized it's better to clean up typos like this in one simple PR instead of many small ones.  Soooo I checked the rest of the repo and this became much larger than anticipated.

Most of these typos are in markdown, comments, test names/messages or log strings; there are just a couple that touch functional code.

There are only a couple changes of note.
1. `test/lib/cross_agent_tests/README.md` said "_These tests cover parsing of Docker container IDs from `/proc/*/mountinfop` on Linux hosts._" which I think was a typo, as far as I can tell `mountinfop` should be `mountinfo`
2. Two tests had a `recod_sql` setting, which was probably non-functional but I wanted to point it out in case it was intended
3. `default.js` had a `logging.diagnostics.deafult: false` setting, this was the first typo I wanted to change and James said it's a safe correction for me to make


amychisholm03:
Thanks for doing this @brnhensley! We could probably create a ticket to add a spell checker to our `eslint` to limit our typos in the future.

amychisholm03 (approved):

Issue #3386, opened 2025-09-23
Add configuration for ratio based sampling


jsumners-nr:
We are working toward supporting alternate sampling algorithms. Some algorithms may require extra configuration. Spec PR 740 details the new configuration we must support.

Basically, we need to update the `sampler` section of this configuration block:

https://github.com/newrelic/node-newrelic/blob/b79be723abd29a81c9fa2b635118de44be78c69b/lib/config/default.js#L1125-L1189

The result should be that we support:

+ `distributed_tracing.sampler.root` (**string** or **object**): Possible string values are `default`, `adaptive_sampling` (the same as "default"), `always_on`, and `always_off`. When set to an object, the key name must be `trace_id_ratio_based` and the object value must support the key `raio` (floating point in the range `(0.0, 1.0]`).
+ `distributed_tracing.sampler.remote_parent_sampled`: Same as `.root`.
+ `distributed_tracing.sampler.remote_parent_not_sampled`: Same as `.root`.

-----

It's not clear to me how we will support this string or object configuration value. Our configuration parser/builder utilizes a bespoke schema system that does not currently support a JSON Schema-style "oneOf" mechanic. The basic case of reading a `newrelic.js` file with the configuration should not be a problem. But when setting configuration from the environment, this could be tricky. Consider if the ratio value should be set from the environment. We would expect to see an environment variable named `NEW_RELIC_DISTRIBUTED_TRACING_SAMPLER_ROOT_TRACE_ID_RATIO_BASED_RATIO` (with value `0.1`). This key value is meant to be parsed by our builder into the configuration:

```
distributed_tracing:
  sampler:
    root:
      trace_id_ratio_based:
        ratio: 0.1
```

I attempted to research what we would need to do in order to support this so that I could write up in this issue, but I couldn't quite sort it out and it started feeling a lot like I was just doing the work. So I think this issue is going to mostly be around how we support this new configuration scenario.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-466398

amychisholm03:
Didn't auto close for some reason

PR #3438, opened 2025-10-20
feat: Added configuration for ratio based sampling for distributed tracing

svetlanabrennan:
## Description

Updated our defaults to accept string or object for distributed samplers: `root`, `remote_parent_sampled`, and `remote_parent_not_sampled`. When a user sets the sampler to `trace_id_ratio_based` object via env, the user must set two env vars like this:

```
NEW_RELIC_DISTRIBUTED_TRACING_SAMPLER_ROOT: 'trace_id_ratio_based'
NEW_RELIC_DISTRIBUTED_TRACING_SAMPLER_ROOT_TRACE_ID_RATIO_BASED_RATIO: '0.1'
```

There is no default value for ratio; if `trace_id_ratio_based` is configured and ratio is not configured, the agent logs an error and falls back to using default for sampling.

## How to Test

Add unit tests. 

## Related Issues

Fixes issue #3386

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -1105,7 +1105,7 @@ Config.prototype._buildTraceIdRatioSamplers = function _buildTraceIdRatioSampler
   const samplers = Object.keys(samplerConfig)
 
   // user can set multiple samplers
-  samplers.forEach((sampler) => {
+  for (const sampler of samplers) {
     if (this._isTraceIdRatioBasedConfig(sampler, samplerConfig)) {
       const ratioValue = samplerConfig[sampler].trace_id_ratio_based.ratio
       if (ratioValue) {
@@ -1119,7 +1119,7 @@ Config.prototype._buildTraceIdRatioSamplers = function _buildTraceIdRatioSampler
         logger.error('Not setting trace id ratio based sampler on %s.', samplers)
       }
     }
-  })
+  }
 }
 
 /**


lib/config/index.js
```diff
@@ -1163,6 +1163,25 @@ function setFromEnv({ config, key, envVar, formatter, paths }) {
   }
 }
 
+function setTraceIDRatioSampler({ config, envVar, formatter }) {
```
jsumners-nr: Please add a jsdoc block here that describes the expected inputs. It will help me understand a bit of the code.

lib/config/index.js
```diff
@@ -1163,6 +1163,31 @@ function setFromEnv({ config, key, envVar, formatter, paths }) {
   }
 }
 
+/**
+ * @param {object} params object passed to fn
+ * @param {object} params.config agent config
+ * @param {string} params.envVar name of env var
```
jsumners-nr: What sort of strings is this function expecting for the value? For example:
```suggestion
 * @param {string} params.envVar Name of environment variable that contains
 * the value for the setting. It should look like `a = b` or `{"foo":"bar"}`.
```

jsumners-nr: I wasn't trying to suggest adding my direct text as a solution. I should not have have used the "suggest" shortcut. I am trying to understand what this function is looking for in order to properly parse the value. It is a mystery to me what the `prefix` and `suffix` lines are trying to accomplish. These are the sort of algorithms that we need to document clearly so that other engineers can easily see how they are supposed to work.

lib/config/index.js
```diff
@@ -1060,10 +1060,13 @@ Config.prototype._fromPassed = function _fromPassed(external, internal, arbitrar
 
   for (const key of Object.keys(external)) {
     // if it's not in the defaults, it doesn't exist
-    if (!arbitrary && !(key in internal)) {
+    if (!arbitrary && (typeof internal === 'object') && !(key in internal)) {
```
svetlanabrennan: Added this check for object because this was causing an error for this [test](https://github.com/newrelic/node-newrelic/pull/3438/files#diff-0f128976da140f1bba4eb1991e863bebb9dc47d908cef65bc6b378eba2065d5dR288) when wrong key is used for trace id ratio based. 

Error was: `Cannot use 'in' operator to search for 'trace_id_ratio_based' in default`. This happens because we set the sampler back to default if trace id ratio based doesn't have the `ratio` key and the iteration continues.

lib/config/index.js
```diff
@@ -1087,6 +1090,52 @@ Config.prototype._fromPassed = function _fromPassed(external, internal, arbitrar
...
+      const ratioValue = samplerConfig[sampler].trace_id_ratio_based.ratio
+      if (ratioValue) {
+        internal[key][sampler] = {
+          trace_id_ratio_based: {
+            ratio: ratioValue
+          }
+        }
+        logger.trace('Setting trace id ratio based sampler on %s', samplers)
+      } else {
+        logger.error('Not setting trace id ratio based sampler on %s.', samplers)
```
svetlanabrennan, line 1119: No need to set sampler back to default since it's already set to that in `internals`

lib/config/index.js
jsumners-nr:
```suggestion
    if (!arbitrary && Object.hasOwn(internal, key) === false) {
```

lib/config/index.js
```diff
@@ -1087,6 +1090,54 @@ Config.prototype._fromPassed = function _fromPassed(external, internal, arbitrar
...
+ */
+Config.prototype._buildTraceIdRatioSamplers = function _buildTraceIdRatioSamplers(key, internal, samplerConfig) {
+  if (key !== 'sampler') {
+    return
+  }
+
+  const samplers = Object.keys(samplerConfig)
+
+  // user can set multiple samplers
+  samplers.forEach((sampler) => {
```
jsumners-nr: We recently migrated all of our `.forEach` instances to traditional loops. We should do the same here, e.g.:

```js
for (const sampler of samplers) {
}
```

PR #3445, opened 2025-10-21
feat: Updated token calculation for `aws-bedrock` LLM Events

svetlanabrennan:
## Description

### Token Calculation

- See "`response.usage.*` attributes: " section below for details on the token usage data we look for. 

These will be set by obtaining the token values from the tokenCallback or from the token usage metadata in the response object or response headers (in this order).

   - If prompt tokens and completion tokens are not obtained from the tokenCallback or the response object or the response headers, don’t set them at all.

  - use the tokenCallback to get the input tokens and set `response.usage.total_tokens`
  - For the embedding event, if total tokens is not obtained from the tokenCallback or the response object or the response headers, don’t set it at all.

### Additional changes
I updated the stream handlers for Claude3 and Llama in order to get the token usage (input and output) from the stream and add to the currentBody before returning the response. The other handlers like Titan already has the token usage in the parsed event so no additional logic had to be added there. 

## How to Test

Run `npm run test`.

## Related Issues

Partially fixes issue #3042

### DIFF
--- a/lib/llm-events/aws-bedrock/utils.js
+++ b/lib/llm-events/aws-bedrock/utils.js
@@ -83,7 +83,7 @@ function setUsageTokens(response, context) {
       Number(response?.usage?.input_tokens ||
        response?.usage?.inputTokens ||
         response?.parsedBody?.results?.[0]?.inputTextTokenCount ||
-         response?.parsedBody?.usage.input_tokens)
+         response?.parsedBody?.usage?.input_tokens)
     const completionTokens =
         Number(response?.usage?.output_tokens ||
          response?.usage?.outputTokens ||


test/unit/llm-events/aws-bedrock/chat-completion-message.test.js
```diff
@@ -136,3 +139,82 @@ test('should not capture content when `ai_monitoring.record_content.enabled` is
...
+  const event = new LlmChatCompletionMessage(t.nr)
+  assert.equal(event.token_count, 0)
+})
+
+test('should not set token_count if response prompt and completion content is undefined', async (t) => {
+  t.nr.bedrockCommand.prompt = undefined
+  t.nr.bedrockResponse.completions = undefined
+  const event = new LlmChatCompletionMessage(t.nr)
+  assert.equal(event.token_count, undefined)
+})
```
mrickard: Should there be test cases for when prompt or completion tokens are defined?

lib/llm-events/aws-bedrock/chat-completion-message.js
```diff
@@ -63,6 +62,35 @@ class LlmChatCompletionMessage extends LlmEvent {
       this.id = `${this.bedrockResponse.id || this.id}-${index}`
     }
   }
+
+  setTokenCount(agent) {
+    const tokenCB = agent.llm?.tokenCountCallback
```
mrickard: Previously (old line 44) this was more defensive, assuming that `agent` may not be defined. Should this declaration retain that defensiveness?

lib/llm-events/aws-bedrock/chat-completion-summary.js
```diff
@@ -36,6 +38,33 @@ class LlmChatCompletionSummary extends LlmEvent {
     this['response.choices.finish_reason'] = res.finishReason
     this['request.temperature'] = cmd.temperature
     this['response.number_of_messages'] = (cmd.prompt.length ?? 0) + (res.completions.length ?? 0)
+
+    this.setTokens(agent)
+  }
+
+  setTokens(agent) {
+    const tokenCB = agent.llm?.tokenCountCallback
```
mrickard: Similar to the above comment, do we want to defend against agent being undefined here?

lib/llm-events/aws-bedrock/embedding.js
```diff
@@ -21,18 +22,50 @@ class LlmEmbedding extends LlmEvent {
...
     this.duration = params.segment.getDurationInMillis()
 
-    // Even if not recording content, we should use the local token counting callback to record token usage
-    if (typeof tokenCb === 'function') {
-      this.token_count = tokenCb(this.bedrockCommand.modelId, input)
+    this.setTotalTokens(agent, input)
+  }
+
+  setTotalTokens(agent, input) {
+    const tokenCB = agent.llm?.tokenCountCallback
```
mrickard: Should we prefer `agent?.llm?.tokenCountCallback` here?

lib/llm-events/aws-bedrock/utils.js
```diff
@@ -67,7 +69,63 @@ function stringifyConverseChunkedMessage(chunks) {
...
+ */
+function setUsageTokens(response, context) {
+  // input and output token counts must available in order to add all usage attributes to response
+  // if total tokens is not available, we can manually add it up (from input and output token count)
+  if (tokenUsageAttributesExist(response)) {
+    const promptTokens =
+      Number(response?.usage?.input_tokens ||
+       response?.usage?.inputTokens ||
+        response?.parsedBody?.results?.[0]?.inputTextTokenCount ||
+         response?.parsedBody?.usage.input_tokens)
```
mrickard: Given the variety in how these are defined, we're on the hook for *some* kind of complex conditional somewhere. I wonder if these could be broken out into helper functions, though. And we may want to allow for `response.usage` to be undefined in some of these.

lib/llm-events/aws-bedrock/chat-completion-message.js
svetlanabrennan: Yep. Good catch. I'll fix that.

lib/llm-events/aws-bedrock/chat-completion-summary.js
svetlanabrennan: Yes I'll get that fixed. Thanks

lib/llm-events/aws-bedrock/utils.js
svetlanabrennan: Do you mean something like:

`const promptTokens = Number(getPromptTokens)`

And move the prompt token logic to getPromptTokens?

svetlanabrennan: `response?.usage?` does allow it to be undefined. Can you clarify what you mean here?

mrickard: @svetlanabrennan `response?.usage?` is the right approach. I should have pointed out the line with `response?.parsedBody?.usage.output_tokens`.

mrickard:
```suggestion
         response?.parsedBody?.usage?.input_tokens)
```

svetlanabrennan: Oh got it. I committed your suggestion.

PR #3527, opened 2025-11-19
refactor: Break samplers into classes and store them on `agent.sampler.*`

amychisholm03:
## Description

Fixes some sampling-related bugs by better encapsulating `remote_parent_sampled` and `remote_parent_not_sampled` sampling decisions with `agent.sampler.remoteParentSampled` and `agent.sampler.remoteParentNotSampled`. Also, replaces `agent.transactionSampler` with `agent.sampler.root`, so quite a few tests were modified to account for this.

Introduces the concept of a `Sampler` interface with exposes `applySamplingDecision`, which updates a given `Transaction`'s `priority` and `sampled` flags.

## How to Test

```
npm run unit
npm run integration
```

## Related Issues

Precursor for core tracing and `TraceIdRatioBasedSampler` work.

### DIFF
--- a/lib/samplers/README.md
+++ b/lib/samplers/README.md
@@ -14,7 +14,7 @@ Customers configure how they would like their transactions to be sampled under o
   - `remote_parent_sampled`: The sampler for when the upstream service has sampled the trace.
   - `remote_parent_not_sampled`: The sampler for when the upstream service has not sampled the trace.
 
-NOTE: `distributed_tracing.sampler` only exists for backward compatiability and will be deprecated in favor of `distributed_tracing.sampler.full_granularity`. For now, `full_granularity` will take precedence over the old path.
+NOTE: `distributed_tracing.sampler` only exists for backward compatiability and may be deprecated in favor of `distributed_tracing.sampler.full_granularity`. For now, `full_granularity` will take precedence over the old path.
 
 ### Full Config in Accordance to Spec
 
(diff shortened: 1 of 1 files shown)


lib/samplers/README.md
```diff
@@ -0,0 +1,131 @@
...
+`agent.sampler` would be defined as:
+
+* `agent.sampler.full_granularity.root`
+* `agent.sampler.full_granularity.remote_parent_sampled`
+* `agent.sampler.full_granularity.remote_parent_not_sampled`
+* `agent.sampler.partial_granularity.root`
+* `agent.sampler.partial_granularity.remote_parent_sampled`
+* `agent.sampler.partial_granularity.remote_parent_not_sampled`
+
+Will be deprecated, `full_granularity` takes precedence:
```
svetlanabrennan: This hasn't been decided yet. The spec says to use both but `agent.sampler.full_granularity.*` takes precedence over `agent.sampler.*`.

lib/samplers/README.md
```diff
@@ -0,0 +1,131 @@
...
+* `agent.sampler.root`
+* `agent.sampler.remote_parent_sampled`
+* `agent.sampler.remote_parent_not_sampled`
+
+`Transaction.prototype._calculatePriority` would be modified like:
+
+```javascript
+...
+ // Decide sampling from w3c data
+  let full_sampler = null
```
svetlanabrennan: Should we make these camel case since that's the preferred way?

lib/agent.js
```diff
@@ -848,10 +850,10 @@ Agent.prototype._listenForConfigChanges = function _listenForConfigChanges() {
     self.txSegmentNormalizer.load.apply(self.txSegmentNormalizer, arguments)
   })
   this.config.on('sampling_target', function updateSamplingTarget(target) {
-    self.transactionSampler.samplingTarget = target
+    self.sampler.root.samplingTarget = target
```
svetlanabrennan, line 853: So the transactionSampler is always the root sampler?

amychisholm03: Yes, it was.

lib/samplers/README.md
amychisholm03: This is more pseudo-codey than anything else, but I can make that change

lib/samplers/README.md
amychisholm03: When I mean "deprecated", I'm referring to the `agent.sampler.root`, `agent.sampler.remoteParentSampled`, and `agent.sampler.remoteParentNotSampled` fields on the agent, not the config. In the future, it will be `agent.sampler.fullGranularity.root`, etc., which assumes the precedence logic of `sampler.fullGranularity` > `sampler` handled by the config initialization.

lib/agent.js
amychisholm03: Though this `updateSamplingTarget` function does reveal another thing we will have to resolve with the sharing of `AdaptiveSampler` states, but that's out of scope for this PR.

Issue #3530, opened 2025-11-21
Update references for full granularity


svetlanabrennan:
According to the spec, if a user sets values for any samplers under `distributed_tracing.sampler.full_granularity.*`, those take precedences over samplers set under `distributed_tracing.sampler.*`. 

If a user left any `distributed_tracing.sampler.full_granularity.*` samplers options as `default`, then we use the values from `distributed_tracing.sampler.*`.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-490840

svetlanabrennan:
Closing this PR. We got consensus that we will use the global path for the sampler as the full granularity sampler. So use `distributed_tracing.sampler.*` and remove the `distributed_tracing.sampler.full_granularity.*` paths (I'll open a separate issue for this).

PR #3531, opened 2025-11-21
feat: Updated global and full granularity samplers based on precedence

svetlanabrennan:
## Description

Updated the config setup logic to reconfigure the full granularity vs global (aka the top level samplers) based on precedence. 

Both the global sampler and the full granularity sampler are suppose to be the same thing: be the full granularity sampler. 

-If both global and full granularity samplers are default, do nothing. 
-If the global sampler has a non default value and the full granularity sampler has a default value, set the full granularity sampler to match the global sampler value and log a warning. 
-If the global sampler has a non default value and the full granularity sampler has a non default value as well, set the global sampler to match the full granularity sampler and log a warning. 

Note: one of these samplers (global or full granularity) might be removed. 

## How to Test

Run `npm run unit`

## Related Issues

Closes issue #3530

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -200,6 +200,9 @@ function Config(config) {
   // 10. Setup labels for both `collector/facts` and application logging
   this.parsedLabels = parseLabels(this.labels, logger)
   this.loggingLabels = this._setApplicationLoggingLabels()
+
+  // 11. Reconfigure DT samplers based on precedence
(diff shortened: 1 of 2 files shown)


svetlanabrennan:
Closing PR as this is no longer needed since all agents decided to use the global path `distributed_tracing.sampler.*` as the definition for full granularity.

PR #3552, opened 2025-12-03
chore: Improve GCP utilization logging

### LINKED ISSUE #3548: Improve GCP utilization logging
NR-478622 is proving difficult to diagnose without better logging around determining the "host" name for the container. We need to verify that all error paths are logging a useful trace message, and that the success path(s) are logging at least an info message of the result.

jsumners-nr:
This PR resolves #3548.

Originally, I was going to add more logging to `gcp-info.js`. But after refactoring the `getVendors` function, I ended up with better logging for _all_ vendors. That is, all vendors now log their success state with the result data.


lib/utilization/index.js
```diff
@@ -17,24 +17,40 @@ const VENDOR_METHODS = {
   kubernetes: require('./kubernetes-info'),
   pcf: require('./pcf-info')
 }
-const VENDOR_NAMES = Object.keys(VENDOR_METHODS)
 
-module.exports.getVendors = getVendors
-function getVendors(agent, callback) {
-  let done = 0
+module.exports.getVendors = async function getVendors(
```
jsumners-nr, line 21: By refactoring this function into an async function, as it was already being used via promisify, we get clearer agent logs. Vendors will be resolved serially, so the logs will be sequential instead of interleaved.

test/unit/utilization/main.test.js
```diff
@@ -4,13 +4,15 @@
  */
 
 'use strict'
-const test = require('node:test')
+
+const { describe, test } = require('node:test')
```
jsumners-nr, line 8: Supposedly, using `describe` gives us some extra benefit: https://github.com/pinojs/pino-pretty/pull/569#discussion_r2083493061

I'm not seeing it with the below test. But it does obviate the need to `await` every sub-test. 🤷‍♂️ I can revert it if we don't like it. I really don't see the need for the nesting to begin with.

<details>
<summary>Test</summary>

```js
'use strict'

const { describe, test } = require('node:test')

test('test parent', async t => {
  await t.test('should be skipped', (t) => {
    t.assert.ok(true)
  })

  await t.test('should not be skipped', { only: true }, (t) => {
    t.assert.ok(true)
  })
})

describe('describe parent', () => {
  test('should be skipped', (t) => {
    t.assert.ok(true)
  })

  test.only('should not be skipped', (t) => {
    t.assert.ok(true)
  })
})
```

</details>

PR #3589, opened 2025-12-10
refactor: Updated `bunyan` instrumentation to subscribe to events emitted

### LINKED ISSUE #3586: Bunyan log forwarding does not occur when you extend Bunyan to create logger
## Description

Bunyan log forwarding does not occur when you extend Bunyan to create logger

## Expected Behavior
Log forwarding occurs on logger that is extended from Bunyan

## Steps to Reproduce
```
const Bunyan = require('bunyan');
class Logger extends Bunyan {}
const logger = new Logger({ name: 'empty-extends', level: 'info' });
console.log('test line');
```

## Addtionial Context
This is because we only instrument `bunyan.createLogger`, not Bunyan as a constructor. Since bunyan.createLogger calls the bunyan constructor, we should be able to update instrumentation to appease both cases.

bizob2828:
## Description

A customer reported that we were not forwarding logs when extending Bunyan. This is because we only instrumented `bunyan.createLogger` which returns a logger instance which we added our stream for forwarding.  It was too difficult to instrument the bunyan export because what's getting exported is a copy of `function Logger` which is not the one being used when you do `class CustomLogger extends Bunyan`. Instead of getting creative with our traditional monkey patching, I migrated the instrumentation to tracing channel, which makes this easy because we can rewrite the `function Logger` which is used both when you call `bunyan.createLogger` or `class CustomLogger extends Bunyan`. I added tests for each appilcation logging mode: metrics, decorating and forwarding with a custom logger. I also added more assertions to ensure the stream is writing the original log messages.

## How to Test

```sh
npm run versioned:internal bunyan
```

## Related Issues
Closes #3586


lib/subscribers/bunyan/base.js
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright 2025 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+const ApplicationLogsSubscriber = require('../application-logs')
+
+class BunyanBaseSubscriber extends ApplicationLogsSubscriber {
+  NAME_FROM_LEVEL = {
```
bizob2828, line 10: in our previous instrumentation we got this constant from the bunyan export, this is no longer possible. I also don't see this changing as we just need a mapping between integers and level(as string)

jsumners-nr: It's unlikely to ever change. The maintainer doesn't want to label the library deprecated or decommission it, but maintenance on it is effectively done.

jsumners-nr (approved):
Looks good to me.

PR #3594, opened 2025-12-11
feat: Add `entity_guid` to Agent Control health file

### LINKED ISSUE #3591: Add `entity_guid` to agent control health file
### Description
[Agent specs PR 742](https://source.datanerd.us/agents/agent-specs/pull/752) requires us to add `entity_guid` to the agent control health file. This is a requirement for the Feb launch.

### Acceptance Criteria

- [ ] `entity_guid` is added to the [`HealthReporter` class](https://github.com/newrelic/node-newrelic/blob/main/lib/health-reporter.js)
- [ ] relevant tests are created

amychisholm03:
## Description

To facilitate establishing relationships between fleet control agent entities and the related APM service that an agent creates, the entity guid will now be part of the health file written out for agent control. Note that this new field MUST be the first entry in the health file.

From the spec:
"| entity_guid           | string    | The **primary** entity GUID received in the connect response. This **MUST** be the first field in the health file  |

This format is based on the [OpAMP protocol](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#componenthealth-message). The entity
GUID is the first field in the file so as not to break the OpAmp compatibility in the remainder of the file.

For a health file written out prior to the entity GUID being available, the agent **should** include the `entity_guid` key with an empty value."

## How to Test

```
npm run unit
```

## Related Issues

Closes #3591

### DIFF
--- a/lib/collector/api.js
+++ b/lib/collector/api.js
@@ -431,9 +431,6 @@ CollectorAPI.prototype._onConnect = function _onConnect(callback, error, res) {
 
   // pass configuration data from the API so automatic reconnect works
   agent.reconfigure(config)
-  if (config.entity_guid) {
-    agent.healthReporter.setEntityGuid(config.entity_guid)
-  }
 
   callback(null, res)
 }
--- a/lib/health-reporter.js
+++ b/lib/health-reporter.js
@@ -65,8 +65,8 @@ function directoryAvailable(dest) {
  * by the environment.
  */
 class HealthReporter {
+  #agentConfig = {}
   #enabled = false
-  #entityGuid = ''
   #status = HealthReporter.STATUS_HEALTHY
   #interval
   #destFile
@@ -122,6 +122,7 @@ class HealthReporter {
     logger = defaultLogger,
     setInterval = global.setInterval
   } = {}) {
+    this.#agentConfig = agentConfig
     const enabled = agentConfig?.agent_control?.enabled
     const checkInterval = parseInt(agentConfig?.agent_control?.health?.frequency, 10) * 1_000
     let outDir = agentConfig?.agent_control?.health?.delivery_location
@@ -163,7 +164,7 @@ class HealthReporter {
     const healthy = this.#status === HealthReporter.STATUS_HEALTHY
     writeStatus({
       file: this.#destFile,
-      entityGuid: this.#entityGuid,
+      entityGuid: this.#agentConfig.entity_guid || '',
       healthy,
       startTime: this.#startTime,
       code: this.#status,
@@ -182,20 +183,6 @@ class HealthReporter {
 
   get destFile() {
     return this.#destFile
-  }
-
-  /**
-   * Update the entity GUID. This entity GUID will be written at
-   * the top of the health file on the next interval.
-   *
-   * @param {string} entityGuid to update
-   */
-  setEntityGuid(entityGuid) {
-    if (this.#enabled === false) {
-      return
-    }
-
-    this.#entityGuid = entityGuid
   }
 
   /**
@@ -258,7 +245,7 @@ class HealthReporter {
 
     writeStatus({
       file: this.#destFile,
-      entityGuid: this.#entityGuid,
+      entityGuid: this.#agentConfig.entity_guid || '',
       startTime: this.#startTime,
       healthy,
       code,
--- a/test/unit/lib/health-reporter.test.js
+++ b/test/unit/lib/health-reporter.test.js
@@ -277,39 +277,6 @@ test('stop logs writing error', async (t) => {
   await plan.completed
 })
 
-test('setEntityGuid updates entity_guid in output', async (t) => {
-  const plan = tspl(t, { plan: 3 })
-  fs.writeFile = (dest, data, options, callback) => {
-    plan.match(dest, /health-\w{32}\.yaml/)
-    plan.equal(
-      data,
-      [
-        'entity_guid: test-entity-guid-123',
-        'healthy: true',
-        "status: 'Healthy.'",
-        'last_error: NR-APM-000',
-        'start_time_unix_nano: 1000000',
-        'status_time_unix_nano: 2000000'
-      ].join('\n')
-    )
-    callback()
-  }
-
-  function localInterval(method) {
-    // Don't call immediately - let test control timing
-    setImmediate(() => method.call())
-    return {
-      unref() {}
-    }
-  }
-
-  const reporter = new HealthReporter({ ...t.nr, setInterval: localInterval })
-  reporter.setEntityGuid('test-entity-guid-123')
-  plan.ok(reporter)
-
-  await plan.completed
-})
-
 test('entity_guid is empty string by default', async (t) => {
   const plan = tspl(t, { plan: 3 })
   fs.writeFile = (dest, data, options, callback) => {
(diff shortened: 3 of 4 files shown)


lib/collector/api.js
```diff
@@ -431,6 +431,9 @@ CollectorAPI.prototype._onConnect = function _onConnect(callback, error, res) {
 
   // pass configuration data from the API so automatic reconnect works
   agent.reconfigure(config)
+  if (config.entity_guid) {
+    agent.healthReporter.setEntityGuid(config.entity_guid)
+  }
```
jsumners-nr: This seems out of place. We handle configuration updates in `lib/config/index.js`. In particular, for the `entity_guid` it is:

https://github.com/newrelic/node-newrelic/blob/898feb9de7bd691992dff1d63467599d18b57348/lib/config/index.js#L596-L599

That should be updated to emit the change event as shown in:

https://github.com/newrelic/node-newrelic/blob/898feb9de7bd691992dff1d63467599d18b57348/lib/config/index.js#L385-L389

Then the health reporter can register for the `change` event and set its internal value on every event.

jsumners-nr (changes requested):

lib/collector/api.js
amychisholm03: Thanks for pointing this out; I updated accordingly.

lib/health-reporter.js
```diff
@@ -154,13 +156,28 @@ class HealthReporter {
     this.#interval.unref()
 
     this.#enabled = true
+
+    // Listen for entity_guid changes from config.
+    // This entity GUID will be written at the
+    // top of the health file on the next interval.
+    if (agentConfig) {
```
jsumners-nr: We don't need this check. The reporter is initialized at:

https://github.com/newrelic/node-newrelic/blob/898feb9de7bd691992dff1d63467599d18b57348/lib/agent.js#L174-L181

Which always happens _after_ the configuration has been successfully loaded:

https://github.com/newrelic/node-newrelic/blob/898feb9de7bd691992dff1d63467599d18b57348/index.js#L97-L117

In fact, we should file a follow up issue to come back and clean up the useless optional chains in this file.

lib/health-reporter.js
```diff
@@ -154,13 +156,28 @@ class HealthReporter {
...
+      agentConfig.on('entity_guid', (entityGuid) => {
+        if (this.#enabled === false) {
+          return
+        }
```
jsumners-nr: We shouldn't need this check. The listener will not get registered unless the reporter is enabled.

lib/health-reporter.js
```diff
@@ -154,13 +156,28 @@ class HealthReporter {
     this.#interval.unref()
 
     this.#enabled = true
+
+    // Listen for entity_guid changes from config.
+    // This entity GUID will be written at the
+    // top of the health file on the next interval.
+    if (agentConfig) {
+      agentConfig.on('entity_guid', (entityGuid) => {
```
jsumners-nr: Unless I'm reading things incorrectly, there isn't an "entity_guid" event. What we want is:

```js
agentConfig.on('change', (config) => {
	this.#entityGuid = config.entity_guid
})
```

amychisholm03: Yeah, I just realized we don't need the `this.emit('change')` under the `entity_guid` case because the `entity_guid` events are emitted via `this.emit(key, val)`, so that's why it's working.

test/unit/lib/health-reporter.test.js
```diff
@@ -173,6 +174,16 @@ test('setStatus and stop do nothing if reporter disabled', (t, end) => {
...
 
+test('entity_guid event listener does nothing if reporter disabled', (t) => {
+  delete t.nr.agentConfig.agent_control.enabled
+  const reporter = new HealthReporter(t.nr)
+  assert.ok(reporter, 'reporter should be created')
+  const entityGuid = 'test-guid-should-not-be-set'
+  t.nr.agentConfig.emit('entity_guid', entityGuid)
+  assert.notEqual(reporter.entityGuid, entityGuid)
+  assert.equal(reporter.enabled, false, 'reporter should not be enabled')
+})
```
jsumners-nr: This test is extraneous now that the prior feedback has been addressed. The reporter can only be enabled or disabled at construction time. If the feature is not enabled, the listener will never be added.

test/unit/lib/health-reporter.test.js
```diff
@@ -273,3 +286,126 @@ test('stop logs writing error', async (t) => {
...
+        'start_time_unix_nano: 1000000',
+        'status_time_unix_nano: 2000000'
+      ].join('\n')
+    )
+    callback()
+  }
+
+  function localInterval(method) {
+    // Don't call immediately - let test control timing
+    setImmediate(() => method.call())
```
jsumners-nr: Is there any need to invoke the method from the timer at all for this test?

lib/health-reporter.js
```diff
@@ -154,13 +156,22 @@ class HealthReporter {
     this.#interval.unref()
 
     this.#enabled = true
+
+    // Listen for entity_guid changes from config.
+    // This entity GUID will be written at the
+    // top of the health file on the next interval.
+    agentConfig.on('entity_guid', (entityGuid) => {
+      this.#entityGuid = entityGuid
+    })
```
jsumners-nr: This event does not currently exist. I think we should just look for the `change` event and add the emission of it in the `_fromServer` block.

lib/health-reporter.js
```diff
@@ -154,13 +156,22 @@ class HealthReporter {
     this.#interval.unref()
 
     this.#enabled = true
+
+    // Listen for entity_guid changes from config.
+    // This entity GUID will be written at the
+    // top of the health file on the next interval.
+    agentConfig.on('change', (entityGuid) => {
+      this.#entityGuid = entityGuid
+    })
```
jsumners-nr: The `change` event receives the configuration object. The parameter and value assignment should match my original suggestion: https://github.com/newrelic/node-newrelic/pull/3594#discussion_r2624312162

test/unit/lib/health-reporter.test.js
amychisholm03: I noticed that if I do the `simpleInterval` like the other tests, the test finishes before the `'change'` event gets handled.

jsumners-nr: So, it does need to invoke the method. But I'd probably make this easier to reason through by doing:

```js
const interval = {
	unref() {}
	doWork() { this.method.call() }
}
function setInterval(method) {
	interval.method = method
	return interval
}

const reporter = new HealthReporter({ ...t.nr, setInterval })
// emit change event
interval.doWork()

await plan
```

lib/health-reporter.js
```diff
@@ -154,13 +156,22 @@ class HealthReporter {
     this.#interval.unref()
 
     this.#enabled = true
+
+    // Listen for entity_guid changes from config.
+    // This entity GUID will be written at the
+    // top of the health file on the next interval.
+    agentConfig.on('change', (config) => {
+      this.#entityGuid = config.entity_guid
+    })
```
jsumners-nr: Sorry about not catching this yesterday, I got hung up on just getting the event stuff "right" instead of questioning if we need it: we don't need it. The reporter is given a reference to the configuration object. That referenced object will be updated _at some point_ before a reporter interval fires. The subsequent reporter interval will then have the value of the guid by way of the configuration object. In short, all we need to do is write the value of `agentConfig.entity_guid` each and every reporter interval. As in:

```js
writeStatus({
	entityGuid: this.#agentConfig.entity_guid || ''
})
```

test/unit/lib/health-reporter.test.js
```diff
@@ -273,3 +276,130 @@ test('stop logs writing error', async (t) => {
...
+    return interval
+  }
+
+  const reporter = new HealthReporter({ ...t.nr, setInterval })
+  t.nr.agentConfig.entity_guid = 'test-entity-guid-123'
+  plan.ok(reporter)
+  interval.doWork()
+
+  await plan.completed
+})
```
jsumners-nr: This test is no longer necessary:

```suggestion
```

Issue #3476, opened 2025-10-29
Support OTEL span events


jsumners-nr:
Following on from #3474, [Span Events](https://opentelemetry.io/docs/concepts/signals/traces/#span-events) are effectively metadata on a span that we should propagate to New Relic. A span with event data will have an `.events` property set to an array of event objects:

+ https://open-telemetry.github.io/opentelemetry-js/interfaces/_opentelemetry_sdk-node.node.ReadableSpan.html#events
+ https://open-telemetry.github.io/opentelemetry-js/interfaces/_opentelemetry_sdk-node.node.TimedEvent.html

We need to update the our processing, as described in the above linked issue, to translate this event data and attach it to NR spans as so:

+ `type` (string): must be set to `SpanEvent`.
+ `timestamp` (number): milliseconds epoch relative. This must be derived from the event data, e.g. `span.events[0].time`. Our [timestamp translator](https://github.com/newrelic/node-newrelic/blob/e383c4efc7171f02aef41d6c657455338b4b7258/lib/otel/logs/normalize-timestamp.js) should be elevated to a general OTEL utility function and utilized to normalize the value.
+ `span.id` (string): the identifier of the containing `Span` (`span.context.spanId`).
+ `trace.id` (string): the trace identifier of the containing `Span` (`span.context.traceId`).
+ `name` (string): the name of the event (e.g. `span.events[0].name`).

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-482852

svetlanabrennan:
Completed.

PR #3600, opened 2025-12-12
feat: Added support for OTel span events

svetlanabrennan:
## Description

Add support for OTel span events.  We set the attributes on span.events as intrinsics attributes and the rest of the attributes are set as agent attributes (`type`, `timestamp`, `span.id`, `trace.id`, and `name`). 

Span events are set on an OTel span during the `onEnd` call so we start the processing there. 

## How to Test

Run versioned test for timed-events. 

## Related Issues

Closes issue #3476

### DIFF
--- a/lib/otel/traces/span-processor.js
+++ b/lib/otel/traces/span-processor.js
@@ -110,6 +110,15 @@ module.exports = class NrSpanProcessor {
     segment.overwriteDurationInMillis(duration)
   }
 
+  /**
+   * Iterates the span events that have been added to the span and attaches
+   * them to the `TraceSegment` in the format expected by the New Relic
+   * collector.
+   *
+   * @param {object} params Function parameters
+   * @param {object} params.span The OTEL span that may contain events.
+   * @param {object} params.segment The target segment to attach the events to.
+   */
   reconcileEvents({ segment, span }) {
     const spanContext = span.spanContext()
     for (let i = 0; i < span.events.length; i++) {


lib/otel/traces/span-processor.js
```diff
@@ -105,6 +110,16 @@ module.exports = class NrSpanProcessor {
...
   }
 
+  reconcileEvents({ segment, span }) {
+    for (let i = 0; i < span.events.length; i++) {
+      const event = new TimedEvent({
+        event: span.events[i],
+        spanContext: span.spanContext(),
+      })
+      segment.addTimedEvent(event)
+    }
```
jsumners-nr, line 130: A small optimization:

```suggestion
    const spanContext = span.spanContext()
    for (let i = 0; i < span.events.length; i++) {
      const event = new TimedEvent({
        event: span.events[i],
        spanContext
      })
      segment.addTimedEvent(event)
    }
```

lib/spans/timed-event.js
```diff
@@ -0,0 +1,52 @@
...
+class TimedEvent {
+  constructor({ event, spanContext }) {
+    this.userAttributes = new Attributes({ scope: Attributes.SCOPE_SEGMENT })
+    this.agentAttributes = new Attributes({ scope: Attributes.SCOPE_SEGMENT })
+    this.intrinsics = Object.create(null)
+
+    const timestamp = normalizeTimestamp(event.time)
+
+    this.intrinsics.type = 'SpanEvent'
+    this.intrinsics.timestamp = timestamp > 0 ? timestamp : Date
```
jsumners-nr:
```suggestion
    this.intrinsics.timestamp = timestamp > 0 ? timestamp : Date.now()
```

test/versioned/otel-bridge/timed-event.test.js
```diff
@@ -0,0 +1,186 @@
...
+      assert.equal(event.agentAttributes.attributes['request.url'].value, `${REQUEST_URL}/post`)
+      assert.equal(event.agentAttributes.attributes['custom.attribute'].value, 'test-value')
+
+      assert.equal(event.intrinsics.type, 'SpanEvent')
+      assert.equal(event.intrinsics['span.id'], httpSpanEvent.intrinsics.guid)
+      assert.equal(event.intrinsics['trace.id'], tx.traceId)
+      // OTEL sets the timestamp to a hrtime tuple. Which can't actually be
+      // converted to an epoch millisecond representation. So the best we can do
+      // is to verify that the two times are within a narrow window.
+      assert.equal(
```
jsumners-nr: This comment isn't completely accurate any longer. We can convert the tuple to an accurate timestamp. But it is still better to check the window since we can't easily set the timestamp on the event.

lib/otel/traces/span-processor.js
```diff
@@ -105,6 +110,17 @@ module.exports = class NrSpanProcessor {
     segment.overwriteDurationInMillis(duration)
   }
 
+  reconcileEvents({ segment, span }) {
```
jsumners-nr, line 122:
```suggestion
  /**
   * Iterates the span events that have been added to the span and attaches
   * them to the `TraceSegment` in the format expected by the New Relic
   * collector.
   *
   * @param {object} params Function parameters
   * @param {object} params.span The OTEL span that may contain events.
   * @param {object} params.segment The target segment to attach the events to.
   */
  reconcileEvents({ segment, span }) {
```

lib/spans/span-event.js
```diff
@@ -217,6 +217,7 @@ class SpanEvent {
       span = new SpanEvent(attributes, customAttributes)
     }
     span.spanLinks = segment.spanLinks ?? []
+    span.timedEvents = segment.timedEvents ?? []
```
jsumners-nr, line 220: We should update this subtest group with tests for events:

https://github.com/newrelic/node-newrelic/blob/9bbb0f8e52eb48f1bbaf452d6a2b21cf8b375d00/test/unit/spans/span-event.test.js#L45-L67

lib/spans/timed-event.js
```diff
@@ -0,0 +1,75 @@
...
+   */
+  constructor({ event, spanContext }) {
+    this.userAttributes = new Attributes({ scope: Attributes.SCOPE_SEGMENT })
+    this.agentAttributes = new Attributes({ scope: Attributes.SCOPE_SEGMENT })
+    this.intrinsics = Object.create(null)
+
+    const timestamp = normalizeTimestamp(event.time)
+
+    this.intrinsics.type = 'SpanEvent'
+    this.intrinsics.timestamp = timestamp > 0 ? timestamp : Date.now()
```
jsumners-nr: `normalizeTimestamp` will never return `0`. In the `SpanLink` implementation we accepted an optional `timestamp` function parameter. So we needed to set the value with the ternary. In this implementation, it's not possible to hit the `=== 0` case.

test/versioned/otel-bridge/package.json
```diff
@@ -42,6 +42,20 @@
       "files": [
         "span-links.test.js"
       ]
+    },
+
+    {
+      "engines": {
+        "node": ">=18"
```
jsumners-nr: Go ahead and update this one, and the links test, to `>= 20`.

PR #3606, opened 2025-12-15
fix: Account for links added during transaction

### LINKED ISSUE #3601: Intercept OTEL span links through `.addLink`
While implementing #3474, I missed that the `Span` object has an [`.addLink`](https://open-telemetry.github.io/opentelemetry-js/interfaces/_opentelemetry_api._opentelemetry_api.Span.html#addlink) method to allow adding of links during the lifetime of the span. The implementation for #3474 handles the links present _at creation_ of the span. While it is recommended to only add links at creation, the reality is that links can be added subsequent to creation and we need to account for those.

We need to update our span `onStart` method to intercept the `.addLink` method on the span to handle in-flight links. It will likely look something like:

```js
onStart(span) {
  span[otelSynthesis] = this.synthesizer.synthesize(span)
  const { segment } = span[otelSynthesis]
  
  const addLink = span.addLink
  span.addLink = function nrAddLink(link) {
    const nrLink = new SpanLink({ link, context: span.spanContext(), timestamp: span.startTime })
    segment.addSpanLink(nrLink)
    addLink.call(span, link)
  }

  const addLinks = span.addLinks
  span.addLinks = function nrAddLinks(links) {
    for (const link of links) { span.addLink(link) }
  }
}
```

https://github.com/newrelic/node-newrelic/blob/ae0074b6fa4caea0be26a2a34ecf4f53387bc48d/lib/otel/traces/span-processor.js#L42-L44

jsumners-nr:
This PR resolves #3601.

### DIFF
--- a/lib/otel/traces/segment-synthesis.js
+++ b/lib/otel/traces/segment-synthesis.js
@@ -6,8 +6,6 @@
 'use strict'
 
 const defaultLogger = require('#agentlib/logger.js').child({ component: 'segment-synthesizer' })
-const SpanLink = require('#agentlib/spans/span-link.js')
-const normalizeTimestamp = require('../normalize-timestamp.js')
 const { RulesEngine } = require('./rules.js')
 const {
   createConsumerSegment,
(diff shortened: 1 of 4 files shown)


lib/otel/traces/segment-synthesis.js
jsumners-nr, line 1: This is a revert to the way it was before I started on the span links.

test/unit/lib/otel/traces/segment-synthesis.test.js
jsumners-nr, line 1: Test is no longer a valid case.

PR #3608, opened 2025-12-16
feat: Added `compact` type for partial granularity traces.

### LINKED ISSUE #3454: Add support for `"compact"` partial granularity

bizob2828:
## Description

This PR adds support for compact partial traces.

The high level requirements are:
  * All spans without entity-synthesis agent attributes will be dropped from the span tree except the entry span and LLM spans.
  * `nr.pg` agent attribute is added to entry span.
  * All span agent attributes that are not entity-synthesis attributes (except error attributes) and custom attributes will be dropped from spans.
  * All spans containing the same entity-synthesis agent attribute values will be combined together under a single span.
  * When combining spans together, the durations will be summed together under the first occurring span (by timestamp) in the span tree and recorded under the attribute `nr.durations` (the original duration value will be left unmodified).
  * When combining spans together, the span ids that were compressed will be added to a new array agent attribute called `nr.ids`. 
  * When combining spans together that have error attributes, one span's error attributes should be prioritized (either first occurring or last occurring) and added to the compressed span.(we have chosen last)
  * All spans will be re-parented onto the entry span.

This PR also does some refactoring to encapsulate all the partial tracing logic.  When a transaction ends and span events are enabled, it used to call `trace.generateSpanEvents`. Now it checks if `transaction.partialType` is not null and create a partial trace and assigns it to the transaction.  All the logic for handling partial traces is now in `lib/transaction/trace/partial-trace.js`.  

I moved all static method on SpanEvent to be instance methods so they can be called at any point.  Most of the instance methods are cached and if called more than once it will serve the cached value. I also moved applying partialTrace rules to the partial trace class to completely separate partial tracing behavior.

## How to Test

```sh
npm run unit
```

## Related Issues
Closes #3454

### DIFF
--- a/lib/spans/span-event.js
+++ b/lib/spans/span-event.js
@@ -169,7 +169,7 @@ class SpanEvent {
    */
   get isLlmSpan() {
     if (this._isLlmSpan === undefined) {
-      this._isLlmSpan = this.intrinsics.name?.startsWith('Llm/')
+      this._isLlmSpan = this.intrinsics.name?.startsWith('Llm/') || false
     }
 
     return this._isLlmSpan
--- a/lib/transaction/trace/partial-trace.js
+++ b/lib/transaction/trace/partial-trace.js
@@ -71,10 +71,10 @@ class PartialTrace {
       // span was not dropped, add to trace until all spans have been processed
       this.spans.push(span)
     } else if (this.type !== 'compact') {
-	    // span was dropped so keep track of its id and parent as any spans
-	    // whose parent id was dropped needs to update to this new id
-	    // unless compact where all parentIds are assigned to the entry span
-	    // in finalizeSpanEvents
+      // span was dropped so keep track of its id and parent as any spans
+      // whose parent id was dropped needs to update to this new id
+      // unless compact where all parentIds are assigned to the entry span
+      // in finalizeSpanEvents
       this.droppedSpans.set(id, parentId)
     }
   }
@@ -83,7 +83,7 @@ class PartialTrace {
    * Iterates over dropped spans and reparents span if its current parent was dropped.
    * It'll traverse until it finds a parent that wasn't dropped or there are no more parents to check.
    *
-   * @param {Span} span the span to potentially reparent
+   * @param {SpanEvent} span the span to potentially reparent
    */
   maybeReparentSpan(span) {
     let result = this.droppedSpans.get(span.parentId)


lib/spans/span-event.js
```diff
@@ -143,73 +150,219 @@ class SpanEvent {
...
+   * Check if span is exit span
+   * @returns {boolean} if span is external http, db, or msg producer(aka exit span)
+   */
+  get isExitSpan() {
+    if (this._isExitSpan === undefined) {
+      const name = this.intrinsics.name
+      this._isExitSpan = REGEXS.CLIENT.EXTERNAL.test(name) || REGEXS.CLIENT.DATASTORE.test(name) || REGEXS.PRODUCER.test(name)
+    }
+
+    return this._isExitSpan
```
jsumners-nr, line 163: Is it possible for this to still be `undefined`? Should it be?: `return this._isExitSpan === true`

lib/spans/span-event.js
```diff
@@ -143,73 +150,219 @@ class SpanEvent {
...
+  /**
+   * Check if span is LLM span
+   * @returns {boolean} if span name starts with `Llm/`
+   */
+  get isLlmSpan() {
+    if (this._isLlmSpan === undefined) {
+      this._isLlmSpan = this.intrinsics.name?.startsWith('Llm/')
+    }
+
+    return this._isLlmSpan
```
jsumners-nr, line 175: Same question as `_isExitSpan`.

lib/transaction/index.js
```diff
@@ -160,8 +161,10 @@ function Transaction(agent, traceId) {
   this.traceContext = new TraceContext(this)
   this.logs = new Logs(agent)
   this.ignoreApdex = false
-  // flag to tell span event aggregator to run partial granularity rules
+  // flag uses to create partial trace when transaction ends
```
jsumners-nr:
```suggestion
  // flag used to create partial trace when transaction ends
```

lib/transaction/trace/partial-trace.js
```diff
@@ -0,0 +1,243 @@
...
+    // span was not dropped, add to trace until all spans have been processed
+    if (span) {
+      this.spans.push(span)
+    // span was dropped so keep track of its id and parent as any spans
+    // whose parent id was dropped needs to update to this new id
+    // unless compact where all parentIds are assigned to the entry span
+    // in finalizeSpanEvents
+    } else if (this.type !== 'compact') {
+      this.droppedSpans.set(id, parentId)
+    }
```
jsumners-nr: This is just clearer to me:

```suggestion
    if (span) {
      // span was not dropped, add to trace until all spans have been processed
      this.spans.push(span)
    } else if (this.type !== 'compact') {
	    // span was dropped so keep track of its id and parent as any spans
	    // whose parent id was dropped needs to update to this new id
	    // unless compact where all parentIds are assigned to the entry span
	    // in finalizeSpanEvents
      this.droppedSpans.set(id, parentId)
    }
```

lib/transaction/trace/partial-trace.js
```diff
@@ -0,0 +1,243 @@
...
+   *
+   * @param {object} meta metadata for a given applyCompaction run
+   * @param {SpanEvent} span an exit span to the same entity as retained span
+   */
+  compactionError(meta, span) {
+    // store the error that occurs in all exit spans
+    if (span.hasErrorAttrs) {
+      if (!meta.errorSpan) {
+        meta.errorSpan = span
+        logger.trace('Partial trace is compact, found an error to use from span %s, error attrs %s', span.intrinsics.name, span.errorAttrs)
```
jsumners-nr, line 114: Isn't `span.errorAttrs` and array? Will `%s` serialize it correctly?

lib/spans/span-event.js
bizob2828: the result of the check is a boolean, i'm not sure I follow.

lib/spans/span-event.js
bizob2828: name is always present i can remove the optional chaining so the result will always be a boolean

lib/transaction/trace/partial-trace.js
bizob2828: no it's an object

jsumners-nr: Either way, I don't think the log is going to impart the desired information.

bizob2828: it's fine. it stringifies the object https://github.com/newrelic/node-newrelic/blob/c465f4c540599b0ded7eac1f27e12443b69d7a81/lib/util/logger.js#L276

jsumners-nr (approved):
We reviewed together over Zoom. Everything seems reasonable to me.

PR #3624, opened 2025-12-23
test: Refactor LangChain versioned tests

### LINKED ISSUE #3623: Refactor LangChain versioned tests
### Description
Currently, the LangChain versioned tests (ran with `npm run versioned:internal:major langchain` in the CI), has a pretty large test version matrix relative to the number of versions of LangChain there are. This is because the dependencies for the underlying LLMs being tested (currently just the OpenAI and AWS Bedrock Converse APIs) are also included. There are also repeated test cases that we should cut.

We should separate the "common" LangChain tests (irrespective of underlying LLM API) from the tests that rely on the LLM API. Perhaps something like: `langchain`, `langchain-openai`, and `langchain-aws-bedrock`, instead of `langchain`, `langchain/openai` and `langchain/bedrock`.

### Acceptance Criteria

- [ ] LangChain tests are reorganized for better readability.
- [ ] Duplicate tests are removed.
- [ ] A smaller test version matrix exists.

amychisholm03:
## Description

This PR refactors the tests in `test/versioned/langchain`. Before it had sub folders, `openai` and `bedrock`, to test the `@langchain/openai` and `@langchain/aws` integration respectively. The `package.json` contained dependencies for both of these, creating a large test version matrix.

To reduce the matrix size, `langchain/openai` was moved to `langchain-openai` and `langchain/bedrock` to `langchain-aws`. Because the `vectorstore`, `runnables`, and `runnables-streaming` tests for both packages test the same LangChain functionality, except for the LLM and client, I created a centralized test suite abstraction, see `vectorstore.js`, `runnables.js`, and `runnables-streaming.js` in the `langchain` folder.

## How to Test

```
```

## Related Issues

Closes #3623

### DIFF
--- a/test/versioned/langchain-aws/runnables-streaming.test.js
+++ b/test/versioned/langchain-aws/runnables-streaming.test.js
@@ -7,7 +7,6 @@
 
 const test = require('node:test')
 const assert = require('node:assert')
-const path = require('node:path')
 
 const { removeModules } = require('../../lib/cache-buster')
 const { match } = require('../../lib/custom-assertions')
@@ -27,7 +26,7 @@ const config = {
     }
   }
 }
-const createAiResponseServer = getAiResponseServer(path.join(__dirname, './'))
+const createAiResponseServer = getAiResponseServer(__dirname)
 
 async function beforeEach({ enabled, ctx }) {
   ctx.nr = {}
--- a/test/versioned/langchain-aws/runnables.test.js
+++ b/test/versioned/langchain-aws/runnables.test.js
@@ -6,7 +6,6 @@
 'use strict'
 
 const test = require('node:test')
-const path = require('node:path')
 
 const { removeModules } = require('../../lib/cache-buster')
 const { runRunnablesTests } = require('../langchain/runnables')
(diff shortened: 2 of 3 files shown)


test/versioned/langchain-aws/runnables-streaming.test.js
```diff
@@ -0,0 +1,118 @@
...
+
+const config = {
+  ai_monitoring: {
+    enabled: true,
+    streaming: {
+      enabled: true
+    }
+  }
+}
+const createAiResponseServer = getAiResponseServer(path.join(__dirname, './'))
```
jsumners-nr: What is the purpose of the `path.join`? Is this held over from the previous state, or a new addition?

amychisholm03: Held over from the previous state -- `getAiResponseServer` needs to know which directory it came from (more specifically which `node_modules` to use), so it can use the right AWS Bedrock version for that test suite.

jsumners-nr: `__dirname` is equivalent to `path.join(__dirname, './')`.

amychisholm03: Ah, yup. It was `'../'`, but when I moved things around, that had to be changed. I'll fix it

PR #3625, opened 2025-12-24
chore: Don't create OpenAI segments if `ai_monitoring` is disabled

### LINKED ISSUE #3622: OpenAI instrumentation should not create spans when AIM is disabled
### Description
While working on other AIM packages, we re-read the AIM spec and noticed that both `Llm` events and spans must not be created if `ai_monitoring.enabled === false`. The OpenAI instrumentation still creates a coordinating span if AIM is disabled, so we need to fix it.

### Acceptance Criteria

- [ ] relevant tests are added to handle this scenario.

amychisholm03:
## Description

OpenAI instrumentation should not create segments if `ai_monitoring.enabled === false`. This PR adds this missing logic and tests to assert it.

## How to Test

```
```

## Related Issues

Closes #3622

### DIFF
--- a/lib/subscribers/openai/chat.js
+++ b/lib/subscribers/openai/chat.js
@@ -32,6 +32,10 @@ class OpenAIChatCompletions extends OpenAISubscriber {
       this.logger.warn(`Instrumenting chat completion streams is only supported with openai version ${MIN_STREAM_VERSION}+.`)
       return ctx
     }
+    if (!this.enabled) {
+      this.logger.debug('OpenAI instrumentation is disabled, not creating segment.')
+      return
+    }
 
     const segment = this.agent.tracer.createSegment({
       name: OPENAI.COMPLETION,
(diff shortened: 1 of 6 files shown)


lib/subscribers/openai/chat.js
jsumners-nr, line 1: I suspect there's a better way of handling this than to replicate checks in multiple methods across multiple implementations. Perhaps something like:

```js
class Subscriber {
	enable() {
		if (this.shouldEnable === false) {
			this.logger.trace(
				{ packageName: this.packageName },
				'Not subscribing to channel events. Instrumentation is disabled.'
			)
			return
		}

		// do the real stuff
	}
}
```

amychisholm03: Ideally, this would be enough. However, because `collect_ai` / `ai_monitoring` can be changed serverside, we have to check the config values after the subscribers have been setup/enabled.

Also, I wanted to maintain parity with the other AIM subscribers, and I'll likely refactor this `ai_monitoring` check logic in #3487

jsumners-nr (approved):

PR #3638, opened 2026-01-06
fix: Updated logic to properly assign content and role in LangChain chat completion messages

### LINKED ISSUE #3583: Include `role` attribute in LangChain `LlmChatCompletionMessage`s
### Description
The `role` field on LangChain `LlmChatCompletionMessage`s is not present in the UI. We need to define this attribute as we do with the other LLM packages.

Customer complaint: "Missing Role Field: Our current logic for identifying request ('user') and response ('assistant') messages relies on the presence of the role field. Unlike other LLM vendors, the New Relic agent is not capturing the role field within the LangChain instrumentation. Because the role field is missing for LangChain messages, our filtering mechanism fails. It defaults to retrieving the latest available message content that does contain a role within the same trace."

### Acceptance Criteria

- [ ] `role` should be present on LangChain `LlmChatCompletionMessage`s in the UI

### Additional context
[Jira link](https://new-relic.atlassian.net/browse/NR-497185)

amychisholm03:
There were a few bugs with our LangChain instrumentation that this PR fixes:
* `role` is missing from LangChain `LlmChatCompletionMessage`
* there are a variety of ways chunks can come through with LangChain streams; I made sure we're assigning the content correctly on `LlmChatCompletionMessage`

The `role` and `content` fix now allows LangChain AI conversations to show up in the UI correctly.

This PR resolves #3583.

### DIFF
--- a/lib/subscribers/langchain/runnable-stream.js
+++ b/lib/subscribers/langchain/runnable-stream.js
@@ -56,7 +56,6 @@ class LangchainRunnableStreamSubscriber extends LangchainSubscriber {
 
     // Extract data.
     const request = data?.arguments?.[0]
-    const userRequest = request?.messages ? request.messages?.[0] : request
     const params = data?.arguments?.[1] || {}
     const metadata = params?.metadata ?? {}
     const tags = params?.tags ?? []
@@ -64,7 +63,7 @@ class LangchainRunnableStreamSubscriber extends LangchainSubscriber {
 
     // Instrument stream.
     if (output?.next) {
-      this.wrapNextHandler({ output, segment, request: userRequest, metadata, tags, transaction, pkgVersion })
+      this.wrapNextHandler({ output, segment, request, metadata, tags, transaction, pkgVersion })
     } else {
       // Input error occurred which means a stream was not created.
       // Skip instrumenting streaming and create Llm Events from
@@ -73,7 +72,7 @@ class LangchainRunnableStreamSubscriber extends LangchainSubscriber {
         transaction,
         segment,
         messages: [],
-        events: [userRequest],
+        events: [request],
         metadata,
         tags,
         err,
--- a/lib/subscribers/langchain/runnable.js
+++ b/lib/subscribers/langchain/runnable.js
@@ -37,7 +37,6 @@ class LangchainRunnableSubscriber extends LangchainSubscriber {
     }
 
     const request = data?.arguments?.[0]
-    const userRequest = request?.messages ? request.messages?.[0] : request
     const params = data?.arguments?.[1] || {}
     const { result, error: err, moduleVersion: pkgVersion } = data
 
@@ -46,7 +45,7 @@ class LangchainRunnableSubscriber extends LangchainSubscriber {
       transaction,
       segment,
       messages: [result],
-      events: [userRequest, result],
+      events: [request, result],
       metadata: params?.metadata ?? {},
       tags: params?.tags ?? [],
       err


lib/subscribers/langchain/base.js
```diff
@@ -131,7 +133,7 @@ class LangchainSubscriber extends Subscriber {
 
       this.recordEvent({
         type: 'LlmChatCompletionMessage',
-        pkgVersion: this.moduleVersion,
+        pkgVersion,
```
amychisholm03, line 137: We never assigned `this.moduleVersion`, so this was never working.

test/versioned/langchain/runnables-streaming.js
```diff
@@ -230,7 +230,7 @@ function runStreamingEnabledTests(config) {
           const stream = await chain.stream(input, options)
           let content = ''
           for await (const chunk of stream) {
-            content += chunk
+            content += chunk?.[0]
```
amychisholm03, line 233: This test was passing because content would be `''` and our instrumentation, because we didn't concat the stream correctly, would also return `''` and thus be equal. Now, the actual response is present.

lib/llm-events/langchain/chat-completion-message.js
```diff
@@ -47,6 +47,8 @@ class LangChainCompletionMessage extends LangChainEvent {
     this.sequence = params.sequence
     this.completion_id = params.completionId
     this.is_response = params.isResponse ?? false
+    // overly simplified for now
```
bizob2828: what does this comment mean?

lib/subscribers/langchain/runnable-stream.js
```diff
@@ -96,53 +97,73 @@ class LangchainRunnableStreamSubscriber extends LangchainSubscriber {
    */
   wrapNextHandler({ output, segment, transaction, request, metadata, tags, pkgVersion }) {
     const self = this
-    const orig = output.next
+    const proto = Object.getPrototypeOf(output)
     const ctx = this.agent.tracer.getContext()
     let content = ''
 
-    async function wrappedIterator(...args) {
```
bizob2828, line 103: what was wrong with this?

lib/subscribers/langchain/runnable-stream.js
```diff
@@ -96,53 +97,73 @@ class LangchainRunnableStreamSubscriber extends LangchainSubscriber {
...
 
-    output.next = this.agent.tracer.bindFunction(wrappedIterator, ctx, false)
+    // Replace the asyncIterator method on the prototype
+    Object.defineProperty(proto, Symbol.asyncIterator, {
+      value: function() {
+        // Call the original method to get the iterator
+        const iterator = originalAsyncIteratorMethod.call(this)
+        const originalNext = iterator.next.bind(iterator)
+
+        return {
```
bizob2828: this is very hard to read. do you think you can define the wrap and then just assign to next? also it's an anonymous function, i would name it like before `wrappedIterator`

bizob2828 (changes requested):
a few questions and I also don't any tests for new logic. I'm still unsure how you found these issues and if the fixes are correct without seeing tests

lib/llm-events/langchain/chat-completion-message.js
amychisholm03: Looks like I missed a commit in my squashing; should be there now

lib/subscribers/langchain/runnable-stream.js
amychisholm03: I included it from my WIP LangGraph instrumentation because I noticed the segment duration in the UI for LangChain streams was very small. I fixed it on that branch, but it could have been another change that fixed it. I'll remove from it this PR because it doesn't need to be here.

lib/subscribers/langchain/runnable-stream.js
```diff
@@ -115,11 +116,16 @@ class LangchainRunnableStreamSubscriber extends LangchainSubscriber {
             pkgVersion
           })
         } else {
+          // Concat the streamed content
```
bizob2828, line 118: where are the various tests to assert these diff shapes?

lib/llm-events/langchain/chat-completion-message.js
```diff
@@ -47,6 +49,13 @@ class LangChainCompletionMessage extends LangChainEvent {
     this.sequence = params.sequence
     this.completion_id = params.completionId
     this.is_response = params.isResponse ?? false
+    if (params.role) {
```
bizob2828, line 52: we have unit tests for this message, you should update them to reflect this new logic

lib/subscribers/langchain/base.js
bizob2828: how were the tests passing then?

lib/subscribers/langchain/runnable-stream.js
```diff
@@ -48,31 +48,32 @@ class LangchainRunnableStreamSubscriber extends LangchainSubscriber {
...
-    // Instrument stream
     const ctx = this.agent.tracer.getContext()
     const { segment, transaction } = ctx
     if (transaction?.isActive() !== true) {
       return
     }
 
+    // Extract data.
     const request = data?.arguments?.[0]
+    const userRequest = request?.messages ? request.messages?.[0] : request
```
bizob2828: are we not asserting the right stuff in streaming tests? I don't see any updates to reflect this.

bizob2828 (changes requested):

lib/subscribers/langchain/runnable-stream.js
amychisholm03: They already exist. The only "new" one is `result?.value?.[0] === 'string'` which the `'should create langchain events for every stream call with parser that returns an array as output'` test covers in `runnables-streaming.js`. I have a comment there on why it was incorrectly passing before.

lib/subscribers/langchain/base.js
amychisholm03: We aren't asserting that the `'Supportability/Nodejs/ML/LangChain/pkgVersion'` metric get screated, only the tracking metric. I can add that now.

lib/subscribers/langchain/runnable-stream.js
amychisholm03: Another thing that's specific to LangGraph<>LangChain instrumentation. I was porting over code changes to LangChain as to not bloat the LangGraph PR. The tests for this different request style are on the LangGraph branch; I can remove it here; there will just be more changes to LangChain on the LangGraph PR.

lib/subscribers/langchain/base.js
amychisholm03: In test/versioned/langchain/runnables.js, we have this test that only checks if it's greater than 0. There is 1, which comes from the summary event, but there are not ones for the individual messages, which this code fixes.

I can change that test to assert 3 rather than just >0.

```
 test('should increment tracking metric for each langchain chat prompt event', (t, end) => {
    const { agent, prompt, outputParser, model, langchainCoreVersion } = t.nr

      const input = inputData
      const options = { metadata: { key: 'value', hello: 'world' }, tags: ['tag1', 'tag2'] }

      const chain = prompt.pipe(model).pipe(outputParser)
      await chain.invoke(input, options)

      const metrics = agent.metrics.getOrCreateMetric(
        `Supportability/Nodejs/ML/LangChain/${langchainCoreVersion}`
      )
      assert.equal(metrics.callCount > 0, true)

      tx.end()
      end()
    })
  })
```

PR #3645, opened 2026-01-08
feat: Add instrumentation support for `@langchain/langgraph`

### LINKED ISSUE #3584: Instrument `@langchain/langgraph` in accordance to Agentic AI spec
### Description
There is a customer and internal request to instrument [`@langchain/langgraph`](https://github.com/langchain-ai/langgraphjs). This instrumentation must utilize the the new `LlmAgent` event, introduced in this [pending spec](https://source.datanerd.us/agents/agent-specs/pull/761). The coordinating segment `Llm/agent/LangGraph/{function_name}/{agent_name}` must be created.

Specifically, for Node, this will involve instrumenting the `Pregel` class, the underlying base class for all `StateGraph` classes. For example, `createReactAgent()` is one of the most common ways to make an "agent" in LangGraph and it returns a `CompiledStateGraph`. The methods we will instrument is `invoke` (like LangChain's tools) and `stream` for streaming invocations.

UPDATE (12/30/25): Individual tool calls do not need to be instrumented (e.g. a `LlmTool` event and a `Llm/tool/*` segment are created) because we already get this granular-level detail from existing LangChain instrumentation. Confirmed by Uma for Python agent as well.

UPDATE (1/7/2026): We also need to add the span attribute: `subcomponent`: `{"type": "APM-AI_AGENT", "name": {agent_name}}` (source: Agentic AI spec).

The attribute value will be a JSON formatted string where the `type` value is static and the `name` value is the name of the AI agent or tool being represented by the span.

Customer request:
Agent should instrument the framework methods according to the Agentic spec:
- It should be possible to measure agent’s latency, errors and tokens
- It should be able to monitor agent calls to LLM/Tool/Agent 
- It should be able to detect there was an agent call within the span
- It should be possible to view agent to agent calls within entity map

### Acceptance Criteria

- [x] `LlmAgent` event is created for agent functions: `invoke` and `stream`
- [x] Coordinating spans `Llm/agent/*`, are created
- [x] The `subcomponent` attribute is added to the `Llm/agent/*` span
- [x] A `LlmError` event with `agent_id` is created if the AI agent invocation fails.
- [x] A test suite is created to test the instrumentation
- [x] Create an example app in `newrelic-node-examples` to test this in a simple real-life scenario: LLM is OpenAi, Tool is LangChain, and Agent is LangGraph.

### Additional context
- [Related spec](https://source.datanerd.us/agents/agent-specs/pull/761) is currently in PR review
- [Jira link](https://new-relic.atlassian.net/browse/NR-497186)

amychisholm03:
## Description 🦜

LangGraph (`@langchain/langgraph`) is built on-top of LangChain to add stateful, multi-step workflows (also known as "AI agents" or "graphs") to help build complex, large-scale AI operations. This PR adds instrumentation for LangGraph AI agent invocations via the methods `invoke` and `stream`. 

We will create a segment named `Llm/agent/LangGraph/stream/<aiAgentName>`, a new LLM event called `LlmAgent`, and `LlmErrorMessage` if an error occurs. We must also add a new attribute, `subcomponent`, to the span for the UI to link together these agentic AI components. Like other LLM instrumentations, the `llm = true` attribute will be added to the transaction trace.

`invoke` is a wrapper around `stream` that abstracts away dealing with the stream directly and returns the completed response asynchronously. Therefore, unlike the Python agent, we will _**not**_ instrument the `invoke` call as it would be redundant information. We will have tests, though, to make sure `invoke` calls trigger the underlying `stream` instrumentation.

Side note: We don't need to instrument tool calls or chat completions. We get tool invocation details from our underlying LangChain instrumentation and `LlmChatCompletion` events from the underlying LLM (currently have only tested with OpenAI).

## How to Test ⚙️

```
npm run versioned:major langgraph
# make sure standard langchain instrumentation didn't break
```

Run the example app [here](https://github.com/newrelic/newrelic-node-examples/pull/359).
* Example `stream` transaction: https://staging.onenr.io/02R5AZ5JPRb

## Related Issues

Closes #3584

### DIFF
--- a/lib/subscribers/langchain/runnable.js
+++ b/lib/subscribers/langchain/runnable.js
@@ -71,7 +71,7 @@ class LangchainRunnableSubscriber extends AiMonitoringChatSubscriber {
     if (response || response === '') {
       // Add the response if it is NOT an outgoing
       // tool call with no result yet
-      if (!(response.content === '' && response?.tool_calls?.length > 0)) {
+      if (!(response?.content === '' && response?.tool_calls?.length > 0)) {
         messages.push(response)
       }
     }


lib/llm-events/error-message.js
```diff
@@ -21,9 +21,10 @@ module.exports = class LlmErrorMessage {
    * @param {LlmVectorStoreSearch} [params.vectorsearch] Details about the vector
    * search if it was a vector search event.
    * @param {LlmTool} [params.tool] Details about the tool event if it was a tool event.
+   * @param {object} [params.agent] Details about the AI agent event if it was an AI agent event.
    * @param {boolean} [params.useNameAsCode] defaults to false, only Bedrock sets it to true so far
    */
-  constructor({ response, cause, summary = {}, embedding = {}, vectorsearch = {}, tool = {}, useNameAsCode = false } = {}) {
+  constructor({ response, cause, summary = {}, embedding = {}, vectorsearch = {}, tool = {}, agent = {}, useNameAsCode = false } = {}) {
```
bizob2828: agent is a loaded term, you may want to consider renaming this to llmAgent or something?

lib/llm-events/langgraph/agent.js
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * @typedef {object} LangGraphAgentEventParams
+ * @property {Agent} agent The New Relic agent instance.
+ * @property {string} name The name of the LangGraph agent, defaults to 'agent'.
+ * @property {object} segment The associated NR segment.
+ * @property {object} transaction The associated NR transaction.
+ * @property {boolean} error A boolean flag to indicate if an error occurred.
+ */
+
+module.exports = class LangGraphAgentEvent extends BaseLlmEvent {
```
bizob2828, line 18: is this correct? if so, where is this documented?

lib/llm-events/error-message.js
amychisholm03: Yeah I can. The JSDoc describes it as an AI agent, but I can further clarify it.

lib/llm-events/langgraph/agent.js
amychisholm03: It is extended from `BaseLlmEvent` so that the `llm.<user_defined_metadata>` field is constructed properly. The `LlmAgent` event is defined in pending agent spec PR 761.

lib/llm-events/langgraph/agent.js
```diff
@@ -0,0 +1,42 @@
...
+   * @param {LangGraphAgentEventParams} params should contain all necessary and optional LangGraph data
+   */
+  constructor(params) {
+    super(params)
+    const { agent, segment, transaction, error = false, name = 'agent' } = params
+
+    this.name = name
+    this.span_id = segment?.id
+    this.trace_id = transaction?.traceId
+    this.error = error
```
bizob2828, line 35: i don't see this in the spec

lib/subscribers/langchain/runnable-stream.js
```diff
@@ -12,41 +12,43 @@ class LangchainRunnableStreamSubscriber extends LangchainRunnableSubscriber {
   }
 
   asyncEnd(data) {
+    // Exit early if disabled.
```
bizob2828: i don't understand these code comments, what value are they providing?

lib/subscribers/langchain/runnable-stream.js
```diff
@@ -12,41 +12,43 @@ class LangchainRunnableStreamSubscriber extends LangchainRunnableSubscriber {
...
     const tags = params?.tags ?? []
     const { result: response, error: err } = data
 
-    // Instrument stream.
-    if (response?.next) {
-      this.wrapNextHandler({ response, ctx, request, metadata, tags })
+    // Note: as of 18.x `ReadableStream` is a global
+    // eslint-disable-next-line n/no-unsupported-features/node-builtins
+    if (response instanceof ReadableStream) {
+      this.wrapNextHandler({ response, ctx, request: userRequest, metadata, tags })
```
bizob2828: I'd change the name of this to instrumentStream or something. this is no longer wrapping the next handler

lib/subscribers/langchain/runnable-stream.js
```diff
@@ -12,41 +12,43 @@ class LangchainRunnableStreamSubscriber extends LangchainRunnableSubscriber {
...
     const ctx = this.agent.tracer.getContext()
-
     const { transaction } = ctx
     if (transaction?.isActive() !== true) {
       return
     }
 
     // Extract data.
     const request = data?.arguments?.[0]
+    const userRequest = request?.messages ? request.messages?.[0] : request
```
bizob2828, line 35: can you put a comment explaining what this is doing, because I don't understand it

lib/subscribers/langchain/runnable-stream.js
```diff
@@ -67,54 +69,69 @@ class LangchainRunnableStreamSubscriber extends LangchainRunnableSubscriber {
...
+      const reader = orig.apply(this, arguments)
+      const origRead = reader.read
+      let content = ''
+      let langgraphMessages = []
+      reader.read = async function wrappedRead(...args) {
+        try {
+          const result = await origRead.apply(this, args)
+          if (result?.done) {
+            // only create Llm events when stream iteration is done
+            const responseMsgs = langgraphMessages.length > 0
```
bizob2828: we have a variable for langgraph in langchain instrumentation? seems like something is off here

lib/subscribers/langchain/runnable-stream.js
```diff
@@ -67,54 +69,69 @@ class LangchainRunnableStreamSubscriber extends LangchainRunnableSubscriber {
...
+              // Base LangChain case
+              content += result.value
+            } else if (typeof result?.value?.[0] === 'string') {
+              // Array parser case
+              content += result.value[0]
+            }
+          }
+          return result
+        } catch (error) {
+          const responseMsgs = langgraphMessages.length > 0
```
bizob2828: this is redundant code from above, you may want to encapsulate this into a function

lib/subscribers/langchain/runnable-stream.js
```diff
@@ -67,54 +69,69 @@ class LangchainRunnableStreamSubscriber extends LangchainRunnableSubscriber {
...
+            self.recordChatCompletionEvents({
+              ctx,
+              response: responseMsgs,
+              request,
+              metadata,
+              tags
+            })
+          } else {
+            // Concat the streamed content
+            if (result?.value?.messages || result?.value?.agent?.messages) {
```
bizob2828: you may want to encapsulate this parsing logic into a function, it's getting quite complex

lib/subscribers/langgraph/graph-stream.js
```diff
@@ -0,0 +1,133 @@
...
+      logger,
+      packageName: '@langchain/langgraph',
+      channelName: 'nr_stream',
+      name: `${LANGGRAPH.AGENT}/stream`,
+      trackingPrefix: LANGGRAPH.TRACKING_PREFIX })
+    this.events = ['asyncEnd']
+  }
+
+  get enabled() {
+    return super.enabled && this.agent.config.ai_monitoring.streaming.enabled
```
bizob2828: I wouldn't conflate these two settings. If streaming is disabled you need to log the streaming disabled metric

lib/subscribers/langgraph/graph-stream.js
```diff
@@ -0,0 +1,133 @@
...
+    // If error already:
+    if (initialErr) {
+      this.recordAiAgentEvent({ aiAgentName, transaction, segment, error: initialErr })
+      return
+    }
+
+    // Note: as of 18.x `ReadableStream` is a global
+    // eslint-disable-next-line n/no-unsupported-features/node-builtins
+    if (stream instanceof ReadableStream) {
+      this.wrapNextHandler({ stream, segment, transaction, aiAgentName })
```
bizob2828: same comment as langchain, this isn't wrapping the next handler. I'd rename this to `instrumentStream`

lib/subscribers/langchain/tool.js
```diff
@@ -38,12 +39,23 @@ class LangchainToolSubscriber extends AiMonitoringSubscriber {
...
     const params = data?.arguments?.[1] || {}
     const { metadata: paramsMeta, tags: paramsTags } = params
     const metadata = this.mergeMetadata(instanceMeta, paramsMeta)
     const tags = this.mergeTags(instanceTags, paramsTags)
 
+    // Get tool output and input
+    const request = data?.arguments?.[0]
+    const output = result?.content ?? result
+    let input
+    if (request?.input) {
```
bizob2828, line 53: are you going to log a ticket to add the `segment.addSpanAttribute('subcomponent', `{"type": "APM-AI_AGENT", "name": ${aiAgentName}}`)` but for tools? the spec updates call this out

lib/subscribers/langgraph/graph-invoke.js
```diff
@@ -0,0 +1,34 @@
...
+const AiMonitoringSubscriber = require('../ai-monitoring/base')
+const { AI: { LANGGRAPH } } = require('#agentlib/metrics/names.js')
+
+/**
+ * Subscribes to `Pregel.invoke()` events. Only exists to
+ * create a segment/span. The related `LangGraphStreamSubscriber`
+ * handles the creation of Llm events because `invoke` is just a
+ * wrapper around `stream`, and we don't want to create duplicates.
+ */
+class LangGraphInvokeSubscriber extends AiMonitoringSubscriber {
```
bizob2828: I don't understand this subscriber, this is just creating a segment name for this call which is accounted for in the `graph-stream` call. I don't see a case where invoke doesn't call stream . I would remove this instrumentation to avoid creating useless segments for customers

lib/subscribers/langgraph/graph-stream.js
```diff
@@ -0,0 +1,133 @@
...
+    if (transaction?.isActive() !== true) {
+      return
+    }
+
+    // Extract data
+    const { error: initialErr, result: stream } = data
+    const aiAgentName = data?.self?.name ?? 'agent'
+
+    // Mark as LLM event
+    transaction.trace.attributes.addAttribute(DESTINATIONS.TRANS_EVENT, 'llm', true)
```
bizob2828: this is done in the handler as it calls `addLlmMeta`

test/versioned/langgraph/graph-stream.test.js
```diff
@@ -0,0 +1,421 @@
...
+})
+
+test('should not create segment or events when ai_monitoring.enabled is false', async (t) => {
+  const { agent, langgraphAgent } = t.nr
+
+  // Disable ai_monitoring
+  agent.config.ai_monitoring.enabled = false
+
+  await helper.runInTransaction(agent, async (tx) => {
+    try {
```
bizob2828, line 418: I don't think you need to wrap this in a `try/catch`. If the stream throws an error, it'll fail the test

test/versioned/langgraph/graph-stream.test.js
```diff
@@ -0,0 +1,421 @@
...
+
+    const [segment] = tx.trace.getChildren(tx.trace.root.id)
+    const attribute = segment?.attributes?.attributes?.subcomponent
+    assert.equal(attribute?.value, '{"type": "APM-AI_AGENT", "name": LangGraphReactAgent}')
+
+    tx.end()
+  })
+})
+
+test('should create LlmError event when stream fails', async (t) => {
```
bizob2828: this test doesn't assert that a stream fails. it immediately is throwing, this test is valid but we're missing a test where the stream fails

lib/subscribers/langgraph/graph-stream.js
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  get enabled() {
+    return super.enabled && this.agent.config.ai_monitoring.streaming.enabled
+  }
+
+  handler(data, ctx) {
+    // Update segment name with LangGraph agent name.
+    const aiAgentName = data?.self?.name ?? 'agent'
+    this.name = `${LANGGRAPH.AGENT}/stream/${aiAgentName}`
```
bizob2828: i see maybe why you did this because I realized the langchain tools have issues. I would define the `name` above as unknown though since you're redefining it every time

test/versioned/langgraph/graph-stream.test.js
```diff
@@ -0,0 +1,421 @@
...
+    }
+  }
+}
+
+function consumeChunk(chunk) {
+  // intentional no-op
+  return chunk
+}
+
+test.beforeEach(async (ctx) => {
```
bizob2828, line 33: missing an edge case where the calls are done outside a transaction. Also the stream tests aren't interacting with stream so the catch on the stream call is never getting hit

lib/llm-events/langgraph/agent.js
amychisholm03: Its under Agentic.md#Error Tracing:

"`LlmAgent` events and `LlmTool` events will still be created in error contexts despite the fact that they could lack the detail of a tool call or agent run that succeeded. All attributes that can be captured without a successful response object **MUST** still be added to respective events. In addition, an `error` attribute **MUST** be added to these events and set to true as: `error`: `True`."

lib/subscribers/langchain/runnable-stream.js
amychisholm03: Just better code sectioning, makes it easier to read to me at least. I can see how it can be redundant, I'll remove it

lib/subscribers/langchain/runnable-stream.js
amychisholm03: LangGraph _is_ LangChain in the sense that `@langchain/langgraph` is part of the monolith package `langchain`. There is going to be some overlap. You can think of LangGraph as another abstraction / API layer over LangChain.

lib/subscribers/langgraph/graph-stream.js
amychisholm03: Spec says to have the default name as `'agent'` so I'll just set it to that.

lib/subscribers/langgraph/graph-stream.js
amychisholm03: Ah, I see why I did this. `AiMonitoringSubscriber.enabled()` will create the segment if `enabled=true && streaming.enabled=false`. I thought in this case it still wouldn't create the segment.

Doubling checking the spec, I got this wrong, let me update the test

lib/subscribers/langchain/tool.js
amychisholm03: Yes I will, thanks for the reminder

lib/subscribers/langgraph/graph-invoke.js
amychisholm03: Yeah, I agree. Python agent is adding support for `invoke` and `stream` explicitly, so I just followed suit. For the Python package, I think it's actually two different functions, but that's not the case for us.

When someone calls `invoke` in our case, they may be confused when the segment is labeled `stream` in the UI because they may not know that `stream` is being called underneath. We could just have the segment name reflect whatever method was actually called by the user.

test/versioned/langgraph/graph-stream.test.js
amychisholm03: Yeah I've had a hard time getting the stream to fail in the middle

test/versioned/langgraph/graph-stream.test.js
amychisholm03: I do have the no-transaction test: https://github.com/amychisholm03/node-newrelic/blob/69888b2fd2028fcdb84dd0324b1744baf85be543/test/versioned/langgraph/graph-stream.test.js#L337-L350

lib/subscribers/langchain/runnable-stream.js
```diff
@@ -65,56 +67,97 @@ class LangchainRunnableStreamSubscriber extends LangchainRunnableSubscriber {
...
+  /**
+   * Extracts response messages from accumulated stream data.
+   * For LangGraph, filters out HumanMessages.
+   * For LangChain, returns concatenated content.
+   *
+   * @param {object[]} langgraphMessages accumulated LangGraph messages
+   * @param {string} content accumulated LangChain content
+   * @returns {object[]|string} the response messages
+   */
+  getResponseMessages(langgraphMessages, content) {
```
bizob2828: why not just pass in the object and read the keys?

lib/llm-events/langgraph/agent.js
```diff
@@ -0,0 +1,42 @@
...
+    this.duration = segment?.getDurationInMillis()
```
bizob2828: I also don't see this in the spec

lib/subscribers/langgraph/graph-stream.js
```diff
@@ -0,0 +1,131 @@
...
+      return
+    }
+    if (!this.streamingEnabled) {
+      logger.debug('LangGraph streaming instrumentation is disabled, not instrumenting stream.')
+      this.agent.metrics.getOrCreateMetric(STREAMING_DISABLED).incrementCallCount()
+      return
+    }
+    const ctx = agent.tracer.getContext()
+    const { segment, transaction } = ctx
+    if (transaction?.isActive() !== true) {
```
bizob2828: if segment is not defined this will crash. we also seem to be mixed on various AIM instrumentaitons about what to check here. some check `If (!(segment || transaction))` or this check

test/versioned/langgraph/graph-stream.test.js
```diff
@@ -0,0 +1,473 @@
...
+    plan.equal(exceptions.length, 2)
+    const lgException = exceptions[1]
+    const str = Object.prototype.toString.call(lgException.customAttributes)
+    plan.equal(str, '[object LlmErrorMessage]', 'should be a LlmErrorMessage')
+    plan.equal(lgException.customAttributes.agent_id, agentEvent.id, 'ai agent_id should match')
+    plan.equal(lgException.customAttributes['error.code'], lgException.error['lc_error_code'], 'error codes should match')
+    plan.equal(lgException.customAttributes['error.message'], lgException.error['message'], 'error messages should match')
+
+    tx.end()
+  })
```
bizob2828, line 351: you want to add after the `runInTransaction` `await plan.completed`

test/versioned/langgraph/graph-stream.test.js
```diff
@@ -0,0 +1,473 @@
...
+    plan.equal(exceptions.length, 2)
+    const lgException = exceptions[1]
+    const str = Object.prototype.toString.call(lgException.customAttributes)
+    plan.equal(str, '[object LlmErrorMessage]', 'should be a LlmErrorMessage')
+    plan.equal(lgException.customAttributes.agent_id, agentEvent.id, 'ai agent_id should match')
+    plan.equal(lgException.customAttributes['error.code'], lgException.error['code'], 'error codes should match')
+    plan.equal(lgException.customAttributes['error.message'], lgException.error['message'], 'error messages should match')
+
+    tx.end()
+  })
```
bizob2828, line 393: same here `await plan.completed`

bizob2828 (changes requested):
A few comments. I also noticed the AI responses are broken in the UI.

lib/subscribers/langchain/runnable-stream.js
amychisholm03: This was ugly, I pushed up a simpler way to do this.

lib/subscribers/langgraph/graph-stream.js
amychisholm03: I'll just check both

lib/subscribers/langchain/runnable.js
```diff
@@ -51,13 +55,20 @@ class LangchainRunnableSubscriber extends AiMonitoringChatSubscriber {
...
+      // via langgraph api, response object lives at `messages[0]`
+      if (response.messages?.length > 0) {
+        response = response.messages[0]
+      }
+      if (response.content === '' && response?.tool_calls?.length > 0) {
+        // Do not create a LlmChatCompletionMessage for an outgoing
+        // tool call with no result yet
+        return [request]
+      }
       messages.push(response)
```
bizob2828:
```suggestion
       // Check if response is not a tools_call or content is empty
       // before adding response to the messages array
      if (response.content !== '' && response?.tool_calls?.length < 0) {
      
        messages.push(response)
      }
```

lib/subscribers/langchain/runnable.js
```diff
@@ -55,13 +55,20 @@ class LangchainRunnableSubscriber extends AiMonitoringChatSubscriber {
 
   getMessages({ request, response }) {
```
bizob2828, line 65: i would add jsdoc explaining what this is doing.  Mostly deciding if the request/response should be stored and later captured as LlmCompletionMessages

lib/subscribers/langchain/runnable.js
bizob2828: or something to that effect. It's weird to see we create a messages array at the front and push to it in certain times or in this case just return a new array with only the request.

test/versioned/langgraph/graph-stream.test.js
```diff
@@ -265,6 +265,11 @@ test('should have LlmTool events from LangChain instrumentation', async (t) => {
     const chatEvents = events.filter((e) => e[0].type === 'LlmChatCompletionMessage' && e[1].role === 'tool')
     assert.equal(chatEvents.length, 2, 'should have one tool message for openai and another for langchain')
 
+    const langchainEvents = events.filter((e) => e[0].type === 'LlmChatCompletionMessage' && e[1].vendor === 'langchain')
```
bizob2828, line 268: should you also assert the length here? we know how many langchain chat completion messages there should be

lib/subscribers/langchain/runnable.js
```diff
@@ -49,16 +54,26 @@ class LangchainRunnableSubscriber extends AiMonitoringChatSubscriber {
...
     if (request || request === '') {
       messages.push(request)
     }
 
-    // check if response is truthy and an empty string
     if (response || response === '') {
-      messages.push(response)
+      // Add the response if it is NOT an outgoing
+      // tool call with no result yet
+      if (!(response.content === '' && response?.tool_calls?.length > 0)) {
```
bizob2828:
```suggestion
      if (!(response?.content === '' && response?.tool_calls?.length > 0)) {
```

lib/subscribers/langchain/runnable.js
```diff
@@ -49,16 +54,26 @@ class LangchainRunnableSubscriber extends AiMonitoringChatSubscriber {
...
   }
 
+  /**
+   * Determines if the LangChain request and response should be stored and
+   * later captured as `LlmChatCompletionMessage`s.
+   * @param {object} params function parameters
+   * @param {object|string} params.request request object (`{ content, role, ... }`) or string
+   * @param {object|string} params.response response object (`{ content, role, ... }`) or string
+   * @returns {object[]} an array with all of the valid request and response objects/strings
+   */
```
bizob2828, line 64: i still don't understand why these check if request/response are empty strings and push them onto the messages array but in the case where response is an object we omit it if content is empty

amychisholm03: In some cases, `response = ''` is valid and we have a LangChain test to assert that in a "circular dependency", `response: ''` is returned with a `LlmChatCompletionMessage`. Now, whether or not this is a valid use case, I'm not sure, but I didn't want to change established LangChain logic.

Issue #3616, opened 2025-12-19
Implement "reduced" mode support for span links


jsumners-nr:
Spec PR 769 defines the actions that need to be taken when the agent is configured for partial tracing and span links are present. For the "reduced" mode, we need to:

1. Determine if a dropped span contains links
2. Re-parent the links to the closest remaining parent span
3. Update the `id` attribute of the links to match the id of the new parent span

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-503464

PR #3649, opened 2026-01-09
feat: Added reduced tracing mode for span links

svetlanabrennan:
## Description

For reducing tracing, we need to move the span links from a dropped span (if there are any spank links) to the nearest parent span that was not dropped and update the id attribute of the links to match the id of the new parent span.

## How to Test

Run `npm run unit`

## Related Issues

Closes issue #3616

### DIFF
--- a/test/unit/spans/span-event-aggregator.test.js
+++ b/test/unit/spans/span-event-aggregator.test.js
@@ -316,6 +316,10 @@ test('SpanAggregator', async (t) => {
         timestamp
       }))
 
+      // span link id is initially set to the span id in the context they are created on
+      assert.equal(child1Segment.spanLinks[0].intrinsics.id, 'span1')
+      assert.equal(child2Segment.spanLinks[0].intrinsics.id, 'span2')
+
       spanEventAggregator.addSegment({ segment: rootSegment, transaction: tx, parent: '1', isEntry: true })
       // root span has no span links
       assert.equal(rootSegment.spanLinks.length, 0)
@@ -404,6 +408,10 @@ test('SpanAggregator', async (t) => {
         },
         timestamp
       }))
+
+      // span link id is initially set to the span id in the context they are created on
+      assert.equal(child1Segment.spanLinks[0].intrinsics.id, 'span1')
+      assert.equal(child2Segment.spanLinks[0].intrinsics.id, 'span2')
 
       spanEventAggregator.addSegment({ segment: rootSegment, transaction: tx, parent: '1', isEntry: true })
       // root span has 1 span link of it's own


lib/transaction/trace/partial-trace.js
```diff
@@ -76,7 +77,32 @@ class PartialTrace {
...
+      // span was dropped but we still need to move its span links to the last kept span
+      // spanLinks were captured before the span was dropped
+      if (spanLinks && spanLinks.length > 0) {
+        this.reparentSpanLinks(spanLinks)
+      }
+    }
+  }
+
+  /**
+   * Iterates over the span links from a drop span and reassigns them to the last kept span.
```
jsumners-nr:
```suggestion
   * Iterates over the span links from a dropped span and reassigns them to the last kept span.
```

lib/transaction/trace/partial-trace.js
```diff
@@ -76,7 +77,32 @@ class PartialTrace {
...
+    const lastSpan = this.spans[this.spans.length - 1]
+
+    // the id instinsics attribute needs to be updated to equal the id of the new span the
+    // span links are moving to
+    for (const link of spanLinks) {
+      link.intrinsics.id = lastSpan.id
     }
+
+    // move the span links events to the last kept span
+    this.spans[this.spans.length - 1].spanLinks = [...lastSpan.spanLinks, ...spanLinks]
```
jsumners-nr:
```suggestion
    const lastSpan = this.spans.at(-1)

    // The id intrinsics attribute needs to be updated to equal the id of the new span the
    // span links are moving to.
    for (const link of spanLinks) {
      link.intrinsics.id = lastSpan.id
    }

    // move the span links events to the last kept span
    Array.prototype.push.apply(lastSpan.spanLinks, spanLinks)
```

test/unit/spans/span-event-aggregator.test.js
```diff
@@ -267,6 +267,176 @@ test('SpanAggregator', async (t) => {
...
+          attributes: { test: 'test2' },
+          context: { spanId: 'parent2', traceId: 'trace1' }
+        },
+        spanContext: {
+          spanId: 'span2',
+          traceId: 'trace2'
+        },
+        timestamp
+      }))
+
```
jsumners-nr, line 318: I think we should verify that the added span links's `id` is pointing to the segments they were added to here. This will show the later assertions for the `id` being updated are correct.

jsumners-nr (commented):
Looking great. A couple of tweaks and we should be good.

test/unit/spans/span-event-aggregator.test.js
svetlanabrennan: Is this what you meant? https://github.com/newrelic/node-newrelic/pull/3649/changes/249799b05ad0848ddc2f6bb2c976d994ae53b42c

jsumners-nr: Yep.

PR #3655, opened 2026-01-13
chore: Updated OpenAI instrumentation to skip creating a `LlmChatCompletionMessage` for an outgoing tool call response

### LINKED ISSUE #3644: LangChain <> OpenAI `tool_calls`  `LlmChatCompletionMessage`s should not have `null` content
### Description
While running a LangGraph<>OpenAI application, I noticed under the 'Token usage' section in the AIM UI two line items (`LlmChatCompletionMessage`s) with `null` for their content. Clicking on them reveals their 'Finish reason' is `tool_calls`. Going through our OpenAI instrumentation, I noticed we do not assign any content for tool calls, and it is a required field for `LlmChatCompletionMessage`. 

I looked around the OpenAI response object and could not find the result of the tool call. The Python agent was able to find one, so it should exist somewhere on the response. As a backup, we could include the tool name and the arguments passed into the tool or some other helpful info.

### Acceptance Criteria

- [x] something that can act as `content` for OpenAI tool calls is assigned correctly to the coordinating `LlmChatCompletionMessage` and is verified in the UI
- [x] appropriate tests are created

### Additional Info

Example of `null` showing up:

Image

Possible solution (tool name and tool call id) if we can't find content:

Image

amychisholm03:
## Description

While debugging a LangGraph <> OpenAI application, I noticed that a `LlmChatCompletionMessage` from OpenAI had `null` content in the UI. Clicking on the message reveals that it's from a tool call (stop reason: `'tool_calls'`). Because this is an outgoing tool call (OpenAI -> LangChain/Graph) , OpenAI has no idea what the result is yet so we cannot use that for `content`.

Following what the Python agent has done, we will not create a `LlmChatCompletionMessage` for this type of response. There will be a second conversation which will contain the user request, the tool message with the result, and then the chat response from the LLM, and we'll create LLM message events normally for that case.

## How to Test

```
```

## Related Issues

Closes #3644

### DIFF
--- a/lib/subscribers/openai/utils.js
+++ b/lib/subscribers/openai/utils.js
@@ -31,14 +31,14 @@ function getMessageFromResponse(response) {
     content = response?.output?.[0]?.content?.[0]?.text
     role = response?.output?.[0]?.role
   } else {
-    const msg = response?.choices?.[0]?.message
-    if (msg?.tool_calls?.[0]) {
+    const choice = response?.choices?.[0]
+    if (choice?.finish_reason === 'tool_calls') {
       // A false response. Don't create a LlmChatCompletionMessage
       // for this -- the full conversation isn't done yet.
       return []
     }
-    content = msg?.content
-    role = msg?.role
+    content = choice?.message?.content
+    role = choice?.message?.role
   }
 
   return [{ content, role }]


lib/llm-events/openai/chat-completion-message.js
```diff
@@ -37,6 +37,15 @@ module.exports = class LlmChatCompletionMessage extends LlmEvent {
...
     if (agent.config.ai_monitoring.record_content.enabled === true) {
       this.content = content
+      if (this.is_response && response?.choices?.[0]?.finish_reason === 'tool_calls') {
+        // `content` is null on outgoing tool calls, and OpenAI doesn't
+        // know the result yet. For a better UI experience, we will
+        // provide a description of the tool being called and assign
+        // it to `content`.
+        const toolCall = response?.choices?.[0]?.message?.tool_calls?.[0]
+        const toolArgs = JSON.parse(toolCall?.function?.arguments || '{}')
+        this.content = `Tool call: ${toolCall?.function?.name}(${JSON.stringify(toolArgs)})`
```
jsumners-nr: Both `JSON.parse` and `JSON.stringify` throw errors that are not accounted for here. Also, unless this behavior has been added to the AIM spec, this will be an inconsistency between language agents. There should at least be an internal discussion to determine the appropriate action here. Do you agree @bizob2828?

bizob2828: I agree. Our goal isn't to always provide content values. It's to provide them when they're available.

amychisholm03: The current spec is ambiguous in this scenario; it just asks for a `string` that is the "content of the message". Normally `content` is defined somewhere on the response, but for this outgoing tool call, that is not the case. Uma told me for Python she found the content "in a different place", but I truly could not find the outgoing tool call result anywhere.

bizob2828 (commented):
This seems like a feature request.  It appears there is an idea of function calling in openai which we don't have a published spec nor any requests for at the moment

lib/llm-events/openai/chat-completion-message.js
```diff
@@ -42,6 +42,15 @@ module.exports = class LlmChatCompletionMessage extends LlmEvent {
     const content = message?.content
     if (agent.config.ai_monitoring.record_content.enabled === true) {
       this.content = content
+      if (this.is_response && response?.choices?.[0]?.finish_reason === 'tool_calls') {
```
bizob2828: based on the python agent logic, we should not create a LlmChatCompletionMessage if the  finish reason is `tool_calls`.

amychisholm03: Implemented this change.

lib/subscribers/openai/utils.js
```diff
@@ -30,11 +31,17 @@ function getMessageFromResponse(response) {
     content = response?.output?.[0]?.content?.[0]?.text
     role = response?.output?.[0]?.role
   } else {
-    content = response?.choices?.[0]?.message?.content
-    role = response?.choices?.[0]?.message?.role
+    const msg = response?.choices?.[0]?.message
+    if (msg?.tool_calls?.[0]) {
```
bizob2828: i think you want to use the `finish_reason` being `tool_calls` to determine this

bizob2828 (approved):

Issue #3582, opened 2025-12-09
Input and output `LlmChatCompletionMessage`s should not have identical timestamps


amychisholm03:
### Description
An issue has been identified with our timestamping of certain `LlmEvents`, namely `LlmChatCompletionMessage` and `LlmChatCompletionSummary`: "The timestamp for input and output messages in LangChain traces is currently the same because both events are recorded at the end of the span. This may also contribute to ordering and display issues and will need a long-term solution (proposed as adding a separate, accurate timestamp attribute to the custom event data)."

`LlmChatCompletionMessage`s should have unique timestamps for when they were created, not when they were aggregated into a `LlmChatCompletionSummary`.

This logic should be applied to all LLMs, not just Langchain.

### Acceptance Criteria

- [ ] UI reflects this timestamp change
- [ ] appropriate tests are created

### Additional context
[Jira link](https://new-relic.atlassian.net/browse/NR-487636)

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-499503

bizob2828:
@amychisholm03 This ticket should be broken up instead of having 1 with a task list of areas to update

PR #3657, opened 2026-01-13
feat: Add `timestamp` to OpenAI input `LlmChatCompletionMessage`s

amychisholm03:
## Description

Adds the `timestamp` field to OpenAI request/input `LlmChatCompletionMessage`s. This value is derived from the context's segment's start time. When the helper function, `recordChatCompletions`, is called at the end of the function, it will use this timestamp instead of `Date.now()` when adding request/input `LlmChatCompletionMessage` events to the event aggregator.

Also fixes segment duration issue by having the OpenAI subscriber create segments with `Subscriber.createSegment` instead of `agent.tracer.createSegment`, so the segment will actually be started.

## How to Test

```
npm run unit
```

## Related Issues

Part of #3582

### DIFF
--- a/test/unit/llm-events/openai/chat-completion-message.test.js
+++ b/test/unit/llm-events/openai/chat-completion-message.test.js
@@ -345,7 +345,6 @@ test('openai.responses.create', async (t) => {
         const summaryId = 'chat-summary-id'
         const content = req.input
         const role = 'user'
-        const inputTimestamp = Date.now()
         const chatMessageEvent = new LlmChatCompletionMessage({
           transaction: tx,
           agent,
@@ -354,12 +353,11 @@ test('openai.responses.create', async (t) => {
           response: chatRes,
           completionId: summaryId,
           message: { content, role }, // lib/instrumentation/openai.js sets this object up
-          index: 0,
-          inputTimestamp
+          index: 0
         })
         const expected = getExpectedResult(tx, { id: 'resp_id-0' }, 'message', summaryId)
         expected.token_count = 0
-        expected.timestamp = inputTimestamp
+        expected.timestamp = segment.timer.start
         assert.deepEqual(chatMessageEvent, expected)
         end()
       })


lib/subscribers/openai/chat.js
```diff
@@ -67,7 +69,8 @@ class OpenAIChatCompletions extends OpenAISubscriber {
         response,
         segment: ctx.segment,
         transaction: ctx.transaction,
-        err
+        err,
+        inputTimestamp: ctx?._extras?.llmChatInputTime
```
jsumners-nr: It doesn't seem possible for `ctx` to be undefined at this point. Is that accurate?

lib/subscribers/openai/chat.js
```diff
@@ -78,7 +81,8 @@ class OpenAIChatCompletions extends OpenAISubscriber {
         request,
         response,
         headers: ctx.extras?.headers,
-        err
+        err,
+        inputTimestamp: ctx?._extras?.llmChatInputTime
```
jsumners-nr: Same as above.

lib/subscribers/openai/utils.js
```diff
@@ -82,9 +85,16 @@ function getMessagesFromRequest(request, logger) {
  */
 function recordEvent({ agent, type, msg }) {
   const llmContext = extractLlmContext(agent)
-
+  // Spec 771: The `timestamp` attribute MUST be used to override
+  // the timestamp value attached to the intrinsics of the custom
+  // LlmChatCompletionMessage event before it is recorded using
+  // record_custom_event().
+  let timestamp = msg?.timestamp
```
jsumners-nr: Isn't `msg` a required parameter? Shouldn't this be?:
```suggestion
  const timestamp = msg.timestamp ?? Date.now()
```

lib/subscribers/openai/chat.js
```diff
@@ -26,6 +26,7 @@ class OpenAIChatCompletions extends OpenAISubscriber {
   }
 
   handler(data, ctx) {
+    const llmChatInputTime = Date.now()
```
bizob2828: as mentioned in the langchain PR, I would just use the segment's start time instead of keeping track of another time

amychisholm03: Would it just be `ctx.segment.timer.start`? I'm looking at it now and it is null in `asyncEnd`. I fixed it by touching the segment beforehand.

amychisholm03: Courtesy of related LangChain PR - will update to use `Subscriber.createSegment` instead of `agent.tracer.createSegment`.

lib/subscribers/openai/chat.js
```diff
@@ -67,7 +64,8 @@ class OpenAIChatCompletions extends OpenAISubscriber {
         response,
         segment: ctx.segment,
         transaction: ctx.transaction,
-        err
+        err,
+        inputTimestamp: ctx.segment.timer.start
```
bizob2828: this is coming from the segment which is a param already. seems like you can just pull that from the segment when creating the LlmChatCompletionMessage and avoid updating a few signatures to pass in _a_ key from an object that is already passed in

amychisholm03: Yeah, makes sense. I'll make that change

test/unit/llm-events/openai/chat-completion-message.test.js
```diff
@@ -352,10 +354,12 @@ test('openai.responses.create', async (t) => {
           response: chatRes,
           completionId: summaryId,
           message: { content, role }, // lib/instrumentation/openai.js sets this object up
-          index: 0
+          index: 0,
+          inputTimestamp
```
bizob2828: this is no longer an arg

PR #3658, opened 2026-01-13
feat: Add `timestamp` to LangChain input `LlmChatCompletionMessage`s

### LINKED ISSUE #3582: Input and output `LlmChatCompletionMessage`s should not have identical timestamps

amychisholm03:
## Description

Adds the timestamp field to LangChain request/input `LlmChatCompletionMessage`s. This value is derived from the coordinating segment's start time. When the helper function, `recordChatCompletionEvents`, is called at the end of the instrumentation, it will use this timestamp instead of `Date.now()` when adding request/input `LlmChatCompletionMessage` events to the event aggregator.

## How to Test

```
npm run unit
```

## Related Issues

Closes #3582


lib/subscribers/langchain/runnable-stream.js
```diff
@@ -25,14 +25,17 @@ class LangchainRunnableStreamSubscriber extends LangchainSubscriber {
       logger.debug('Langchain instrumentation is disabled, not creating segment.')
       return ctx
     }
+    const llmChatInputTime = Date.now()
 
     // Create segment
     const segment = agent.tracer.createSegment({
```
bizob2828: you can just use the segment start time instead of creating another time and passing it along in the context extras

lib/subscribers/langchain/runnable-stream.js
```diff
@@ -32,6 +32,7 @@ class LangchainRunnableStreamSubscriber extends LangchainSubscriber {
       parent: ctx.segment,
       transaction: ctx.transaction
     })
+    segment.touch()
```
bizob2828: you don't `touch` the segment. you should be starting the segment but we have this already handled for you in the subscriber base `createSegment` method it also handles other things. we shouldn't be using `agent.tracer.createSegment` directly in any of these subscribers, that's prob why durations are wonky

amychisholm03: I'll update the LangChain subscribers to use `Subscriber.createSegment` instead.

PR #3665, opened 2026-01-15
fix: Updated openai instrumentation to properly parse headers when a call fails

bizob2828:
## Description

I noticed while triaging the fix from #3664 that we weren't properly assigning keys on Llm events if the call failed.  This is because when we migrated this to a subscriber based call, we didn't handle the difference betweeen good and bad calls with openai. If an openai call fails the headers live on the error not the result.  This PR fixes that and also handles the difference in the shape of the response headers in v4 vs v5 of openai.  I also added 2 assertions to error calls to ensure it's assigning a relevant header value to a LLm event.

## How to Test

```sh
npm run versioned:internal openai
```

### DIFF
--- a/test/versioned/openai/mock-responses-api-responses.js
+++ b/test/versioned/openai/mock-responses-api-responses.js
@@ -10,6 +10,7 @@ module.exports = responses
 
 responses.set('Invalid API key.', {
   headers: {
+    'content-type': 'application/json',
     'x-request-id': 'req_f3b1353d6a35554bcc6d6e0cbf07ad4b'
   },
   code: 401,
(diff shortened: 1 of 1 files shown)


test/versioned/openai/mock-responses-api-responses.js
amychisholm03, line 33:
```suggestion
  headers: {
    'x-request-id': 'req_f3b1353d6a35554bcc6d6e0cbf07ad4b'
  },
  code: 400,
```

amychisholm03, line 102:
```suggestion
headers: {
  'x-request-id': 'req_f3b1353d6a35554bcc6d6e0cbf07ad4b'
},
```

bizob2828:
> Should have headers in the other error mock responses.

Sure but I wasn't using this PR to update all the mocks

bizob2828:
Ok there aren't many mocks missing headers ill update here

amychisholm03 (approved):

Issue #3617, opened 2025-12-19
Implement "essential" mode support for span links


jsumners-nr:
Spec PR 769 defines the actions that need to be taken when the agent is configured for partial tracing and span links are present. For the "reduced" mode, we need to:

1. Perform all actions as described in #3616 
2. Remove all agent and user attributes attached to the link

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-503466

PR #3670, opened 2026-01-16
feat: Added essential tracing mode for span links

svetlanabrennan:
## Description

For essential tracing, we need to do two things:

1. move the span links from a dropped span (if there are any spank links) to the nearest parent span that was not dropped and update the id attribute of the links to match the id of the new parent span.
2. remove all agent and user attributes attached to the span links

## How to Test

Run `npm run unit`

## Related Issues

Closes issue #3617

### DIFF
--- a/test/unit/spans/span-event-aggregator.test.js
+++ b/test/unit/spans/span-event-aggregator.test.js
@@ -13,7 +13,7 @@ const SpanEventAggregator = require('../../../lib/spans/span-event-aggregator')
 const Metrics = require('../../../lib/metrics')
 const SpanLink = require('#agentlib/spans/span-link.js')
 const { PARTIAL_TYPES } = require('#agentlib/transaction/index.js')
-const SpanEvent = require('../../../lib/spans/span-event')
+const SpanEvent = require('#agentlib/spans/span-event.js')
 
 const RUN_ID = 1337
 const DEFAULT_LIMIT = 2000


lib/transaction/trace/partial-trace.js
```diff
@@ -86,6 +94,21 @@ class PartialTrace {
...
+   * @returns {SpanLink[]} spanLinks with only intrinsic attributes
+   */
+  removeNonIntrinsicsAttrs(spanLinks) {
+    for (const link of spanLinks) {
+      link.userAttributes = new Attributes({ scope: Attributes.SCOPE_SEGMENT })
+      link.agentAttributes = new Attributes({ scope: Attributes.SCOPE_SEGMENT })
+    }
+
+    return spanLinks
+  }
```
jsumners-nr, line 107: Unless I'm mistaken, this function mutates the passed in `spanLinks`. So there isn't any need to `return spanLinks`.

svetlanabrennan: You're right. Good catch. I'll change that right now.

test/unit/spans/span-event-aggregator.test.js
```diff
@@ -13,6 +13,7 @@ const SpanEventAggregator = require('../../../lib/spans/span-event-aggregator')
 const Metrics = require('../../../lib/metrics')
 const SpanLink = require('#agentlib/spans/span-link.js')
 const { PARTIAL_TYPES } = require('#agentlib/transaction/index.js')
+const SpanEvent = require('../../../lib/spans/span-event')
```
jsumners-nr:
```suggestion
const SpanEvent = require('#agentlib/spans/span-event.js')
```

jsumners-nr (approved):

PR #3674, opened 2026-01-20
chore: Revert mongodb+bson pin

jsumners-nr:
This PR reverts #3671. They released `bson@7.1.1` to correct the issue.

### DIFF
--- a/test/versioned/disabled-instrumentation/package.json
+++ b/test/versioned/disabled-instrumentation/package.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "express": "latest",
         "ioredis": "latest",
-        "mongodb": "latest",
+        "mongodb": "latest"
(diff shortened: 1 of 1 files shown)


test/versioned/disabled-instrumentation/package.json
```diff
@@ -12,7 +12,7 @@
         "express": "latest",
         "ioredis": "latest",
         "mongodb": "latest",
-        "bson": "7.0.0"
+        "bson": "latest"
```
amychisholm03:
```suggestion
```

`bson` dep wasn't explicitly specified before; should be safe to remove entirely.

amychisholm03 (changes requested):

amychisholm03 (dismissed):

PR #3676, opened 2026-01-20
refactor: Added AIM abstract subscribers and updated all AIM subscribers to leverage them

### LINKED ISSUE #3487: Create a base AIM subscriber to abstract handling `Llm*` events

bizob2828:
## Description

This PR creates abstractions around AIM.  Most AIM libraries at the very least check if enabled and create a segment from a constant(or slightly dynamic at times).  The chat AIM subscriber adds a simple recordCompletionEvents which requires the inherited subscriber to implement 3 methods: getting the prompt/responses from call, creating llm summary and creating n llm completion messages.  The embedding subscriber adds a simple recordEmbeddingEvent which requires the inherited subscrber to implement a method to create the embedding.

## How to Test

```sh
npm run versioned:internal:major
```

## Related Issues
Closes #3487

### DIFF
--- a/lib/subscribers/langchain/runnable-stream.js
+++ b/lib/subscribers/langchain/runnable-stream.js
@@ -17,13 +17,13 @@ class LangchainRunnableStreamSubscriber extends LangchainRunnableSubscriber {
       return
     }
 
-    const ctx = this.agent.tracer.getContext()
-
     if (!this.streamingEnabled) {
       this.logger.debug('`ai_monitoring.streaming.enabled` is set to false, stream will not be instrumented.')
       this.agent.metrics.getOrCreateMetric(STREAMING_DISABLED).incrementCallCount()
       return
     }
+
+    const ctx = this.agent.tracer.getContext()
 
     const { transaction } = ctx
     if (transaction?.isActive() !== true) {
--- a/lib/subscribers/mcp-sdk/client-request.js
+++ b/lib/subscribers/mcp-sdk/client-request.js
@@ -11,7 +11,7 @@ const { MCP } = require('../../metrics/names')
 class McpClientRequestSubscriber extends AiMonitoringSubscriber {
   constructor({ agent, logger }) {
     // name is assigned but overwritten in handler as the name is dynamic
-    super({ agent, logger, packageName: '@modelcontextprotocol/sdk', channelName: 'nr_request', trackingPrefix: MCP.TRACKING_PREFIX, name: 'temp' })
+    super({ agent, logger, packageName: '@modelcontextprotocol/sdk', channelName: 'nr_request', trackingPrefix: MCP.TRACKING_PREFIX, name: 'unknown' })
     this.events = ['asyncEnd']
   }
 
--- a/lib/subscribers/openai/chat.js
+++ b/lib/subscribers/openai/chat.js
@@ -45,9 +45,18 @@ class OpenAIChatCompletions extends AiMonitoringChatSubscriber {
       return
     }
 
-    const ctx = this.agent.tracer.getContext()
     const { result: response, arguments: args, error: err } = data
     const [request] = args
+
+    if (request.stream && !this.streamingEnabled) {
+      this.logger.debug(
+        '`ai_monitoring.streaming.enabled` is set to `false`, stream will not be instrumented.'
+      )
+      this.agent.metrics.getOrCreateMetric(AI.STREAMING_DISABLED).incrementCallCount()
+      return
+    }
+
+    const ctx = this.agent.tracer.getContext()
 
     if (request.stream) {
       this.instrumentStream({
@@ -80,14 +89,6 @@ class OpenAIChatCompletions extends AiMonitoringChatSubscriber {
    * @param {Error} [params.err] error if it exists
    */
   instrumentStream({ ctx, request, response, err = null }) {
-    if (!this.streamingEnabled) {
-      this.logger.debug(
-        '`ai_monitoring.streaming.enabled` is set to `false`, stream will not be instrumented.'
-      )
-      this.agent.metrics.getOrCreateMetric(AI.STREAMING_DISABLED).incrementCallCount()
-      return
-    }
-
     const self = this
     if (!(ctx?.segment || ctx?.transaction)) {
       this.logger.debug('Empty context, not instrumenting stream')


lib/subscribers/ai-monitoring/chat.js
```diff
@@ -0,0 +1,73 @@
...
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+const AiMonitoringSubscriber = require('./base')
+const LlmErrorMessage = require('#agentlib/llm-events/error-message.js')
+
+class AiMonitoringChatSubscriber extends AiMonitoringSubscriber {
+  constructor({ agent, logger, packageName, channelName }) {
```
amychisholm03: A note about `trackingPrefix` and `name` needing to be defined on the child subscriber would be helpful.

lib/subscribers/ai-monitoring/base.js
```diff
@@ -0,0 +1,76 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+const Subscriber = require('../base')
+const { DESTINATIONS } = require('#agentlib/config/attribute-filter.js')
+const { extractLlmContext } = require('#agentlib/util/llm-utils.js')
+
+class AiMonitoringSubscriber extends Subscriber {
+  constructor({ agent, logger, packageName, channelName }) {
```
amychisholm03: You should have a constructor description or a comment that explains that `trackingPrefix` needs to be defined.

lib/subscribers/ai-monitoring/chat.js
```diff
@@ -0,0 +1,73 @@
...
+    throw new Error('createCompletionMessage must be implemented by your subscriber')
+  }
+
+  createCompletionSummary({ ctx, request, response, err, metadata, tags }) {
+    throw new Error('createCompletionSummary must be implemented by your subscriber')
+  }
+
+  getMessages({ request, response }) {
+    throw new Error('getMessages must be implemented by your subscriber')
+  }
```
amychisholm03: Function descriptions outlining the expected result of these functions would be helpful. If we refactor `llm-events` to have one class for e.g. `LlmChatCompletionMessage`, then we could more accurately describe these functions, but that can be done in a later PR.

lib/subscribers/ai-monitoring/embedding.js
```diff
@@ -0,0 +1,52 @@
...
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+const AiMonitoringSubscriber = require('./base')
+const LlmErrorMessage = require('#agentlib/llm-events/error-message.js')
+
+class AiMonitoringEmbeddingSubscriber extends AiMonitoringSubscriber {
+  constructor({ agent, logger, packageName, channelName }) {
```
amychisholm03: You should have a constructor description or a comment that explains that `name` and `trackingPrefix` need to be defined.

lib/subscribers/ai-monitoring/base.js
```diff
@@ -0,0 +1,76 @@
...
+    this.agent.customEventAggregator.add([
+      { type, timestamp },
+      Object.assign({}, msg, llmContext)
+    ])
+  }
+
+  /**
+   * Defines the common handler for all AIM calls.
+   * All of these calls simply create a segment with a name
+   * The `this.name` must be defined in the inherited subscriber
```
jsumners-nr, line 93: Should this class define?:

```js
get name () { throw Error('not implemented') }
```

-----

A quick test shows me that this might help us track down errors in implementations:

```js
'use strict'

class Foo {
  greet () {
    console.log(`hello ${this.name}`)
  }

  get name () { throw Error('not implemented') }
}

class Bar extends Foo {}

class Baz extends Foo {
  get name () { return 'world' }
}

const bar = new Bar()
const baz = new Baz()

//bar.greet()
baz.greet()
```

lib/subscribers/ai-monitoring/embedding.js
```diff
@@ -0,0 +1,52 @@
...
+    super({ agent, logger, packageName, channelName })
+  }
+
+  createEmbedding({ ctx, request, response, err }) {
+    throw new Error('createEmbedding must be implemented by your subscriber')
+  }
+
+  /**
+   * Function to create embedding
+   *  as well as assign errors to transaction if present
```
amychisholm03:
```suggestion
   * Function to create embedding as well as assign errors to transaction if present
```

lib/subscribers/ai-monitoring/chat.js
```diff
@@ -0,0 +1,73 @@
...
+
+    if (!(ctx?.segment || ctx?.transaction)) {
+      this.logger.debug('Empty context, not creating completion events.')
+      return
+    }
+
+    // Explicitly end segment to provide consistent duration
+    // for both LLM events and the segment
+    ctx.segment.end()
+    const completionSummary = this.createCompletionSummary({ ctx, request, response, err, metadata, tags })
```
amychisholm03, line 100: Nit - I think `createCompletionSummary` should be called right before `this.recordEvent({ type: 'LlmChatCompletionSummary', msg: completionSummary })`, just easier to follow

lib/subscribers/google-genai/embed-content.js
```diff
@@ -3,67 +3,46 @@
...
-    // Explicitly end segment to get consistent duration
-    // for both LLM events and the segment
-    ctx.segment.end()
-
-    const embedding = new LlmEmbedding({
-      agent,
-      segment: ctx.segment,
-      transaction: ctx.transaction,
+    this.addLlmMeta({ ctx, version: data.moduleVersion })
+    this.recordEmbedding({
```
amychisholm03: Because `recordEmbedding` is checking for proper `ctx`, it should be called before `this.addLlmMeta`.

lib/subscribers/ai-monitoring/embedding.js
```diff
@@ -0,0 +1,52 @@
...
+   * @param {object} params to function
+   * @param {Context} params.ctx active context
+   * @param {object} params.request request made to method on a given llm library
+   * @param {object} params.response response from method on a given llm library
+   * @param {object} params.err error if present
+   */
+  recordEmbedding({ ctx, request, response, err }) {
+    if (!this.enabled) {
+      this.logger.debug('config.ai_monitoring.enabled is set to false, not creating embedding event.')
```
jsumners-nr, line 50: Do you think we should tag these logs? As things are, we won't see in the logs which specific thing is disabled. Something like:

```js
this.logger.debug({ aimLib: this }, 'blah blah blah')
```

-----

```js
'use strict'

class Foo {
  greet () {
    console.log(`hello ${this}`)
  }

  // Pick one. 🤷‍♂️
  get [Symbol.toStringTag] () { return 'Foo' }
  toString () { return 'Foo' }
}

class Bar extends Foo {
  get name () { return 'world' }

  get [Symbol.toStringTag] () { return 'Bar' }
  toString () { return 'Bar' }
}

const foo = new Foo()
const bar = new Bar()

foo.greet()
bar.greet()
```

lib/subscribers/langchain/tool.js
```diff
@@ -44,7 +36,7 @@ class LangchainToolSubscriber extends LangchainSubscriber {
     if (transaction?.isActive() !== true) {
       return
     }
-    const { moduleVersion: pkgVersion, result, error: err } = data
+    const { moduleVersion: result, error: err } = data
```
amychisholm03:
```suggestion
    const { result, error: err } = data
```
typo - this will fix your langchain tool test failing

amychisholm03 (changes requested):
Just some documentation nits mainly; I also found the typo that's resulting in the langchain tool test failing (lib/subscribers/langchain/tool.js‎ Line 39)

lib/subscribers/utils.js
```diff
@@ -0,0 +1,35 @@
...
+ * We are now wrapping all openai functions in a `traceSync` call.
+ * Then we attach resolve/rejection handlers to indicate a promise is finished.
+ * We are not re-throwing the error because that would create an unhandled promise rejection, since the error is getting handled in openai code.
+ * OpenAI has a [custom promise](https://github.com/openai/openai-node/blob/master/src/core/api-promise.ts).
+ * see: https://github.com/newrelic/node-newrelic/issues/3379
+ * see: https://github.com/nodejs/node/issues/59936
+ * @param {object} data the data associated with the `end` event
+ * @returns {Promise|void} If not a thenable then returns the "promise", otherwise wraps promise in tracing channel methods
+ */
+function wrapPromise(data) {
```
jsumners-nr, line 21: This seems to be OpenAI specific. If so, I'd home it under `subscribers/openai/`.

bizob2828: it's not, it's used by ioredis as well

bizob2828:
thanks for the review y'all. this PR looks sloppy in some cases and that's due to a last min refactor I had. I'll address all the comments and ask for another round of review

lib/subscribers/ai-monitoring/chat.js
bizob2828: we will never have 1 class for LLM messages. there's too much nuance which is why they're abstracted currently

lib/subscribers/ai-monitoring/chat.js
bizob2828: so im digging into this because I borrowed the abstracted flow from openai which had this and not sure why

bizob2828: ok i see why this was done, i can remove it now. it was because we added headers to the response object then removed it. I'm now spreading the response and headers so it's a new object

lib/subscribers/ai-monitoring/embedding.js
bizob2828: the logger lines will have component `${packageName}-subscriber` already so this is not needed, [see](https://github.com/newrelic/node-newrelic/blob/23da6547e72f49bf798046330e8cfa74a0b4d2af/lib/subscribers/base.js#L77)

lib/subscribers/ai-monitoring/chat.js
```diff
@@ -0,0 +1,117 @@
...
+   * Function that must be implemented by inherited subscriber to create an llm completion message.
+   *
+   * @param {object} params to function
+   * @param {Context} params.ctx active context
+   * @param {object} params.request request made to method on a given llm library
+   * @param {object} params.response response from method on a given llm library
+   * @param {object|string} params.message the message object/string used to create llm completion message
+   * @param {string} params.completionId the id of the llm completion summary for a given conversation
+   * @param {number} params.index index of message for a given conversation
+   * returns {object} a llm completion message instance for the given LLM
```
bizob2828, line 40: these returns aren't jsdoc because this method doesn't return anything but wanted to document what it expects

test/versioned/openai/embeddings.test.js
```diff
@@ -201,8 +202,7 @@ test('should not create segment or llm events when ai_monitoring.enabled is fals
     const events = agent.customEventAggregator.events.toArray()
     assert.equal(events.length, 0, 'should not create llm events when ai_monitoring is disabled')
 
-    const children = tx.trace.segments.root.children
-    assert.equal(children.length, 0, 'should not create OpenAI completion segment')
+    assert.ok(!findSegment(tx.trace, tx.trace.root, OPENAI.EMBEDDING))
```
bizob2828, line 205: this was a bug that was fixed. basically if ai monitoring was enabled, it didn't return the current context so it never recorded the http external call

test/versioned/openai/chat-completions.test.js
```diff
@@ -542,8 +543,7 @@ test('chat.completions.create', async (t) => {
       const events = agent.customEventAggregator.events.toArray()
       assert.equal(events.length, 0, 'should not create llm events when ai_monitoring is disabled')
 
-      const children = tx.trace.segments.root.children
-      assert.equal(children.length, 0, 'should not create OpenAI completion segment')
+      assert.ok(!findSegment(tx.trace, tx.trace.root, OPENAI.COMPLETION))
```
bizob2828, line 546: this was a bug that was fixed. basically if ai monitoring was enabled, it didn't return the current context so it never recorded the http external call

test/versioned/openai/chat-completions-res-api.test.js
```diff
@@ -202,8 +203,7 @@ test('responses.create', async (t) => {
       const events = agent.customEventAggregator.events.toArray()
       assert.equal(events.length, 0, 'should not create llm events when ai_monitoring is disabled')
 
-      const children = tx.trace.segments.root.children
-      assert.equal(children.length, 0, 'should not create OpenAI completion segment')
+      assert.ok(!findSegment(tx.trace, tx.trace.root, OPENAI.COMPLETION))
```
bizob2828, line 206: this was a bug that was fixed. basically if ai monitoring was enabled, it didn't return the current context so it never recorded the http external call

test/versioned/langchain/runnables.js
```diff
@@ -455,7 +455,7 @@ function runRunnablesTests(config) {
         const [, chainEvent] = event
         return chainEvent.vendor === 'langchain'
       })
-      assert.equal(langchainEvents.length, 3, 'should create 3 langchain events')
+      assert.equal(langchainEvents.length, 2, 'should create 2 langchain events')
```
bizob2828, line 458: this was also a bug where it created a llm completion message for a response that didn't exist

test/versioned/langchain/runnables-streaming.js
```diff
@@ -114,7 +114,7 @@ function runStreamingEnabledTests(config) {
           const metrics = agent.metrics.getOrCreateMetric(
             `Supportability/Nodejs/ML/LangChain/${langchainCoreVersion}`
           )
-          assert.equal(metrics.callCount, 3)
+          assert.equal(metrics.callCount, 1)
```
bizob2828, line 117: this fixes logging this metric too frequently. now it only creates it when the segment is created

test/versioned/langchain/runnables.js
```diff
@@ -94,7 +94,7 @@ function runRunnablesTests(config) {
       const metrics = agent.metrics.getOrCreateMetric(
         `Supportability/Nodejs/ML/LangChain/${langchainCoreVersion}`
       )
-      assert.equal(metrics.callCount, 3)
+      assert.equal(metrics.callCount, 1)
```
bizob2828, line 97: this fixes logging this metric too frequently. now it only creates it when the segment is created

lib/subscribers/google-genai/generate-content-stream.js
```diff
@@ -95,29 +86,16 @@ class GoogleGenAIGenerateContentStreamSubscriber extends GoogleGenAIGenerateCont
       this.agent.metrics.getOrCreateMetric(AI.STREAMING_DISABLED).incrementCallCount()
       return
     }
```
amychisholm03, line 88: A nested approach for lines 76-88 may be better.

```
// Check config to see if ai_monitoring is still enabled
if (!this.enabled) {
   
    if (!this.streamingEnabled) {
      this.logger.warn(
        '`ai_monitoring.streaming.enabled` is set to `false`, stream will not be instrumented.'
      )
      this.agent.metrics.getOrCreateMetric(AI.STREAMING_DISABLED).incrementCallCount()
      return
   }

     this.logger.debug('`ai_monitoring.enabled` is set to false, stream will not be instrumented.')
      return
}
```

lib/subscribers/langchain/runnable-stream.js
```diff
@@ -3,53 +3,29 @@
...
-      if (!this.streamingEnabled) {
-        logger.warn('Langchain streaming instrumentation is disabled, stream will not be instrumented.')
-        agent.metrics.getOrCreateMetric(STREAMING_DISABLED).incrementCallCount()
-      }
-      agent.metrics
-        .getOrCreateMetric(`${LANGCHAIN.TRACKING_PREFIX}/${data?.moduleVersion}`)
-        .incrementCallCount()
+    if (!this.streamingEnabled) {
+      this.logger.debug('`ai_monitoring.streaming.enabled` is set to false, stream will not be instrumented.')
+      this.agent.metrics.getOrCreateMetric(STREAMING_DISABLED).incrementCallCount()
```
amychisholm03, line 22: `ctx` assignment should be after `this.streamingEnabled` check.

Also, I think the nested enabled check is more readable:
```
 // Exit early if need be.
    if (!this.enabled) {
      if (!this.streamingEnabled) {
        logger.warn('Langchain streaming instrumentation is disabled, stream will not be instrumented.')
        agent.metrics.getOrCreateMetric(STREAMING_DISABLED).incrementCallCount()
      }
      agent.metrics
        .getOrCreateMetric(`${LANGCHAIN.TRACKING_PREFIX}/${data?.moduleVersion}`)
        .incrementCallCount()
      return
    }
```

lib/subscribers/mcp-sdk/client-request.js
```diff
@@ -5,19 +5,16 @@
...
-const Subscriber = require('../base')
+const { AiMonitoringSubscriber } = require('../ai-monitoring')
 const { MCP } = require('../../metrics/names')
 
-class McpClientRequestSubscriber extends Subscriber {
+class McpClientRequestSubscriber extends AiMonitoringSubscriber {
   constructor({ agent, logger }) {
-    super({ agent, logger, packageName: '@modelcontextprotocol/sdk', channelName: 'nr_request' })
+    // name is assigned but overwritten in handler as the name is dynamic
+    super({ agent, logger, packageName: '@modelcontextprotocol/sdk', channelName: 'nr_request', trackingPrefix: MCP.TRACKING_PREFIX, name: 'temp' })
```
amychisholm03:
```suggestion
    super({ agent, logger, packageName: '@modelcontextprotocol/sdk', channelName: 'nr_request', trackingPrefix: MCP.TRACKING_PREFIX, name: 'Unknown' })
```
Nit - but I'm not sure `temp` is the best placeholder we want to use

test/versioned/google-genai/chat-completions.test.js
```diff
@@ -379,11 +379,14 @@ test('should not create llm events when ai_monitoring.streaming.enabled is false
       'Supportability/Nodejs/ML/Streaming/Disabled'
     )
     assert.equal(streamingDisabled.callCount > 0, true)
-
-    const activeSeg = agent.tracer.getSegment()
-    assert.equal(activeSeg?.isRoot, true)
-    const children = tx.trace.getChildren(activeSeg.id)
-    assert.notEqual(children?.[0]?.name, GEMINI.COMPLETION)
+    const name = `External/${host}:${port}/v1beta/models/${model}:streamGenerateContent`
+    // Should still create the Gemini segment since ai_monitoring is enabled
```
amychisholm03, line 383: Is there a Google Gen AI test that asserts no segments and no llm events if ai_monitoring.enabled and ai_monitoring.streaming.enabled are both false?

amychisholm03 (changes requested):
A few more suggestions

lib/subscribers/ai-monitoring/base.js
```diff
@@ -0,0 +1,113 @@
...
+  get name() {
+    return this._name
+  }
+
+  set trackingPrefix(prefix) {
+    if (!prefix) {
+      throw new Error('subscriber must pass in `trackingPrefix`')
+    }
+    this._prefix = prefix
+  }
```
jsumners-nr, line 49: 👍

lib/subscribers/ai-monitoring/chat.js
jsumners-nr: I understand what you're saying, but when I look at the OpenAI implementation, the override method does not include its own docblock. So this docblock is going to be used by tooling to aid a developer working with this method. So I'd still:

```suggestion
   * @returns {object} a llm completion message instance for the given LLM
```

Along with the other such instances.

bizob2828: linting fails though because this actually doesn't return anything

lib/subscribers/google-genai/generate-content-stream.js
bizob2828: i don't think your suggestion is logically correct.  the streaming enabled flag is only in play if ai monitoring is enabled. so there's no reason to check it if ai monitoring is disabled

lib/subscribers/langchain/runnable-stream.js
bizob2828: i will move the context retrieval but nesting this isn't technically correct and i explained above. I have issue with the fact we have to repeat the `!this.enabled` and `!this.streamingEnabled` checks in certain `asyncEnd` calls but couldn't figure out a reasonable way to abstract this

test/versioned/google-genai/chat-completions.test.js
bizob2828: no but none of the libraries have that check because they are mutually exclusive.  if ai monitoring is not enabled it exits

PR #3677, opened 2026-01-21
chore: Fix CI issue when sec-agent present

jsumners-nr:
When the security agent is present, the version of `undici` that ships with it will be resolved first (see https://github.com/newrelic/node-newrelic/actions/runs/21203677985/job/60994941999). This means that when we are, for example, testing against version 5 then the version 7 that the sec-agent ships with will be found instead. This PR avoids the problem by just skipping the metrics test when the security agent is loaded.

### DIFF
--- a/test/versioned/undici/requests.test.js
+++ b/test/versioned/undici/requests.test.js
@@ -32,10 +32,12 @@ test.beforeEach((ctx) => {
   }
 })
 
-test.afterEach((ctx) => {
+test.afterEach(afterEach)
+
+function afterEach(ctx) {
   helper.unloadAgent(ctx.nr.agent)
   ctx.nr.server.close()
-})
+}
 
 test('should not fail if request not in a transaction', async (t) => {
   const { undici, REQUEST_URL } = t.nr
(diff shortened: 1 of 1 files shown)


test/versioned/undici/requests.test.js
```diff
@@ -52,8 +54,16 @@ test('should not fail if request not in a transaction', async (t) => {
 })
 
 test('should create tracking metrics', (t, end) => {
-  t.plan(5)
   const { agent, pkgVersion, undici, REQUEST_URL } = t.nr
+  if (helper.isSecurityAgentEnabled(agent) === true) {
```
bizob2828, line 58: this is a shame we can't just use `test('should create tracking metrics', { skip: helper.isSecurityAgentEnabled }`

bizob2828 (approved):

test/versioned/undici/requests.test.js
jsumners-nr: Agreed. That's what I wanted to do at first, but then I looked at the implementation 😢.

PR #3680, opened 2026-01-21
test: Assert that span events drop in partial traces

### LINKED ISSUE #3647: Account for span events in core tracing
Spec PR 770 defines the actions that need to be taken when the agent is configured for partial tracing and span events are present. For all supported reduced trace granularity modes (reduced, essential, & compact), any events on a span must be dropped.

amychisholm03:
## Description
According to agent spec 770, we need to drop all `SpanEvent`s (aka timed events) generated from the OTel Bridge for partial traces when partial granularity is enabled. We already do this (see below), so I just wrote some tests to assert the behavior.

https://github.com/newrelic/node-newrelic/blob/23da6547e72f49bf798046330e8cfa74a0b4d2af/lib/spans/span-event-aggregator.js#L101-L106

## How to Test

```
npm run versioned:major otel-bridge
```

## Related Issues

Closes #3647

### DIFF
--- a/test/versioned/otel-bridge/timed-event.test.js
+++ b/test/versioned/otel-bridge/timed-event.test.js
@@ -184,19 +184,19 @@ test('should properly attach span event data for two span events', async (t) => 
 })
 
 test('should drop all span events if partial granularity is enabled with reduced type', async (t) => {
-  const { agent, REQUEST_URL, requestHook, registerInstrumentations, UndiciInstrumentation } = t.nr
+  const { agent, REQUEST_URL, requestHook, responseHook, registerInstrumentations, UndiciInstrumentation } = t.nr
   agent.config.distributed_tracing.sampler.partial_granularity.enabled = true
   agent.config.distributed_tracing.sampler.partial_granularity.type = 'reduced'
 
   registerInstrumentations([
     new UndiciInstrumentation({
-      requestHook
+      requestHook,
+      responseHook
     })
   ])
 
   await helper.runInTransaction(agent, async (tx) => {
     // Force tx to be partial
-    tx.priority = 42
     tx.partialTrace = true
     tx.partialType = PARTIAL_TYPES.REDUCED
     tx.createPartialTrace()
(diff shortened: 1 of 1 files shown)


test/versioned/otel-bridge/timed-event.test.js
```diff
@@ -181,3 +182,111 @@ test('should properly attach span event data for two span events', async (t) =>
...
+      body: Buffer.from('{"key":"value"}')
+    })
+    assert.equal(status, 200)
+    tx.end()
+    const segment = tx.trace.root
+    const segmentChildren = tx.trace.getChildren(segment.id)
+    const foundSegment = segmentChildren.find((c) => c.name === 'External/localhost/post')
+    assert.ok(foundSegment, 'should still create segment')
+    const spanEvents = agent.spanEventAggregator.getEvents()
+    assert.equal(spanEvents.length, 0, 'should not have span events')
```
svetlanabrennan, line 219: This should be asserting that timedEvents (aka span events in otel language) should not be present. So `spanEvents.timedEvents.length`. 

`agent.spanEventAggregator.getEvents()` gets the Span (we just call it spanEvent at the moment). 

If I understood the spec correctly, we can keep a span for a partial tracing mode but we have to drop the timedEvent (aka spanEvent).

Same note for the other tests.

amychisholm03: I don't see `spanEvents.timedEvents` because `spanEvents` is empty, but I do see `foundSegment.timedEvents`: https://github.com/newrelic/node-newrelic/blob/23da6547e72f49bf798046330e8cfa74a0b4d2af/lib/spans/span-event.js#L378

amychisholm03: The spec says: "By storing `SpanEvent` events on `Span` events, it is assured that if a `Span` event is dropped from the event reservoir, then its `SpanEvent` events will also be dropped, minimizing scenarios that could result in only one of those pieces of data getting stored in NRDB." Because we are not adding the `Span` event to the aggregator, neither are the `SpanEvent` events.

svetlanabrennan (approved):

PR #3681, opened 2026-01-21
feat: Added compact mode for span links

### LINKED ISSUE #3618: Implement "compact" mode support for span links
Spec PR 769 defines the actions that need to be taken when the agent is configured for partial tracing and span links are present. For the "compact" mode, we need to:

1. Perform all actions as described in #3617
2. Add a max limit of how many spanLink events are stored on a span event. 
3. Move the span links from all "compressed" spans to the new compressed (aggregate) span

svetlanabrennan:
## Description

For compact tracing, we do everthing we did in essential:

1. move the span links from a dropped span (if there are any spank links). If a span is dropped, move the span links to the nearest span (such as dropping a non exit span, or span with no entity relationship attributes). If a span is an duplicate exit span, move span links to the retained span.  And update the id attribute of the links to match the id of the new parent span. 
3. remove all agent and user attributes attached to the span links

And if we are keeping 1 span per entity, move the span links from the other spans we're dropping for that entity to the compressed span that we're keeping. 

I also moved the partial tracing span linking tests from the span aggregator test file to the the span links test file. 

## How to Test

Run `npm run unit`. 

## Related Issues

Closes #3618 

## TODO

- [x] Move all span linking partial tracing tests it's own test suite in the span-links.test.js file

### DIFF
--- a/test/unit/spans/helpers.js
+++ b/test/unit/spans/helpers.js
@@ -68,7 +68,7 @@ function setupPartialTrace({ agent, partialType, tx, addRootSpanLink = false, ad
     createSpanLink({ segment: rootSegment, spanId: rootSegment.id, traceId: 'trace1', linkSpanId: rootSegment.id, linkTraceId: 'trace1', testAttr: 'rootTest' })
   }
 
-  // setup child 2 segment with span link
+  // setup child 1 segment with span link
   const child1Name = addExitSpan ? 'MessageBroker/api.example.com/users' : 'child1-segment'
   const child1Segment = createSegment(agent, 'child1', child1Name, rootSegment, tx)
   createSpanLink({ segment: child1Segment, spanId: 'span1', traceId: 'trace1', linkSpanId: 'parent1', linkTraceId: 'trace1', testAttr: 'test1' })
@@ -77,7 +77,7 @@ function setupPartialTrace({ agent, partialType, tx, addRootSpanLink = false, ad
     return { rootSegment, child1Segment }
   }
 
-  // setup child 2 sgement with span links if needed
+  // setup child 2 segment with span links if needed
   const child2Segment = createSegment(agent, 'child2', 'child2-segment', child1Segment, tx)
   createSpanLink({ segment: child2Segment, spanId: 'span2', traceId: 'trace2', linkSpanId: 'parent2', linkTraceId: 'trace1', testAttr: 'test2' })
 


lib/transaction/trace/partial-trace.js
```diff
@@ -118,11 +121,15 @@ class PartialTrace {
     // The id intrinsics attribute needs to be updated to equal the id of the new span the
     // span links are moving to.
     for (const link of spanLinks) {
+      if (lastSpan.spanLinks.length === 100) {
```
svetlanabrennan: This requirement was added to the spec recently for all partial modes - not just compact.

lib/transaction/trace/partial-trace.js
```diff
@@ -255,6 +261,11 @@ class PartialTrace {
         span.addAttribute(key, value, truncateExempt)
       }
     }
+
+    if (meta.spanLinks.length > 0) {
+      this.removeNonIntrinsicsAttrs(meta.spanLinks)
```
bizob2828: feels like this method should live on the span links class. and be done in the applyPartialTraceRules within span event, no?

svetlanabrennan: Sure that works. I put it at the top of the applyPartialTraceRules because it needs to remove those attributes for span links for dropped spans.

lib/spans/span-link.js
```diff
@@ -86,6 +86,11 @@ class SpanLink {
       filterNulls(this.agentAttributes.get(DESTINATIONS.TRANS_SEGMENT))
     ]
   }
+
+  removeNonIntrinsicsAttrs() {
+    this.userAttributes = new Attributes({ scope: Attributes.SCOPE_SEGMENT })
```
bizob2828: there's no reason to create a new empty attributes class. you can borrow how we do it in span event by doing [this](https://github.com/newrelic/node-newrelic/blob/f4dd09ec332c47396277f6aaeff2f6f15daad97a/lib/spans/span-event.js#L351) for both of these

lib/spans/span-event.js
```diff
@@ -318,6 +318,12 @@ class SpanEvent {
    *  @returns {SpanEvent|null} the span after applying the rules, or null if dropped
    */
   applyPartialTraceRules({ isEntry, partialTrace }) {
+    if (partialTrace.type !== PARTIAL_TYPES.REDUCED) {
```
bizob2828, line 323: this function also is sequential to avoid doing things that are not necessary.  I would just put this code after the reduced check below

lib/transaction/trace/partial-trace.js
```diff
@@ -89,40 +81,30 @@ class PartialTrace {
...
-    const lastSpan = this.spans.at(-1)
-
+  reparentSpanLinks(span, spanLinksToReparent) {
     // The id intrinsics attribute needs to be updated to equal the id of the new span the
     // span links are moving to.
-    for (const link of spanLinks) {
-      link.intrinsics.id = lastSpan.id
-    }
+    for (const link of spanLinksToReparent) {
+      if (span.spanLinks.length === 100) {
```
bizob2828, line 109: hmm this is going to pose a problem where we have sprawl of the metric i just added in #3688

lib/spans/span-event.js
svetlanabrennan: I tried doing that already and it was causing one of the span link tests for essential to fail because there's a span link on a non exit span (it's hitting this check and returning null `if (!this.isExitSpan)` so it doesn't filter the attributes before dropping the span.

bizob2828: I would say the test is invalid then. My understanding of core tracing is it's only retaining entry and exit spans. But for exit spans there are special rules that apply

lib/transaction/trace/partial-trace.js
```diff
@@ -66,16 +66,15 @@ class PartialTrace {
   addSpan({ span, isEntry }) {
     const id = span.id
     const parentId = span.parentId
+
+    // capture necessary information for reparenting span links before span is potentially dropped
     const spanLinks = span.spanLinks
+    const exitSpan = span.isExitSpan
```
svetlanabrennan, line 72: @bizob2828 You mentioned adding a property on Span Events that determines whether span link reparenting needs to happen. Were you talking about moving these two check (is exit span and has entity relationships) to happen when a span is created?

lib/transaction/trace/partial-trace.js
```diff
@@ -89,40 +88,37 @@ class PartialTrace {
       // span was dropped but we still need to move its span links to the last kept span
       // spanLinks were captured before the span was dropped
       if (spanLinks && spanLinks.length > 0) {
-        this.reparentSpanLinks(spanLinks)
+        this.reparentSpanLinks(this.spans.at(-1), spanLinks)
```
bizob2828: I would just update `reparentSpanLinks` where you default the 2nd arg to an empty array so you don't have to check it first before calling it

lib/transaction/trace/partial-trace.js
```diff
@@ -89,40 +88,37 @@ class PartialTrace {
...
+  reparentSpanLinks(span, spanLinksToReparent) {
     // The id intrinsics attribute needs to be updated to equal the id of the new span the
     // span links are moving to.
-    for (const link of spanLinks) {
-      link.intrinsics.id = lastSpan.id
-    }
+    for (const link of spanLinksToReparent) {
+      if (span.spanLinks.length === 100) {
+        logger.trace('Span links limit reached. Not moving additional span links.')
+        this.transaction.metrics.getOrCreateMetric(SPAN_EVENTS.LINKS_DROPPED).incrementCallCount()
```
bizob2828: i don't see you asserting this metric in the test that asserts it only retains 100

lib/transaction/trace/partial-trace.js
```diff
@@ -238,13 +231,19 @@ class PartialTrace {
...
       }
 
       this.compactionError(meta, sameEntitySpan)
       this.calculateDuration(meta, sameEntitySpan)
     }
 
     this.assignCompactAttrs({ span, meta })
+
+    if (meta.spanLinks.length > 0) {
+      this.reparentSpanLinks(span, meta.spanLinks)
```
bizob2828: you can just call this, not check the length

test/unit/spans/span-link.test.js
```diff
@@ -9,6 +9,61 @@ const test = require('node:test')
...
+      traceId
+    },
+    timestamp: TIMESTAMP
+  })
+  segment.spanLinks.push(link)
+  return link
+}
+
+function addRootSegment(spanEventAggregator, tx, rootSegment) {
+  spanEventAggregator.addSegment({ segment: rootSegment, transaction: tx, parent: '1', isEntry: true })
```
bizob2828: this is incorrect. should be `parentId`. I also don't see much of a difference between function and `addChildSegment` when i meant create fixtures I meant more than just encapsulate certain calls in a function.  You have common scenarios where you add n segments with/without links.

test/unit/spans/span-link.test.js
```diff
@@ -9,6 +9,61 @@ const test = require('node:test')
...
+const TIMESTAMP = 1765285200000 // 2025-12-09T09:00:00.000-04:00
+
+function setupPartialTransaction(tx, partialType) {
+  tx.priority = 42
+  tx.sampled = true
+  tx.partialType = partialType
+  tx.createPartialTrace()
+}
+
+function createSpanLink(segment, spanId, traceId, linkSpanId, linkTraceId, testAttr) {
```
bizob2828: when you have more than 2 arguments it's prob easier to pass an object so you understand the keys that have to be passed.

Issue #3682, opened 2026-01-22
Add supportability metric when links and events are dropped


amychisholm03:
### Description
Agent spec PR 775 requires us to add the following supportability metrics for when we drop OTel `SpanEvent`s (timed events) in core tracing:

```
Supportability/Nodejs/SpanEvent/Links/Dropped
Supportability/Nodejs/SpanEvent/Events/Dropped
```

### Acceptance Criteria

- [ ] span event aggregator tests are added to assert the correct call count for these new metrics

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-515052

PR #3683, opened 2026-01-22
chore: add timed event limit of 100

### LINKED ISSUE #3646: Limit number of span events
Spec PR 770 clarifies that we should be limiting the number of span events to 100. Thus, the following code:

https://github.com/newrelic/node-newrelic/blob/8929ab3ae6cc28edbef4bee41a2084154361d6e0/lib/transaction/trace/segment.js#L307-L309

Should be updated to match:

https://github.com/newrelic/node-newrelic/blob/8929ab3ae6cc28edbef4bee41a2084154361d6e0/lib/transaction/trace/segment.js#L299-L305

amychisholm03:
## Description

Agent spec PR 775 states that like OTel span links, OTel timed (span) events also need to have a limit of 100.

## How to Test

```
npm run unit
```

## Related Issues

Closes #3646

### DIFF
--- a/lib/transaction/trace/segment.js
+++ b/lib/transaction/trace/segment.js
@@ -305,6 +305,10 @@ TraceSegment.prototype.addSpanLink = function addSpanLink(spanLink) {
 }
 
 TraceSegment.prototype.addTimedEvent = function addTimedEvent(timedEvent) {
+  if (this.timedEvents.length === 100) {
+    this.logger.trace({ timedEvent }, 'Timed event limit reached. Not adding new event.')
+    return
+  }
   this.timedEvents.push(timedEvent)
 }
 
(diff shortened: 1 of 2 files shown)


lib/transaction/trace/segment.js
```diff
@@ -305,6 +305,10 @@ TraceSegment.prototype.addSpanLink = function addSpanLink(spanLink) {
 }
 
 TraceSegment.prototype.addTimedEvent = function addTimedEvent(timedEvent) {
+  if (this.timedEvents.length === 100) {
+    this.logger.trace({ timedEvent }, 'Timed event limit reached. Not adding new event.')
+    return
```
bizob2828, line 310: is there some requirement to be tracking some metric where dropping of these events occurs?

amychisholm03: Yes, they will be added in #3682.

PR #3684, opened 2026-01-22
fix: Updated `Supportability/Nodejs/PartialGranularity/NrIds/Dropped` metric to only keep track of spans that got dropped when `nr.ids` execeeded 63 spans.

### LINKED ISSUE #3640: Add core tracing cross agent tests(partial trace span behavior)
### Description
Cross agent tests are being developed for 3 areas of core tracing: tracecontext, sampler rates/config, and partial tracing span behavior.  This ticket will focus on the updates to partial tracing span behavior(location in cross agent tests is TBD)

bizob2828:
## Description
This PR adds the final cross agent tests for core tracing. I did discover one issue where we were incorrectly incrementing `Supportability/Nodejs/PartialGranularity/NrIds/Dropped`. The intention of this metric is to only increment when we exceed the max size of `nr.ids` which is 63 ids.  This PR fixes that and also brings in the tests around core tracing logic.  

## How to Test

```sh
node test/integration/distributed-tracing/core-tracing.test.js
node test/unit/transaction/trace/partial-trace.test.js
```

## Related Issues
Closes #3640

### DIFF
--- a/lib/transaction/trace/partial-trace.js
+++ b/lib/transaction/trace/partial-trace.js
@@ -217,6 +217,7 @@ class PartialTrace {
 
     const meta = {
       ids: [],
+      droppedIds: 0,
       totalDuration: 0,
       currentStart: null,
       currentEnd: null,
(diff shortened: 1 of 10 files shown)


test/lib/cross_agent_tests/samplers/sampler_configuration.json
```diff
@@ -87,7 +87,6 @@
         },
         "remote_parent_sampled": {
           "trace_id_ratio_based": {
-            "ratio": ""
```
bizob2828, line 90: i re-synced the cross agent tests and this was changed since I integrated these tests

test/lib/cross_agent_tests/distributed_tracing/trace_context.json
```diff
@@ -633,7 +633,7 @@
     }
   },
   {
-    "test_name": "newrelic_remote_parent_sampled_uses_partial_ratio_sampler_not_smapled",
+    "test_name": "newrelic_remote_parent_sampled_uses_partial_ratio_sampler_not_sampled",
```
bizob2828, line 636: i re-synced the cross agent tests and this was changed since I integrated these tests

lib/transaction/trace/partial-trace.js
```diff
@@ -231,23 +232,34 @@ class PartialTrace {
       // do not push its own id to `nr.ids`
       // first span in array is always the retained exit span
       if (i !== 0) {
-        meta.ids.push(sameEntitySpan.id)
+        // Max size for `nr.ids` = 1024. Max length = 63 (each span id is 16 bytes + 8 bytes for list type).
+        if (meta.ids.length < 63) {
```
jsumners-nr, line 236: I don't understand this comment:

`(16+8)63 = 1512`

Is my math incorrect?

bizob2828: It was a comment from other agents but no it's 16 * 63(for each span) + 8 for the array. So it's 1016. 64 would put it over 1024

PR #3692, opened 2026-01-27
chore: remove eslint disable next line for `sonarjs/no-dead-store`

amychisholm03:
Fixes linting error

### DIFF
--- a/lib/instrumentation/when/index.js
+++ b/lib/instrumentation/when/index.js
@@ -296,7 +296,6 @@ module.exports = function initialize(shim, when) {
         ret = agent.tracer.bindFunction(fn, context, true).apply(this, arguments)
       } finally {
         if (ret && typeof ret.then === 'function') {
-          // eslint-disable-next-line sonarjs/no-dead-store
           ret = ctx.next[symbols.context].continue(ret)
         }
       }


lib/instrumentation/when/index.js
```diff
@@ -296,7 +296,6 @@ module.exports = function initialize(shim, when) {
         ret = agent.tracer.bindFunction(fn, context, true).apply(this, arguments)
       } finally {
         if (ret && typeof ret.then === 'function') {
-          // eslint-disable-next-line sonarjs/no-dead-store
           ret = ctx.next[symbols.context].continue(ret)
```
jsumners-nr, line 299: I'm really not clear what is going on here. The `finally` will _always_ be evaluated (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#description). But without a `catch` block, if the `agent.tracer.bindFunction` throws, the error will bubble up and prevent further evaluation:

```js
'use strict'

const res = foo()
console.log('res:', res)

function foo () {
  let res
  try {
    res = bar()
  } catch (error) {
    console.log('got error')
  } finally {
    if (!res) {
      res = 'from finally'
    }
  }
  return res
}

function bar () {
  throw Error('boom')
}
```

With that, we get:

```
got error
res: from finally
```

But the following will result in an uncaught exception and program termination:

```js
'use strict'

const res = foo()
console.log('res:', res)

function foo () {
  let res
  try {
    res = bar()
  } finally {
    if (!res) {
      res = 'from finally'
    }
  }
  return res
}

function bar () {
  throw Error('boom')
}
```

bizob2828: regardless, this is to just unblock CI

Issue #3656, opened 2026-01-13
Add `timestamp` field to all `LlmChatCompletionSummary`s


amychisholm03:
### Description
Spec 771 adds the requirement for `timestamp` to be added to all `LlmChatCompletionSummary`s:

* `timestamp` - Timestamp captured at the time of the LLM request with millisecond precision, Integer
* The `timestamp` attribute **MUST** be used to override the timestamp value attached to the intrinsics of the custom `LlmChatCompletionSummary` event before it is recorded using the `record_custom_event()` API. This is required because the UI relies on accurate timestamping based on time of request to support the Model Inventory view.

This will involve modifying all the `LlmChatCompletionSummary` classes defined in `lib/llm-events`.

### Acceptance Criteria

- [ ] AWS Bedrock `LlmChatCompletionSummary` is updated to include `timestamp`
- [ ] Google GenAI ^^
- [ ] LangChain ^^
- [ ] OpenAI ^^
- [ ] All of these tests are updated to check for `timestamp`.

### Additional context
[Spec link](https://source.datanerd.us/agents/agent-specs/pull/771)
[Related Jira link](https://new-relic.atlassian.net/browse/NR-487636)

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-509795

PR #3701, opened 2026-01-29
feat: Added `timestamp` to LangChain `LlmChatCompletionSummary`

amychisholm03:
## Description

Adds `timestamp` to LangChain `LlmChatCompletionSummary`s and relevant tests.

## How to Test

```
npm run unit
```

## Related Issues

Part of #3656

### DIFF
--- a/lib/llm-events/langchain/chat-completion-summary.js
+++ b/lib/llm-events/langchain/chat-completion-summary.js
@@ -33,7 +33,7 @@ class LangChainCompletionSummary extends LangChainEvent {
     super(params)
     const { segment } = params
 
-    this.duration = segment?.getDurationInMillis()
+    this.duration = segment.getDurationInMillis()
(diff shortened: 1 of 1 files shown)


lib/llm-events/langchain/chat-completion-summary.js
```diff
@@ -35,6 +35,7 @@ class LangChainCompletionSummary extends LangChainEvent {
 
     this.duration = segment?.getDurationInMillis()
```
bizob2828: Not sure why this is here. This optional chaining may be unnecessary. Do we check if the segment is defined anywhere in the instrumentation before calling this?

amychisholm03: Yeah the segment should always be defined because we take it off `ctx` in the RunnableSubscriber. I'll fix

PR #3706, opened 2026-02-02
chore: Add subcomponent attribute to LangChain tool traces

### LINKED ISSUE #3693: Create `APM-AI_TOOL` span attribute to AI tool calls
### Description
Agent PR 761 requires to add a `subcomponent` attribute to spans related to AI/LLM tool calls:

"When `Llm` spans are created to denote an agent or tool, APM agents **MUST** also include an agent attribute to denote that the span is the start of a component. The span attribute will identify the type of component as well as the name of the component.
The `subcomponent` attributes must be defined as follows:

Tools: `subcomponent`: `{"type": "APM-AI_TOOL", "name": {tool_name}}`

The attribute value will be a JSON formatted string where the `type` value is static and the `name` value is the name of the AI agent or tool being represented by the span."

So far, this only applies to our LangChain instrumentation. ~~MCP is excluded because it does not create `LlmTool` events.~~ (MCP `subcomponent` work was done here: https://github.com/newrelic/node-newrelic/issues/3744)

### Acceptance Criteria

- [ ] The `subcomponent` attribute is addd to `lib/subscribers/langchain/tool.js`
- [ ] Tests are written to assert this behavior.

jsumners-nr:
This PR resolves #3693.

### DIFF
--- a/lib/subscribers/langchain/tool.js
+++ b/lib/subscribers/langchain/tool.js
@@ -56,7 +56,7 @@ class LangchainToolSubscriber extends AiMonitoringSubscriber {
       input = JSON.stringify(request)
     }
 
-    segment.addAttribute('subcomponent', `{"type":"APM-AI_TOOL","name":"${name}"}`)
+    segment.addSpanAttribute('subcomponent', `{"type":"APM-AI_TOOL","name":"${name}"}`)
     segment.end()
 
     const toolEvent = new LangChainTool({


lib/subscribers/langchain/tool.js
```diff
@@ -56,6 +56,7 @@ class LangchainToolSubscriber extends AiMonitoringSubscriber {
       input = JSON.stringify(request)
     }
 
+    segment.addAttribute('subcomponent', `{"type":"APM-AI_TOOL","name":"${name}"}`)
```
amychisholm03: I used `segment.addSpanAttribute` for `subcomponent: AI_AGENT`. I'm not sure if there is a difference.

PR #3713, opened 2026-02-04
security: Bumped minimum version of security agent

jsumners-nr:
This PR bumps the minimum allowed version of the NR security agent to resolve CVE warnings on the 12.x agent release line.

All other changes were necessary in order to get CI to complete successfully. These changes are a result of the ever changing ecosystem. Cherry picked backports were used where possible. Otherwise, manual changes were made.

### DIFF
--- a/test/versioned/koa/package.json
+++ b/test/versioned/koa/package.json
@@ -53,7 +53,8 @@
           "samples": 5
         },
         "@koa/router": {
-          "versions": ">=11.0.2",
+          "comment": "15.0.0 uses Array.prototype.toReversed which is only available in Node >=20",
+          "versions": ">=11.0.2 <15.0.0",
           "samples": 5
         }
       },


esm-loader.mjs
```diff
@@ -3,4 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-export * from 'import-in-the-middle/hook.mjs'
+// eslint-disable-next-line n/no-unsupported-features/node-builtins
+import { register } from 'node:module'
+
+register('import-in-the-middle/hook.mjs', import.meta.url, {})
```
bizob2828: this doesn't need to be migrated right? i'm sure you cherry picked commits but trying to minimize the blast radius here

jsumners-nr: This one was not cherry picked. Our `express-esm` tests fail without this change. The loader just wouldn't work.

bizob2828: I think it's something else. this changes the behavior of the esm-loader IIRC, ill dig a little

bizob2828: I restored the original file and ran `npm run versioned:internal:major` and it all works fine. can you please revert this?

jsumners-nr: Sorry, I had forgotten the test suite that is the actual problem. See https://github.com/newrelic/node-newrelic/actions/runs/21836448309/job/63008779573?pr=3713#step:6:16

bizob2828: it passes if you run like this though `node --loader ./esm-loader.mjs test/integration/issue-2155/test.test.mjs`.

bizob2828: it seems like `NODE_OPTIONS` is called many times and the newer register style can handle this? but this still feels test related, and I want to minimize the blast radius of this PR

bizob2828: the reporter is hiding the actual error:

```sh
TypeError [ERR_INVALID_RETURN_PROPERTY_VALUE]: Expected string, array buffer, or typed array to be returned for the "source" from the "load" hook but got undefined.
    at assertBufferSource (node:internal/modules/helpers:381:9)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:104:3)
    at #translate (node:internal/modules/esm/loader:564:20)
    at afterLoad (node:internal/modules/esm/loader:614:29)
    at async #link (node:internal/modules/esm/module_job:180:19) {
  code: 'ERR_INVALID_RETURN_PROPERTY_VALUE'
}
```

jsumners-nr: Yeah, I recall trying to debug that error for this specific case and ultimately decided it was the change in loader API.

bizob2828: ok so i think this is a combination of things. i went back to 12.25.0 and this test fails on the latest node. when we released that version it was node 24.4.0 and it passes. 24.8.0 is where this started to become a problem. we can either upgrade iitm to at least 2.x or just keep your code. I think keeping your code is fine as it is backwards compatible. can you open up a separate PR that upgrades iitm to latest off of main?

bizob2828:
I'll post this here for posterity in case someone in the future releases and out of band patch on a dead release line.  We aren't going to support this 12.25.1 for a year from release date. We're only supporting it until the 12.x EOL which is 7/9/26.  It looks like we can update the PR to docs-website to include `eolDate` in the front matter and it will do the right thing when rendering this version on our EOL page.  [see EolPage component](https://github.com/newrelic/docs-website/blob/62aac6b06582ae839b29d8f6490d0b3ada099d23/src/components/EolPage.js#L37)

Issue #3687, opened 2026-01-23
Refactor `LlmEvent` classes to provide better abstraction/reusability


amychisholm03:
### Description
The agent spec outlines several different objects to encapsulate LLM data that we want to record: `LlmChatCompletionMessage`, `LlmChatCompletionSummary`, `LlmEmbedding`, `LlmErrorMessage `, and with Agentic AI: `LlmTool` and `LlmAgent`. 

With each different LLM/AI package (and sometimes there is nuance within the package itself), the logic to extract the required fields for these objects differs, but at the end of the day, the result needs to be the same. We should have a parent class that expects all of the required fields to be defined and then children classes that can handle the logic within the package. Some of the LLM events already do this (somewhat). 

### Acceptance Criteria

- [x] `LlmChatCompletionMessage` parent class is created and all relevant classes inherit from it. This will allow us to reference this data type in the now abstracted AIM subscribers.
- [x] `LlmChatCompletionSummary` ^^
- [x] `LlmEmbedding` ^^
- [ ] `LlmErrorMessage` ^^
- [x] `LlmAgent` ^^ (currently only used by LangGraph)
- [x] `LlmTool` ^^ (currently only used by LangChain)

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-515458

PR #3726, opened 2026-02-05
refactor: OpenAI LLM event refactor

amychisholm03:
## Description

Refactors OpenAI LLM events to extend from `LlmChatCompletionMessage`, `LlmChatCompletionSummary`, and `LlmEmbedding`, instead of a base OpenAI LLM event class.

* According to the AIM spec, `error` and `isResponse` should only be defined if they are true
* `id` was changed to 32 characters (I'm not sure why it was 36)
* `appName` is not listed in the current AIM spec, so I removed it

## How to Test

```
npm run unit
```

## Related Issues

Part of #3687

### DIFF
--- a/lib/subscribers/openai/chat.js
+++ b/lib/subscribers/openai/chat.js
@@ -153,6 +153,15 @@ class OpenAIChatCompletions extends AiMonitoringChatSubscriber {
     }
   }
 
+  /**
+   * Creates the OpenAI LlmChatCompletionSummary.
+   * @param {object} params Function parameters.
+   * @param {Context} params.ctx Current context.
+   * @param {object} params.request OpenAI request object.
+   * @param {object} [params.response] OpenAI response object, defaults to `{}`.
+   * @param {object} [params.err] Error object if one occurred.
+   * @returns {LlmChatCompletionSummary} The OpenAI LlmChatCompletionSummary instance.
+   */
   createCompletionSummary({ ctx, request, response = {}, err = null }) {
     const { transaction, segment } = ctx
     const headers = ctx?.extras?.headers || {}
@@ -162,7 +171,7 @@ class OpenAIChatCompletions extends AiMonitoringChatSubscriber {
       transaction,
       request,
       response: { ...response, headers },
-      error: err !== null
+      error: !!err
     })
     return summary
   }
--- a/lib/llm-events/openai/chat-completion-summary.js
+++ b/lib/llm-events/openai/chat-completion-summary.js
@@ -13,7 +13,16 @@ const getUsageTokens = require('./get-usage-tokens')
  * Encapsulates an OpenAI `LlmChatCompletionSummary` event.
  */
 module.exports = class OpenAiLlmChatCompletionSummary extends LlmChatCompletionSummary {
-  constructor({ agent, segment, transaction, request = {}, response = {}, error }) {
+  /**
+   * @param {object} params Constructor parameters.
+   * @param {Agent} params.agent New Relic agent instance.
+   * @param {TraceSegment} params.segment Current segment.
+   * @param {Transaction} params.transaction Current and active transaction.
+   * @param {object} params.request OpenAI request object.
+   * @param {object} params.response OpenAI response object.
+   * @param {boolean} [params.error] Set to `true` if an error occurred, can be omitted in false cases.
+   */
+  constructor({ agent, segment, transaction, request, response, error }) {
     super({
       agent,
       segment,
--- a/test/versioned/openai/common-chat-api.js
+++ b/test/versioned/openai/common-chat-api.js
@@ -64,8 +64,22 @@ function assertChatCompletionMessages(
   })
 }
 
+/**
+ * Asserts that the OpenAI LlmChatCompletionSummary has the expected properties.
+ * @param {object} params1 main params object
+ * @param {Transaction} params1.tx associated transaction
+ * @param {string} params1.model LLM id
+ * @param {LlmChatCompletionSummary} params1.chatSummary The `LlmChatCompletionSummary` to check.
+ * @param {boolean} [params1.error] Should `chatSummary.error` equal `true`? Defaults to `false`.
+ * @param {number} [params1.promptTokens] Prompt tokens, defaults to 11.
+ * @param {number} [params1.completionTokens] Completion tokens, defaults to 53.
+ * @param {number} [params1.totalTokens] Total tokens, defaults to 64.
+ * @param {boolean} [params1.noUsageTokens] Should the `chatSummary` have NOT recorded `response.usage.*tokens`? Defaults to `false`.
+ * @param {object} [params2] params object to contain assert library
+ * @param {object} [params2.assert] assert library to use
+ */
 function assertChatCompletionSummary(
-  { tx, model, chatSummary, error, promptTokens = 53, completionTokens = 11, totalTokens = 64, noUsageTokens = false },
+  { tx, model, chatSummary, error = false, promptTokens = 53, completionTokens = 11, totalTokens = 64, noUsageTokens = false },
   { assert = require('node:assert') } = {}
 ) {
   const [segment] = tx.trace.getChildren(tx.trace.root.id)
--- a/test/versioned/openai/common-responses-api.js
+++ b/test/versioned/openai/common-responses-api.js
@@ -80,8 +80,23 @@ function assertChatCompletionMessages(
   }
 }
 
+/**
+ * Asserts that the OpenAI LlmChatCompletionSummary has the expected properties.
+ * @param {object} params1 main params object
+ * @param {Transaction} params1.tx associated transaction
+ * @param {string} params1.model LLM id
+ * @param {LlmChatCompletionSummary} params1.chatSummary The `LlmChatCompletionSummary` to check.
+ * @param {boolean} [params1.error] Should `chatSummary.error` equal `true`? Defaults to `false`.
+ * @param {boolean} [params1.singleInput] Does this chatSummary have a single input/request message? Defaults to `false`.
+ * @param {boolean} [params1.streaming] Was this created via a streaming API? Defaults to `false`.
+ * @param {number} [params1.promptTokens] Prompt tokens, defaults to 11.
+ * @param {number} [params1.completionTokens] Completion tokens, defaults to 53.
+ * @param {number} [params1.totalTokens] Total tokens, defaults to 64.
+ * @param {object} [params2] params object to contain assert library
+ * @param {object} [params2.assert] assert library to use
+ */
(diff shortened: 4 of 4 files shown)


lib/llm-events/openai/get-usage-tokens.js
jsumners-nr, line 1: I'd like to avoid creating yet another "utils.js" file. Every such file is a mystery box that means we have to remember what is where. I'd rather have a simple `get-usage-tokens.js` file that looks like:

```js
'use strict'

module.exports = foo

function foo() {}
```

lib/llm-events/base.js
jsumners-nr, line 1: I think we should move away from naming files `base.js`. It's pretty much the same issue as with `utils.js`. With a single file, it imparts an understanding, but when we get multiple types we see multiple "base"s in the same directory (look at `lib/subscribers/`). I'd rather we name files for their exports. As written, I'd rename this file to `llm-event.js`. But importing `llm-events/llm-event.js` may be a bit much for some folks, so I might rename the export as well. Possibly:

```js
module.exports = class EventBase {}
```

With filename ending up being `llm-events/event-base.js`.

lib/llm-events/base.js
```diff
@@ -0,0 +1,182 @@
...
+ */
+
+'use strict'
+
+const { DESTINATIONS } = require('../config/attribute-filter')
+const { makeId } = require('../util/hashes')
+
+/**
+ * The base LLM event class that contains logic and properties
+ * (e.g. `trace_id`, `vendor`) that are common to all LLM events.
```
jsumners-nr: Please expand this to highlight that properties are defined publicly because the object itself will be serialized and the property names are intended to be included as-is in the serialization.

lib/llm-events/base.js
```diff
@@ -0,0 +1,182 @@
...
+    this.id = makeId(32)
+    this.span_id = segment.id
+    this.trace_id = transaction.traceId
+    this.vendor = vendor
+    this.metadata = agent
+
+    // Omit `error` property if no error occurred
+    if (error === true) {
+      this.error = error
+    }
```
jsumners-nr: If one understands that the directly attached public properties to this object are meant to be included in serialization, then this comment makes sense. Otherwise, it is a bit odd.

Does our serializer omit properties that have been explicitly set to `undefined`? If so, then it'll be easier to understand if we do: `this.error = error || undefined`.

lib/llm-events/base.js
```diff
@@ -0,0 +1,182 @@
...
+
+    // If a certain attribute value is not accessible via instrumentation,
+    // it can be omitted from the event.
```
jsumners-nr: Same question as with `.error`.

lib/llm-events/base.js
```diff
@@ -0,0 +1,182 @@
...
+    if (requestId) this.request_id = requestId
+    if (responseModel) this['response.model'] = responseModel
+  }
+
+  // eslint-disable-next-line accessor-pairs
```
jsumners-nr:
```suggestion
  // eslint-disable-next-line accessor-pairs: `.metadata` should be omitted from serialization
```

lib/llm-events/base.js
```diff
@@ -0,0 +1,182 @@
...
+  set metadata(agent) {
```
jsumners-nr: I'd like to see a doc block that clarifies this method is intended to be used to attach `llm.` prefixed custom attributes to the event object so that they will be shipped to the back end upon the object being serialized.

lib/llm-events/base.js
```diff
@@ -0,0 +1,182 @@
...
+    }
+  }
+
+  /**
+   * Determines if the provided token count is valid.
+   * A valid token count is greater than 0 and not null.
+   * @param {number} tokenCount The token count obtained from the token callback
+   * @returns {boolean} Whether the token count is valid
+   */
+  validTokenCount(tokenCount) {
```
jsumners-nr: This will be easier to understand when reading implementation code:

```suggestion
  isValidTokenCount(tokenCount) {
```

lib/llm-events/base.js
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   * If `totalTokens` is valid, assigns it to
+   * `this['response.usage.total_tokens']`.
+   * @param {number} totalTokens total tokens on embedding message
+   */
+  setTokensOnEmbeddingMessage(totalTokens) {
+    if (this.validTokenCount(totalTokens)) {
+      this['response.usage.total_tokens'] = totalTokens
+    }
+  }
```
jsumners-nr: I'm commenting on both of these at once because they seem tied together, but also I am confused about that being true.

1. I would make `getTotalTokenCount` a getter: `get totalTokenCount() {}`
2. The name of the setter is confusing. The object is about LLM "events". How does that relate to embeddings? If this is only for embedding events, shouldn't it be on the embedding class? Looking at the implementation, I would make this an associated setter with the above getter: `set totalTokenCount(value) {}`

amychisholm03: I copied this over from the old base class in `lib/llm-events/event.js`. There appears to be different getter/setter like functions for embeddings, summaries, and completions. 

I will move to them to their more specific classes.

lib/llm-events/base.js
amychisholm03: I'm not sure where exactly serialization takes place, but when we send up the `Llm` custom events with the `EventAggregator`, a property, e.g. `error`, explicitly set to `undefined` will be sent up. 

Then, when looking at the event in NRQL, the property is not there, so I think we are fine to implement your approach.

jsumners-nr: No, that would be a change in bytes sent. Leave the code as you have it, but the comment could use a bit of clarification.

lib/llm-events/event-base.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /**
+   * Attaches `llm.` prefixed custom attributes to the LLM event object.
+   *
+   * @param {Agent} agent New Relic agent instance.
+   */
+  // `this.metadata` should be omitted from serialization
+  // eslint-disable-next-line accessor-pairs
+  set metadata(agent) {
```
jsumners-nr, line 70:
```suggestion
  /**
   * Attaches `llm.` prefixed custom attributes to the LLM event object.
   *
   * @param {Agent} agent New Relic agent instance.
   */
  set metadata(agent) {
```

lib/subscribers/openai/chat.js
```diff
@@ -162,7 +162,7 @@ class OpenAIChatCompletions extends AiMonitoringChatSubscriber {
       transaction,
       request,
       response: { ...response, headers },
-      withError: !!err
+      error: !!err
```
jsumners-nr, line 174: Since we are changing this line, I'd prefer to see it get changed to something easier to read:

```suggestion
      error: err != null
```

That is, if `err` is meant to be an instance of `Error` or not defined. If it is a boolean,:


```suggestion
      error: err === true
```

lib/subscribers/openai/chat.js
```diff
@@ -237,15 +237,28 @@ class OpenAIChatCompletions extends AiMonitoringChatSubscriber {
...
+    // Check if the given message is the response.
+    // The response object differs based on the API called.
+    // If it's `responses.create`, we check against `response.output`.
+    // If it's `chat.completions.create`, we check against `response.choices`.
+    let isResponse
+    if (response?.object === 'response') {
+      isResponse = message.content === response?.output?.[0]?.content?.[0]?.text
+    } else if (response?.object?.includes('chat.completion')) {
+      isResponse = message.content === response?.choices?.[0]?.message?.content
+    }
```
jsumners-nr, line 259:
```suggestion
    if (response?.object === 'response') {
      isResponse = message.content === response.output?.[0]?.content?.[0]?.text
    } else if (response?.object?.includes('chat.completion')) {
      isResponse = message.content === response.choices?.[0]?.message?.content
    }
```

The top level conditionals have already asserted that `response` is defined. The extra `?.` is unnecessary overhead.

lib/subscribers/openai/embeddings.js
```diff
@@ -29,7 +29,7 @@ class OpenAIEmbeddings extends AiMonitoringEmbeddingSubscriber {
       transaction,
       request,
       response: { ...response, headers },
-      withError: !!err
+      error: !!err
```
jsumners-nr: See previous suggestion:

```suggestion
      error: err != null
```

lib/llm-events/event-base.js
amychisholm03: I'll have to wait until this is merged: https://github.com/newrelic/eslint-config-newrelic/pull/43?

lib/subscribers/openai/chat.js
amychisholm03: `err` is an `Error` object, so I'll do `error: err !== null`.

lib/subscribers/openai/chat.js
amychisholm03: Though extremely unlikely, `response.object` could be defined but for some reason OpenAI may not define `response.output` or `response.choices`, so I think we should keep this.

jsumners-nr: But `response` _will_ be defined as I outlined.

lib/subscribers/openai/chat.js
jsumners-nr: I see tests that are passing in booleans.

amychisholm03: Are your referring to the helper method `assertChatCompletionSummary`? In that function, `error` does not represent the literal error, but whether or not to check if `LlmChatCompletionSummary.error = true`.

jsumners-nr: See line 134 of `test/versioned/openai/common-responses-api.js` and line 96 of `test/unit/llm-events/openai/embedding.test.js` plus others in this change set.

amychisholm03: When constructing the `LlmEvent`, yes `LlmEvent.error` will be a boolean (true). It is omitted (or `undefined`) in the case there is no error. In the tests, we just supply `error: true` instead of doing something like `error: new Error() !== null`

jsumners-nr: So we can't make my suggested change unless we are consistent with the value of the parameter:

```
❯ node
Welcome to Node.js v24.13.1.
Type ".help" for more information.
> true != null
true
> false != null
true
>
```

amychisholm03: `err` (what you are showing as `true` or `false` above) is never a boolean. It is an `Error` object. The test lines you mentioned is a different variable, `error`, which is used to determine if we need to check `LlmEvent.error = true`. I pushed up some documentation to hopefully make it clearer.

PR #3732, opened 2026-02-06
feat: Added `ProfilingAggregator` that will be used to collect and send `pprof_data` telemetry

### LINKED ISSUE #3716: [Profiling] Add ProfilingAggregator
### Description
We need an ability to collect compressed pprof binary data on an interval. This is done by creating a new aggregator in `lib/aggregators` called ProfilingAggregator.  This should inherit from `lib/aggregators/base-aggregator`.  The method should `pprof_data`.  The enabled flag should `config.profiling.enabled`. The `periodMs` should be `config.profiling.interval`.  This should also be registered in `lib/agent` and stored on the agent to be used in the Profiler class to enqueue telemetry

bizob2828:
## Description

This is the first attempt at creating an "aggregator" to collect and send `pprof_data` to the collector.
Aggregator is in quotes because it doesn't really aggregate data but extends the `BaseAggregator` as a means
of sending data to collector.  As this project moves along, this code may change but for now is good enough
to keep us moving.

## How to Test

```sh
node test/unit/aggregators/profiling-aggregator.test.js
```

## Related Issues
Closes #3716

### DIFF
--- a/lib/collector/api.js
+++ b/lib/collector/api.js
@@ -103,7 +103,8 @@ function CollectorAPI(agent) {
     'sql_trace_data',
     'error_event_data',
     'span_event_data',
-    'log_event_data'
+    'log_event_data',
+    'pprof_data'
   ]) {
     const method = new RemoteMethod(name, agent, initialEndpoint)
     this._methods[name] = method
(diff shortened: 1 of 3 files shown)


lib/collector/api.js
```diff
@@ -103,7 +103,8 @@ function CollectorAPI(agent) {
     'sql_trace_data',
     'error_event_data',
     'span_event_data',
-    'log_event_data'
+    'log_event_data',
+    'pprof_data'
```
bizob2828, line 107: I had to add this so the tests would work

PR #3734, opened 2026-02-09
feat: Updated `RemoteMethod` to handle sending `pprof_data` through the collector

### LINKED ISSUE #3714: [Profiling] Provide ability to send `pprof_data` through collector
### Description
As part of the continuous profiling effort, we will be collecting a new type of data. This pprof data is gzip encoded, binary data format.  

### Acceptance Criteria
As a Node.js engineer I want the ability to send profiling data to the collector so that it can be processed, stored and rendered in Profiling UX.

### Additional context
 * We need to add `pprof_data` as a new method [here](https://github.com/newrelic/node-newrelic/blob/4f591088888ae724136d409ee93438ecd81a5539/lib/collector/api.js#L106)
 * The [RemoteMethod](https://github.com/newrelic/node-newrelic/blob/4f591088888ae724136d409ee93438ecd81a5539/lib/collector/remote-method.js#L37) class needs to allow a content type to be passed in, it should default to `application/json` but for pprof_data it should be `application/octet-stream`
     * We should update the `RemoteMethod` signature to pass in an object and also update the configuration in `lib/collector/api` to pass in an object.  
     * Then when `_headers` is called on the `RemoteMethod` it should read the `this._contentType` property [here](https://github.com/newrelic/node-newrelic/blob/4f591088888ae724136d409ee93438ecd81a5539/lib/collector/remote-method.js#L380)

svetlanabrennan:
## Description

Updated the `RemoteMethod` class to allow content type to be passed in but default to `application/json` but when the method is `pprof_data` the content type will be `application/octet-stream`. 

## How to Test

Run `npm run test`

## Related Issues

Closes #3714

### DIFF
--- a/lib/collector/api.js
+++ b/lib/collector/api.js
@@ -90,24 +90,31 @@ function CollectorAPI(agent) {
    * save some GC time?
    */
   this._methods = {}
-  for (const name of [
-    'preconnect',
-    'connect',
-    'agent_settings',
-    'error_data',
-    'metric_data',
-    'transaction_sample_data',
-    'shutdown',
-    'analytic_event_data',
-    'custom_event_data',
-    'sql_trace_data',
-    'error_event_data',
-    'span_event_data',
-    'log_event_data',
-    'pprof_data'
+  for (const methodOptions of [
+    { name: 'preconnect' },
+    { name: 'connect' },
+    { name: 'agent_settings' },
+    { name: 'error_data' },
+    { name: 'metric_data' },
+    { name: 'transaction_sample_data' },
+    { name: 'shutdown' },
+    { name: 'analytic_event_data' },
+    { name: 'custom_event_data' },
+    { name: 'sql_trace_data' },
+    { name: 'error_event_data' },
+    { name: 'span_event_data' },
+    { name: 'log_event_data' },
+    { name: 'pprof_data', contentType: 'application/octet-stream' },
   ]) {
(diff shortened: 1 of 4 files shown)


lib/collector/api.js
```diff
@@ -106,7 +106,7 @@ function CollectorAPI(agent) {
     'log_event_data',
     'pprof_data'
```
bizob2828: I would change this array to be a collection and you can pass in the override of contentType. 

So a small snippet is:

```
[
  { endpoint: 'pprof_data', contentType: 'application/octet-stream' },
  { endpoint: 'log_event_data' }
]
```

lib/collector/remote-method.js
```diff
@@ -48,6 +48,7 @@ function RemoteMethod(name, agent, endpoint) {
   this._protocolVersion = 17
 
   this.endpoint = endpoint
+  this.contentType = name === 'pprof_data' ? 'application/octet-stream' : contentType
```
bizob2828: shift this logic to `lib/collector/api.js`

lib/collector/api.js
svetlanabrennan: Is this what you meant? https://github.com/newrelic/node-newrelic/pull/3734/changes#diff-e2ba01732f1669e0e2821d95a482cf95d3cef725932a0ce6048d61904d0afd26R94-R107

Or should I add `endpoint` and `agent` each method option?

PR #3736, opened 2026-02-10
chore: Added npm tag support to CI

jsumners-nr:
This PR adds a new option when triggering the "create release" workflow to support specifying a dist-tag when publishing to npm.


bizob2828:
We agreed to not do this because we have a few post release jobs: azure site extensions and creating tags for lambda layers and init containers that would also have to be inspected. It's not worth the risk

PR #3737, opened 2026-02-10
chore: Removed `@langchain/core` peer dep from LangGraph test matrix

amychisholm03:
I removed the `@langchain/core` dependency in the LangGraph test matrix because it is a peer dependency for `@langchain/langgraph`. If the version for the explicitly listed `@langchain/core` module defers from what `@langchain/langgraph` is expecting, the tests were failing.


bizob2828:
can you explain why you have to remove it?

PR #3741, opened 2026-02-11
chore: Fix`@langchain/community` peer dep conflict

amychisholm03:
`@langchain/community` and `@langchain/core` now have to be the same minor version when testing or else there will be a peer dep conflict. `@langchain/community` is only needed for vectorstores, so I created a separate test matrix for that.

### DIFF
--- a/test/versioned/langchain-openai/package.json
+++ b/test/versioned/langchain-openai/package.json
@@ -17,17 +17,27 @@
       "engines": {
         "node": ">=20"
       },
-      "comment": "because @langchain/core and @langchain/openai need to be the same major version, we will only test >=1.0.0 to avoid v0 vs. v1 conflict",
       "dependencies": {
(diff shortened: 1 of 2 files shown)


test/versioned/langchain-openai/package.json
```diff
@@ -17,17 +17,27 @@
       "engines": {
         "node": ">=20"
       },
-      "comment": "because @langchain/core and @langchain/openai need to be the same major version, we will only test >=1.0.0 to avoid v0 vs. v1 conflict",
       "dependencies": {
         "@langchain/core": ">=1.0.0",
-        "@langchain/community": ">=1.0.0",
-        "@langchain/openai": ">=1.0.0",
-        "openai": "4.90.0",
```
amychisholm03, line 25: `openai` was never needed because the api is provided by `@langchain/openai` and we're using a mock client

PR #3748, opened 2026-02-12
refactor: Google Gen AI LLM event refactor

amychisholm03:
## Description

Refactors Google Gen AI LLM events to use the new base `LlmEmbedding`, `LlmChatCompletionMessage`, and `LlmChatCompletionSummary` classes.

## How to Test

```
npm run unit
npm run versioned google-genai
```

## Related Issues

Part of #3687

### DIFF
--- a/lib/llm-events/google-genai/get-usage-tokens.js
+++ b/lib/llm-events/google-genai/get-usage-tokens.js
@@ -14,8 +14,13 @@ module.exports = getUsageTokens
  * @returns {object} { promptTokens, completionTokens, totalTokens }
  */
 function getUsageTokens(response) {
-  const promptTokens = Number(response?.usageMetadata?.promptTokenCount)
-  const completionTokens = Number(response?.usageMetadata?.candidatesTokenCount)
-  const totalTokens = Number(response?.usageMetadata?.totalTokenCount)
-  return { promptTokens, completionTokens, totalTokens }
+  const { usageMetadata } = response
+  if (!usageMetadata) {
+    return { promptTokens: 0, completionTokens: 0, totalTokens: 0 }
+  }
+  return {
+    promptTokens: Number(usageMetadata.promptTokenCount),
+    completionTokens: Number(usageMetadata.candidatesTokenCount),
+    totalTokens: Number(usageMetadata.totalTokenCount)
+  }
 }
--- a/lib/subscribers/google-genai/embed-content.js
+++ b/lib/subscribers/google-genai/embed-content.js
@@ -14,6 +14,16 @@ class GoogleGenAIEmbedContentSubscriber extends AiMonitoringEmbeddingSubscriber 
     this.events = ['asyncEnd']
   }
 
+  /**
+   * Creates a Google Gen AI LlmEmbedding instance.
+   *
+   * @param {object} params to function
+   * @param {Context} params.ctx active context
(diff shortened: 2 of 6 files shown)


lib/llm-events/google-genai/get-usage-tokens.js
```diff
@@ -0,0 +1,21 @@
...
+ * Grabs the prompt, completion, and total token count from the
+ * given response object.
+ * @param {object} response Google Gen AI response object
+ * @returns {object} { promptTokens, completionTokens, totalTokens }
+ */
+function getUsageTokens(response) {
+  const promptTokens = Number(response?.usageMetadata?.promptTokenCount)
+  const completionTokens = Number(response?.usageMetadata?.candidatesTokenCount)
+  const totalTokens = Number(response?.usageMetadata?.totalTokenCount)
+  return { promptTokens, completionTokens, totalTokens }
```
jsumners-nr: There's no need to assert that response is defined when the function requires it as a parameter. I'd also avoid the cost of "is metadata defined" by:

```js
const { usageMetadata } = response
if (Object.prototype.toString.call(usageMetadata) !== '[object Object]') {
  return { promptTokens: 0, completionTokens: 0, totalTokens: 0 }
}
```

lib/subscribers/google-genai/embed-content.js
```diff
@@ -22,7 +22,7 @@ class GoogleGenAIEmbedContentSubscriber extends AiMonitoringEmbeddingSubscriber
       transaction,
       request,
       response,
-      withError: !!err
+      error: !!err
```
jsumners-nr, line 35:
```js
error: err === true
```

lib/subscribers/google-genai/generate-content.js
```diff
@@ -27,13 +27,26 @@ class GoogleGenAIGenerateContentSubscriber extends AiMonitoringChatSubscriber {
...
     // request.contents can be a string or an array of strings
-    // response.candidates is an array of candidates (choices); we only take the first one
-    const inputMessages = Array.isArray(request.contents) ? request.contents : [request.contents]
-    const responseMessage = response?.candidates?.[0]?.content
-    return responseMessage !== undefined ? [...inputMessages, responseMessage] : inputMessages
+    const contents = Array.isArray(request.contents) ? request.contents : [request.contents]
+    const messages = contents.map((item) => {
+      return { content: item, role: 'user' }
+    })
+    const responseContent = response?.text ?? response?.candidates?.[0]?.content?.parts?.[0]?.text
```
jsumners-nr: The function signature and docblock indicates that `response` will always be defined. Is that correct?

lib/subscribers/google-genai/generate-content.js
```diff
@@ -44,21 +57,26 @@ class GoogleGenAIGenerateContentSubscriber extends AiMonitoringChatSubscriber {
       transaction,
       request,
       response,
-      withError: !!err
+      error: !!err
```
jsumners-nr, line 70:
```
error: err === true
```

amychisholm03: `err` is an object, updated doc block to reflect

lib/subscribers/google-genai/embed-content.js
amychisholm03: err is an object, updated doc block to reflect

PR #3749, opened 2026-02-12
chore: Added subcomponent attribute to mcp-sdk invocations

### LINKED ISSUE #3744: [AIM] Add agentic tool `subcomponent` span attribute for MCP tool calls
### Description
When Llm spans are created to denote an agent or tool, APM agents MUST also include an agent attribute to denote that the span is the start of a component. The span attribute will identify the type of component as well as the name of the component. The subcomponent attributes must be defined as follows:
Tools:

`subcomponent: {"type": "APM-AI_TOOL", "name": {tool_name}}`

The attribute value will be a JSON formatted string where the type value is static and the name value is the name of the AI agent or tool being represented by the span.

### Acceptance Criteria

- [ ] `subcomponent` is added to the MCP `callTool` segment only (not `readResource` or `getPrompt`)
- [ ] relevant tests are created

jsumners-nr:
This PR resolves #3744.

### DIFF
--- a/lib/subscribers/mcp-sdk/client-request.js
+++ b/lib/subscribers/mcp-sdk/client-request.js
@@ -55,7 +55,7 @@ class McpClientRequestSubscriber extends AiMonitoringSubscriber {
     this.name = `${mcpPrefix}/${functionName}/${primitiveName}`
 
     const newCtx = super.handler(data, ctx)
-    if (newCtx?.segment) {
+    if (newCtx?.segment && functionName === 'callTool') {
       newCtx.segment.addSpanAttribute('subcomponent', `{"type":"APM-AI_TOOL","name":"${primitiveName}"}`)
     }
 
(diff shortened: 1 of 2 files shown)


lib/subscribers/mcp-sdk/client-request.js
```diff
@@ -53,7 +53,13 @@ class McpClientRequestSubscriber extends AiMonitoringSubscriber {
     }
 
     this.name = `${mcpPrefix}/${functionName}/${primitiveName}`
-    return super.handler(data, ctx)
+
+    const newCtx = super.handler(data, ctx)
+    if (newCtx?.segment) {
+      newCtx.segment.addSpanAttribute('subcomponent', `{"type":"APM-AI_TOOL","name":"${primitiveName}"}`)
```
bizob2828, line 59: my understanding is that we only create this attribute if the call is a `tools/call`, right @amychisholm03 ?

amychisholm03: Yes, it was listed in the issue.

@jsumners-nr Suggested change:

```
if(newCtx?.segment && functionName === 'callTool'){
  newCtx.segment.addSpanAttribute('subcomponent', `{"type":"APM-AI_TOOL","name":"${primitiveName}"}`)
}
```

PR #3758, opened 2026-02-17
feat: Added ability to delay the start of the Profiler and run for n milliseconds before shutting down.

### LINKED ISSUE #3753: [Profiling] Apply the delay and duration configs for profiling
Configure the profiling to apply delay and duration configs after this issue #3752 is completed. 

These configurations are for the profiling feature as a whole (and not per profiler).

bizob2828:
## Description
This PR adds the ability to register an aggregator with a delay and duration property.  This in turn will be used when registering the aggregators with the harvester class. If a delay is specified it will delay the start. If a duration is specified it will schedule the aggregator to be stopped.  I also tweaked the profiling aggregator slightly to return if the profiling manager actually started any profilers before scheduling the profiling aggregator from running. This still isn't completely wired up so the unit tests only test the functionality with an aggregator in the harvester unit tests.

## How to Test

```sh
npm run unit
```

## Related Issues
Closes #3753

### DIFF
--- a/lib/aggregators/profiling-aggregator.js
+++ b/lib/aggregators/profiling-aggregator.js
@@ -43,8 +43,13 @@ class ProfilingAggregator extends BaseAggregator {
    * and send the gzipped binary encoded data for each profiler
    */
   start() {
-    logger.trace(`${this.method} aggregator started.`)
     this.profilingManager.register()
+    const started = this.profilingManager.start()
+    if (!started) {
+      return
+    }
+
+    logger.trace(`${this.method} aggregator started.`)
 
     if (!this.sendTimer) {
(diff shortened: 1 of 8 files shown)


lib/aggregators/profiling-aggregator.js
```diff
@@ -43,8 +43,13 @@ class ProfilingAggregator extends BaseAggregator {
    * and send the gzipped binary encoded data for each profiler
    */
   start() {
-    logger.trace(`${this.method} aggregator started.`)
     this.profilingManager.register()
+    const started = this.profilingManager.start()
+    if (!started) {
```
bizob2828, line 48: I'm not logging any messages as the profiling manager logs a message if there are no registered profilers

svetlanabrennan (approved):

PR #3769, opened 2026-02-19
feat: Added support for error causes

### LINKED ISSUE #2971: Error `cause` argument is not logged
## Description

I opened an issue in the NR Browser Agent repository to report that the `cause` argument passed to `Error` is not logged to NR.
I've received confirmation this is an issue in that package, please see https://github.com/newrelic/newrelic-browser-agent/issues/1398#issuecomment-2688593596

I'm opening this issue to ask if this is also happening in this package please.
If so, this means I need to refactor my back-end code to include the `cause` text in the first argument on `Error` too, as anybody else expecting NR to save that text.

Thanks.

jsumners-nr:
This PR resolves #2971.

I tested this with the following web app:

```js
'use strict'

const server = require('fastify')({ logger: true })

server.route({
  path: '/error-causes',
  method: 'get',
  handler (req, res) {
    const parent = Error('parent')
    const child = Error('child')
    parent.cause = child

    try {
      throw Error('grand-child')
    } catch (error) {
      child.cause = error
      throw parent
    }
  }
})

server.listen({ port: 8000 })
```

The result is an errors inbox entry with `error.cause` in the attributes list view:

image

### DIFF
--- a/test/unit/lib/errors/error-collector.test.js
+++ b/test/unit/lib/errors/error-collector.test.js
@@ -8,18 +8,18 @@
 const test = require('node:test')
 const assert = require('node:assert')
 
-const helper = require('../../lib/agent_helper')
-const Exception = require('../../../lib/errors').Exception
-const ErrorCollector = require('../../../lib/errors/error-collector')
-const ErrorTraceAggregator = require('../../../lib/errors/error-trace-aggregator')
-const ErrorEventAggregator = require('../../../lib/errors/error-event-aggregator')
+const helper = require('#testlib/agent_helper.js')
+const Exception = require('#agentlib/errors/index.js').Exception
+const ErrorCollector = require('#agentlib/errors/error-collector.js')
+const ErrorTraceAggregator = require('#agentlib/errors/error-trace-aggregator.js')
+const ErrorEventAggregator = require('#agentlib/errors/error-event-aggregator.js')
 
-const Transaction = require('../../../lib/transaction')
-const Metrics = require('../../../lib/metrics')
+const Transaction = require('#agentlib/transaction/index.js')
+const Metrics = require('#agentlib/metrics/index.js')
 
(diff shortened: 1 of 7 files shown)


lib/errors/index.js
```diff
@@ -110,6 +168,7 @@ function createError(transaction, exception, config) {
 
   maybeAddAgentAttributes(params, exception)
 
+  process._rawDebug('!!! params:', params)
```
bizob2828: debug code, please remove

lib/errors/index.js
```diff
@@ -26,6 +39,21 @@ class Exception {
     this.agentAttributes = agentAttributes || {}
     this._expected = expected
     this.errorGroupCallback = null
+
+    // We need to detect error causes and do something with them. We don't have
+    // a spec, or ui expectation, for them. So we are opting to embed them as
+    // agent attributes. This does not account for `AggregateError` instances.
+    if (Object.hasOwn(error ?? {}, 'cause') === true) {
```
bizob2828: looking at the structure of this code, it seems like most of the logic lives in `createError` around the  error that gets sent over the wire

test/unit/lib/errors/exception.test.js
```diff
@@ -0,0 +1,186 @@
+/*
+ * Copyright 2024 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const test = require('node:test')
```
bizob2828, line 8: looks like most of the unit tests for errors live in `test/unit/errors/`. i know you're just reflecting the location but this may cause more confusion as its the only one with tests. based on my comment in `lib/errors/index.js`. you can probably shift the tests to another file, or at the very least just move this to the folder where all the other error tests live

PR #3772, opened 2026-02-23
feat: added functionality to ingest logs from Azure Functions App automatically

### LINKED ISSUE #3767: Ingest logs from Azure Function Apps automatically
As a user, when instrumenting an Azure Function App, I want to automatically ingest logs. 

## Additional Context
Looking at the [azure functions otel instrumentation](https://github.com/Azure/azure-functions-nodejs-opentelemetry/blob/655f89365d70e09256f478886132cdf768eaee1a/src/instrumentation.ts#L44), there appears to be a log hook we can register to see when `context.log` is called within an azure function. We obviously have to add our linking metadata as well before adding logs to agent.

svetlanabrennan:
## Description

Update our Azure functions app instrumentation to automatically capture logs if application logging is enabled and log forwarding is enabled. 

Azure has 5 logging [levels](https://learn.microsoft.com/en-us/azure/azure-functions/functions-reference-node?tabs=javascript%2Cwindows%2Cazure-cli&pivots=nodejs-model-v4#log-levels). Two of those levels had to be mapped to meet our naming for those levels. 
- mapped "information" to "info"
- mapped "warning to "warn". 

We capture the following for a log:
- message, level and category (from azure logs)
- timestamp and linking metadata (we add this on)

After we capture the azure logs, we update the logging metrics. 

## How to Test

Run versioned test for azure.

## Related Issues

Closes #3767

### DIFF
--- a/lib/instrumentation/@azure/functions.js
+++ b/lib/instrumentation/@azure/functions.js
@@ -197,10 +197,10 @@ function registerAzureLogs(agent, azureFunctions, logger) {
 
   azureFunctions.app.hook?.log((context) => {
     const logLevel = AZURE_TO_NR_LEVEL[context.level] ?? context.level
-    
+
     if (isLogForwardingEnabled(agent.config, agent)) {
       const meta = agent.getLinkingMetadata(true)
-    
+
       const logData = {
         message: context.message,
         level: logLevel,
@@ -208,11 +208,10 @@ function registerAzureLogs(agent, azureFunctions, logger) {
         category: context.category,
         ...meta
       }
-      
+
       agent.logs.add(logData)
     }
 
-   
     if (isMetricsEnabled(agent.config)) {
       incrementLoggingLinesMetrics(logLevel, agent.metrics)
     }
--- a/lib/instrumentation/@azure/functions.js
+++ b/lib/instrumentation/@azure/functions.js
@@ -85,9 +85,10 @@ module.exports = function initialize(_agent, azureFunctions, _moduleName, shim, 
 
   azureFunctions.app.hook.log((context) => {
     const meta = _agent.getLinkingMetadata(true)
+    const logLevel = AZURE_TO_NR_LEVEL[context.level] ?? context.level
     const logData = {
       message: context.message,
-      level: context.level,
+      level: logLevel,
       timestamp: Date.now(),
       attributes: {
         CategoryName: context.category
@@ -97,8 +98,7 @@ module.exports = function initialize(_agent, azureFunctions, _moduleName, shim, 
     _agent.logs.add(logData)
 
     if (isMetricsEnabled(_agent.config)) {
-      const level = AZURE_TO_NR_LEVEL[context.level] ?? context.level
-      incrementLoggingLinesMetrics(level, _agent.metrics)
+      incrementLoggingLinesMetrics(logLevel, _agent.metrics)
     }
   })
 }
--- a/lib/instrumentation/@azure/functions.js
+++ b/lib/instrumentation/@azure/functions.js
@@ -41,6 +41,16 @@ module.exports = function initialize(_agent, azureFunctions, _moduleName, shim, 
     return
   }
 
+  azureFunctions.app.hook.log((context) => {
+    const meta = _agent.getLinkingMetadata(true)
+    const logData = {
+      message: context.message,
+      level: context.level,
+      ...meta
+    }
+    _agent.logs.add(logData)
+  })
+
   const httpMethods = ['http', 'get', 'put', 'post', 'patch', 'deleteRequest']
   shim.wrap(azureFunctions.app, httpMethods, wrapAzureHttpMethods)
 
--- a/lib/instrumentation/@azure/functions.js
+++ b/lib/instrumentation/@azure/functions.js
@@ -40,7 +40,7 @@ const AZURE_TO_NR_LEVEL = {
 
 let coldStart = true
 
-module.exports = function initialize(_agent, azureFunctions, _moduleName, shim, { logger = defaultLogger } = {}) {
+module.exports = function initialize(agent, azureFunctions, _moduleName, shim, { logger = defaultLogger } = {}) {
   if (!SUBSCRIPTION_ID || !RESOURCE_GROUP_NAME || !AZURE_FUNCTION_APP_NAME) {
     logger.warn(
       {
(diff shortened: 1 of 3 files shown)


lib/instrumentation/@azure/functions.js
```diff
@@ -59,6 +72,35 @@ module.exports = function initialize(_agent, azureFunctions, _moduleName, shim,
     'webPubSub'
   ]
   shim.wrap(azureFunctions.app, backgroundMethods, wrapAzureBackgroundMethods)
+
+  if (!isApplicationLoggingEnabled(_agent.config)) {
+    logger.debug('Application logging not enabled. Not auto capturing logs from Azure Functions.')
+    return
+  }
+
+  if (!isLogForwardingEnabled(_agent.config, _agent)) {
```
bizob2828: this check needs to be done within the hook. you only want to omit defining the hook if application logging is enabled. if you enable metrics but not forwarding, you won't get the metrics

test/versioned/azure-functions/logs.test.js
```diff
@@ -0,0 +1,290 @@
...
+        trace(message) { fireLogHook(message, 'trace') }
+      })
+    },
+    app: {
+      hook: {
+        log(callback) {
+          logHookCallbacks.push(callback)
+        }
+      },
+      get(name, options) {
```
svetlanabrennan, line 119: I'm only running `GET` requests for these tests. Didn't see a reason to run these tests for all request types.

lib/instrumentation/@azure/functions.js
```diff
@@ -59,6 +72,35 @@ module.exports = function initialize(_agent, azureFunctions, _moduleName, shim,
...
+  }
+
+  azureFunctions.app.hook?.log((context) => {
+    const meta = _agent.getLinkingMetadata(true)
+    const logLevel = AZURE_TO_NR_LEVEL[context.level] ?? context.level
+    const logData = {
+      message: context.message,
+      level: logLevel,
+      timestamp: Date.now(),
+      attributes: {
```
bizob2828: i wouldn't put this in an attributes key, i would just specify `category: context.category` in the root of the the object that gets forwarded

lib/instrumentation/@azure/functions.js
```diff
@@ -25,6 +31,13 @@ const SUBSCRIPTION_ID = WEBSITE_OWNER_NAME?.split('+').shift()
...
+// Azure Functions uses 'information' as the context level for context.log and context.info and uses 'warning' as the context level for context.warn.
+// These map to the NR standard names that LOGGING.LEVELS recognizes ('info', 'warn').
+const AZURE_TO_NR_LEVEL = {
+  information: 'info',
+  warning: 'warn'
+}
+
 let coldStart = true
 
 module.exports = function initialize(_agent, azureFunctions, _moduleName, shim, { logger = defaultLogger } = {}) {
```
bizob2828: since you now use agent i would change this var to `agent`. `_<var>` implies it's unused

bizob2828 (changes requested):
I haven't tested with the live app yet but wanted to submti this feedback first.

lib/instrumentation/@azure/functions.js
```diff
@@ -59,6 +72,35 @@ module.exports = function initialize(_agent, azureFunctions, _moduleName, shim,
     'webPubSub'
   ]
   shim.wrap(azureFunctions.app, backgroundMethods, wrapAzureBackgroundMethods)
+
+  if (!isApplicationLoggingEnabled(_agent.config)) {
```
bizob2828: you also may want to consider wrapping all of the logging instrumentation into a function

lib/instrumentation/@azure/functions.js
```diff
@@ -216,6 +189,35 @@ function wrapAzureBackgroundMethods(shim, appMethod) {
...
 
+function registerAzureLogs(agent, azureFunctions, logger) {
+  if (!isApplicationLoggingEnabled(agent.config)) {
+    logger.debug('Application logging not enabled. Not auto capturing logs from Azure Functions.')
+    return
+  }
+
+  azureFunctions.app.hook?.log((context) => {
+    if (!isLogForwardingEnabled(agent.config, agent)) {
+      logger.debug('Log forwarding not enabled. Logs from Azure Functions will not be forwarded to New Relic.')
```
bizob2828: you don't want to exit early if log forwarding is enabled. you simply want to encapulate the log forwarding logic within this check. metrics and forwarding operate indepdendently

bizob2828 (changes requested):

lib/instrumentation/@azure/functions.js
svetlanabrennan: Oh ok. I didn't realize they are separate. I'll update this.

lib/instrumentation/@azure/functions.js
```diff
@@ -174,6 +189,35 @@ function wrapAzureBackgroundMethods(shim, appMethod) {
...
+      timestamp: Date.now(),
+      category: context.category,
+      ...meta
+    }
+    agent.logs.add(logData)
+
+    if (isMetricsEnabled(agent.config)) {
+      incrementLoggingLinesMetrics(logLevel, agent.metrics)
+    }
+  })
```
bizob2828:
```suggestion
    const logLevel = AZURE_TO_NR_LEVEL[context.level] ?? context.level
    
    if (isLogForwardingEnabled(agent.config, agent)) {
      const meta = agent.getLinkingMetadata(true)
    
      const logData = {
        message: context.message,
        level: logLevel,
        timestamp: Date.now(),
        category: context.category,
        ...meta
      }
      
      agent.logs.add(logData)
    }

   
    if (isMetricsEnabled(agent.config)) {
      incrementLoggingLinesMetrics(logLevel, agent.metrics)
    }
  })
```

svetlanabrennan: This checks if log forwarding is enabled, then adds the log and then logs a debug message saying not forwarding not enabled so logs won't be forwarded. Is it a typo or am I misunderstanding the logic here?

svetlanabrennan: Looks like you updated it so nevermind.

svetlanabrennan:
I'll update tests now to reflect the updated logic

bizob2828 (approved):

sarinmsari:
Do you usually capture decisions and internal technical documentation somewhere, or does it mostly stay in PR threads?

sarinmsari:
This PR has a lot of useful context. I tried structuring it into a short doc so it’s easier to revisit later:
[https://gist.github.com/doopstech/d88005be8f34cc67a8c736143e73091f](https://gist.github.com/doopstech/d88005be8f34cc67a8c736143e73091f)

If this isn’t appropriate here, feel free to ignore, just experimenting with better ways to preserve PR context.

PR #3773, opened 2026-02-23
refactor: Migrated `redis` 4+ instrumentation to subscribe to events emitted

### LINKED ISSUE #3731: Refactor `redis` instrumentation to subscribe to events emitted
### Description
Refactor `redis` instrumentation to subscriber-based instead of shim-based.

### Acceptance Criteria

- [ ] all of the redis tests still pass and example apps continue to work as expected

amychisholm03:
## Description

The `redis` package contains `@redis/client` which contains the Redis client code we actually want to instrument. So for subscribers, we will instrument `@redis/client` instead of `redis`. `@node-redis/client` is the same package just under a different name.

This only applies to `redis@>=4`. I decided to keep the current `redis` shim-based instrumentation for v3 because it is so old.

## How to Test

```
npm run versioned redis
```

## Related Issues

Closes #3731


lib/subscriber-configs.js
```diff
@@ -22,9 +22,11 @@ const subscribers = {
   ...require('./subscribers/mcp-sdk/config'),
   ...require('./subscribers/mysql/config'),
   ...require('./subscribers/mysql2/config'),
+  ...require('./subscribers/node-redis-client/config'),
```
bizob2828, line 26: so it looks like this is here because redis 4-4.1.0 has the client in `@node-redis/client`

lib/instrumentation/redis.js
```diff
@@ -14,6 +14,7 @@ const {
 module.exports = function initialize(_agent, redis, _moduleName, shim) {
   const proto = redis?.RedisClient?.prototype
   if (!proto) {
+    // This will happen when redis >= 4.0.0
     shim.logger.warn('Skipping redis instrumentation due to unrecognized module shape')
```
bizob2828, line 18: can you log another ticket to migrate redis <4 instrumentation? now that we figured out all the nuances of 4+ this doesn't seem like it'll be a lot of work

PR #3778, opened 2026-02-24
chore: Updated version format in agent-metadata.yml

amychisholm03:
Agent metadata workflow failed with `13.14.0` release because the workflow assumed the version would be passed in as `'v13.14.0'`.

### DIFF
--- a/.github/workflows/agent-metadata.yml
+++ b/.github/workflows/agent-metadata.yml
@@ -59,6 +59,7 @@ jobs:
     uses: newrelic/newrelic-agent-init-container/.github/workflows/agent-metadata.yml@main
     with:
       agent-type: ${{ needs.get-version.outputs.agent-type }}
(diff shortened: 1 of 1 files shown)


.github/workflows/agent-metadata.yml
```diff
@@ -59,7 +59,7 @@ jobs:
     uses: newrelic/newrelic-agent-init-container/.github/workflows/agent-metadata.yml@main
     with:
       agent-type: ${{ needs.get-version.outputs.agent-type }}
-      version: ${{ needs.get-version.outputs.version }}
+      version: v${{ needs.get-version.outputs.version }}
```
bizob2828, line 63: i would put a comment explaining this.

maybe: "Our version in package.json omits a `v`, This reusable workflow checks out a tag which is prefix with a `v`"

Issue #3581, opened 2025-12-09
Add `first_time_to_token` attribute to `LlmChatCompletionSummary`


amychisholm03:
### Description
“Time To First token” is an important and required metric to measure the model performance. We need to add this metric to AIM via `LlmChatCompletionSummary.first_time_to_token`.

- This metric should only be captured for streaming requests
- The metric name should be captured as `LlmChatCompletionSummary.time_to_first_token`
- The metric value should be stored as the number of milliseconds between the time when the request was issued and the first token is received

### Pending Questions to Clarify
- Is this for all LLMs or just a select few?
- What is the default/error value? `0`, `undefined`?

### Acceptance Criteria

- [ ] `first_time_to_token` is added to the `LlmChatCompletionSummary` class via `this['response.first_time_to_token']`
- [ ] this attribute can be seen in the UI
- [ ] Appropriate tests are created

### Additional context
[Jira link](https://new-relic.atlassian.net/browse/NR-497191)

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-499484

PR #3781, opened 2026-02-25
feat: Add `time_to_first_token` for `LlmChatCompletionSummary`

amychisholm03:
This provides a base for #3581. Subsequent PRs for this issue will include vendor-specific logic to procure this value.

Also adds unit tests for `LlmChatCompletionSummary`.

### DIFF
--- a/test/unit/llm-events/chat-completion-summary.test.js
+++ b/test/unit/llm-events/chat-completion-summary.test.js
@@ -1,0 +1,298 @@
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const test = require('node:test')
+const assert = require('node:assert')
+const LlmChatCompletionSummary = require('#agentlib/llm-events/chat-completion-summary.js')
+const helper = require('../../lib/agent_helper')
+
+test.beforeEach((ctx) => {
+  ctx.nr = {}
+  ctx.nr.agent = helper.loadMockedAgent()
+})
+
+test.afterEach((ctx) => {
+  helper.unloadAgent(ctx.nr.agent)
+})
+
+test('should set all constructor properties when all params are provided', (t, end) => {
+  const { agent } = t.nr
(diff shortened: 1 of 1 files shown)


lib/llm-events/chat-completion-summary.js
```diff
@@ -47,17 +47,19 @@ module.exports = class LlmChatCompletionSummary extends LlmEvent {
...
-    responseOrg, temperature, maxTokens, numMsgs = 0, finishReason, error }) {
+    responseOrg, temperature, maxTokens, numMsgs = 0, finishReason, timeToFirstToken, error }) {
     super({ agent, segment, transaction, vendor, responseModel, requestId, error })
 
     if (requestModel) this['request.model'] = requestModel
     if (maxTokens) this['request.max_tokens'] = maxTokens
     if (temperature) this['request.temperature'] = temperature
     if (finishReason) this['response.choices.finish_reason'] = finishReason
     if (responseOrg) this['response.organization'] = responseOrg
+    if (timeToFirstToken) this['time_to_first_token'] = timeToFirstToken
```
bizob2828: do we want to do the calculation here instead of in every llm? meaning this is `firstToken - timestamp`?

amychisholm03: Yeah that makes sense

lib/llm-events/chat-completion-summary.js
```diff
@@ -61,6 +65,7 @@ module.exports = class LlmChatCompletionSummary extends LlmEvent {
 
     this['response.number_of_messages'] = numMsgs
     this.timestamp = segment.timer.start
+    if (timeOfFirstToken) this['time_to_first_token'] = timeOfFirstToken - this.timestamp
```
bizob2828, line 68: aren't there unit tests that should include this?

amychisholm03: There actually aren't any just for the base `LlmChatCompletionSummary`. For the vendor-specific classes that extend from `LlmChatCompletionSummary`, yes, there are.

PR #3786, opened 2026-03-02
chore: Removed unneccessary `generic-pool` context propagation

### LINKED ISSUE #3783: Determine if `generic-pool.js` is necessary with new context manager
### Description
The instrumentation for generic-pool just wrapps a segment so that the legacy context manager could propagate context correctly. With our new context manager, this is likely no longer necessary. We can confirm this by deleting `generic-pool.js` and then running the tests and asserting that they still pass.

We must go back to a version of agent where legacy context manager is default, remove instrumentation, and verify the tests fail.

### Acceptance Criteria

- [ ] generic-pool instrumentation is removed
- [ ] we verify it was necessary at some point in the agent with the legacy context manager

amychisholm03:
## Description

We have been reviewing some of our older instrumentation, and `lib/generic-pool.js` was one of them. Its only purpose was to wrap a segment for the legacy context manager. With the new context manger via `AsyncLocalStorage`, this is no longer required. This was confirmed by deleting `generic-pool.js` and running the current tests for it and having them still pass.

To assert that this was needed with the legacy context manager, I ran the same experiment with agent version `10.6.2` and Node version 14.21.3. Sure enough, the generic-pool tests failed without this instrumentation.

Because of the new context manager, we can safely removed this unnecessary code. Tests will stay.

## How to Test

```
npm run verisoned generic-pool
```

## Related Issues

closes #3783

### DIFF
--- a/lib/instrumentations.js
+++ b/lib/instrumentations.js
@@ -22,7 +22,6 @@ module.exports = function instrumentations() {
     bluebird: { type: InstrumentationDescriptor.TYPE_PROMISE },
     connect: { type: InstrumentationDescriptor.TYPE_WEB_FRAMEWORK },
     fastify: { type: InstrumentationDescriptor.TYPE_WEB_FRAMEWORK },
-    'generic-pool': { type: InstrumentationDescriptor.TYPE_GENERIC },
     kafkajs: { type: InstrumentationDescriptor.TYPE_MESSAGE },
     koa: { module: './instrumentation/koa' },
     memcached: { type: InstrumentationDescriptor.TYPE_DATASTORE },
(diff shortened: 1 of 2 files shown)


lib/instrumentation/generic-pool.js
```diff
@@ -1,33 +0,0 @@
-/*
- * Copyright 2020 New Relic Corporation. All rights reserved.
- * SPDX-License-Identifier: Apache-2.0
- */
-
-'use strict'
-
-module.exports = function initialize(agent, generic, moduleName, shim) {
```
bizob2828, line 8: you also have to remove it in `lib/instrumentations.js`

PR #3788, opened 2026-03-02
chore: Remove `@langchain/core` peer dep from langchain openai and aws tests

amychisholm03:
## Description

`@langchain/aws@1.3.0` calls `this._addVersion()` in the`ChatBedrockConverse` constructor, but that method was only introduced in `@langchain/core@1.1.29`. Both packages were released on the same day, and `@langchain/aws` declares a loose peer dep of `@langchain/core: ^1.0.0`, so the versioned test runner could pair them with an incompatible older `@langchain/core`.

By removing the peer dep `@langchain/core`, we can resolve this issue and potentially future ones. I also modified the `@langchain/openai` test package.json to match.

## How to Test

```
npm run versioned langchain-aws
npm run versioned langchain-openai
```

## Related Issues

None. CI fix.

### DIFF
--- a/test/versioned/langchain-aws/package.json
+++ b/test/versioned/langchain-aws/package.json
@@ -18,7 +18,7 @@
         "node": ">=20"
       },
       "dependencies": {
-        "@langchain/aws": ">=0.2.0"
+        "@langchain/aws": ">=0.1.3"
       },
       "files": [
         "runnables.test.js",


test/versioned/langchain-aws/package.json
```diff
@@ -18,14 +18,28 @@
         "node": ">=20"
       },
       "dependencies": {
-        "@langchain/aws": ">=1.0.0",
+        "@langchain/aws": ">=1.0.0 <1.3.0",
         "@langchain/core": ">=1.0.0"
```
bizob2828: it's worth noting i don't think we need `@langchain/core` as it is a peer dep of `@langchain/aws`. this could prob be solved by just removing that

test/versioned/langchain-openai/package.json
```diff
@@ -19,7 +19,6 @@
       },
       "comment": "@langchain/openai@>=1.2.8 imports @langchain/core/errors which was not exported until @langchain/core@1.1.18",
       "dependencies": {
-        "@langchain/core": ">=1.1.18",
         "@langchain/openai": ">=1.0.0"
```
bizob2828: you may want to explore reducing this to `>=0.2.0`, i just ran it locally with `--patch` and all versions pass.

test/versioned/langchain-aws/package.json
```diff
@@ -18,8 +18,7 @@
         "node": ">=20"
       },
       "dependencies": {
-        "@langchain/aws": ">=1.0.0",
-        "@langchain/core": ">=1.0.0"
+        "@langchain/aws": ">=1.0.0"
```
bizob2828: same here, not sure which version we want to make as the min tested version but can be less than the 1.x line now that you removed langchain/core as a dep under test

test/versioned/langchain-aws/package.json
```diff
@@ -18,8 +18,7 @@
         "node": ">=20"
       },
       "dependencies": {
-        "@langchain/aws": ">=1.0.0",
-        "@langchain/core": ">=1.0.0"
+        "@langchain/aws": ">=0.2.0"
```
bizob2828: there isn't a 0.2.x line of aws. looking at usage and peer dep of langchain core i think `>=0.1.3` is appropriate

test/versioned/langchain-openai/package.json
bizob2828, line 34: I think you can remove this as well

test/versioned/langchain-aws/package.json
bizob2828, line 35: I think you can remove this as well

PR #3790, opened 2026-03-02
feat: Added server-side configuration for `profiling.enabled`

### LINKED ISSUE #3721: [Profiling] server side configuration support
### Acceptance Criteria
As a Node.js agent user, I want the ability to enable/disable profiling from the NR1 UI.

bizob2828:
## Description

Adds server-side configuration for `profiling.enabled`. I had to change a few things around in the profiling manager. The profilers is now a Map, to avoid adding a profiler again when the profiling is re-enabled.  Now that the profilers is a map the name of the profiler is the key in the map, so i removed the name property from a profiler

## How to Test

```sh
npm run unit
```

## Related Issues
Closes #3721

### DIFF
--- a/lib/profiling/index.js
+++ b/lib/profiling/index.js
@@ -12,43 +12,43 @@ class ProfilingManager {
   constructor(agent, { logger = defaultLogger } = {}) {
     this.logger = logger
     this.config = agent.config.profiling
-    this.profilers = []
+    this.profilers = new Map()
     this.outputDir = process.cwd() + '/profiler-data'
   }
 
   register() {
-    if (this.config.include.includes('heap')) {
+    if (this.config.include.includes('heap') && !this.profilers.has('HeapProfiler')) {
       const { HeapProfiler } = require('./profilers')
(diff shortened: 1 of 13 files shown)


svetlanabrennan (approved):

lib/profiling/index.js
```diff
@@ -12,43 +12,43 @@ class ProfilingManager {
...
     this.logger = logger
     this.config = agent.config.profiling
-    this.profilers = []
+    this.profilers = new Map()
     this.outputDir = process.cwd() + '/profiler-data'
   }
 
   register() {
-    if (this.config.include.includes('heap')) {
+    if (this.config.include.includes('heap') && !this.profilers.has('HeapProfiler')) {
```
jsumners-nr, line 20: Will code outside of this file need to know about names like `HeapProfiler`?

bizob2828: I don't think so

bizob2828: why do you ask though?

PR #3802, opened 2026-03-03
refactor: Updated `redis` v3 instrumentation to subscribe to events emitted

### LINKED ISSUE #3796: Refactor `redis` `3.x` to subscribe to events emitted
### Description
We already migrated `redis@>=4` via `@redis/client` and `@node-redis/client` in #3773. Now, we should migrate the legacy 3.x line too.

### Acceptance Criteria

- [ ] `redis` subscribers are created
- [ ] tests still pass

amychisholm03:
## Description

Refactors `redis >=3 <4` instrumentation to use the subscriber model. Because we longer will have legacy `Callback` segments, the segment tree that the tests were asserting had to be modified slightly.

## How to Test

```
npm run versioned redis 
npm run unit
```

## Related Issues

closes #3796


lib/subscribers/redis/config.js
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const internalSendCommand = {
+  path: './redis/internal-send-command',
+  instrumentations: [{
+    channelName: 'nr_internalSendCommand',
+    module: { name: 'redis', versionRange: '>=3 <4', filePath: 'index.js' },
```
bizob2828: i know we only test on 3.1.0+ but we only dropped <2.6.0 in 12.x, we never announced we don't support 2.6.0-3.1.0. so i would change this:

```suggestion
    module: { name: 'redis', versionRange: '>=2.6.0 <4', filePath: 'index.js' },
```

lib/subscribers/redis/internal-send-command.js
```diff
@@ -0,0 +1,60 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const DbOperationSubscriber = require('../db-operation')
+const stringify = require('json-stringify-safe')
+
+module.exports = class RedisInternalSendCommandSubscriber extends DbOperationSubscriber {
+  constructor({ agent, logger }) {
+    super({ agent, logger, packageName: 'redis', channelName: 'nr_internalSendCommand', system: 'Redis' })
+    this.events = ['end']
```
bizob2828: this isn't propagating the time as it wraps it in callback so you must subscribe to `asyncEnd` as well

bizob2828: I think we need to add a test that can assert the time like this:

```js
  await t.test('should time segments accordingly', async function (t) {
    const { agent, client } = t.nr
    await helper.runInTransaction(agent, async function transactionInScope() {
      const transaction = agent.getTransaction()
      assert.ok(transaction, 'transaction should be visible')

      const WAIT = 2000
        client.wait(1, WAIT, function (error, ok) {
          assert.ok(!error)
          assert.ok(agent.getTransaction(), 'transaction should still be visible')
          assert.equal(ok, 0)
          resolve()
        })
      })

      const trace = transaction.trace
      const children = trace.getChildren(trace.root.id)
      assert.equal(children.length, 1, 'there should be only one child of the root')

      const [waitSegment] = children
      assert.ok(waitSegment, 'trace segment for set should exist')
      assert.equal(waitSegment.name, 'Datastore/operation/Redis/wait', 'should register the set')
      assert.ok(waitSegment.timer.hrDuration, 'trace segment should have ended')
      assert.ok(waitSegment.getDurationInMillis() > WAIT)
      assert.ok(waitSegment.getDurationInMillis() < WAIT + 100)
    })
  })
```

In fact we should have a similar test in the v4 but that will be slightly diff and can be in a different PR

test/versioned/redis/redis-v3.test.js
```diff
@@ -371,4 +364,33 @@ test('Redis instrumentation', { timeout: 20000 }, async function (t) {
...
+      const trace = transaction.trace
+      const children = trace.getChildren(trace.root.id)
+      assert.equal(children.length, 1, 'there should be only one child of the root')
+
+      const [waitSegment] = children
+      assert.ok(waitSegment, 'trace segment for wait should exist')
+      assert.equal(waitSegment.name, 'Datastore/operation/Redis/wait', 'should register the wait')
+      assert.ok(waitSegment.timer.hrDuration, 'trace segment should have ended')
+      assert.ok(waitSegment.getDurationInMillis() > WAIT)
+      assert.ok(waitSegment.getDurationInMillis() < WAIT + 100)
```
bizob2828: you might want to remove this. i ran the suite a few times and it was failing for being a few ms off

amychisholm03: Yeah, I removed it for #3810, but forgot to here.

PR #3805, opened 2026-03-04
chore: Refactored grpc-js server instrumentation to subscriber type

### LINKED ISSUE #3799: Migrate grpc (server) to subscriber type
As titled.

jsumners-nr:
This PR resolves #3799.

### DIFF
--- a/test/versioned/grpc/grpc-server.cjs
+++ b/test/versioned/grpc/grpc-server.cjs
@@ -44,6 +44,7 @@ function sayHello({ metadata, request: { name } }, cb) {
 }
 
 function sayHelloClientStream(call, cb) {
+  const ctx = AGENT.tracer.getContext()
   const { metadata } = call
   const names = []
   call.on('data', function (clientStream) {
@@ -53,7 +54,9 @@ function sayHelloClientStream(call, cb) {
   })
   call.on('end', function () {
     cb(null, {
-      message: `Hello ${names.join(', ')}`
+      message: `Hello ${names.join(', ')}`,
+      transaction_id: ctx.transaction?.id,
+      segment_name: ctx.segment?.name
     })
   })
 }
@@ -77,12 +80,17 @@ function sayHelloServerStream(call) {
 }
 
 function sayHelloBidiStream(call) {
+  const ctx = AGENT.tracer.getContext()
   const { metadata } = call
   call.on('data', (clientStream) => {
     const { name } = clientStream
     // add the metadata from client that the server receives so we can assert DT functionality
     SERVER.metadataMap.set(name, metadata.internalRepr)
-    call.write({ message: `Hello ${name}` })
+    call.write({
+      message: `Hello ${name}`,
+      transaction_id: ctx.transaction?.id,
+      segment_name: ctx.segment?.name
+    })
   })
   call.on('end', () => {
     call.end()
--- a/test/versioned/grpc/server-bidi-streaming.test.js
+++ b/test/versioned/grpc/server-bidi-streaming.test.js
@@ -66,8 +66,8 @@ test('should track bidirectional requests', async (t) => {
   names.forEach(({ name }, i) => {
     const response = responses[i]
     assert.equal(response.message, `Hello ${name}`, 'response stream message should be correct')
-    assert.ok(Object.hasOwn(response, 'transaction_id'))
-    assert.ok(Object.hasOwn(response, 'segment_name'))
+    assert.equal(response.transaction_id, transaction.id)
+    assert.equal(response.segment_name, '/helloworld.Greeter/SayHelloBidiStream')
   })
 
   assert.ok(transaction, 'transaction exists')
--- a/test/versioned/grpc/server-client-streaming.test.js
+++ b/test/versioned/grpc/server-client-streaming.test.js
@@ -68,8 +68,8 @@ test('should track client streaming requests', async (t) => {
     `Hello ${names.map(({ name }) => name).join(', ')}`,
     'response message is correct'
   )
-  assert.ok(Object.hasOwn(response, 'transaction_id'))
-  assert.ok(Object.hasOwn(response, 'segment_name'))
+  assert.equal(response.transaction_id, transaction.id)
+  assert.equal(response.segment_name, '/helloworld.Greeter/SayHelloClientStream')
   assertServerTransaction({ transaction, fnName: 'SayHelloClientStream' })
   assertServerMetrics({ agentMetrics: agent.metrics._metrics, fnName: 'SayHelloClientStream' })
 })
--- a/test/versioned/grpc/server-streaming.test.js
+++ b/test/versioned/grpc/server-streaming.test.js
@@ -66,8 +66,8 @@ test('should track server-streaming requests', async (t) => {
   names.forEach((name, i) => {
     const response = responses[i]
     assert.equal(response.message, `Hello ${name}`, 'response stream message should be correct')
-    assert.ok(Object.hasOwn(response, 'transaction_id'))
-    assert.ok(Object.hasOwn(response, 'segment_name'))
+    assert.equal(response.transaction_id, transaction.id)
+    assert.equal(response.segment_name, '/helloworld.Greeter/SayHelloServerStream')
   })
   assert.ok(transaction, 'transaction exists')
   assertServerTransaction({ transaction, fnName: 'SayHelloServerStream' })


lib/subscribers/grpcjs/server.js
```diff
@@ -0,0 +1,206 @@
...
+        const error = Error(`gRPC status code ${statusCode}: ${status}`)
+        agent.errors.add(transaction, error)
+      }
+      transaction.end()
+      return onCallEnd.apply(stream.call.callEventTracker, args)
+    }
+  }
+
+  #loadConstants() {
+    const modPath = path.dirname(require.resolve('@grpc/grpc-js'))
```
bizob2828: i wouldn't do this. FWIW this isn't the dependency under test, it's our grpc dep in the root of the agent. This mapping is only being used to provide both the id and status name in errors. I would just drop that in the message and just log what we get

lib/subscribers/grpcjs/server.js
```diff
@@ -0,0 +1,191 @@
...
+    headers.tracestate = metadata.get('tracestate').join(',')
+    headers.traceparent = metadata.get('traceparent').join(',')
+    headers.newrelic = metadata.get('newrelic').join(',')
+    transaction.acceptDistributedTraceHeaders('HTTP', headers)
+  }
+
+  #createContext(instance, handlerName, handlerType) {
+    let ctx = instance.agent.tracer.getContext()
+    let transaction = ctx.transaction
+    if (!transaction) {
```
bizob2828: based on the [legacy code shim code](https://github.com/newrelic/node-newrelic/blob/d8f6d0f6ac7f3ff17a667391269c41e6e57dc2fb/lib/shim/transaction-shim.js#L411). this actually logs a trace level log and calls the original function. it does not proceed with creating a new baseSegment. since this is responsible for update the context with a new transaction/segment you may have to tweak this because even if you just return the context here, it's going to attempt to wrap functions and accept DT, etc

lib/subscribers/grpcjs/server.js
```diff
@@ -0,0 +1,191 @@
...
+      const stream = args[0]
+      self.#acceptDtHeaders(self, transaction, stream)
+
+      if (semver.gte(moduleVersion, '1.10.0') === true) {
+        self.#instrumentInterceptors(self, transaction, stream)
+      } else {
+        self.#instrumentEventListeners(self, transaction, stream)
+      }
+
+      return handlerFn.apply(server, args)
```
bizob2828: this function has to be run in context, line 53 is returning context but you're only getting transaction from it
```suggestion
      return self.agent.tracerBindFunction(handlerFn, <context>).apply(server, args)
```

lib/subscribers/grpcjs/server.js
```diff
@@ -0,0 +1,191 @@
...
+    if (server.handlers.has(handlerName) === true) {
+      this.logger.debug(
+        `Not re-instrumenting gRPC method handler for ${handlerName}: it is already registered in the server.`
+      )
+      return server.register(...arguments)
+    }
+
+    const self = this
+    _arguments[1] = function wrappedGrpcHandler(...args) {
+      const { transaction } = self.#createContext(self, handlerName, handlerType)
```
bizob2828:
```suggestion
      const ctx = self.#createContext(self, handlerName, handlerType)
      const { transaction } = ctx
```

test/versioned/grpc/package.json
```diff
@@ -9,7 +9,7 @@
         "node": ">=20"
       },
       "dependencies": {
-        "@grpc/grpc-js": ">=1.4.0"
+        "@grpc/grpc-js": ">=1.10.0"
```
bizob2828: this should be reverted

lib/subscribers/grpcjs/server.js
bizob2828: I think we should refactor the grpc test server to pass in agent and assert the correct tx and segment exist in the handler functions. If we had this you would've known this is missing. Also it appears the older instrumentation is also partially broken, at least in the event listeners. Here's what i'm proposing you change in the test server(note: this is only updating a few handlers but we should assert the tx is active and the proper segment name in various areas):

```sh
diff --git a/test/versioned/grpc/grpc-server.cjs b/test/versioned/grpc/grpc-server.cjs
index 9664bda00..69273bd56 100644
--- a/test/versioned/grpc/grpc-server.cjs
+++ b/test/versioned/grpc/grpc-server.cjs
@@ -5,23 +5,37 @@
 
 'use strict'
 const { ERR_MSG, ERR_CODE, HALT_SERVER_ERR_MSG, HALT_CODE } = require('./constants.cjs')
-module.exports = function createServerMethods(server) {
+module.exports = function createServerMethods(server, agent) {
   return {
     sayHello: function sayHello({ metadata, request: { name } }, cb) {
+      const ctx = agent.tracer.getContext()
+      assert.ok(ctx.transaction)
+      assert.equal(ctx.transaction.isActive(), true)
+      assert.equal(ctx.segment.name, '/helloworld.Greeter/SayHello')
       // add the metadata from client that the server receives so we can assert DT functionality
       server.metadataMap.set(name, metadata.internalRepr)
       const message = `Hello ${name}`
       cb(null, { message })
     },
     sayHelloClientStream: function sayHelloCStream(call, cb) {
+      debugger
+      const ctx = agent.tracer.getContext()
+      console.log('ctx in handler', ctx?.segment?.name)
       const { metadata } = call
       const names = []
       call.on('data', function (clientStream) {
+        debugger
+        const ctx = agent.tracer.getContext()
+        console.log('ctx on data', ctx?.segment?.name)
         const { name } = clientStream
         server.metadataMap.set(name, metadata.internalRepr)
         names.push(name)
       })
       call.on('end', function () {
+        debugger
+        const ctx = agent.tracer.getContext()
+        console.log('ctx onEnd', ctx?.segment?.name)

diff --git a/test/versioned/grpc/server-client-streaming.test.js b/test/versioned/grpc/server-client-streaming.test.js
index c9aed692c..e3bfaa606 100644
--- a/test/versioned/grpc/server-client-streaming.test.js
+++ b/test/versioned/grpc/server-client-streaming.test.js
@@ -32,7 +32,7 @@ test.beforeEach(async (ctx) => {
   ctx.nr.agent = helper.instrumentMockedAgent()
   ctx.nr.grpc = require('@grpc/grpc-js')
 
-  const { port, proto, server } = await createServer(ctx.nr.grpc)
+  const { port, proto, server } = await createServer(ctx.nr.grpc, ctx.nr.agent)
   ctx.nr.port = port
   ctx.nr.proto = proto
   ctx.nr.server = server
```

bizob2828: I'm ok if we defer properly wrapping the stream.emit and running it on context as it's currently broken but if we had these kind of tests before we would've known to add this instrumentation.

bizob2828 (changes requested):
first pass. The lack of assertions in our example server did not expose some of my comments i called out when porting migration. let me know if you have any questions

lib/subscribers/grpcjs/server.js
```diff
@@ -0,0 +1,191 @@
...
+      const stream = args[0]
```
bizob2828: there's prob more for bidi streaming but for one way streaming at least from client, this code binds the context where in the `data` `end` events it has the right context:

```js
      const emit = stream.emit
      stream.emit = function wrappedEmit(...args) {
        const [ eventName ] = args
        console.log('event name', eventName)
        debugger
        return self.agent.tracer.bindFunction(emit, ctx).apply(this, args)
      }
```

bizob2828: this is also prob fine but this is technically not a stream at all times.  for unary it's just a call

lib/subscribers/grpcjs/server.js
jsumners-nr: I think I have addressed this issue in https://github.com/newrelic/node-newrelic/pull/3805/commits/751651d9d119bd16d8873d37da1f4b6866a0986b.

lib/subscribers/grpcjs/server.js
jsumners-nr: I believe I have this addressed in https://github.com/newrelic/node-newrelic/pull/3805/commits/6eda7f0e8bd8534cac4c3b313177388212f06f09.

test/versioned/grpc/grpc-server.cjs
```diff
@@ -4,89 +4,116 @@
...
+  SERVER.metadataMap.set(name, metadata.internalRepr)
+  const message = `Hello ${name}`
+  cb(null, {
+    message,
+    transaction_id: ctx.transaction?.id,
+    segment_name: ctx.segment?.name
+  })
+}
+
+function sayHelloClientStream(call, cb) {
```
bizob2828, line 46: all of these functions could pass in context. it would have to be something like:

```
const ctx = AGENT.tracer.getContext()
let dataCtx
call.on('data', function (clienStream) {
  dataCtx = AGENT.tracer.getContext()
})

call.on('end, function () {
  const endCtx = AGENT.tracer.getContext()
  cb(null, {
    message: `Hello ${names.join(', ')}`,
    cbCtx: {
     transaction_id: ctx.transaction.id,
     segment_name: ctx.segment.name
   },
   dataCtx: {
     transaction_id: dataCtx.transaction.id,
     segment_name: dataCtx.segment.name
   },
   endCtx: {
     transaction_id: endCtx.transaction.id,
     segment_name: endCtx.segment.name
   }
  }
})
```

At this point since you aren't binding the stream to the context the data/end won't work, which i think we agreed to defer but you still should and can assert the ctx in the callback

test/versioned/grpc/example.proto
```diff
@@ -28,4 +28,6 @@ message ArrayHelloRequest {
 
 message HelloReply {
   string message = 1;
+  string transaction_id = 2;
```
bizob2828, line 31: in the follow up PR we'll have to change this to also include the tx id and segment name from the data and end events on stream

lib/subscribers/grpcjs/server.js
```diff
@@ -0,0 +1,215 @@
...
+      return server.register(...arguments)
+    }
+
+    const self = this
+    _arguments[1] = function wrappedGrpcHandler(...args) {
+      let ctx = self.agent.tracer.getContext()
+      if (ctx.transaction != null) {
+        // If a transaction already exists, we run the function under that
+        // transaction.
+        return self.agent.tracer.bindFunction(handlerFn, ctx).apply(server, args)
```
bizob2828: I don't think this is hurting anything but I don't think you have to bind the function to the current context, it already is bound to that context

lib/subscribers/grpcjs/server.js
```diff
@@ -0,0 +1,215 @@
...
+   * @param {string} handlerName The name of the handler. Used to name the
+   * segment.
+   * @param {string} handlerType The gRPC server handler type as enumerated at
+   * https://github.com/grpc/grpc-node/blob/c9f8f93/packages/grpc-js/src/server-call.ts#L397
+   *
+   * @returns {AsyncContext} A new context to be utilized for the current
+   * request.
+   */
+  #createContext(instance, handlerName, handlerType) {
+    let ctx = instance.agent.tracer.getContext()
```
bizob2828: since you're getting the context above in the caller function you can just pass in the context  and also remove this redundant check for the transaction being null

bizob2828 (changes requested):
you're missing asserting the tx and segment in all types of servers. we can skip the assertions of tx and segment in the stream handlers as that instrumentation will be done in a follow up. However, it is still good to assert the tx/segment in every grpc server handler callback

test/versioned/grpc/grpc-server.cjs
jsumners-nr: I purposefully left the streaming changes out of this PR because we agreed to focus on those separately.

test/versioned/grpc/grpc-server.cjs
bizob2828, line 84: need to get the context and add  `transaction_id` and `segment_name` in the `call.write`

test/versioned/grpc/server-bidi-streaming.test.js
```diff
@@ -48,19 +49,25 @@ test.afterEach((ctx) => {
...
+  const responses = request.value
+  assert.ok(trace.value, 'transaction exists')
+  const [transaction] = trace.value
+
   names.forEach(({ name }, i) => {
-    assert.equal(responses[i], `Hello ${name}`, 'response stream message should be correct')
+    const response = responses[i]
+    assert.equal(response.message, `Hello ${name}`, 'response stream message should be correct')
+    assert.ok(Object.hasOwn(response, 'transaction_id'))
+    assert.ok(Object.hasOwn(response, 'segment_name'))
```
bizob2828:
```suggestion
    assert.equal(response.transaction_id, transaction.id)
    assert.equal(response.segment_name, '/helloworld.Greeter/SayHelloBidiStream')
```

test/versioned/grpc/server-client-streaming.test.js
```diff
@@ -48,23 +49,27 @@ test.afterEach((ctx) => {
...
+  assert.ok(trace.value, 'transaction exists')
+  const [transaction] = trace.value
+
   assert.equal(
     response.message,
     `Hello ${names.map(({ name }) => name).join(', ')}`,
     'response message is correct'
   )
+  assert.ok(Object.hasOwn(response, 'transaction_id'))
+  assert.ok(Object.hasOwn(response, 'segment_name'))
```
bizob2828:
```suggestion
    assert.equal(response.transaction_id, transaction.id)
    assert.equal(response.segment_name, '/helloworld.Greeter/SayHelloClientStream')
```

test/versioned/grpc/server-streaming.test.js
```diff
@@ -48,19 +49,25 @@ test.afterEach((ctx) => {
...
+  const responses = request.value
+  assert.ok(trace.value, 'transaction exists')
+  const [transaction] = trace.value
+
   names.forEach((name, i) => {
-    assert.equal(responses[i], `Hello ${name}`, 'response stream message should be correct')
+    const response = responses[i]
+    assert.equal(response.message, `Hello ${name}`, 'response stream message should be correct')
+    assert.ok(Object.hasOwn(response, 'transaction_id'))
+    assert.ok(Object.hasOwn(response, 'segment_name'))
```
bizob2828:
```suggestion
     assert.equal(response.transaction_id, transaction.id)
    assert.equal(response.segment_name, '/helloworld.Greeter/SayHelloServerStream')
```

test/versioned/grpc/grpc-server.cjs
```diff
@@ -4,89 +4,121 @@
...
+
+function sayError(call, cb) {
+  return cb({
+    code: ERR_CODE,
+    message: ERR_MSG
+  })
+}
+
+function sayErrorClientStream(call, cb) {
+  call.on('data', function (stream) {
```
bizob2828, line 108: need to get the context and add `transaction_id` and `segment_name` in the `cb` response

test/versioned/grpc/grpc-server.cjs
bizob2828: streaming changes are getting the context in `call.on('data'` and `call.on('end'`

PR #3808, opened 2026-03-05
refactor: Refactor `winston` instrumentation to subscribe to events emitted

### LINKED ISSUE #3443: Update winston instrumentation to rely on tracing channel instrumentation
### Description
Update instrumentation to create a subscriber for winston and a instrumentator for orchestrion.  

### Acceptance Criteria
winston instrumentation still functions as expected after traditional monkey patching and shim helpers are used.

amychisholm03:
## Description

Refactors our `winston` instrumentation to subscribe to events emitted.

## How to Test

```
npm run versioned winston
```

## Related Issues

Closes #3443

### DIFF
--- a/lib/subscribers/winston/configure.js
+++ b/lib/subscribers/winston/configure.js
@@ -5,16 +5,13 @@
 
 const ApplicationLogsSubscriber = require('../application-logs')
 const NrTransport = require('./nr-winston-transport.js')
-const { format } = require('logform')
 
 /**
- * This is the subscriber for Logger.configure.
- * Logger.constructor calls this.configure(options), so this fires
+ * This is the subscriber for `Logger.configure`.
+ * `Logger.constructor` calls `this.configure(options)`, so this fires
  * during initial construction and on any subsequent reconfiguration.
  *
- * Handles both:
- * - NR formatter registration (before configure runs, via handler)
- * - NrTransport injection (after configure runs, via end)
+ * Handles `NrTransport` injection (after `configure` runs, via `end`).
  */
 module.exports = class WinstonConfigure extends ApplicationLogsSubscriber {
   constructor({ agent, logger }) {
@@ -22,23 +19,8 @@ module.exports = class WinstonConfigure extends ApplicationLogsSubscriber {
     this.events = ['end']
   }
 
-  /**
-   * Runs before Logger.configure executes.
-   * Registers the NR formatter on opts so configure sets this.format correctly.
-   * This works for both initial construction and reconfiguration since
-   * configure() always sets this.format from the options.
-   * @param {object} data event data
-   * @param {Context} ctx active context
-   */
   handler(data, ctx) {
-    const { arguments: args } = data
-    const opts = (args[0] = args[0] || {})
-
     this.createModuleUsageMetric('winston')
-
-    if (this.isLocalDecoratingEnabled() || this.isMetricsEnabled()) {
-      this.#registerFormatter({ opts })
-    }
     return ctx
   }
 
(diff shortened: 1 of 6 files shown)


lib/subscribers/winston/configure.js
```diff
@@ -0,0 +1,116 @@
...
+   * Apply a formatter to keep track of logging metrics, and in the case of local decorating appending
+   * the NR-LINKING metadata to message.  We want to do this first so any formatter that is transforming
+   * data will have the changes.
+   *
+   * @param {object} params object passed to function
+   * @param {object} params.opts options from configure()
+   */
+  #registerFormatter({ opts }) {
+    // eslint-disable-next-line n/no-extraneous-require
+    const { format } = require('logform')
```
bizob2828: We can't require a module in instrumentation, especially one that's not there. This was one of the problems I faced originally.

lib/subscribers/winston/configure.js
```diff
@@ -0,0 +1,115 @@
...
+   *
+   * @param {object} data event data
+   */
+  end(data) {
+    const logger = data.self
+    if (!this.isLogForwardingEnabled()) {
+      return
+    }
+
+    const hasNrTransport = logger.transports?.some((t) => t.name === 'newrelic')
```
bizob2828, line 39: can you explain this? i'm reading it's only adding a logger if one isn't there but is this possible? I know in our old instrumentation we instrumented a layer above and added this transport but when configure was called we'd have to add it back in because it could get removed. This is getting run after configure is getting called so it's not the same situation

lib/subscribers/winston/configure.js
```diff
@@ -0,0 +1,115 @@
...
+   * @param {Context} ctx active context
+   */
+  handler(data, ctx) {
+    const { arguments: args } = data
+    const opts = (args[0] = args[0] || {})
+
+    this.createModuleUsageMetric('winston')
+
+    if (this.isLocalDecoratingEnabled() || this.isMetricsEnabled()) {
+      this.#registerFormatter({ opts })
```
bizob2828: i might be wrong but we may be able to simplify this now and run this after the logger is constructed in the `end` method below

bizob2828 (changes requested):
I gave you patch to remove our dependence on logform

lib/subscribers/winston/configure.js
amychisholm03: We still need it or else we'll double log i.e. fail this test:

```
test at winston.test.js:354:11
✖ should not double log nor instrument composed logger (16.350958ms)
  AssertionError [ERR_ASSERTION]: should only add 3 log to aggregator
      at /Users/achisholm/Desktop/amychisholm03-node-newrelic/test/versioned/winston/helpers.js:79:12
      at logWithAggregator (/Users/achisholm/Desktop/amychisholm03-node-newrelic/test/versioned/winston/helpers.js:75:11)
      at TestContext.<anonymous> (/Users/achisholm/Desktop/amychisholm03-node-newrelic/test/versioned/winston/winston.test.js:380:5)
      at Test.runInAsyncScope (node:async_hooks:214:14)
      at Test.run (node:internal/test_runner/test:1090:21)
      at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
      at async TestContext.<anonymous> (/Users/achisholm/Desktop/amychisholm03-node-newrelic/test/versioned/winston/winston.test.js:354:3)
      at async Test.run (node:internal/test_runner/test:1110:7)
      at async Test.processPendingSubtests (node:internal/test_runner/test:785:7) {
    generatedMessage: false,
    code: 'ERR_ASSERTION',
    actual: 4,
    expected: 3,
    operator: '==',
    diff: 'simple'
  }
```

bizob2828: ah ok, the previous context was within `winston.loggers.add` where we had a set to compare against with an id where as now we don't but we do have a name for our transport

PR #3809, opened 2026-03-06
fix: Reduce per-request overhead in middleware wrapper and context manager

malwarebo:
## Problem
_The image shows the spike in latency after we deployed the newrelic update in production._
Screenshot 2026-03-03 at 1 54 18 PM

After upgrading the New Relic Node.js agent from v13.5.0 to v13.6.0 and the latest version which is v13.14.0, we observed a noticeable increase in latency at the Node.js level in our Express application. The latency increase appeared immediately after the upgrade and was visible in the Node.js portion of web transaction time in APM.

The root cause is per-request overhead in the subscriber-based instrumentation introduced in v13.6.0, specifically in the middleware wrapper and the underlying context manager's `runInContext` method.

### Changes 
_Context manager (`async-local-context-manager.js`)_

Simplify `runInContext` to stop allocating a new bound function (`.bind()`) and stop spreading the `arguments` object on every call. Both are expensive in V8 and were happening on every instrumented function across the entire agent and not just Express, but MySQL, Redis, HTTP, etc. Replaced with a single `Reflect.apply` passthrough that achieves the same result with zero allocations.

_Middleware wrapper (`middleware-wrapper.js`)_

Bypassed the `bindFunction` to `_makeWrapped` to `runInContext` indirection chain for both the handler and done/next callback. Previously each middleware invocation created multiple intermediate closures, assigned symbol properties, and went through several layers of wrapping just to call `AsyncLocalStorage.run()`. Now both paths call `ALS.run` directly.
Also removed a per-request `Object.defineProperties()` call on the done handler and added an explicit `segment.touch()` in a `finally` block to ensure segment timing is always recorded.

_Net effect from our benchmarks_

For a typical request hitting ~13 middleware, this eliminates roughly 65-70 unnecessary heap allocations per request (closures, bound functions, symbol assignments) compared to the pre-fix state.

### DIFF
--- a/test/benchmark/shim/webframework-shim/middlewareWrapper.bench.js
+++ b/test/benchmark/shim/webframework-shim/middlewareWrapper.bench.js
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const benchmark = require('#testlib/benchmark.js')
+const helper = require('#testlib/agent_helper.js')
+
+const agent = helper.loadMockedAgent()
+const tracer = helper.getTracer()
+const als = tracer._contextManager._asyncLocalStorage
+
+const suite = benchmark.createBenchmark({ name: 'handler invocation', runs: 10000 })
+
+const transaction = helper.runInTransaction(agent, function (tx) {
+  return tx
+})
+
+function handler(req, res, next) {
+  return next()
+}
+
+function noop() {}
+
+tracer.setSegment({ transaction, segment: transaction.trace.root })
+const ctx = tracer.getContext()
+const segment = tracer.createSegment({
+  name: 'Nodejs/Middleware/Benchmarks/handler',
+  parent: ctx.segment,
+  transaction
+})
+const newCtx = ctx.enterSegment({ segment })
+
+// warmup both paths
+for (let i = 0; i < 100000; ++i) {
+  // eslint-disable-next-line no-useless-call
+  tracer.bindFunction(handler, newCtx, true).apply(null, [{}, {}, noop])
+  als.run(newCtx, Reflect.apply, handler, null, [{}, {}, noop])
+}
+
+setTimeout(function () {
+  suite.add({
+    name: 'bindFunction (full=true)',
+    fn: function () {
+      tracer.setSegment({ transaction, segment: transaction.trace.root })
+      for (let i = 0; i < 100; ++i) {
+        // eslint-disable-next-line no-useless-call
+        tracer.bindFunction(handler, newCtx, true).apply(null, [{}, {}, noop])
+      }
+    }
+  })
+
+  suite.add({
+    name: 'bindFunction (full=false)',
+    fn: function () {
+      tracer.setSegment({ transaction, segment: transaction.trace.root })
+      for (let i = 0; i < 100; ++i) {
+        // eslint-disable-next-line no-useless-call
+        tracer.bindFunction(handler, newCtx, false).apply(null, [{}, {}, noop])
+      }
+    }
+  })
+
+  suite.add({
+    name: 'direct ALS.run + reflect.apply',
+    fn: function () {
+      tracer.setSegment({ transaction, segment: transaction.trace.root })
+      for (let i = 0; i < 100; ++i) {
+        als.run(newCtx, Reflect.apply, handler, null, [{}, {}, noop])
+      }
+    }
+  })
+
+  suite.run()
+}, 15)
--- a/test/benchmark/tracer/runInContext.bench.js
+++ b/test/benchmark/tracer/runInContext.bench.js
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const helper = require('#testlib/agent_helper.js')
+const shared = require('./shared')
+
+const s = shared.makeSuite('runInContext')
+const suite = s.suite
+const tracer = helper.getTracer()
+const ctxManager = tracer._contextManager
+const als = ctxManager._asyncLocalStorage
+
+const tx = helper.runInTransaction(s.agent, function (_tx) {
+  return _tx
+})
+tracer.setSegment({ transaction: tx, segment: tx.trace.root })
+const ctx = tracer.getContext()
+
+function handler(a, b, c) {
+  return a + b + c
+}
+
+function runInContextWithBind(context, callback, cbThis, args) {
+  const toInvoke = cbThis ? callback.bind(cbThis) : callback
+  if (args) {
+    return als.run(context, toInvoke, ...args)
+  }
+  return als.run(context, toInvoke)
+}
+
+function runInContextWithReflect(context, callback, cbThis, args) {
+  if (cbThis || args) {
+    return als.run(context, Reflect.apply, callback, cbThis ?? null, args ?? [])
+  }
+  return als.run(context, callback)
+}
+
+const thisArg = {}
+
+// warmup
+for (let i = 0; i < 100000; ++i) {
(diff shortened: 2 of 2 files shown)


bizob2828:
@malwarebo we will take a look. At first glance it makes sense but we may have change parts. Can you share the benchmark tests you can please?

malwarebo:
> @malwarebo we will take a look. At first glance it makes sense but we may have change parts. Can you share the benchmark tests you can please?

Note: These are not from a production env but a test environemnt

**`runInContext`  : core context propagation method**

Called on every instrumented function across the entire agent (express, etc.). The `arguments` object case is the real-world code path from `_makeWrapped`.

Runs: 10,000 × 100 calls each

| Approach | Mean (ms) | Median (ms) | p95 (ms) |
|----------|-----------|-------------|----------|
| Old: `.bind(cbThis)` + `...arguments` spread | 0.0160 | 0.016 | 0.019 |
| New: `Reflect.apply` passthrough | 0.0015 | 0.001 | 0.002 |

**~10.9 times faster** for the hot path.

______________________________________________________________________________

**Handler invocation : per-middleware execution**

Measures the cost of invoking a middleware handler within a traced context.

Runs: 10,000 × 100 calls each

| Approach | Mean (ms) | Median (ms) | p95 (ms) |
|----------|-----------|-------------|----------|
| `bindFunction(full=true).apply()` | 0.0190 | 0.017 | 0.026 |
| `bindFunction(full=false).apply()` | 0.0186 | 0.017 | 0.022 |
| Direct `ALS.run(ctx, Reflect.apply, ...)` | 0.0144 | 0.014 | 0.017 |

**~25% faster** per handler call, compounding across 13+ middleware per request.

The express 5 service I tested against has the following middleware executin on every request:

1. Auth/context configuration middleware
2. Body parser (`express.json`)
3. Async context restoration middleware
4. Local storage context setup
5. Request ID propagation
6. Error tracking request tagging
7. CORS
8. Route dispatch layer
9. Logging decorator middleware
10. App/consumer authorization
11. Content-type enforcement
12. Request body validation
13. Route handler wrapper

That's **13 middleware per request**, each triggering the `wrappedHandler` path twice (once for the handler, once for the done/next callback). Combined with MySQL, Redis, and external HTTP calls to payment gateways, a single request
generates **40–60 `runInContext` calls** — all of which benefit from the 10.9x improvement in the core path.

These results can be reproduced by benchmarking `runInContext` with an `arguments` object (old `.bind()` + spread vs new `Reflect.apply`) and by benchmarking `bindFunction().apply()` vs direct `ALS.run()` for handler invocation within a traced transaction context.

bizob2828:
@malwarebo can you please post the code you used to run the benchmarks, a simple link to a github repo with the example app is fine. I understand the root cause fix but I want to experiment myself

malwarebo:
> @malwarebo can you please post the code you used to run the benchmarks, a simple link to a github repo with the example app is fine. I understand the root cause fix but I want to experiment myself

added with the PR. If you are wondering about, Why test in isolation, not end-to-end?

Because segment creation overhead is unchanged. Isolating the changed codepath gives a cleaner signal. The cumulative impact scales with middleware depth.

bizob2828:
thanks for updating. there's a larger change here that we have yet to explore but has existed for a few years now. the overhead of `tracer.bindFunction` was known but the side effect of it wasn't caught when migrating certain instrumentation. again, thanks for the PR, if we don't land it, a more holistic fix will be done as this affects more than just middleware

bizob2828:
@malwarebo do you mind providing an application to reproduce this issue? I understand running the benchmark shows it's faster, and I see that as well. However when I test with an express application that has middleware that either make redis calls, http calls or just sleep, I don't see a different in the `Node.js` time within that graph you linked above between 13.5.0 to 13.6.0 to my latest PR

malwarebo:
> @malwarebo do you mind providing an application to reproduce this issue? I understand running the benchmark shows it's faster, and I see that as well. However when I test with an express application that has middleware that either make redis calls, http calls or just sleep, I don't see a different in the `Node.js` time within that graph you linked above between 13.5.0 to 13.6.0 to my latest PR

I can put together a minimal repro app (maybe using autocannon), but before that I want to clarify something that might explain why you're not seeing the difference. 

The key is middleware composition, in our case I think. The service runs around 13 middleware per request and most of them are synchronous and lightweight as in CORS headers, security headers, auth token extraction, rate limiting, body validation, request logging. Each one does microseconds of actual work. With lightweight middleware, the instrumentation overhead stands out. With I/O-heavy middleware, it probably gets buried under redis/HTTP latency and likely won't show up in the graph.

Please correct me if I am missing anything here.

bizob2828:
> > @malwarebo do you mind providing an application to reproduce this issue? I understand running the benchmark shows it's faster, and I see that as well. However when I test with an express application that has middleware that either make redis calls, http calls or just sleep, I don't see a different in the `Node.js` time within that graph you linked above between 13.5.0 to 13.6.0 to my latest PR
> 
> I can put together a minimal repro app (maybe using autocannon), but before that I want to clarify something that might explain why you're not seeing the difference.
> 
> The key is middleware composition, in our case I think. The service runs around 13 middleware per request and most of them are synchronous and lightweight as in CORS headers, security headers, auth token extraction, rate limiting, body validation, request logging. Each one does microseconds of actual work. With lightweight middleware, the instrumentation overhead stands out. With I/O-heavy middleware, it probably gets buried under redis/HTTP latency and likely won't show up in the graph.
> 
> Please correct me if I am missing anything here.

Ok. I had 13 middleware functions but they are doing the redis calls and http calls. So let me just create 13 bare bones middleware that just call next and see if that makes a difference

bizob2828:
@malwarebo I ran a test with simple sync middleware and I see the issue but both my PR(#3811) and your PR still exhibit the issue. I'm going to dig into how the graph you linked is being queried.  You can see the response times are consistent which indicates that there isn't a hit to perf. I suspect this is just a benign bookkeeping error.

The first chunk is 13.5.0 of agent, the 1st half of 2nd chunk is 13.14.0, the 2nd of 2nd chunk is my PR, the 3rd chunk is your PR, and then 4th chunk is back to 13.5.0.

screenshot 2026-03-09 at 2 51 09 PM

malwarebo:
> @malwarebo I ran a test with simple sync middleware and I see the issue but both my PR(#3811) and your PR still exhibit the issue. I'm going to dig into how the graph you linked is being queried. You can see the response times are consistent which indicates that there isn't a hit to perf. I suspect this is just a benign bookkeeping error.
> 
> The first chunk is 13.5.0 of agent, the 1st half of 2nd chunk is 13.14.0, the 2nd of 2nd chunk is my PR, the 3rd chunk is your PR, and then 4th chunk is back to 13.5.0.
> screenshot 2026-03-09 at 2 51 09 PM

Thanks for digging into this.

So the graph I shared is basically being queried as:
"Show the average web response time (in milliseconds -> apm.service.overview.web * 1000) broken down by segment, for the last 30 minutes up to now, plotted over time."

bizob2828:
correct but the solid blue line doesn't change between versions. just the segment durations. the root cause of this change is in the legacy instrumentation we were ending a segment before the next function was actually called. Whereas now we're touching the segment.  That is why the durations are different.  In fact, the durations are actually now more accurate and we fixed a long standing issue with this instrumentation migration.  I ran a test but timing the actual function

```js
function testMiddleware(req, res, next) {
  const now = Date.now()
  try {
    fs.readFile('./foo')
  } catch {}
  next()
  console.log('mw time', Date.now() - now)
}
```

In < 13.6.0 - the segment time for `testMiddleware` was microseconds. Where as now it is the correct time.

Before 13.6.0:

```sh
mw time 5
recording segment duration and exclusive 0.18375 0.18375
```

After 13.6.0:

```sh
mw time 6
recording segment duration and exclusive 6.042125 6.042125
```

So in summary, you did identify some possible bottlenecks to fix, it is not the source of your issue. In fact post 13.5.0 we are more accurately recording the times it takes to execute middleware in express.  You can see the agent is not leading to any more overhead than before based on the response time line being consistent through your upgrades.  I'm going to close this PR

PR #3810, opened 2026-03-06
test: Add `redis-client` and `node-redis-client` segment duration timing tests

amychisholm03:
## Description

Title

## How to Test

```
npm run versioned redis
```


test/versioned/redis/redis-v4-legacy-mode.test.js
```diff
@@ -324,4 +324,32 @@ test('Redis instrumentation', async function (t) {
...
+
+      const trace = transaction.trace
+      const children = trace.getChildren(trace.root.id)
+      assert.equal(children.length, 1, 'there should be only one child of the root')
+
+      const [waitSegment] = children
+      assert.ok(waitSegment, 'trace segment for wait should exist')
+      assert.equal(waitSegment.name, 'Datastore/operation/Redis/wait', 'should register the wait')
+      assert.ok(waitSegment.timer.hrDuration, 'trace segment should have ended')
+      assert.ok(waitSegment.getDurationInMillis() > WAIT)
```
jsumners-nr: I think each of these lines should use `>=` instead of a strict `>`. It's unlikely they'd ever be equal, but theoretically possible.

PR #3815, opened 2026-03-10
chore: Added delay + duration to ProfilingAggregator

bizob2828:
## Description
When testing delay + duration, I realized it wasn't working. This is because we were not passing in the configurations from `config.profiling` to the aggregator.  This PR passes that in and adds tests to assert this is working.

### DIFF
--- a/test/unit/agent/agent.test.js
+++ b/test/unit/agent/agent.test.js
@@ -945,7 +945,7 @@ test('when connected', async (t) => {
 })
 
 test('should set up delay + duration for profilingAggregator', async (t) => {
-  const plan = tspl(t, { plan: 13 })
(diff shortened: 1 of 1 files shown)


lib/aggregators/base-aggregator.js
```diff
@@ -211,6 +213,16 @@ class Aggregator extends EventEmitter {
 
       logger.trace(`${this.method} aggregator stopped.`)
     }
+
+    if (this.delayTimeout) {
+      clearTimeout(this.delayTimeout)
+      this.delayTimeout = null
```
jsumners-nr, line 219: Interesting. `clearTimeout` only removes the associated `_onTimeout` callback from the timeout object. I always thought it completely wiped out the object.

PR #3875, opened 2026-03-30
test: Added `assertSegmentDuration` custom assertion and added it to `memcached` tests

amychisholm03:
## Description

As we migrate our instrumentation to use the tracing channel subscribers, we need a way to be able to check if our segment durations are expected (e.g. not ending immediately). However, previous attempts have been extremely flaky because our CI has much more constrained environments vs our PCs running locally. 

This PR adds a new custom assertion, `assertSegmentDuration` and adds it to just `memcached` for now.

The `actualTime` and the segment's duration time both have `segment.timer.hrstart` as thier starting point because there could be some time elapsed between test code and the subscriber calling `timer.begin()`. This means that this assertion primarily tests that the segment **ended** correctly, not that it started correctly. 

## How to Test

```
npm run versioned memcached
```

## Related Issues

Closes #3862

### DIFF
--- a/test/lib/custom-assertions/assert-segment-duration.js
+++ b/test/lib/custom-assertions/assert-segment-duration.js
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2025 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+/**
+ * Asserts that a segment ended at approximately the right time.
+ *
+ * Both measurements use `segment.timer.hrstart` as their reference point, so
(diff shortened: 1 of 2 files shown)


bizob2828 (commented):
i see the CI failed and the actual duration vs record durations seem off like our timing is wrong

amychisholm03:
> i see the CI failed and the actual duration vs record durations seem off like our timing is wrong

Yeah, there are some large time differences. However, the failures are different packages every time, so I don't think it has to deal with how we are timing the `actualTime`, but just general flakiness for some reason.

bizob2828 (approved):

PR #3890, opened 2026-04-07
refactor: Updated Lambda instrumentation to a subscriber middleware

bizob2828:
## Description

Migrates lambda instrumentation to middleware of subscriber. I also added a `getSegment` method on the smithy client subscriber which pulls the segment from the symbol on request or context manager. This is used in at least 2 places now and wanted to centralize it.

## How to Test

```sh
npm run versioned:internal aws-sdk-v3
```

## Related Issues
Closes #3840

PR #3891, opened 2026-04-08
refactor: Updated `aws-sdk/dynamodb` instrumentation to a subscriber middleware

### LINKED ISSUE #3290: Update dynamodb to rely on tracing channel instrumentation
## Description
Update instrumentation to create a subscriber for `@aws-sdk/client-dynamodb` and `@aws-sdk/lib-dynamodb` and a instrumentator for orchestrion.

## Acceptance Criteria
dynamodb instrumentation still functions as expected after traditional monkey patching and shim helpers are used.

svetlanabrennan:
## Description

Migrated aws-sdk/dynamodb instrumentation to use subscribers. 

## How to Test

Run `npm run versioned`

## Related Issues

Closes #3290


lib/subscribers/aws-sdk/middleware/dynamodb/index.js
```diff
@@ -0,0 +1,162 @@
...
+ * Middleware hook that records the middleware chain
+ * when command is in a list of monitored commands.
+ *
+ * @param {Subscriber} subscriber subscriber instance of subscriber
+ * @param {object} config AWS SDK client configuration
+ * @param {Function} next middleware function
+ * @param {object} context Context for the running command
+ * @returns {Function}
+ */
+function dynamoMiddleware(subscriber, config, next, context) {
```
jsumners-nr, line 47: See this comment: https://github.com/newrelic/node-newrelic/pull/3890#discussion_r3051128609

lib/instrumentation/aws-sdk/v3/smithy-client.js
amychisholm03, line 1: I do not see the old instrumentation file (`lib/instrumentation/aws-sdk-v3/dynamodb.js`) deleted.

svetlanabrennan: I deleted the v3 one. Are we leaving everthing in v2 for now? I still see sns and sqs shim based instrumentation files in v2 directory.

jsumners-nr: Correct. This work is focused on v3.

lib/subscribers/aws-sdk/middleware/dynamodb/index.js
```diff
@@ -0,0 +1,155 @@
...
+  },
+  {
+    fn: resourceIdMiddleware,
+    config: {
+      name: 'NewRelicCloudResource',
+      step: 'deserialize',
+      priority: 'low',
+      override: true
+    }
+  }
```
jsumners-nr, line 38: Seeing both of these being exported in the same array makes it clear that we should just omit the `init` method when all it does is `return true`. I recommend removing it here, and also in the SNS instrumentation.

lib/subscribers/aws-sdk/middleware/dynamodb/index.js
```diff
@@ -0,0 +1,155 @@
...
+      subscriber.logger.debug(err, 'Failed to get the endpoint.')
+    }
+
+    const [command] = args
+    const params = setDynamoParameters(endpoint, command.input)
+
+    const name = `${DB.OPERATION}/DynamoDB/${commandName}`
+    const newCtx = subscriber.createSegment({
+      name,
+      recorder: recordOperationMetrics.bind(DYNAMO_METRICS),
```
jsumners-nr, line 71: Kind of strange that `recordOperationMetrics` requires a `bound this`. It looks like both of the database focused recorders are written this way. I think it would be a good idea to create a follow-up issue for us to implement new versions of those recorders such that they accept a parameter in place of a `bound this`. Even if we end up currying the value here we can extract some more performance by not binding:

```sh
bound this                                    x 90,367,306 ops/sec (11 runs sampled) min..max=(10.82ns...11.37ns)
direct param                                  x 91,852,686 ops/sec (11 runs sampled) min..max=(10.69ns...11.31ns)
curried                                       x 91,954,607 ops/sec (12 runs sampled) min..max=(10.64ns...11.33ns)
```

<details>
<summary>Benchmark</summary>

```js
'use strict'

const suite = new Suite()

suite.add('bound this', () => {
  const fn = boundThis.bind({ foo: 'foo' })
  fn()
})

suite.add('direct param', () => {
  foo('foo')
})

suite.add('curried', () => {
  const input = 'foo'
  const fn = () => foo(input)
  fn()
})

suite.run()

function boundThis () {
  assert.equal(this.foo, 'foo')
}

function foo (a) {
  assert.equal(a, 'foo')
}
```

</details>

jsumners-nr (changes requested):

lib/instrumentation/aws-sdk/v3/smithy-client.js
```diff
@@ -7,13 +7,10 @@
 
 const { middlewareConfig } = require('./common')
 const { sqsMiddlewareConfig } = require('./sqs')
-const { dynamoMiddlewareConfig } = require('./dynamodb')
```
bizob2828, line 9: since this is the last PR now. you can remove all the aws sdk v3 instrumentation

lib/subscribers/aws-sdk/middleware/dynamodb/index.js
bizob2828: I really despise the fact that this metric recorder as well as the sql db record expects a `this` reference to something else.

lib/subscribers/aws-sdk/send.js
```diff
@@ -12,15 +12,15 @@ const MIDDLEWARE = Symbol('nrMiddleware')
...
 const middlewareByClient = {
   BedrockRuntime: [...nrMiddleware, require('./middleware/bedrock/index.js')],
+  DynamoDB: [...nrMiddleware, ...require('./middleware/dynamodb/index.js')],
+  DynamoDBDocument: [...nrMiddleware, ...require('./middleware/dynamodb/index.js')],
   Lambda: [...nrMiddleware, require('./middleware/lambda/index.js')],
   SNS: [...nrMiddleware, require('./middleware/sns/index.js')]
 }
 
 // clients handled by legacy instrumentation from lib/instrumentation/aws-sdk/v3
 const legacyClients = new Set([
```
bizob2828: this code can now be removed as well since the other migrations have been merged

lib/subscribers/aws-sdk/middleware/dynamodb/index.js
```diff
@@ -0,0 +1,155 @@
...
+
+    return next(...args)
+  }
+}
+
+const INSTANCE_KEYS = ['host', 'port_path_or_id']
+const DB_NAME_KEY = 'database_name'
+const HOST_KEY = 'host'
+
+function addDatastoreAttributes(segment, params, agentConfig) {
```
bizob2828, line 121: i think we need to do something similar to the `lib/message-broker-description` eventually as this code is the same as `lib/subscribers/db.js`

bizob2828 (changes requested):
you need to rebase and remove all the legacy aws sdk v3 instrumentation now that this is the last middleware to be migrated

lib/subscribers/aws-sdk/middleware/dynamodb/index.js
jsumners-nr: I'm guilty of crafting such functions. It just does not seem necessary here and an easy micro-optimization win.

lib/instrumentation/aws-sdk/v3/smithy-client.js
svetlanabrennan: Ok. Are you talking about removing `v3/common.js` and `v3/smithy-client.js` files?

bizob2828: it might be more than that. we can chat after standup

lib/subscribers/aws-sdk/middleware/dynamodb/index.js
svetlanabrennan: Added comment with a future TODO for this

bizob2828 (approved):

bizob2828 (changes requested):
turns out there was a test that was missed when bedrock was migrated, now that the old instrumentation was removed it's failing. this will fix it:

```sh
diff --git a/test/versioned/aws-sdk-v3/bedrock-converse-api.test.js b/test/versioned/aws-sdk-v3/bedrock-converse-api.test.js
index 1e9939086..1f845521c 100644
--- a/test/versioned/aws-sdk-v3/bedrock-converse-api.test.js
+++ b/test/versioned/aws-sdk-v3/bedrock-converse-api.test.js
@@ -53,12 +53,8 @@ test('Converse API', { skip: semver.lt(bedrockVersion, '3.587.0') }, async (t) =
     agent.llm.tokenCountCallback = null
   })
 
-  await t.test('should log tracking metrics', function(t) {
-    const { version } = require('@smithy/smithy-client/package.json')
-    assertPackageMetrics({ agent, pkg: '@smithy/smithy-client', version })
-  })
-
   await t.test('should properly create completion segment', async (t) => {
+    const { version } = require('@smithy/smithy-client/package.json')
     const prompt = 'text converse ultimate question'
     const input = {
       modelId,
@@ -80,6 +76,7 @@ test('Converse API', { skip: semver.lt(bedrockVersion, '3.587.0') }, async (t) =
         ['Llm/completion/Bedrock/ConverseCommand', [expectedExternalPath(modelId)]],
         { exact: false }
       )
+      assertPackageMetrics({ agent, pkg: '@smithy/smithy-client', version, subscriberType: true })
       tx.end()
     })
```

jsumners-nr (dismissed):

bizob2828 (approved):

PR #3893, opened 2026-04-09
chore: Added distributed tracing propagation to SNS

### LINKED ISSUE #3892: Add DT propagation to SNS
In #3876 we skipped propagating distributed tracing headers because we weren't sending them previously and we weren't sure the feature works correctly with SNS. In #3880 we added DT propagation to SQS and concluded that SNS also needs it. We need to copy the SQS implementation over and round out the SNS tests.

jsumners-nr:
This PR resolves #3892.


jsumners-nr (commented):
This PR does not attempt to handle receiving the headers back from SNS. According to https://docs.aws.amazon.com/sns/latest/dg/welcome.html#how-it-works it is most likely that an SQS poll will receive the message. We support receiving the headers with SQS.

lib/subscribers/aws-sdk/middleware/sns/index.js
```diff
@@ -63,10 +64,12 @@ function snsMiddleware(subscriber, config, next, context) {
...
-    // In a typical producer type instrumentation we'd invoke
-    // `subscriber.insertDTHeaders` here. But since SNS is not really a message
-    // storage destination, headers cannot be propagated. Therefore, we do not
-    // need to attach them here.
+    const headers = Object.create(null)
+    subscriber.insertDTHeaders({ headers, ctx: newCtx, useMqNames: true })
+    if (Object.hasOwn(command.input, 'MessageAttributes') === false) {
+      command.input.MessageAttributes = {}
+    }
+    attachHeaders({ message: command.input, headers })
```
bizob2828: think the util should insert the headers as well as attach? these 6 lines are identical in SQS

bizob2828 (dismissed):
LGTM, should a tiny comment on repeated code that could just go into the shared function

lib/subscribers/aws-sdk/middleware/sns/index.js
jsumners-nr: Yeah, that's a fine idea. Done.

bizob2828 (approved):

PR #3894, opened 2026-04-09
refactor: Removed reliance on `AsyncLocalStorage.enterWith` for undici, domain and `process._fatalException` instrumentation

bizob2828:
## Description

We received a report about a possible memory leak with `enterWith`(#3870). We are waiting for a reproduction, however after looking at the code this method is not needed.

With undici we were using it to set the active segment, but we are storing that context on the request object. In fact this was an issue with concurrent undici requests where the parent of the subsequent undici calls was the previous one, which is technically not correct.  

The other two places this was used was to set the context to null: `domain` and `process._fatalException`. Our testing around this isn't great and I'm not convinced we need it.

I did keep the `setContext` on context manager, which calls `enterWith` and the tracer method `setSegment` which calls `setContext` after entering a new context. This should only be used for testing and I added comments to state that.

## Related Issues
Closes #3870


jsumners-nr:
Looks like this may have resulted in Apollo's segment tree changing.

lib/shim/shim.js
```diff
@@ -1188,8 +1188,7 @@ function getActiveSegment(obj) {
  * @returns {TraceSegment} - The segment set as active on the context.
  */
 function setActiveSegment(segment) {
-  const transaction = this.tracer.getTransaction()
-  this.tracer.setSegment({ segment, transaction })
+  this._logger.warn('setActiveSegment is deprecated and will be removed in the next major release.')
```
jsumners-nr, line 1191: We'll need an issue for this one. It's not currently tagged for the release coming up soon.

bizob2828:
> Looks like this may have resulted in Apollo's segment tree changing.

Lovely. I hope I can also remove its reliance on setting an active segment

test/versioned-external/external-repos.js
```diff
@@ -15,7 +15,7 @@ const repos = [
   {
     name: 'apollo-server',
     repository: 'https://github.com/newrelic/newrelic-node-apollo-server-plugin.git',
-    branch: 'remove-set-timeout',
+    branch: 'enable-timeout-instrumentation',
```
bizob2828: I will restore this to main, once https://github.com/newrelic/newrelic-node-apollo-server-plugin/pull/365 lands

jsumners-nr (dismissed):

jsumners-nr (approved):

PR #3905, opened 2026-04-15
feat: Added instrumentation for Google ADK `BaseAgent.runAsync`

### LINKED ISSUE #3817: Instrument `@google/adk` `BaseAgent.runAsync` for `Llm` spans and `LlmAgent` events
### Description
For FY27Q1 (Apr - Jun 2026), one of our AIM commits is Google Agent Development Kit (ADK) instrumentation ([`@google/adk`](https://github.com/google/adk-js)). ADK is similar to LangChain/LangGraph in that it allows you to easily create and manage complex AI agent ecosystems. We will start with agent invocations. As such, we will create `Llm/agent` spans with coordinating `LlmAgent` events. 

The relevant methods appear to be:
* For Agent events - [`BaseAgent.prototype.runAsync`](https://github.com/google/adk-js/blob/6d13c61c0f0a06d1c026700e9021c7fd2630e5f3/core/src/agents/base_agent.ts#L169): This is the public method on every agent — `LlmAgent`, `SequentialAgent`, `LoopAgent`, `ParallelAgent`, and custom agents all inherit it from `BaseAgent`. It's the method that gets called both by the `Runner` and by parent agents when invoking sub-agents, so wrapping it gives us coverage at every level of a multi-agent hierarchy.

### Acceptance Criteria
* For agent invocations, an `LlmAgent` event and a `Llm/agent/ADK/runAsync/<agent_name>` segment are created
* Relevant versioned tests are created testing the above basic funcitonality
* Relevant unit tests are created for the `LlmAgent` event

### More Info
* [Agentic AI spec for spans](https://source.datanerd.us/agents/agent-specs/blob/main/artificial_intelligence/Agentic.md#named-spans)
* [Spec for LlmAgent events](https://source.datanerd.us/agents/agent-specs/blob/main/artificial_intelligence/Agentic.md#llmagent-events)

amychisholm03:
## Description

This PR adds subscriber-based instrumentation for Google ADK's `BaseAgent.runAsync` method, which is the main method and base class for all agent invocations in the SDK. Tool calls are out of scope for this PR. A `Llm/agent/GoogleADK/runAsync/<agent_name>` segment is created for each `runAsync` call along with a `LlmAgent` custom event with `vendor: 'google_adk'`.

The `LlmAgent` event class that was only being used by LangGraph has been repurposed to be used for both Google ADK and LangGraph because it is very simple; the only difference from the base `LlmEvent` class is the AI agent name property.

Also, when I was testing it with the [example app](https://github.com/newrelic/newrelic-node-examples/pull/369), I noticed that Gemini (primary LLM to be used with ADK) `LlmChatCompletionMessage` events had missing `content`, so there is a one-line fix for that - it just lives in a different location on the object.

## How to Test

```
npm run versioned google-adk
npm run unit
```

## Related Issues

Closes #3817


lib/subscribers/google-adk/config.js
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// A single subscriber handles both bundled and unbundled versions of @google/adk.
+// Only 0.4.0 ships unbundled separate files; all other versions use a bundled index.js.
+const agentRunAsync = {
```
jsumners-nr: This is a completely new instrumentation, correct? Do we have a specific request to support older versions? Typically, when we add a new instrumentation, we only support the latest current version.

amychisholm03: Correct, new instrumentation, and I just arbitrarily set the minimum version to what I could instrument. I can change it to latest which is I think 0.6.1, but the instrumentation works for all old and current versions of `@google/adk`. The package itself is very new, with 0.1.0 being released only 6 months ago.

jsumners-nr: We should target the current release. This will eliminate the caveats about file location and simplify the instrumentation.

lib/llm-events/google-genai/chat-completion-message.js
```diff
@@ -21,16 +21,19 @@ module.exports = class GoogleGenAiLlmChatCompletionMessage extends LlmChatComple
     content, role,
     completionId,
     isResponse }) {
-    super({ agent,
+    super({
+      agent,
```
jsumners-nr, line 25: I appreciate this subtle change. I prefer the opening and closing braces on their own line when we have to write long parameter blocks like this. But it's definitely nothing to nit pick over; hence why I haven't commented on it previously.

lib/metrics/names.js
```diff
@@ -217,6 +217,11 @@ AI.LANGGRAPH = {
   AGENT: `${AI.AGENT}/LangGraph`
 }
 
+AI.GOOGLE_ADK = {
```
jsumners-nr: I see that the existing `AI.*` objects are not in alphabetical order, but this is a good opportunity to correct that. Please move this one up and re-order the others into alphabetical order.

lib/subscribers/google-adk/agent-run-async.js
```diff
@@ -0,0 +1,114 @@
...
+    this.name = `${GOOGLE_ADK.AGENT}/runAsync/${this.aiAgentName}`
+    return super.handler(data, ctx)
+  }
+
+  // BaseAgent.runAsync is an async generator. The tracePromise wrapper returns
+  // the generator object synchronously (it has no .then), so `asyncEnd` is
+  // never published. We use `end` instead, which fires right after the
+  // generator is created but before it is consumed by the caller.
+  end(data) {
+    const { agent, logger, aiAgentName } = this
```
jsumners-nr: Have you tried doing this in the `handler`? It looks to me like the handler is receiving the same data.

lib/subscribers/google-adk/agent-run-async.js
```diff
@@ -0,0 +1,114 @@
...
+   * child instrumentation (e.g. Gemini, FunctionTool) runs within the
+   * agent segment context. Records the LlmAgent event when the generator
+   * completes or throws.
+   *
+   * @param {object} params function parameters
+   * @param {object} params.ctx The NR async context to bind generator calls to
+   * @param {AsyncGenerator} params.generator The async generator from runAsync
+   * @param {object} params.segment The current NR segment
+   * @param {object} params.transaction The active NR transaction
+   * @param {string} params.aiAgentName The AI agent name
```
jsumners-nr: Aren't the segment and transaction objects members of the ctx object?

test/versioned/google-adk/agent-run-async.test.js
```diff
@@ -0,0 +1,275 @@
...
+    assertPackageMetrics({
+      agent,
+      pkg: '@google/adk',
+      version: adkVersion,
+      subscriberType: true
+    }, { assert: t.assert })
+    end()
+  })
+})
+
```
jsumners-nr, line 97: It would be better to include a `t.plan(5)` (I think that's the right number for that assert function) as the first line in this test. It gives the reader a clear idea of how many assertions the test is expected to pass, and reinforces that they are covered (since we are dealing with callbacks in this test).

test/versioned/google-adk/agent-run-async.test.js
```diff
@@ -0,0 +1,275 @@
...
+  await helper.runInTransaction(agent, async (tx) => {
+    const results = await consumeGenerator(testAgent.runAsync({}))
+    assert.equal(results.length, 1)
+    assertSegments(tx.trace, tx.trace.root, ['Llm/agent/ADK/runAsync/span_agent'], {
+      exact: false
+    })
+
+    tx.end()
+  })
+})
```
jsumners-nr, line 117: This is another case where I would establish an expected plan. Basically, any time we have to wrap stuff in a callback handler we should specify how many assertions are expected. If we don't, it's very easy to think a test is working correctly when in reality most of the required assertions have not been hit.

jsumners-nr (commented):
Do we need to update `ai-support.json`?

lib/subscribers/google-adk/agent-run-async.js
amychisholm03: `data.result` aka the generator is not available in the `handler` since the function hasn't finished yet when `handler` fires.

jsumners-nr (dismissed):
Looks good to me.

lib/metrics/names.js
```diff
@@ -187,20 +187,19 @@ AI.ANTHROPIC = {
...
-AI.OPENAI = {
-  TRACKING_PREFIX: `${AI.TRACKING_PREFIX}/OpenAI`,
-  EMBEDDING: `${AI.EMBEDDING}/OpenAI/create`,
-  COMPLETION: `${AI.COMPLETION}/OpenAI/create`
-}
-
-AI.BEDROCK = {
-  TRACKING_PREFIX: `${AI.TRACKING_PREFIX}/Bedrock`
+AI.GOOGLE_ADK = {
+  TRACKING_PREFIX: `${AI.TRACKING_PREFIX}/ADK`,
```
bizob2828: this seems very generic especially since they even reference it as Google ADK

lib/metrics/names.js
```diff
@@ -187,20 +187,19 @@ AI.ANTHROPIC = {
...
-AI.OPENAI = {
-  TRACKING_PREFIX: `${AI.TRACKING_PREFIX}/OpenAI`,
-  EMBEDDING: `${AI.EMBEDDING}/OpenAI/create`,
-  COMPLETION: `${AI.COMPLETION}/OpenAI/create`
-}
-
-AI.BEDROCK = {
-  TRACKING_PREFIX: `${AI.TRACKING_PREFIX}/Bedrock`
+AI.GOOGLE_ADK = {
+  TRACKING_PREFIX: `${AI.TRACKING_PREFIX}/GOOGLE_ADK`,
```
bizob2828: why is this in all caps when all the other metrics are properly cased?

test/versioned/google-adk/agent-run-async.test.js
```diff
@@ -0,0 +1,290 @@
...
+  const assert = t.assert
+
+  const testAgent = createTestAgent(BaseAgent, {
+    name: 'span_agent',
+    events: [{ id: 'evt-1', author: 'span_agent', content: { parts: [{ text: 'hello' }] } }]
+  })
+
+  await helper.runInTransaction(agent, async (tx) => {
+    const results = await consumeGenerator(testAgent.runAsync({}))
+    assert.equal(results.length, 1)
```
bizob2828, line 110: this isn't using `t.assert` so it's not part of the plan

bizob2828: a lot of tests in this file are doing this

lib/subscribers/google-adk/agent-run-async.js
```diff
@@ -0,0 +1,108 @@
...
+   *
+   * @param {object} params function parameters
+   * @param {object} params.ctx The NR async context to bind generator calls to
+   * @param {AsyncGenerator} params.generator The async generator from runAsync
+   */
+  #instrumentGenerator({ ctx, generator }) {
+    const self = this
+    const origNext = generator.next
+
+    generator.next = async function wrappedNext(...args) {
```
bizob2828, line 69: you're wrapping a generator function with technically incorrect syntax. Generators are `async function * <nameOfFunction>() {}`

bizob2828: I'm also not sure this is semantically correct aside from the lack of `*`. I think you should be yielding back results as they are present. I don't think this will work for a common consumer of ADK

bizob2828 (commented):
I haven't studied a lot of this PR just wanted to submit some things I noticed

lib/subscribers/google-adk/agent-run-async.js
bizob2828: oh wait n/m this is the `.next` property of the generator itself

test/versioned/google-adk/agent-run-async.test.js
amychisholm03: I assigned `const assert = t.assert` on line 101: https://github.com/amychisholm03/node-newrelic/blob/1d0f8e0f10c6b04f0edfe71a9edee92ef2bc8e66/test/versioned/google-adk/agent-run-async.test.js#L101

If it's confusing, I can just change it to `t.assert`.

lib/llm-events/google-genai/chat-completion-message.js
bizob2828: this should be part of the linting ruleset if we want to enforce, otherwise it's just personal preference

lib/subscribers/google-adk/agent-run-async.js
```diff
@@ -0,0 +1,108 @@
...
+    this.events = ['end']
+  }
+
+  handler(data, ctx) {
+    this.aiAgentName = data?.self?.name ?? 'agent'
+    this.name = `${GOOGLE_ADK.AGENT}/runAsync/${this.aiAgentName}`
+
+    const newCtx = super.handler(data, ctx)
+    newCtx.segment.addSpanAttribute('subcomponent', `{"type": "APM-AI_AGENT", "name": "${this.aiAgentName}"}`)
+    return newCtx
```
bizob2828:
```suggestion
    return super.handler(data, ctx)
```

lib/subscribers/google-adk/agent-run-async.js
```diff
@@ -0,0 +1,108 @@
...
+  // generator is created but before it is consumed by the caller.
+  end(data) {
+    const { agent, logger } = this
+    if (!this.enabled) {
+      logger.debug('Google ADK instrumentation is disabled, not instrumenting runAsync.')
+      return
+    }
+
+    const ctx = agent.tracer.getContext()
+    const { segment, transaction } = ctx
```
bizob2828, line 43:
```suggestion
    const { segment, transaction } = ctx
    segment.addSpanAttribute('subcomponent', `{"type": "APM-AI_AGENT", "name": "${this.aiAgentName}"}`)
```

bizob2828: suggested moving this to the end even so the handler just returns `super.handle(data, ctx)` for consistency with other handlers

lib/subscribers/google-adk/agent-run-async.js
```diff
@@ -0,0 +1,108 @@
...
+    const agentEvent = new LlmAgent({
+      agent,
+      segment,
+      transaction,
+      aiAgentName,
+      vendor: 'google_adk',
+      error: error || undefined
+    })
+
+    this.recordEvent({ type: 'LlmAgent', msg: agentEvent })
```
bizob2828, line 106: i noticed in langgraph we also add the error to the transaction, this isn't happening here. is that intentional?

bizob2828 (changes requested):
overall looks good. one big thing i see missing is adding the errors to the transaction

lib/subscribers/google-adk/agent-run-async.js
amychisholm03: Nope, I forgot it; I'll add it now

lib/subscribers/google-adk/agent-run-async.js
```diff
@@ -0,0 +1,120 @@
...
+
+    if (error) {
```
bizob2828, line 108: now that we have 2 libraries that create `LlmAgent` events, is it worth doing a follow up to create a base subscriber for agent.js like we do for chat/embedding?

lib/subscribers/google-adk/config.js
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+const agentRunAsync = {
+  path: './google-adk/agent-run-async.js',
+  instrumentations: [
+    // >= 0.5.x < 1.1.x bundles everything into dist/*/index.js with minified class names.
+    // BaseAgent.runAsync is the first `runAsync` method in the bundle (index 0).
+    {
+      channelName: 'nr_runAsync',
+      module: { name: '@google/adk', versionRange: '>=0.5.0 <1.1.0', filePath: 'dist/cjs/index.js' },
```
bizob2828: is there a reason why we would support a `0.x` line of a library? it really causes issues for us.  Why don't we just support `1.1.0`+? I know it was only released 2 days ago, but so what

test/versioned/google-adk/agent-run-async.test.js
bizob2828: oic, it's a little confusing but i guess it's fine

bizob2828 (commented):
The big comment is min supported version. I wouldn't attempt to support `0.x` line as it could cause us grief in the future. Given how new this library is I don't think it matters we'd only support `1.1.0`+. The weekly downloads are so small for this package and a majority of folks are using `0.6.1` probably because they aren't auto upgrading to `1.x`

lib/subscribers/google-adk/agent-run-async.js
amychisholm03: Not a bad idea. We could also do one for `LlmTool`.

lib/subscribers/google-adk/config.js
amychisholm03: We support `0.x` for LangChain just because our instrumentation still works that far back, so I figured that would apply here. I can change it to `>=1.1.0` though.

bizob2828 (approved):

PR #3907, opened 2026-04-16
fix: Fixed `child_process` instrumentation to avoid leaking event handlers subscribed to once

### LINKED ISSUE #3906: `child_process` instrumentation mem leak when using `.once()`

## Description

Event listeners added using `.once()` are not properly removed when calling `.off()`.

```typescript
import { spawn } from 'node:child_process';

const proc = spawn('some_bin', ...);

proc.once('error', onError);
proc.once('close', onClose);

proc.off('error', onError); // not removed
proc.off('close', onClose); // not removed
```

It works as expected with `.on()`

## Expected Behavior

Calling `.off('event', fn)` should work when `fn` is added using `.once('event', fn)`.

bizob2828:
## Description

This PR fixes an issue where you try to call `.removeListener` or `.off` on an event handler from a child process method.
It could not locate the original as our previous instrumentation was binding the handler to the context.  It turns out we do
not need this code at all because event handlers are synchronous operations.  We have tests that assert the context in an event
handler and by removing the code, it still works.  I first refactored it to just wrap the `emit` method but realized this wasn't needed.
This was probably an artifact of our legacy context manager that was instrumenting the event loop.

I also added more tests that showed the issue the customer reported but they are now fixed.

## How to Test
```sh
node test/integration/core/child-process.test.js
```

## Related Issues
Closes #3906

### DIFF
--- a/lib/instrumentation/core/child_process.js
+++ b/lib/instrumentation/core/child_process.js
@@ -21,36 +21,6 @@ function initialize(agent, childProcess, moduleName, shim) {
     return new RecorderSpec({ name: 'child_process.' + name, callback: shim.LAST })
   })
 
-  if (childProcess.ChildProcess) {
-    wrapChildProcessClass(shim, childProcess.ChildProcess)
-  } else {
-    shim.logger.warn('childProcess.ChildProcess should be available in v2.2.0 or higher')
-  }
-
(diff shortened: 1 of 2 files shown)


jsumners-nr (approved):
Reviewed over Zoom 👍

PR #3912, opened 2026-04-21
chore: Refactored Connect instrumentation to subscriber type

### LINKED ISSUE #3910: Migrate `connect` to subscriber instrumentation
As titled.

jsumners-nr:
This PR resolves #3910.


test/unit/instrumentation/connect.test.js
jsumners-nr, line 1: This suite seems to be exercising shim paths through the `connect` instrumentation. This is no longer relevant.

test/versioned/connect/route.test.js
```diff
@@ -112,21 +114,10 @@ test('should default to `/` when no route is specified', async (t) => {
...
-  // you had to manually call app.handle
-  if (semver.satisfies(pkgVersion, '<2')) {
-    requestListener = function (req, res) {
-      app.handle(req, res)
-    }
-  }
-
-  const server = http.createServer(requestListener).listen(0, function () {
+function createServerAndMakeRequest({ url, expectedData, plan, app }) {
+  const server = http.createServer(app).listen(0, function () {
```
jsumners-nr, line 120: The removed bits in here were for supporting `connect@1` and `connect@2`. We haven't supported those since at least `newrelic@13`.

test/versioned/connect/error-intercept.test.js
```diff
@@ -41,7 +41,7 @@ test('should wrap handlers with proxies', (t) => {
   assert.equal(wrapNop.route, '', 'nop handler defaults to all routes')
   assert.ok(wrapNop.handle, 'have nop handle passed above')
   assert.equal(wrapNop.handle.name, 'nop', 'nop name is unchanged')
-  assert.equal(wrapNop.handle[symbols.original], nop, 'nop is wrapped')
+  assert.equal(wrapNop.handle === nop, false, 'nop is wrapped')
```
jsumners-nr, line 43: We no longer decorate with a symbol. But the wrapper object is not the same object as the original one. This indicates the wrapping has been successful.

lib/subscribers/connect/config.js
```diff
@@ -0,0 +1,28 @@
...
+'use strict'
+
+const modName = 'connect'
+
+module.exports = {
+  [modName]: [
+    {
+      path: './connect/create-server.js',
+      instrumentations: [{
+        channelName: 'nr_create_server',
```
amychisholm03, line 15:
```suggestion
        channelName: 'nr_createServer',
```

Nit, but `channelName` should be in the format `nr_{functionName}` to match the recommendation in [`lib/subscribers/README.md`](https://github.com/newrelic/node-newrelic/tree/main/lib/subscribers#config-template)

lib/subscribers/connect/create-server.js
```diff
@@ -0,0 +1,47 @@
...
+'use strict'
+
+const MiddlewareSubscriber = require('../middleware.js')
+
+module.exports = class CreateServerSubscriber extends MiddlewareSubscriber {
+  constructor({ agent, logger }) {
+    super({
+      agent,
+      logger,
+      channelName: 'nr_create_server',
```
amychisholm03, line 15:
```suggestion
      channelName: 'nr_createServer',
```

test/versioned/connect/route.test.js
```diff
@@ -31,12 +30,17 @@ test.afterEach((ctx) => {
 
 test('should log tracking metrics', function(t) {
```
bizob2828: These tests are very sparse.  what is missing:
 1. asserting the middleware segments are getting created
 2. asserting the middleware metrics

The plan changes from 10 to 27

```js
    assertSegments(tx.trace, web, ['Nodejs/Middleware/Connect/middleware//foo'], { exact: true }, { assert: plan }) 
    const expectedMetrics = [
      [{ name: 'WebTransaction' }],
      [{ name: 'WebTransactionTotalTime' }],
      [{ name: 'HttpDispatcher' }],
      [{ name: 'WebTransaction/Connect/GET//foo' }],
      [{ name: 'WebTransactionTotalTime/Connect/GET//foo' }],
      [{ name: 'DurationByCaller/Unknown/Unknown/Unknown/Unknown/all' }],
      [{ name: 'DurationByCaller/Unknown/Unknown/Unknown/Unknown/allWeb' }],
      [{ name: 'Apdex/Connect/GET//foo' }],
      [{ name: 'Apdex' }],
      [{ name: 'Nodejs/Middleware/Connect/middleware//foo'}],
      [{ name: 'Nodejs/Middleware/Connect/middleware//foo', scope: 'WebTransaction/Connect/GET//foo'}],
    ]
    assertMetrics(tx.metrics, expectedMetrics, false, false, { assert: plan })
```

test/versioned/connect/error-intercept.test.js
```diff
@@ -11,7 +11,6 @@ const http = require('node:http')
 
 const { removeModules } = require('../../lib/cache-buster')
 const helper = require('../../lib/agent_helper')
-const symbols = require('../../../lib/symbols')
```
bizob2828, line 14: we're missing tests where we have an error handler middleware and it either swallows error or returns:

```js
test('should trace errors that occur while executing middleware when error middleware passes error along', (t, end) => {
  const { agent, app } = t.nr
  let server

  agent.once('transactionFinished', () => {
    const errors = agent.errors.traceAggregator.errors
    assert.equal(errors.length, 1, 'the error got traced')
    server.close()
    end()
  })

  helper.runInTransaction(agent, () => {
    function wiggleware(req, res, next) {
      const harbl = null
      harbl.bargl() // Induce error.

      return next() // Will never get here.
    }

    function errorMiddleware(err, req, res, next) {
      assert.ok(err)
      next(err)
    }

    app.use(wiggleware)
    app.use(errorMiddleware)

    server = http
      .createServer(function (req, res) {
        app.handle(req, res)
      })
      .listen(0, function () {
        const req = http.request(
          {
            port: server.address().port,
            host: 'localhost',
            path: '/asdf',
            method: 'GET'
          },
          (res) => {
            res.on('data', () => {})
          }
        )
        req.end()
      })
  })
})

test('should not trace errors that occur while executing middleware', (t, end) => {
  const { agent, app } = t.nr
  let server

  agent.once('transactionFinished', () => {
    const errors = agent.errors.traceAggregator.errors
    assert.equal(errors.length, 0, 'the error got swallowed')
    server.close()
    end()
  })

  helper.runInTransaction(agent, () => {
    function wiggleware(req, res, next) {
      const harbl = null
      harbl.bargl() // Induce error.

      return next() // Will never get here.
    }

    function errorMiddleware(_, req, res, next) {
      next()
    }

    app.use(wiggleware)
    app.use(errorMiddleware)

    server = http
      .createServer(function (req, res) {
        app.handle(req, res)
      })
      .listen(0, function () {
        const req = http.request(
          {
            port: server.address().port,
            host: 'localhost',
            path: '/asdf',
            method: 'GET'
          },
          (res) => {
            res.on('data', () => {})
          }
        )
        req.end()
      })
  })
})
```

lib/subscribers/connect/config.js
bizob2828: this is wrapped the createServer, however you could just wrap the `proto.use` directly:

```
        functionQuery: {
          expressionName: 'use',
          kind: 'Sync'
        }
```

Then the subscriber would be something like this:

```
  handler(data, ctx) {
    const { arguments: args } = data
    const [route, mw ] = args
    if (typeof route === 'string') {
      const wrappedMw = this.wrapper.wrap({ handler: mw, route })
      data.arguments[1] = wrappedMw
    } else {
      const wrappedMw = this.wrapper.wrap({ handler: route, route: '/' })
      data.arguments[0] = wrappedMw
    }
    return ctx
  }
```

bizob2828 (changes requested):
A few test addition suggestions to ensure proper test coverage. Also suggested wrapping the `use` method directly instead of the app instance. lmk if you have any questions

bizob2828 (dismissed):
LGTM. The subscriber file is called `create-server` which may be confusing later but i don't really mind

jsumners-nr:
> LGTM. The subscriber file is called `create-server` which may be confusing later but i don't really mind

That was a missed update in the last commit. 🤦‍♂️

bizob2828 (approved):

PR #3918, opened 2026-04-24
feat!: Migrated `@apollo/sever` instrumentation from a plugin(`@newrelic/apollo-server-plugin`) to traditional instrumentation

### LINKED ISSUE #3911: Migrate apollo server plugin to subscriber instrumentation
### Description
The apollo server plugin has a few issues. It cannot properly bind context when creating operation and resolver segments.  It also maintains its own, non-agent config.  You also have to install a separate package and register the plugin with your apollo server

### Acceptance Criteria
As an apollo server user, I want to be able to automatically instrument my apollo server with the Node.js agent so that I can easily get telemetry about the server actions.

### Additional Context
This must go along with 14.0.0 as our apollo server plugin will conflict. You technically could still use the plugin but we'd like to discourage folks from using it.

bizob2828:
## Description

This PR migrates the apollo server plugin logic to traditional subscriber based instrumentation.  Since the apollo plugin was the last external package we ran in CI, I was also able to remove that job and simplified the npm scripts. `npm run versioned`, `npm run versioned:major` or `npm run versioned <pkg>` `npm run versioned:major <pkg`.

Aside from moving the code to a subscriber based model, it also removed its reliance on temporarily stomping on the async context by using `enterWith`.

Since the plugin was external it had its own configuration. I would've been able to work around the plugin co-existing with this instrumentation but cannot handle mapipng the old plugin config, thus a breaking change. But it is timely because we plan on shipping 14.0.0 in mid may.  This will target `next` for now so I can get some eyes on it.

## Additional Link
Closes #3911

## Migration Plan
Starting with 14.0.0 of the agent, apollo server customers will no longer need to install and use `@newrelic/apollo-server-plugin`.  The instrumentation will be automatically applied to every apollo server instance.  Howerver, if customers configured the plugin, this will detail the migration plan.

### Plugin Configuration
```js
const createPlugin = require('@newrelic/apollo-server-plugin')
const plugin = createPlugin({
  captureScalars: true,
  captureIntrospectionQueries: true,
  captureServiceDefinitionQueries: true,
  captureHealthCheckQueries: true,
  captureFieldMetrics: true,
  customResolverAttributes: () => { return { exampleAttribute: 'exampleValue' }},
  customOperationAttributes: () => { return { exampleAttribute: 'exampleValue' }}
})
```
### Agent Configuration
* `captureScalars` is now `config.apollo_server.scalars` or `NEW_RELIC_APOLLO_SERVER_SCALARS`
* `captureIntrospectionQueries` is now `config.apollo_server.introspection_queries` or `NEW_RELIC_APOLLO_SERVER_INTROSPECTION_QUERIES`
* `captureServiceDefinitionQueries` is now `config.apollo_server.service_definition_queries` or `NEW_RELIC_APOLLO_SERVER_SERVICE_DEFINITION_QUERIES`
* `captureHealthCheckQueries` is now `config.apollo_server.health_check_queries` or `NEW_RELIC_APOLLO_SERVER_HEALTH_CHECK_QUERIES`
* `captureFieldMetrics`  is now config.apollo_server.field_metrics` or `NEW_RELIC_APOLLO_SERVER_FIELD_METRICS`

### API calls
There were two previous configurations that registered functions to be called that added the key/values as custom attributes on either the operation or resolver segments.  The agent follows a pattern where customers can register callbacks via api methods, and these two use cases fit that.

#### customOperationAttributes

This is now `newrelic.setApolloOperationAttributesCallback`. It takes a callback and when present the callback gets passed the same argument as before `requestContext`

**Previous**:
```js
createPlugin({
  customOperationAttributes: (requestContext) => { 
    return {
      clientName: requestContext.request.http.headers.get('graphql-client-name')
    }
   }
})
```
**Now**:
```js
function operationAttributes(requestContext) {
 return {
   clientName: requestContext.request.http.headers.get('graphql-client-name')
 }
}
newrelic.setApolloOperationAttributesCallback(operationAttributes)
```

#### customResolverAttributes
This is now `newrelic.setApolloResolverAttributesCallback`. It takes a callback and when present the callback gets passed the same argument as before `{ source, args, contextValue, info }`

**Previous**:
```js
createPlugin({
  customResolverAttributes: ({ source, args, context, info }) => { 
    return {
      allArgs: Object.keys(args).join(','),
      returnType: info.returnType.name,
      sourceBranch: source?.branch
      stage: context.event.requestContext.stage
    }
  }
})
```

**Now**:
```js
function customResolverAttributes({ source, args, contextValue, info }}) {
  return {
    allArgs: Object.keys(args).join(','),
    returnType: info.returnType.name,
    sourceBranch: source?.branch
    stage: contextValue.event.requestContext.stage
  }
}
newrelic.setApolloResolverAttributesCallback(customResolverAttributes)
```


documentation/apollo-server/metrics.md
jsumners-nr, line 1: We should probably link, duplicate, or move this documentation to https://docs.newrelic.com/docs/apm/agents/nodejs-agent/extend-your-instrumentation/apollo-server-plugin-nodejs/#github

documentation/apollo-server/segments-and-spans.md
jsumners-nr, line 1: Ditto https://docs.newrelic.com/docs/apm/agents/nodejs-agent/extend-your-instrumentation/apollo-server-plugin-nodejs/#github

documentation/apollo-server/transactions.md
jsumners-nr, line 1: Ditto https://docs.newrelic.com/docs/apm/agents/nodejs-agent/extend-your-instrumentation/apollo-server-plugin-nodejs/#github

lib/agent.js
```diff
@@ -322,6 +322,7 @@ function Agent(config) {
 
   this.initCounters()
   this.llm = {}
+  this.apollo = {}
```
jsumners-nr: I wonder if we should file an issue to refactor these into something like:

```js
this.extensions = {
	llm: {},
	apollo: {}
}
```

Or at least inline some documentation about what these fields are here and why they need their own namespace on the agent object.

lib/config/default.js
```diff
@@ -1782,6 +1782,44 @@ defaultConfig.definition = () => {
           formatter: int
         }
       }
+    },
+    apollo_server: {
```
jsumners-nr, line 1790: Separating root blocks with a little bit of whitespace helps with readability:

```suggestion
    },

    apollo_server: {
```

lib/config/default.js
```diff
@@ -1782,6 +1782,44 @@ defaultConfig.definition = () => {
...
+        default: false
+      },
+      /**
+       * Enable capture of metrics for every field and resolver argument seen for an Apollo query. This is intended to be used to check for any unused fields in your graphql schema.
+       */
+      field_metrics: {
+        formatter: boolean,
+        default: false
+      }
     }
```
jsumners-nr, line 1827: Case and point, I'm not sure this suggestion is matching up with the opening brace given they are all crammed together:
```suggestion
      }

    }
```

lib/metrics/recorders/apollo-resolver.js
jsumners-nr, line 1: What is your opinion on this living in `lib/subscribers/apollo-server/metrics/recoder.js`? I find there's some cognitive overhead in remembering that metric recorders live in a separate location. It makes sense for the recorders shared among multiple instrumentations, but in a case like this it would make more sense to me if all the code were organized together.

lib/subscribers/apollo-server/request.js
```diff
@@ -0,0 +1,355 @@
...
+  handler(data, ctx) {
+    return this.createSegment({
+      name: DEFAULT_OPERATION_NAME,
+      recorder: genericRecorder,
+      ctx
+    })
+  }
+
+  asyncEnd(data) {
+    const [, , , requestContext] = data.arguments
```
jsumners-nr, line 48: Just an FYI, as reading this line made me wonder about the tradeoff:

```
destructure                                   x 89,562,833 ops/sec (9 runs sampled) min..max=(10.82ns...11.27ns)
at method                                     x 93,350,203 ops/sec (13 runs sampled) min..max=(10.48ns...11.00ns)
subscript                                     x 97,678,056 ops/sec (12 runs sampled) min..max=(10.04ns...10.54ns)
```

Subsequent runs had `destructure` and `at method` be either roughly equivalent or `destructrue` being faster. Seems dependent on how many runs the benchmark framework decided to use.

🤷‍♂️ just thought it is interesting.

<details>
<summary>Benchmark script</summary>

```js
'use strict'

const suite = new Suite()

const testData = ['1', '2', '3', 'get-me']

suite.add('destructure', () => {
  const [, , , val] = testData
  assert.equal(val, 'get-me')
})

suite.add('at method', () => {
  const val = testData.at(-1)
  assert.equal(val, 'get-me')
})

suite.add('subscript', () => {
  const val = testData[3]
  assert.equal(val, 'get-me')
})

suite.run()
```

</details>

lib/subscribers/apollo-server/request.js
```diff
@@ -0,0 +1,355 @@
...
+  /**
+   * Attempts to extract the document from the request context and
+   * add attributes for the query, operation type, operation name and
+   * update the transaction name based on operation name as well
+   *
+   * @param {object} params to function
+   * @param {object} params.requestContext apollo request context
+   * @param {object} params.ctx active agent context
+   */
+  updateOperationSegmentName({ requestContext, ctx }) {
```
jsumners-nr, line 87: This method is tied to the comment on `.handler`, correct?

lib/subscribers/apollo-server/request.js
```diff
@@ -0,0 +1,355 @@
...
+  asyncEnd(data) {
+    const [, , , requestContext] = data.arguments
+    const ctx = this.agent.tracer.getContext()
+    const operationSegment = ctx.segment
+    const transaction = ctx.transaction
+    this.updateOperationSegmentName({ requestContext, ctx })
+
+    if (this.shouldIgnoreTransaction(requestContext.operation)) {
+      transaction.setForceIgnore(true)
+    }
```
jsumners-nr, line 56: Should this come before line 54 in order to avoid unnecessary work? My impression by the name is that the transaction will be completely dropped if the configuration says so, and thus we don't need to update the segment at all.

lib/subscribers/apollo-server/request.js
```diff
@@ -0,0 +1,355 @@
...
+
+  handleErrors({ requestContext, transaction }) {
+    if (Array.isArray(requestContext.errors)) {
+      for (let error of requestContext.errors) {
+        error = error.originalError || error
+        if (!error[apolloErrorHandled]) {
+          this.agent.errors.add(transaction, error, error.extensions)
+        }
+      }
+    }
```
jsumners-nr, line 75:
```suggestion
    if (Array.isArray(requestContext.errors) === false) {
      return
    }
    for (let error of requestContext.errors) {
      error = error.originalError || error
      if (!error[apolloErrorHandled]) {
        this.agent.errors.add(transaction, error, error.extensions)
      }
    }
```

lib/subscribers/apollo-server/request.js
```diff
@@ -0,0 +1,355 @@
...
+   * @param {object} definition of graphql field
+   * @returns {object} { deepestPath, argLocations }
+   */
+  getDeepestPathAndQueryArguments(definition) {
+    const self = this
+    let deepestPath = []
+    let foundDeepestPath = false
+    const argLocations = []
+
+    definition?.selectionSet?.selections?.forEach((selection) => {
```
jsumners-nr, line 249: In https://github.com/newrelic/node-newrelic/pull/3463 we opted to move away from the `forEach` iterator where possible. Are there reasons for this one and the ones on lines 270, 302, & 344?

test/lib/apollo/test-client.js
```diff
@@ -0,0 +1,101 @@
+/*
+ * Copyright 2020 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const http = require('http')
```
jsumners-nr: Only because the next file uses the prefix for its imports:
```suggestion
const http = require('node:http')
```

test/lib/apollo/test-tools.js
```diff
@@ -0,0 +1,128 @@
...
+  if (server) {
+    await server.stop()
+  }
+
+  helper.unloadAgent(agent)
+  unloadModules(testDir)
+}
+
+async function setupCoreTest({ t, testDir, agentConfig = {} } = {}) {
+  agentConfig.instrumentation = { timers: { enabled: true } }
```
jsumners-nr, line 36: Will we need to call this out in the documentation? I don't recall reading about it in the migration guide.

test/versioned/apollo-federation/federated-gateway-server-setup.js
```diff
@@ -0,0 +1,156 @@
...
+    allowBatchedHttpRequests: true,
+    gateway,
+    subscriptions: false,
+    plugins
+  })
+
+  const { url } = await startStandaloneServer(server, { listen: { port: 0 } })
+
+  // eslint-disable-next-line no-console
+  console.log(`${name} ready at ${url}`)
```
jsumners-nr: Why do we need a `console.log`?

test/versioned/apollo-federation/federated-gateway-server-setup.js
```diff
@@ -0,0 +1,156 @@
...
+
+  const server = new ApolloServer({
+    schema: buildSubgraphSchema([{ typeDefs, resolvers }]),
+    plugins
+  })
+
+  const { url } = await startStandaloneServer(server, { listen: { port: 0 } })
+
+  // eslint-disable-next-line no-console
+  console.log(`${name} service ready at ${url}`)
```
jsumners-nr: Ditto.

test/versioned/apollo-server/attributes.test.js
```diff
@@ -0,0 +1,712 @@
...
+      assertCustomAttributes(operationSegment, { clientName: 'ApolloTestClient' })
+
+      const expectedOperationAttributes = {
+        'graphql.operation.type': 'query'
+      }
+
+      const operationAttributes = operationSegment.attributes.get(SEGMENT_DESTINATION)
+      match(operationAttributes, expectedOperationAttributes)
+
+      const hasAttribute = Object.hasOwnProperty.bind(operationAttributes)
```
jsumners-nr, line 61: Crafty!

lib/agent.js
bizob2828: yea this is prob a good idea. I'll make the change for apollo now but follow up to move llm. These are more customCallbacks. so maybe `this.customCallback = { llm: {}, apollo: {}}`?

test/versioned/apollo-federation/federated-gateway-server-setup.js
bizob2828: ¯\_(ツ)_/¯. This is ported code from extension

lib/agent.js
jsumners-nr: Add and `s` to it and sounds good to me. Inline document it too for clarity.

documentation/apollo-server/metrics.md
bizob2828: i'll log a ticket to update this page to keep it simple and link out to the docs from here

test/versioned/apollo-federation/federated-gateway-server-setup.js
jsumners-nr: I'd remove them. They can interfere with reading standard test framework output and we really don't need a log to tell us they are ready.

documentation/apollo-server/metrics.md
bizob2828: https://github.com/newrelic/node-newrelic/issues/3919

test/lib/apollo/test-tools.js
bizob2828: no this is only enabled because we have setTimeouts in resolvers to ensure proper parent/child relationships.

test/versioned/apollo-server/attributes.test.js
bizob2828: maybe i did this long ago but this is just copied from the plugin

test/lib/apollo/test-tools.js
jsumners-nr: Cool.

jsumners-nr (approved):

PR #3929, opened 2026-04-28
feat!: Updated `config.distributed_tracing.exclude_newrelic_header` to be set to `true` by default

### LINKED ISSUE #3924: Change `exclude_newrelic_header` to default to `true`
See spec on this for more details. https://source.datanerd.us/agents/agent-specs/pull/790

bizob2828:
## Description
This defaults to `exclude_newrelic_header` to `true`. Turns out we had more tests than expected that assumed this was also going to be inserted into outgoing headers.  I cleaned up the tests as best as possible.  I had to set the value to `false` for some cross agent tests as it either exclusive relied on a `newrelic` header or tested behavior where both `newrelic` and `traceparent`/`tracestate` were being set.

## Related Issues

Closes #3924

### DIFF
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -1140,7 +1140,7 @@ defaultConfig.definition = () => {
        */
       exclude_newrelic_header: {
         formatter: boolean,
-        default: false
+        default: true
       },
 
       sampler: {
--- a/test/integration/distributed-tracing/background.test.js
+++ b/test/integration/distributed-tracing/background.test.js
@@ -14,7 +14,7 @@ const API = require('../../../api')
 let compareSampled = null
 
 test('background transactions should not blow up with DT', async (t) => {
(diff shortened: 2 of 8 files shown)


jsumners-nr (approved):
It's unclear to me how this stacks with the CAT removal PR. I would think they conflict. Isn't the `newrelic` header part of CAT?

bizob2828:
> It's unclear to me how this stacks with the CAT removal PR. I would think they conflict. Isn't the `newrelic` header part of CAT?

No CAT pre-dates the initial attempt at distributed tracing via the `newrelic` header.  Cat uses a combination of headers which you can see in [here](https://github.com/newrelic/node-newrelic/pull/3925/changes#diff-7cfd9be2b58c11288245371ca336cfc889d20d20b52bf60303154bbc9da3581b)

PR #3935, opened 2026-04-30
feat: Instrument Google ADK `FunctionTool.runAsync`

### LINKED ISSUE #3902: Instrument `@google/adk` tool calls to make `Llm` spans and `LlmTool` events
### Description
For FY27Q1 (Apr - Jun 2026), one of our AIM commits is Google Agent Development Kit (ADK) instrumentation ([`@google/adk`](https://github.com/google/adk-js)). ADK is similar to LangChain/LangGraph in that it allows you to easily create and manage complex AI agent ecosystems. We already have the instrumentation for agent invocations (which create `Llm/agent` spans and `LlmAgent` events) in #3817. Now, we will create `Llm/tool` and spans with coordinating `LlmTool` events. 

The relevant methods appear to be:
* For Tool events - [`BaseTool.prototype.runAsync`](https://github.com/google/adk-js/blob/6d13c61c0f0a06d1c026700e9021c7fd2630e5f3/core/src/tools/base_tool.ts#L109): `BaseTool` is the tool class that more specific tool classes (e.g. `FunctionTool`, `AgentTool`, etc.) extend from. However, this might be problematic because the more specific tool classes e.g. `AgentTool` _override_ `runAsync` ([example](https://github.com/google/adk-js/blob/6d13c61c0f0a06d1c026700e9021c7fd2630e5f3/core/src/tools/agent_tool.ts#L119)) instead of just implementing it.

### Acceptance Criteria
* For tool calls, a `LlmTool` event and `Llm/tool/ADK/runAsync/<tool_name>` segment are created

- [ ] How will we subscribe to a function (`BaseTool.prototype.runAsync`) that is being overriden? Maybe we should just start with instrumenting `FunctionTool.runAsync` because that is the most commonly used way of defining tools by the user?

### More Info
* [Agentic AI spec for spans](https://source.datanerd.us/agents/agent-specs/blob/main/artificial_intelligence/Agentic.md#named-spans)
* [Spec for LlmTool events](https://source.datanerd.us/agents/agent-specs/blob/main/artificial_intelligence/Agentic.md#llmtool-events)

amychisholm03:
## Description

This PR builds off of #3905, which introduced instrumentation for the Google ADK (`@google/adk`) library. That PR wrapped AI agent invocations, and this PR instruments the tool calls that those AI agents make.

Following the agentic AI spec, a span called `Llm/tool/GoogleADK/runAsync/<tool_name>` and a coordinating `LlmTool` event will be created. A special `subcomponent` attribute is added to the span, so that the UI can create the agentic AI graph view (see below). If an error occurs, the error will be added to the transaction.

image

## How to Test

```
npm run versioned google-adk
```

[Run the example app here](https://github.com/newrelic/newrelic-node-examples/tree/main/ai-monitoring/google-adk#readme).

## Related Issues

Closes #3902


lib/subscribers/google-adk/tool-run-async.js
```diff
@@ -0,0 +1,83 @@
...
+    })
+    this.events = ['asyncEnd']
+  }
+
+  handler(data, ctx) {
+    const toolName = data?.self?.name ?? 'unknown'
+    this.toolName = toolName
+    this.toolDescription = data?.self?.description
+    this.toolInput = data?.arguments?.[0]?.args
+      ? JSON.stringify(data.arguments[0].args)
```
jsumners-nr: `data.self` and `data.arguments` should always be defined.
```suggestion
    const toolName = data.self.name ?? 'unknown'
    this.toolName = toolName
    this.toolDescription = data.self.description
    this.toolInput = data.arguments?.[0]?.args
      ? JSON.stringify(data.arguments[0].args)
```

lib/subscribers/google-adk/tool-run-async.js
```diff
@@ -0,0 +1,83 @@
...
+    if (!(segment || transaction) || (transaction?.isActive() !== true)) {
+      return
+    }
+
+    const { error: err, result } = data
+
+    segment.addSpanAttribute('subcomponent', `{"type": "APM-AI_TOOL", "name": "${toolName}"}`)
+    segment.end()
+
+    const output = result != null ? JSON.stringify(result) : undefined
```
jsumners-nr: Are we certain `result` will always stringify without error?

amychisholm03: No, a user can define their own `FunctionalTool.execute` function that could return whatever, so I'll add a guard.

test/versioned/google-adk/tool-run-async.test.js
```diff
@@ -0,0 +1,308 @@
...
+    })
+
+    assert.deepStrictEqual(result, { temperature: '72F', unit: 'fahrenheit' })
+
+    const events = agent.customEventAggregator.events.toArray()
+    const toolEvents = events.filter((e) => e[0].type === 'LlmTool')
+    assert.equal(toolEvents.length, 1, 'should have exactly 1 LlmTool event')
+
+    const [[{ type }, toolEvent]] = toolEvents
+    assert.equal(type, 'LlmTool')
```
bizob2828, line 62: I would assert the entire tool event like we do in other libraries. This isn't asserting a lot of the keys

test/versioned/google-adk/tool-run-async.test.js
```diff
@@ -0,0 +1,308 @@
...
+
+const { removeModules } = require('../../lib/cache-buster')
+const { assertSegments } = require('../../lib/custom-assertions')
+const helper = require('../../lib/agent_helper')
+const { DESTINATIONS } = require('../../../lib/config/attribute-filter')
+
+const config = {
+  ai_monitoring: {
+    enabled: true,
+    streaming: {
```
bizob2828: is there a reason why you have streaming enabled here? I don't see that affect any logic in the instrumentation

lib/subscribers/google-adk/config.js
```diff
@@ -27,8 +27,33 @@ const agentRunAsync = {
...
+      module: { name: '@google/adk', versionRange: '>=1.1.0', filePath: 'dist/cjs/tools/function_tool.js' },
+      functionQuery: {
+        className: 'FunctionTool',
+        methodName: 'runAsync',
+        kind: 'Async'
+      }
+    },
+    {
+      channelName: 'nr_toolRunAsync',
+      module: { name: '@google/adk', versionRange: '>=1.1.0', filePath: 'dist/esm/tools/function_tool.js' },
```
bizob2828, line 44: i don't think this is the correct file to instrument. you're only wrapping function tool calls, however there are a lot more.  This might be a case where you wrap base_tool with the `runAsync` method name which will create a placeholder method for all inherited children.

bizob2828 (changes requested):

lib/subscribers/google-adk/config.js
amychisholm03: `BaseTool.runAsync` is an `abstract` method (https://github.com/google/adk-js/blob/main/core/src/tools/base_tool.ts#L109), so when it complies to JS (`node_modules/@google/adk/dist/cjs/tools/base_tool.js`), the method is nowhere to be seen. 

Each subclass (`FunctionTool`, `AgentTool`, `GoogleSearchTool`, etc.) implements its own `runAsync`. `FunctionTool` is the right target because that is the only entry point for user-defined tool functions; the rest of them are internal.

bizob2828: correct but orchestrion was enhanced to support adding non-existent methods to base classes for this exact reason. are you sure FunctionTool is the only once for user-defined functions? all the tools get exported on `@google/adk`

amychisholm03: It looks like `FunctionTool` is the only user-based one in [their example](https://github.com/google/adk-samples/blob/main/typescript/agents/customer_service/customer_service/tools/function_tools.ts).

`GoogleSearchTool` for example explicitly says it is a "built-in tool on server side". It immediately resolves a `Promise`, so I'm not sure how useful that would be for instrumentation: https://github.com/google/adk-js/blob/53ee7eb0f301b8bf617afb4cf7b0e9b1d7f8c9a9/core/src/tools/google_search_tool.ts#L12-L28

[`AgentTool`](https://github.com/google/adk-js/blob/53ee7eb0f301b8bf617afb4cf7b0e9b1d7f8c9a9/core/src/tools/agent_tool.ts#L56-L64) is just a `BaseTool` wrapper around an agent and we already have the agent instrumented, so again, I don't know how useful that would be for instrumentation.

Which orchestrion PR added that feature?

bizob2828 (approved):
LGTM, we can adjust the tool instrumentation as customers use it and lack of expected functionality

PR #3945, opened 2026-05-05
feat: Updated how license keys are redacted in logs

### LINKED ISSUE #3944: Update how we redact license keys in log
This [merged spec 793](https://source.datanerd.us/agents/agent-specs/pull/793/files) updates how we log the license key in log outputs. 

Previously we redacted with `"REDACTED"` but now we need to:
- Keep the first 10 characters visible and the remaining characters are replaced by `*`. 
- keys must retain their original length after reduction
- if a key is 10 or fewer characters, fully redact it.

svetlanabrennan:
## Description

Updated how we redact license keys when we log the license key in the logs. 

Previously we redacted with `"REDACTED"` but now we need to:

- Keep the first 10 characters visible and the remaining characters are replaced by *.
- keys must retain their original length after reduction
- if a key is 10 or fewer characters, fully redact it.
- if license key is empty, redacted license key is also an empty string

## How to Test

Run `npm run unit`

## Related Issues

Closes #3944

### DIFF
--- a/test/unit/collector/remote-method.test.js
+++ b/test/unit/collector/remote-method.test.js
@@ -751,13 +751,10 @@ test('_safeRequest logging', async (t) => {
       body: 'test-body',
       path: '/nonexistent'
     }
-    ctx.nr.config = {
-      license_key: 'shhh-dont-tell',
-      max_payload_size_in_bytes: 10_000
-    }
+    ctx.nr.config = new Config({ license_key: 'shhh-dont-tell' })
   })
 
-  await t.test('should redact license key in logs', (t) => {
+  await t.test('should use redacted license key in logs', (t) => {
     const { RemoteMethod, options, config } = t.nr
     const method = new RemoteMethod({ name: 'test', agent: { config }, endpoint: {} })
     method._safeRequest(options)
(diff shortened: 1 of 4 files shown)


test/unit/collector/remote-method.test.js
```diff
@@ -751,13 +751,10 @@ test('_safeRequest logging', async (t) => {
       body: 'test-body',
       path: '/nonexistent'
     }
-    ctx.nr.config = {
-      license_key: 'shhh-dont-tell',
-      max_payload_size_in_bytes: 10_000
-    }
+    ctx.nr.config = new Config({ license_key: 'shhh-dont-tell' })
```
bizob2828: why did you remove the `max_payload_size_in_bytes`?

bizob2828 (dismissed):

test/unit/collector/remote-method.test.js
svetlanabrennan: Oh shoot. Didn't mean to.

svetlanabrennan: Weird that the tests are still passing even though that was removed.

bizob2828 (approved):

PR #3946, opened 2026-05-05
refactor: Updated how exclusive time and trace total time is calculated

### LINKED ISSUE #2695: Move root segment to context manager

bizob2828:
## Description

This PR refactors a few hot paths around traces:
 * Maintaining the tree of segments within a trace
 * calculating the total duration of a trace 
 * calculating the exclusive duration of a given segment
 * serializing the trace on end

I guess I could split this up into smaller chunks, so whoever reviews, let me know.  The reason it's all in one is because I was trying to address bottlenecks in the hot paths of managing a trace.  

The biggest issue was to calculate the total time of a trace, we passed in the root segment to all segments and as a segment was touched/ended/finalized we incrementally updated the duration of root.  This was an issue for two reasons: we had to have a reference to the root segment in every segment but also constantly updating root time during the life of a trace.  Instead, I updated it to calculate the root duration as we are processing all the segments to get their exclusive durations when a trace ends.  This is also getting called if you call `segment.getExclusiveDurationInMillis()` but since the values are cached, it is a no-op.  The same is still true for `trace.getTotalTimeDurationInMillis()`, it is cached.  These methods are called in metric recorders(which are only called when a trace ends, as well as in `lib/agent.js` to get the trace duration.

The 2nd issue I resolved was maintaining the segment tree.  Every time a child segment was created, it traversed the tree to find the parent. If a trace was large, this could add some overhead.  Instead, I updated the SegmentTree to keep track of all created nodes in a map so you can find the parent node by one lookup.

The 3rd issue was how the trace was getting serialized. I removed a `getCollectedChildren` as it was creating a new function on the filter which was inefficient as well as maintain the results and stack of nodes in separate arrays.

I also ran several example applications, and controlled the durations of segments and asserted that the segment duration, exclusive duration and total trace times were the same after the refactor.

It's worth noting, i didn't do extensive cpu/memory testing but compare the Node VMs page, it is using slightly less memory, and cpu after the refactor as well as GC metrics were more stable.

## How to Test

```sh
npm run unit
npm run integration
```

## Related Issues
Closes #2695


lib/transaction/trace/index.js
```diff
@@ -225,29 +225,88 @@ Trace.prototype.getTotalTimeDurationInMillis = function getTotalTimeDurationInMi
...
+    for (const child of segmentTree[i].children) {
+      segmentTree.push(child)
+    }
   }
 
-  let totalTimeInMillis = 0
+  return segmentTree
+}
+
+Trace.prototype._updateExclusiveDurationForSegment = function _updateExclusiveDurationForSegment() {
```
svetlanabrennan: Is this leftover from the refactor?

lib/transaction/trace/index.js
```diff
@@ -225,29 +225,88 @@ Trace.prototype.getTotalTimeDurationInMillis = function getTotalTimeDurationInMi
...
-    const node = children.pop()
-    const { segment, children: childChildren } = node
-    totalTimeInMillis += segment.getExclusiveDurationInMillis(this)
-    for (const child of childChildren) {
-      children.push(child)
+/**
+ * Computes total time and exclusive durations for all segments in a single
+ * post-order traversal of the segment tree.  Avoids repeated tree lookups by
+ * working directly with SegmentTree nodes and propagating merged time ranges
+ * up the tree, mirroring the ExclusiveCalculator algorithm.
```
jsumners-nr: This reference to "mirroring the ExclusiveCalculator algorithm" isn't going to make sense to future readers because that algorithm has been removed.
```suggestion
 * up the tree.
```

lib/transaction/trace/segment.js
```diff
@@ -244,6 +220,27 @@ function _setExclusiveDurationInMillis(duration) {
   this._exclusiveDuration = duration
 }
 
+/**
+ * Calculated exclusive duration for a segment which is segment duration - overlap with merged child ranges
```
jsumners-nr:
```suggestion
 * Calculate, and cache, exclusive duration for a segment which is segment duration - overlap with merged child ranges
```

lib/transaction/trace/index.js
```diff
@@ -11,6 +11,7 @@ const TraceSegment = require('./segment')
 const { Attributes, MAXIMUM_CUSTOM_ATTRIBUTES } = require('../../attributes')
 const logger = require('../../logger').child({ component: 'trace' })
 const { DESTINATIONS } = require('../../config/attribute-filter')
+const { mergeRanges } = require('./exclusive-time-calculator')
```
jsumners-nr: Given that this function is only used withing this module, should we move it wholesale into this module as a private function? There isn't a dedicated unit test for it, and it comes with a scary warning about not re-using the mutated input arrays. So it seems like something to keep internal in order to reduce the temptation for someone to use it elsewhere (who knows what future people will attempt).

jsumners-nr (dismissed):
This is a read-through review. Everything seems simple enough. I have a couple of suggestions and non-blocking questions.

lib/transaction/trace/index.js
bizob2828: yea i can move this, good call

lib/transaction/trace/index.js
bizob2828: ah i was working on extracting out common code into their own functions. I'm not sure what this was going to house, so i'll just remove

jsumners-nr (approved):

svetlanabrennan (approved):

PR #3948, opened 2026-05-06
chore: Removed `@newrelic/test-utilities` as it is no longer in use

bizob2828:
## Description
Since we folded the apollo server instrumentation into the agent, `@newrelic/test-utilities` is no longer needed in the agent.  In fact, we no longer need to maintain this package, so we can archive it after we merge the `next` branch with main.

## How to Test


bizob2828:
well i guess this is needed for the versioned test runner. We should probably just move the code for the runner into `bin`

PR #3951, opened 2026-05-07
ci: Updated versions of actions to ensure they are running on node 24

bizob2828:
## Description

This PR updates the verisions of all actions to remove the warnings of 

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36, jsumners-nr/gha-node-deps-divergent@643628fe0da51ec025e984c4644f17fd9f9e93f6. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

I also removed the `repolinter.yml` as it is no longer maintained by New Relic and we got approval to remove as it will soon stop working.


jsumners-nr (approved):
🎉 looking forward to removing repolinter everywhere.

PR #3955, opened 2026-05-07
feat!: Updated minimum supported version of `@nestjs/cli` and `@nestjs/core` to `10.0.0`

svetlanabrennan:
## Description

Updated min supported version to `10.0.0`

## How to Test

Run `npm run versioned`


svetlanabrennan:
Need to check where in the subscriber to update range for this library

bizob2828 (changes requested):
You forgot to update the `lib/subscribers/nestjs/config.js`. We test with `@nestjs/cli` but wrap `@nestjs/core`

bizob2828 (dismissed):

bizob2828 (approved):

PR #3958, opened 2026-05-07
chore: Refactored kafkajs instrumentation to subscriber type

### LINKED ISSUE #3914: Migrate `kafkajs` consumer instrumentation
The `kafkajs` consumer instrumentation should be migrated to the subscriber method.

jsumners-nr:
This PR resolves #3914.


svetlanabrennan (dismissed):

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -0,0 +1,349 @@
...
+   * @param {object} instance Consumer client instance.
+   * @param {string} methodName Name of the method being patched.
+   */
+  #wrapConsumerMethod(instance, methodName) {
+    const self = this
+    const orig = instance[methodName]
+
+    instance[methodName] = function nrWrappedConsumerMethod(...args) {
+      let ctx = self.agent.tracer.getContext()
+      if (ctx.transaction == null) {
```
bizob2828:
```suggestion
      if (ctx.transaction == null || ctx.transaction.isActive() === false) {
```

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -0,0 +1,349 @@
...
+        includeMessageBrokerPrefix: false
+      })
+
+      // We need to process the received message ourselves before bubbling
+      // it up to the original `eachMessage` handler. The outer
+      // `nrWrappedEachMessage` is used to create a new transaction for this
+      // handler to be executed under. So we need yet another wrapper for
+      // the actual logic.
+      const handler = function nrHandler() {
+        const ctx = tools.startConsumeSegment({
```
bizob2828, line 272: I know this is feature parity with the existing instrumentation. But the lack of `host`, `port` as parameters will not allow entity linking between the application and the kafka server.  The net of this is the relationship will be lost. We can file a follow up ticket to pass in `host` and `port`

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -0,0 +1,349 @@
...
+   * @param {object} instance Consumer client instance.
+   */
+  #wrapRunForBatches(instance) {
+    const self = this
+    // In this case, `orig` is actually our first wrapper of the method.
+    const orig = instance.run
+
+    instance.run = function nrWrappedRunForBatches(...args) {
+      let ctx = self.agent.tracer.getContext()
+      if (!ctx.transaction) {
```
bizob2828:
```suggestion
      if (!ctx.transaction || ctx.transaction.isActive() === false) {
```

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -0,0 +1,349 @@
...
+   */
+  #wrapProducerMethod(instance, methodName) {
+    const self = this
+    const orig = instance[methodName]
+    const batch = methodName === 'sendBatch'
+    return function nrWrappedMethod(...args) {
+      const [data] = args
+      const topic = batch === false ? data.topic : data.topicMessages[0].topic
+      let ctx = self.agent.tracer.getContext()
+      if (ctx.transaction == null) {
```
bizob2828:
```suggestion
      if (ctx.transaction == null || ctx.transaction.isActive() === false) {
```

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -0,0 +1,349 @@
...
+        return orig.apply(instance, args)
+      }
+
+      const mbd = new MBD({
+        libraryName: MBD.LIB_KAFKA,
+        destinationName: topic,
+        destinationType: MBD.DESTINATION_TYPE_TOPIC
+      })
+
+      ctx = self.createSegment({
```
bizob2828, line 188: I think the same is true for produces. Without the `host`, and `port` attributes it cannot properly link between the app and the kafka server. I would file a follow up to address this as the lack of these attributes is missing from the original instrumentation

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -0,0 +1,349 @@
...
+      recordLinkingMetrics({
+        agent: self.agent,
+        brokers: instance[kafkaCtx].brokers,
+        topic
+      })
+
+      if (batch === false) {
+        for (const msg of data.messages) {
+          const headers = msg.headers ?? {}
+          self.insertDTHeaders({ ctx, headers, useMqNames: true })
```
bizob2828, line 203: this is fine for now. but we will have to address after we merge `next` with `main`. The `useMqNames` is getting removed because we're ripping out CAT

lib/subscribers/amqplib/consume.js
```diff
@@ -38,6 +38,7 @@ class ConsumeSubscriber extends MessageConsumerSubscriber {
     this.headers = message?.properties?.headers
     this.destination = this.getExchange(message?.fields)
     this.queue = queue
```
bizob2828: this can be removed because it's no longer being used and has been moved to `this.consumerParameters`

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -0,0 +1,349 @@
...
+   *
+   * @returns {Function} The wrapped method.
+   */
+  #wrapProducerMethod(instance, methodName) {
+    const self = this
+    const orig = instance[methodName]
+    const batch = methodName === 'sendBatch'
+    return function nrWrappedMethod(...args) {
+      const [data] = args
+      const topic = batch === false ? data.topic : data.topicMessages[0].topic
```
bizob2828: slight nit, but if transaction isn't active you don't have to parse this. I would move to below the transaction active check

bizob2828 (changes requested):
Looking good, just a few things I noticed

lib/subscribers/kafkajs/client-constructor.js
jsumners-nr: I imagine we skipped it before because we don't actually have that information. We have cached the _set_ of servers the client will communicate with, but we don't have a single host and port combination that is 100% reliable for accuracy.

bizob2828: yea I think in those use cases we typically just take the first one

bizob2828 (approved):

PR #3959, opened 2026-05-07
refactor: Migrate `koa` and `@koa/router` instrumentation to subscriber-based

### LINKED ISSUE #3943: Migrate `koa` and `@koa/router` instrumentation to subscriber-based
### Description
Now that we removed `koa-route` and `koa-router` as part of our semver major `14.0.0` release, it would be a good idea to migrate the more modern `koa` and `@koa/router` packages to subscriber-based instrumentation.

### Acceptance Criteria

- [ ] Shim-based instrumentation in `lib/instrumentation/koa` is removed
- [ ] Subscriber(s) and config are created in `lib/subscribers/koa` 
- [ ] Existing tests pass and there are no lapses in instrumentation. Included segment duration timing tests to make sure there's no regressions there

amychisholm03:
## Description

This PR migrates over the `koa` and `@koa/router` instrumentation from shim-based to subscriber-based using the `MiddlewareWrapper` class that `express`, `connect`, and `fastify` use. 

However, Koa required its own `KoaMiddewareWrapper` subclass because Koa's middleware model differed by a lot and jsut overriding `MiddlewareWrapper.extractTxInfo` wasn't enough. These differences include path stack timing, a path stack entry for each Koa middleware, and error handling caught upstream.

There's quite a few subscribers, for `koa`:
- `use.js` — the main subscriber. Every middleware in a Koa app passes through `app.use()`, so wrapping it here is where segments are created and the path stack is managed for transaction naming.
- `create-context.js` — required because transaction naming can't be driven by `use.js` alone. Koa middleware sets `ctx.body` or `ctx.status` after returning, at which point paths may already be popping off the stack. So, we need to intercept it to get the correct path.

For `@koa/router`:
- `router-register.js` — the main subscriber. Every route (`router.get`, `router.post`, etc.) goes through `register()` internally, which returns a `Layer`. Wrapping the handlers in `layer.stack` here is what gives route middleware their correct segment name and route path.
- `router-param.js` — required because param handlers (`router.param('id', fn)`) have a different signature (value, ctx, next) than regular middleware, so they can't be wrapped by `register` alone. They need their own wrapper with `ctxArgIndex: 1`.
- `router-allowed-methods.js` — required to tag the function returned by `router.allowedMethods()` before it hits `app.use()`. Without the tag, `use.js `would wrap it like any other middleware and append a path, which would prevent `405`/`501` responses from being named by status code.

## How to Test

```
npm run versioned koa
```

## Related Issues

Closes #3943


bizob2828:
This needs rebased with next

PR #3961, opened 2026-05-07
ci: Applied least-privilege model to all workflows to appease CodeQL

bizob2828:
## Description

I used claude to do this.  I'm not sure if it's 100% correct but I challenged a few and it said what is there is correct. 

## Related Issues
Should resolve all the `Workflow does not contain permissions` CodeQL issues [here](https://github.com/newrelic/node-newrelic/security/code-scanning)

### DIFF
--- a/.github/workflows/azure-site-extension.yml
+++ b/.github/workflows/azure-site-extension.yml
@@ -11,6 +11,8 @@ env:
   SPEC_FILE_TEMPLATE: 'NewRelic.Azure.WebSites.Extension.NodeAgent.nuspec'
 
 permissions:
+  # contents:read is required for actions/checkout.
+  contents: read
   # id-token must be set to `write` for OIDC publishing.
   id-token: write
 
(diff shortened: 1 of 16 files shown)


jsumners-nr (dismissed):

.github/workflows/azure-site-extension.yml
```diff
@@ -10,12 +13,6 @@ on:
 env:
   SPEC_FILE_TEMPLATE: 'NewRelic.Azure.WebSites.Extension.NodeAgent.nuspec'
 
-permissions:
```
jsumners-nr, line 13: Why has this been removed?

bizob2828: i had @TimPansino take a look as he recently did this. basically every workflow has contents: read at root. then for given jobs he assigned the appropriate perms. that way if we ever add more jobs, they just inherit the `contents: read` and not any other write operations t

jsumners-nr (approved):

TimPansino (approved):

PR #3967, opened 2026-05-08
fix: Updated `tracer.bindFunction` to touch segment if a promise is fulfilled or rejected

### LINKED ISSUE #3966: Durations are off on aws-sdk v3
## Description

We got reports that the DynamoDB times were drastically reduced.  This is because we aren't touching the segment once the promise resolves.

## Expected Behavior
AWS sdk v3 async calls are properly timed

## Steps to Reproduce

Run aws versioned tests and verify the segment durations.

bizob2828:
## Description

When we started migrating to subscriber based instrumentation, we missed a key bit around duration timing of segment.
Looking at our existing migrations, the only affetcted library is aws-sdk-v3.  We were not touching the segment when a promise is resolved/rejected.
This caused very short durations for sns, sqs, and dynamodb segments.  I added tests that assert the segment durations. Without the code in tracer, they are failing.

**Note**: This logic is identical to `lib/shim/shim.js` `interceptPromise`.  I didn't remove that from shim so it's attaching duplicate `onFulfilled` and `onRejected` handlers.
There was some tests that were failing in shim when I removed it from there.

## How to Test

```sh
npm run versioned
npm run unit
npm run integration
```

## Related Issues
Closes #3966

### DIFF
--- a/lib/subscribers/aws-sdk/middleware/sqs/index.js
+++ b/lib/subscribers/aws-sdk/middleware/sqs/index.js
@@ -47,7 +47,7 @@ function sqsMiddleware(subscriber, config, next, context) {
   }
 
   subscriber.opaque = true
-  return function nrSqsMiddleware(...args) {
+  return async function nrSqsMiddleware(...args) {
     const ctx = agent.tracer.getContext()
     if (!ctx.transaction || ctx.transaction.isActive() === false) {
       return next(...args)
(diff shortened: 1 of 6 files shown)


svetlanabrennan (approved):

jsumners-nr (approved):
We discussed these changes over Zoom on Friday. LGTM.

PR #3975, opened 2026-05-12
chore: Fix prepare-release workflow attempt 2

jsumners-nr:


.github/workflows/prep-release.yml
```diff
@@ -1,7 +1,9 @@
 name: Prepare Release PR
 
+# prepare-release.js uses GITHUB_TOKEN to push a release-notes branch and open a PR.
```
bizob2828: this was done intentionally to scope the relevant permissions to the job it needs

.github/workflows/prepare-release.yml
```diff
@@ -2,7 +2,7 @@ name: Prepare Agent Release PR
 
 # Reusable callee prep-release.yml declares its own contents/pull-requests write perms.
```
bizob2828:
```suggestion

```

.github/workflows/prep-release.yml
jsumners-nr: It didn't work.

.github/workflows/prepare-release.yml
```diff
@@ -2,7 +2,7 @@ name: Prepare Agent Release PR
...
 permissions:
-  contents: read
+  contents: write
   pull-requests: write
```
bizob2828: Move these to the calling job below. The approach to the permissions was contents read on workflow and specific perms for the given jobs. This was to prevent opening permissions if new jobs were added to workflows later that didn't need more than `contents: read`

bizob2828 (changes requested):

.github/workflows/prepare-release.yml
jsumners-nr: > The workflow is not valid. .github/workflows/prepare-release.yml (Line: 24, Col: 3): Error calling workflow 'newrelic/node-newrelic/.github/workflows/prep-release.yml@main'. The nested job 'generate-release-notes' is requesting 'contents: write', but is only allowed 'contents: read'.

That says to me that the _workflow_ needs to set the permissions.

bizob2828: i don't read that as such.

.github/workflows/prep-release.yml
bizob2828: it's not working because the caller doesn't have the right permissions set on the job

.github/workflows/prepare-release.yml
```diff
@@ -1,10 +1,5 @@
 name: Prepare Agent Release PR
 
-# Reusable callee prep-release.yml declares its own contents/pull-requests write perms.
-permissions:
-  contents: read
```
bizob2828, line 5: i guess you don't need the `contents: read` as this is the only job currently in the workflow

.github/workflows/prep-release.yml
bizob2828: this still should be moved to the relevant job and just keep contents: read at root of workflow.  This is to prevent if we should ever add more jobs to the workflow that do not need write perms on contents/pull-requests

.github/workflows/prep-release.yml
```diff
@@ -1,7 +1,7 @@
 name: Prepare Release PR
 
 permissions:
-  contents: read
+  contents: write
```
bizob2828, line 4: i'm not trying to be difficult here so i'll just approve but this should be `contents: read`

bizob2828 (approved):

PR #3986, opened 2026-05-13
chore: Decrease smoke-like esm tests' sample size

amychisholm03:
## Description

We have a few smoke-like ESM tests that just test that ESM instrumentation was loaded correctly or minimal instrumentation cases. The default sample size of 10 is not necessary for these tests. This PR reduces the sample size for applicable ESM test suites down to 1.

## How to Test

```
npm run versioned
```

## Related Issues

None

### DIFF
--- a/test/versioned/amqplib-esm/package.json
+++ b/test/versioned/amqplib-esm/package.json
@@ -9,7 +9,10 @@
         "node": ">=22"
       },
       "dependencies": {
-        "amqplib": ">=0.5.0"
+        "amqplib": {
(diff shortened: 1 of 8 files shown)


bizob2828 (changes requested):
i would change these to `samples: 1`, these tests are just to ensure that our esm instrumentation is working. The non esm tests will randomly sample versions of the module

bizob2828 (commented):
also I would include the two packages you skipped. there are no special tests for these

bizob2828 (approved):

PR #3988, opened 2026-05-13
chore: Removed `husky` in lieu of git hoooks, added security hardened options to `npm install`

### LINKED ISSUE #3987: Harden our security posture while using npm
### Description
It seems like every week there is a new supply chain attack with npm modules. There are a few items we can address to protect ourselves, [see](https://lavamoat.github.io/guides/hardening-dev/)

bizob2828:
## Description

This PR adds some security hardened features into our `.npmrc` with the intention of reducing our risk to supply chain attacks.  It also adds capability in versioned tests to override the defaults which are done for `pg` and `pg-esm` because it has a native module that gets built in a post install step. I also passed in `--min-release-age=0` so that we can test on the latest versions of npm packages we instrument. To support some of these npm options(which were released in 11.10.0 of npm cli, I upgrade npm to latest when you're on node 22). I had to do a 2-staged upgrade because of a bug in `11.11.0`.

Lastly, I removed `husky` in lieu of a pre-commit git hook that must be set on every fork/clone of your repo by running:

```
git config --local core.hooksPath .githooks
```

Otherwise you wont get linting running before committing nor updating third party notices if changes to package.json occur.

## Related Issues
Closes #3987


.npmrc
```diff
@@ -1 +1,13 @@
+# Do not write a package-lock file
 package-lock=false
+# Ignore all lifecycle scripts by default. You can
+# still allow them on a per-package basis with
```
jsumners-nr: I'm not sure we can allow list specific packages with a switch. But if we can, this sentence is incomplete.

package.json
jsumners-nr, line 1: Let's add a new script:

```json
"addhook:precommit": "git config --local core.hooksPath .githooks"
```

jsumners-nr (changes requested):
Looks great. Sorry for the snipe 😬 

A couple of suggestions and we can get this added.

jsumners-nr (approved):

PR #3991, opened 2026-05-13
chore: Updated a few versioned test stanzas to reduce the number of combinations it runs

bizob2828:
## Description

This PR reduces the test matrix by applying sampling, or combining tests into one stanza.
I also fixed a bug with `samples` on `groupedDependencies`. It wasn't using the min because if samples wasn't defined it was not sampling at all.
By defaulting the samples to `Infinity` it resolves the issue.  

After this PR we went from:

prisma: 52 different versions to 19
openai: 19 to 10
fastify: 20 to 12
undici: 13 to 10
aws sdk v2: 11 to 2

The idea is for 3rd party libraries that have multiple test stanzas and those versions are EOLd, I'm sampling at a low rate because they are not changing.

### DIFF
--- a/test/versioned/fastify/package.json
+++ b/test/versioned/fastify/package.json
@@ -29,7 +29,10 @@
         "node": ">=22"
       },
       "dependencies": {
-        "fastify": "^4.0.0",
+        "fastify": {
+          "versions": "^4.0.0",
+          "samples": 1 
+        },
         "@fastify/middie": ">=7.0.0 && <9.0.0"
       },
       "files": [
--- a/test/versioned/openai/chat-completions-res-api.test.js
+++ b/test/versioned/openai/chat-completions-res-api.test.js
@@ -7,6 +7,7 @@
 
 const test = require('node:test')
 const assert = require('node:assert')
+const semver = require('semver')
 const { tspl } = require('@matteo.collina/tspl')
 
 const { removeModules } = require('../../lib/cache-buster')
@@ -25,7 +26,7 @@ const TRACKING_METRIC = `Supportability/Nodejs/ML/OpenAI/${pkgVersion}`
 const responses = require('./mock-responses-api-responses')
 const { assertChatCompletionMessages, assertChatCompletionSummary } = require('./common-responses-api')
 
-test('responses.create', async (t) => {
+test('responses.create', { skip: semver.lte(pkgVersion, '4.87.0') }, async (t) => {
   t.beforeEach(async (ctx) => {
     ctx.nr = {}
     const { host, port, server } = await createOpenAIMockServer(responses)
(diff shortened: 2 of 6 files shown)


amychisholm03:
What about `bluebird`? I know we were talking about reducing the number of versions it tests too.

test/versioned/fastify/package.json
```diff
@@ -29,7 +29,10 @@
         "node": ">=22"
       },
       "dependencies": {
-        "fastify": "^4.0.0",
+        "fastify": {
+          "versions": "^4.0.0",
```
bizob2828, line 33: this line is EOLd, no sense in testing more than one versions

test/versioned/openai/chat-completions-res-api.test.js
```diff
@@ -25,7 +26,7 @@ const TRACKING_METRIC = `Supportability/Nodejs/ML/OpenAI/${pkgVersion}`
 const responses = require('./mock-responses-api-responses')
 const { assertChatCompletionMessages, assertChatCompletionSummary } = require('./common-responses-api')
 
-test('responses.create', async (t) => {
+test('responses.create', { skip: semver.lte(pkgVersion, '4.87.0') }, async (t) => {
```
bizob2828, line 29: this will reduce openai stanzas from 2 to 1 and just skip this file if <4.87.0

test/versioned/undici/package.json
```diff
@@ -9,7 +9,10 @@
         "node": ">=22"
       },
       "dependencies": {
-        "undici": ">=5.0.0 <8.0.0"
```
bizob2828, line 12: I'm now wondering why there was ever 2 stanzas. I may just combine these

bizob2828:
> What about `bluebird`? I know we were talking about reducing the number of versions it tests too.

bluebird is only testing on 8 different versions. it's just that there are many tests that make it appear like a lot of different tests runs

bizob2828:
i have one more change to make

test/versioned/openai/chat-completions-res-api.test.js
jsumners-nr: I think that comment is worth adding to the file.

bizob2828: What do you mean?

jsumners-nr:
```suggestion
// This will reduce openai stanzas from 2 to 1 and just skip this file if <4.87.0.
test('responses.create', { skip: semver.lte(pkgVersion, '4.87.0') }, async (t) => {
```

bizob2828: Not sure this makes sense for future eyes

jsumners-nr (approved):

PR #3992, opened 2026-05-14
chore: Fixed version lookup issue in aws-sdk-v3 tests

jsumners-nr:
As titled.


test/lib/get-package-version.js
jsumners-nr, line 1: I know we already have a `getPackageVersion` in `lib/util/get-package-version.js`. But:

1. It requires a bit more setup to use correctly (have to resolve the path to the package directory before invoking it).
2. It's used in shimmer. So I didn't want to improve the ergonomics of it and potentially affect shimmer.

So I added this one specifically for our tests.

bizob2828 (approved):

PR #3995, opened 2026-05-14
refactor: Migrate `@hapi/hapi` and `@hapi/vision` to subscriber-based

### LINKED ISSUE #3971: Migrate `@hapi/hapi` and `@hapi/vision` instrumentation to subscriber-based
### Description
Refactor our existing `@hapi/hapi` and `@hapi/vision` instrumentation to use tracing channel subscribers.

### Acceptance Criteria

- [ ] No backwards regression in instrumentation - all of the tests in `test/versioned/hapi` still pass. Any `assertPackageMetrics` will have to be changed to `subscriberType: true`.
- [ ] Old instrumentation is deleted - all files in `lib/instrumentation/@hapi` are deleted and `@hapi/*` is removed from `lib/instrumentations.js`

amychisholm03:
## Description

This PR migrates over the shim-based `@hapi/hapi` and adjacent `@hapi/vision` instrumentation over to subscriber-based. 

The test file `test/versioned/hapi/hapi.test.js` was deleted because it just had two tests: one to check if the `Shim` wrap didn't mess up the export and the second to assert package metrics. The former was removed because it is no longer relevant and the later was moved to `test/versioned/hapi/segments.test.js`. 

Also `@hapi/vision` wasn't actually listed in `test/versioned/hapi/package.json` so it never showed up in our compatibility list. It is included in this PR.

## How to Test

```
npm run versioned hapi
```

## Related Issues

Closes #3971


lib/subscribers/middleware-wrapper.js
```diff
@@ -116,20 +118,24 @@ class MiddlewareWrapper {
       self.wrapDoneHandler({ segment, ctx, args, handler, txInfo, route, nextIdx })
 
       try {
-        const result = self.agent.tracer.runInContext({ handler, context: newCtx, thisArg: this, args, full: true })
+        const result = self.agent.tracer.bindFunction(handler, newCtx, true).apply(this, args)
         if (result?.then) {
           return result.then(function onThen(val) {
+            if (popPathOnExit && route) transaction.nameState.popPath(route)
```
bizob2828: i know this is in draft but i was looking at the changes. so if don't have this `popPathOnExit` what happens? is it that we're expecting middleware to be nested where as if you don't do this they're siblings? If so, I'd say this is the correct and the old way was wrong

amychisholm03: Yes for Express it's not needed because each sibling middleware sees a clean stack before appending its own route, but for `hapi`  its sibling route is still on the `nameState` stack

bizob2828 (commented):
this idea can remove the whole popPath on exit. feel free to clean up but the gist is that you want to override `constructRecorder`


```
diff --git a/lib/subscribers/hapi/ext.js b/lib/subscribers/hapi/ext.js
index a2c256585..82143f263 100644
--- a/lib/subscribers/hapi/ext.js
+++ b/lib/subscribers/hapi/ext.js
@@ -7,6 +7,7 @@
 
 const MiddlewareSubscriber = require('../middleware')
 const { transactionInfo } = require('#agentlib/symbols.js')
+const middlewareRecorder = require('#agentlib/metrics/recorders/middleware.js')
 
 // Request-scoped events (true) are instrumented; server lifecycle events (false) are not.
 const ROUTE_EVENTS = {
@@ -26,6 +27,9 @@ const ROUTE_EVENTS = {
 class HapiExtSubscriber extends MiddlewareSubscriber {
     super({ agent, logger, packageName: '@hapi/hapi', channelName: 'nr_ext', system: 'Hapi' })
+    this.wrapper.constructRecorder = function hapiExtConstructRecorder({ segmentName }) {
+      return middlewareRecorder(segmentName)
+    }
   }
 
   handler(data) {
@@ -48,7 +52,8 @@ class HapiExtSubscriber extends MiddlewareSubscriber {
       return middleware
     }
 
-    const wrapped = this.wrapper.wrap({ handler: middleware, route: '/' + event, popPathOnExit: true })
+    const mwName = middleware.name || '<anonymous>'
+    const wrapped = this.wrapper.wrap({ handler: middleware, segmentName: this.wrapper.prefix + '/' + mwName + '//' + event })
 
```

lib/subscribers/hapi/decorate.js
```diff
@@ -0,0 +1,37 @@
...
+const MiddlewareSubscriber = require('../middleware')
+
+class HapiDecorateSubscriber extends MiddlewareSubscriber {
+  constructor({ agent, logger }) {
+    super({ agent, logger, packageName: '@hapi/hapi', channelName: 'nr_decorate', system: 'Hapi' })
+  }
+
+  handler(data) {
+    const { arguments: args } = data
+    const [type,, fn] = args
```
bizob2828, line 17: Should all of this just be in an end event that way you don't have to wrap the handler and call it before wrapping?

test/versioned/hapi/package.json
```diff
@@ -11,13 +20,12 @@
       "dependencies": {
         "ejs": "2.5.5",
         "@hapi/hapi": ">=20.1.2",
-        "@hapi/vision": "latest"
+        "@hapi/vision": ">=5.0.0"
```
bizob2828: This is going to change the test matrix

lib/subscribers/hapi/ext.js
```diff
@@ -0,0 +1,72 @@
...
+      args[1] = this.#wrapMiddleware(method, events)
+    }
+  }
+
+  #wrapMiddleware(middleware, event) {
+    if (!ROUTE_EVENTS[event]) {
+      return middleware
+    }
+
+    const wrapped = this.wrapper.wrap({ handler: middleware, route: '/' + event, appendToPath: false })
```
bizob2828: This technically isn't true. You're passing in the route + appendToPath false just for naming.

bizob2828 (changes requested):
Overall looks good. A couple comments

test/versioned/hapi/package.json
amychisholm03: Yes, it adds to it but without it `compatibility.md` will only show support for the latest version which is technically not true. Though the latest version of `@hapi/version` is a few years old, so it might not matter only supporting the latest.

lib/subscribers/hapi/ext.js
amychisholm03: Yeah, I could tweak this to be more clear. I think I'll leave `route` alone and have like an `eventName` parameter that can be used for naming purposes only.

lib/subscribers/hapi/decorate.js
amychisholm03: `end` fires after `server.decorate()` has already run and stored the original reference. Changing `args[2]` does nothing at this point.

test/versioned/hapi/package.json
bizob2828: you can add `minSupported` to the `targets` object above

lib/subscribers/hapi/ext.js
bizob2828: what's wrong with the diff i posted? or updated middleware-subscriber to pass in a middleware wrapper and you just overload the `constructRecorder`

amychisholm03: Nothing, this was just a different approach. I can implement your approach verbatim. Only thing I see is that `segmentName` will be used as `metricName` for `makeMiddlewareRecorder` when the base implementation of `constructRecorder` actually returns `null` (instead of `makeMiddlewareRecorder(metricName)`) if `segmentName` is defined, so it may be a bit confusing.

bizob2828: it's fine. as long as you don't pass in `route`

bizob2828 (approved):
nice work

PR #3996, opened 2026-05-15
chore: Updated production dependencies to the latest

bizob2828:
## Description

Since we are doing a major release next week, I ran `npm outdated`. This PR upgrades all prod deps. It also removes an unnecessary `readable-stream` in lieu of `node:stream`.

**Note**: I will be opening a few more PRs to get dev deps up to date but that can happen at any time.


test/unit/util/logger.test.js
```diff
@@ -462,34 +463,27 @@ test('logger write queue should buffer writes', function (t, end) {
     hostname: 'my-host'
   })
 
+  const logEntries = ['b', 'c', 'd']
+  const queuedStrings = []
```
bizob2828, line 467: when triaging the tests, this wasn't actually ever asserting the lines. I fixed it and added a plan based style to ensure it won't regress

jsumners-nr (dismissed):

lib/collector/http-agents.js
```diff
@@ -64,6 +63,10 @@ exports.proxyAgent = function proxyAgent(config) {
   }
 
   logger.info(`using proxy: ${proxyUrl}`)
+  // lazy-load `https-proxy-agent` as it is ESM
+  // if required at top it would not allow us to instrument
+  // the core `node:http` correctly.
+  const { HttpsProxyAgent } = require('https-proxy-agent')
```
jsumners-nr, line 69: I like the call out. It will definitely help with maintenance.

jsumners-nr (approved):

PR #4007, opened 2026-05-18
chore: Upgraded `got` and `rimraf` to latest

bizob2828:
## Description
Per PR title


amychisholm03 (dismissed):

test/integration/instrumentation/http-outbound.test.js
```diff
@@ -179,13 +179,9 @@ test('external requests', async function (t) {
       nock.enableNetConnect()
     })
     nock('https://example.com').get('/').reply(200)
-    const got = require('got')
+    const got = require('got').default
```
jsumners-nr, line 182: Is this test even relevant any longer? I can't find a "NODE-1647" issue that describes the problem being tested. I am assuming it was in a prior JIRA instance that I don't have access to.

test/integration/logger.test.js
```diff
@@ -21,13 +21,9 @@ test('logger', async (t) => {
...
 
-    await new Promise((resolve) => {
-      if (fs.existsSync(dirPath)) {
-        rimraf(dirPath, resolve)
-      } else {
-        resolve()
-      }
-    })
+    if (fs.existsSync(dirPath)) {
+      await rimraf(dirPath)
```
jsumners-nr: We could use `fs.rm` at this point to accomplish this task: https://nodejs.org/api/fs.html#fsrmpath-options-callback

bizob2828: Cool. I'll refactor to remove rimraf

test/integration/instrumentation/http-outbound.test.js
bizob2828: No idea

bizob2828: Looking at the history it was a [bug](https://github.com/newrelic/node-newrelic/commit/cb4300ab79#diff-3652916a02fb30b6c705b2f2d3a6820d25faf3e4e11c2e30f7d1f2c57141740c) that was fixed in 2.6.0:

```sh
  The agent's HTTP instrumentation previously did not play well with the way
  `mimic-response` copied properties from an `http.IncomingMessage`. This caused
  modules that relied on that, such as `got`, to hang.
```

I'm going to keep

amychisholm03 (approved):

PR #4008, opened 2026-05-18
chore: Upgraded `glob` to 13

bizob2828:
## Description

glob is promised based. This was more involved than the others, just that I rewrote most of `test/bin/install_sub_deps`

### DIFF
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -4830,30 +4830,72 @@ SOFTWARE.
 
 ### glob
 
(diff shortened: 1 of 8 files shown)


amychisholm03 (approved):

jsumners-nr (commented):
https://nodejs.org/api/fs.html#fsglobpattern-options-callback

bizob2828:
> https://nodejs.org/api/fs.html#fsglobpattern-options-callback

Oh interesting. Let me do this. Thank me less 3rd party dep to manage

PR #4010, opened 2026-05-18
test: Add stable `http.request` shim so AWS test agent setup works with smithy >=4.7.3

amychisholm03:
## Description

Restores the per-test agent setup pattern (`beforeEach`/`afterEach`) for the AWS SDK v3 lambda, sns, and bedrock versioned tests, reverting the structural workarounds introduced in #3836 and #3838.

### Problem
The aforementioned PRs collapsed those specific AWS tests' setup into a single agent-per-file because of a bug surfaced by `@smithy/node-http-handler` ≥ 4.7.3 (ships with `@aws-sdk/*` ≥ ~3.700). At that version smithy lazily resolves `node:http` via:
```
const { Agent, request } = await import('node:http')
```

Node's ESM loader snapshots namespace bindings at first import, so once smithy captures `http.request`, subsequent `shimmer.unwrap` / re-wrap cycles are invisible to it. The second test in a `beforeEach` pattern would call into a wrapper bound to the first (now-unloaded) agent, which short-circuits to the un-instrumented request and produces no segments which will fail the tests.

### Fix

Adds a stable forwarding shim for `http.request` in `agent_helper.instrumentMockedAgent`. Installed once per process, the shim never changes, so smithy's ESM-snapshotted reference always resolves to it. Each `instrumentMockedAgent` call only swaps `_activeHttpRequest` to point at the current shimmer-wrapped function. We could probably remove this when we move over `core` instrumentation to subscriber-based.

With the shim in place, all five test files are restored to per-test agent construction.

## How to Test

```
npm run versioned aws-sdk-v3
```

## Related Issues

Closes #3837

### DIFF
--- a/test/lib/agent_helper.js
+++ b/test/lib/agent_helper.js
@@ -193,8 +193,28 @@ helper.instrumentMockedAgent = (conf, setState = true, shimmer = require('../../
   shimmer.bootstrapInstrumentation(agent)
   shimmer.registerHooks(agent)
   helper.maybeLoadSecurityAgent(agent)
+  installStableHttpRequest()
 
   return agent
+}
+
+// @smithy/node-http-handler ≥ 4.7.3 captures http.request via
+// `await import('node:http')`, which snapshots the ESM namespace at first
+// import. Subsequent shimmer.unwrap/re-wrap cycles (per-test agent setup)
+// are invisible to that snapshot, leaving smithy bound to a defunct agent.
+// Install a stable forwarding shim once; each instrumentMockedAgent call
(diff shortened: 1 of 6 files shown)


jsumners-nr (approved):
Good work. But I think we should also get @bizob2828's input.

bizob2828:
I won't have the time to review for a while. Don't wait on my input. Based on the PR style it already appears like Claude was used to analyze/solve

jsumners-nr:
We can certainly give it a trial and revert if we have problems.

PR #4011, opened 2026-05-19
security: Updated `@opentelemetry/exporter-metrics-otlp-proto` to `0.218.0` to remove reliance on a vulnerable `protobufjs@8.0.1`

bizob2828:
## Description

Turns out in #4006 we updated the dev dep which wasn't actually affected. But `@opentelemetry/exporter-metrics-otlp-proto` was affected and they released a new version to remove their reliance on `protobufjs`.


bizob2828:
I'm losing my mind. It turns out the version of `@opentelemtry/exporter-metrics-otlp-proto` we were on is ok. the affeted protobufjs was 8.0.0, not 8.0.1

PR #4013, opened 2026-05-19
chore: Added OTEL rule to support Prisma 7

jsumners-nr:
This PR is part of solving #3525. In order to utilize `@prisma/instrumentation` effectively, we have to add a rule that matches Prisma's incomplete spans. Specifically, they only attach the `db.system.name` and `db.query.text` attributes to their spans. Which means the lack the required attributes to match the `OtelDbClient1_40` rule (i.e. they are not fully following OTEL's semantic conventions).

With this PR merged, and a new agent version released, I will be able to add an example app that shows how to use our OTEL support to instrument `prisma@7` via their JavaScript variant and get dashboard results like:

image

### DIFF
--- a/lib/otel/traces/transformation-rules/311-OtelDbClientPrisma1_40.json
+++ b/lib/otel/traces/transformation-rules/311-OtelDbClientPrisma1_40.json
@@ -1,0 +1,28 @@
+{
+  "name": "OtelDbClientPrisma1_40",
+  "comment": "This rule supports the JS variant of Prisma 7. Their @prisma/instrumentation otel spans only include `db.system.name` and `db.query.text`. This means we were unable to match the standard rule.",
+  "type": "db",
+  "matcher": {
+    "required_span_kinds": [
+      "client"
+    ],
+    "required_attribute_keys": [
+      "db.system.name"
+    ],
+    "scope_name": "prisma"
+  },
+  "attributes": [
+    {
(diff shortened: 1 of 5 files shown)


svetlanabrennan (approved):

lib/otel/traces/transformation-rules/311-OtelDbClientPrisma1_40.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "OtelDbClientPrisma1_40",
+  "comment": "This rule supports the JS variant of Prisma 7. Their @prisma/instrumentation otel spans only include `db.system.name` and `db.query.text`. This means we were unable to match the standard rule.",
+  "type": "db",
+  "matcher": {
+    "required_span_kinds": [
+      "client"
+    ],
+    "required_attribute_keys": [
+      "db.system.name"
```
bizob2828, line 10: Keep in mind that without server.address the ingest platform will not be able to build an entity relationship between the application and the db server

jsumners-nr: I don't know what we can do about that.

bizob2828: I know they haven't been responsive but opening a PR to fix their otel span attrs would help

PR #4016, opened 2026-05-20
refactor: Refactored `koa` instrumentation to subscribe to events emitted

### LINKED ISSUE #3943: Migrate `koa` and `@koa/router` instrumentation to subscriber-based

bizob2828:
## Description

This supersedes #3983. I based the work off of amy's investigations.  I still have to clean up some things but it simplifies the koa instrumentation.

## How to Test

```sh
npm run versioned koa
```

## Related Issues
Closes #3943


lib/subscribers/koa/use.js
```diff
@@ -0,0 +1,58 @@
...
+    const { arguments: args } = data
+    const [mw] = args
+    if (mw.router) {
+      data.arguments[0] = this.#patchRouterDispatch(mw)
+    } else {
+      data.arguments[0] = this.wrapper.wrap({ handler: mw, route: '/' })
+    }
+  }
+
+  #patchRouterDispatch(layer) {
```
bizob2828, line 38: abstract this further

lib/symbols.js
```diff
@@ -14,10 +14,8 @@ module.exports = {
   disableDT: Symbol('Disable distributed tracing'), // description for backwards compatibility
   executorContext: Symbol('executorContext'),
   kafkaCtx: Symbol('kafkaCtx'),
-  koaBody: Symbol('body'),
```
bizob2828, line 17: remove most of the koa symbols as they are no longer used

test/versioned/koa/koa.test.js
```diff
@@ -107,7 +107,7 @@ test('Should name after koa framework and verb when body set', async (t) => {
   await plan.completed
 })
 
-test('Should name (not found) when no work is performed', async (t) => {
+test('Should name `/` when no work is performed', async (t) => {
```
bizob2828: is this correct? i need to dig

test/versioned/koa/koa.test.js
```diff
@@ -133,139 +133,6 @@ test('Should name (not found) when no work is performed', async (t) => {
   await plan.completed
 })
 
-test('names the transaction after the middleware that sets the body', async (t) => {
```
bizob2828, line 136: these tests were totally bogus and just here to assert behavior around setting `ctx.body` or `ctx.status`.  We should only be recording the middleware and timing/handling naming accordingly. Customers would never use `tx.nameState.appendPath` to change the transaction name

test/versioned/koa/scoped-koa-router.test.js
```diff
@@ -518,7 +499,7 @@ test(`${pkg} instrumentation`, async (t) => {
         // resolution of a recursively returned promise.
         // https://github.com/koajs/compose/blob/e754ca3c13e9248b3f453d98ea0b618e09578e2d/index.js#L42-L44
         assertSegments(tx.trace, tx.trace.root, [
-          'WebTransaction/WebFrameworkUri/Koa/GET//:second',
+          'WebTransaction/WebFrameworkUri/Koa/GET//:first/:second',
```
bizob2828, line 508: this behavior is diff but is now consistent with express. These routes are on different routers. it should use both for naming, [see](https://github.com/newrelic/node-newrelic/blob/54c60a841346737f6e6b42c4910381c0b98552a2/test/versioned/express/transaction-naming.test.js#L289)

test/versioned/koa/scoped-koa-router.test.js
```diff
@@ -558,7 +539,7 @@ test(`${pkg} instrumentation`, async (t) => {
         // resolution of a recursively returned promise.
         // https://github.com/koajs/compose/blob/e754ca3c13e9248b3f453d98ea0b618e09578e2d/index.js#L42-L44
         assertSegments(tx.trace, tx.trace.root, [
-          'WebTransaction/WebFrameworkUri/Koa/GET//:second',
+          'WebTransaction/WebFrameworkUri/Koa/GET//first/:second',
```
bizob2828, line 548: this behavior is diff but is now consistent with express. These routes are on different routers. it should use both for naming, [see](https://github.com/newrelic/node-newrelic/blob/54c60a841346737f6e6b42c4910381c0b98552a2/test/versioned/express/transaction-naming.test.js#L289)

lib/symbols.js
```diff
@@ -14,10 +14,8 @@ module.exports = {
...
-  koaBodySet: Symbol('bodySet'),
+  koaAllowedMethods: Symbol('koaAllowedMethods'),
```
jsumners-nr: I wonder if we should revisit adding instrumentation symbols here. I think the purpose of putting them here is so that they can be used across the code base. In particular, so that that test can gain access to them. With the subscriber migration, I think it makes sense to keep things closer to their implementation. So, with instrumentations, we could keep their specific symbols within the instrumentation and expose them to tests via a static property. For example:

```js
class FooInstrumentation extends Base {
	static internalSym = Symbol('internal-to-foo')
}

// test.js
const FooImplementation = require('/lib/subscribers/foo/whatever.js')
const sym = FooImplementation.internalSym
```

lib/subscribers/koa/middleware-wrapper.js
```diff
@@ -0,0 +1,57 @@
...
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+const MiddlewareWrapper = require('../middleware-wrapper')
+const ALLOWED_METHODS = ['MethodNotAllowedError', 'NotImplementedError']
+const { transactionInfo } = require('#agentlib/symbols.js')
+
+class KoaMiddlewareWrapper extends MiddlewareWrapper {
```
jsumners-nr, line 11: Just to be clear and get it on the record: the `MiddlewareWrapper` is intended to solve the majority of use cases, but for special cases like Koa we are able to extend its functionality accordingly. Correct?

lib/subscribers/middleware-wrapper.js
```diff
@@ -68,9 +68,11 @@ class MiddlewareWrapper {
    * @param {string} params.route route that is serving the handler
    * @param {string} params.segmentName name of segment, only used if not derived from prefix + function name + route
    * @param {number} params.nextIdx index of next handler, defaults to -1
+   * @param {boolean} params.asyncDoneHandler flag to indicate if the next/done handler is async(koa handlers are async but defined without `async` keyword)
+   * @param {boolean} params.isHandler flag to indicate if middleware is coming from a router with multiple middleware
```
jsumners-nr: I think I understand this doc. It's saying that the value of `handler` can be _not_ a handler function?

lib/symbols.js
bizob2828: Yea, prob

lib/subscribers/koa/middleware-wrapper.js
bizob2828: Correct. There are a few issues with koa: their router is different than others, how so, I'm not 100%. They have no error handling middleware, their IncomingMessage lives on a different key(however most frameworks are like this and I want to have follow up PRs to extend middleware wrapper to handle bespoke logic in the relevant library). Their next/done handler is async but doesn't return the async keyword. 

The general gist to handle all the weird transaction naming/routing nuances across all libraries is: if a middleware is part of a router, add the router path to the name state. When the done/next handler is called pop the path from name state unless it's the final handler. This will take care of a more accurate name based on multiple routers. Frankly I find this confusing but that naming decision was made well before us. I'm more than happy to chat more about this. 

I basically want to get away from spaghetti code like shim where you had know idea of all the branches.

lib/subscribers/middleware-wrapper.js
bizob2828: No. Maybe I change the name of this. It's a flag to indicate it's the final middleware in a router. This flag will ensure we don't pop the path for this handler because naming will get messed up

lib/subscribers/koa/middleware-wrapper.js
jsumners-nr: Thank you.

lib/subscribers/middleware-wrapper.js
bizob2828: let me clean this up

lib/symbols.js
```diff
@@ -14,10 +14,8 @@ module.exports = {
...
   koaRouter: Symbol('koaRouter'),
```
amychisholm03: `koaAllowedMethods` and `koaRouter` can be removed too.

lib/subscribers/middleware-wrapper.js
amychisholm03: Yeah, the koa middleware isn't even setting `asyncDoneHandler`, so that parameter should be safe to remove.

lib/subscriber-configs.js
```diff
@@ -17,6 +17,7 @@ const subscribers = {
   ...require('./subscribers/bunyan/config'),
   ...require('./subscribers/cassandra-driver/config'),
   ...require('./subscribers/connect/config'),
+  ...require('./subscribers/koa/config'),
```
amychisholm03: I'm not sure why I didn't put this in alphabetical order

lib/subscribers/middleware-wrapper.js
jsumners-nr: I like the improvement on `asyncDoneHandler`, but this line comment was on the `isHandler` line. The doc on this line is confusing since the first parameter is named `handler`. So `handler` is maybe a handler?

lib/subscribers/koa/middleware-wrapper.js
```diff
@@ -0,0 +1,58 @@
...
+  extractTxInfo(args, route) {
```
amychisholm03: `route` technically isn't needed here.

lib/subscribers/middleware-wrapper.js
```diff
@@ -68,9 +68,10 @@ class MiddlewareWrapper {
    * @param {string} params.route route that is serving the handler
    * @param {string} params.segmentName name of segment, only used if not derived from prefix + function name + route
    * @param {number} params.nextIdx index of next handler, defaults to -1
+   * @param {boolean} params.isHandler flag to indicate the middleware is the final one in a router layer stack
```
amychisholm03: I think `isLastInRouter` would be a more accurate name for this parameter.

bizob2828: what about `isRouteHandler`, or change the `handler` property to `middleware`

bizob2828: i'm terrible at naming. `isLastInRouter` is fine to me

lib/subscribers/koa/use.js
```diff
@@ -0,0 +1,86 @@
...
+    const router = layer.router
+    const stack = router?.stack ?? []
+    for (let i = 0; i < stack.length; i++) {
+      const pathLayer = stack[i]
+      const pathStack = pathLayer.stack
+      const path = pathLayer.path
+      for (let j = 0; j < pathStack.length; j++) {
+        const routedMiddleware = pathStack[j]
+        if (routedMiddleware.param) {
+          pathStack[j] = this.#patchParamMw(routedMiddleware, path)
```
amychisholm03, line 51: I'm not sure if these are valid use cases, but if router gets mounted twice we could double-wrap here. Otel adds a symbol to not rewrap the layer: https://github.com/open-telemetry/opentelemetry-js-contrib/blob/e697d7f351964ffb6f59abfcd0f05ea270a5fa5a/packages/instrumentation-koa/src/instrumentation.ts#L127-L132 so maybe we can implement something similar.

```
// user calls routes() twice for whatever reason
const router = new Router()
router.get('/foo', fooHandler)
app.use(router.routes())  // first call to _patchRouterDispatch
app.use(router.routes())  // second call to _patchRouterDispatch

// double app case
const sharedRouter = new Router()
sharedRouter.get('/api', apiHandler)
const app1 = new Koa()
const app2 = new Koa()
app1.use(sharedRouter.routes())  // wraps apiHandler in router.stack
app2.use(sharedRouter.routes())  // sees wrapped, wraps again
```

bizob2828: hmm, is that valid?

amychisholm03: No idea if someone would actually need to do this but it is valid code that runs; I wrote a test for both scenarios and it does double wrap

test/unit/lib/subscribers/middleware-wrapper.test.js
amychisholm03, line 1: We should probably rename this to middleware-wrapper.test.js

lib/subscribers/koa/use.js
bizob2828: ok i'm handling this for all middleware. I'm doing analysis if any other subscriber based patching would be subject to double wrapping and open a separate PR if need be

jsumners-nr (approved):

amychisholm03 (approved):
LGTM, we'll address double-wrapping for all web middleware in separate PR

bizob2828:
> LGTM, we'll address double-wrapping for all web middleware in separate PR

what do you mean? This fix handles all web middleware

PR #4018, opened 2026-05-21
chore: Added audit logging for OTEL metrics

### LINKED ISSUE #4015: Add audit logging support to OTEL metrics feature
Spec PR 795 adds the requirement that our OTEL metrics feature support audit logging. In particular, we are supposed to:

1. Log the request URI with any sensitive data in the query parameters obscured.
2. Log the outgoing request body as a base64 encoded string along with the byte count for the unencoded data.
3. Log the response code and the base64 encoded response body.

However, we can't accomplish logging the actual response code. We are unable to access the actual HTTP request. This means we also can't log the actual request URI, but we can log the configured URI.

To implement this feature, we need to:

1. Write a class that extends `OTLPMetricExporterBase`. The purpose of this class is to create the actual delegate and a proxying delegate. The proxying delegate is then used to create the super instance.
2. Write a proxying serializer that utilizes `ProtobufMetricsSerializer` to serialize and deserialize the data, and write logs.
3. Write a proxying delegate that will wrap the operations of the actual delegate and provide access to the response code.

We then need to replace the `OTLPMetricExporter` instance with an instance of the class we created in step 1.

Note: the agent logger should be used, and logs should only be written if audit logging or trace level logging is enabled.

jsumners-nr:
This PR resolves #4015.

Technically, we are supposed to log the URI for every outgoing request. But we can't access the transport. So the best we'd be able to do is to log the URL that we pass in for configuration. I omitted that because I don't see the benefit.

Additionally, I brought up the fact that we can't access the transport in our Slack channel and had a discussion about the "MUST" requirements in the spec PR. The consensus is that some agents may not be able to meet those requirements, and should be able to make a best effort. We can't meet any of the "MUST" requirements as they are outlined, so this is our best effort.


lib/util/logger.js
```diff
@@ -199,6 +200,38 @@ for (const _level of Object.keys(LEVELS)) {
...
+   * @returns {*} The result of the logging function or `undefined`.
+   */
+  loggingFunctions.audit = function audit(...args) {
+    if (this.options.auditLogging === true) {
+      return this.info.apply(this, args)
+    }
+    if (this.traceEnabled() === true) {
+      return this.trace.apply(this, args)
+    }
+  }
```
jsumners-nr, line 222: I added this because the only place I could find where we use `config.audit_log` is at https://github.com/newrelic/node-newrelic/blob/6289502ba49d88b8746ac8a42a34e5c2069f0d84/lib/collector/remote-method.js#L258-L267

I thought it would be rather silly to duplicate the above logic multiple times in this new code. So I centralized it in the logger.

lib/otel/metrics/nr-proxying-serializer.js
```diff
@@ -0,0 +1,61 @@
...
+class NRProxyingSerializer {
+  #logger
+
+  constructor(logger = defaultLogger) {
+    this.#logger = logger.child({ subcomponent: this.constructor.name })
+  }
+
+  serializeRequest(metrics) {
+    const serialized = ProtobufMetricsSerializer.serializeRequest(metrics)
+    if (this.#logger.auditEnabled() === true) {
```
bizob2828, line 31: may i ask why you're checking this here but the logger is also checking if audit logging is enabled?

lib/otel/metrics/nr-proxying-serializer.js
```diff
@@ -0,0 +1,61 @@
...
+          bytes: buffer.byteLength
+        },
+        'Serialized metrics data.'
+      )
+    }
+    return serialized
+  }
+
+  deserializeResponse(data) {
+    if (this.#logger.auditEnabled() === true) {
```
bizob2828, line 54: may i ask why you're checking this here but the logger is also checking if audit logging is enabled?

lib/otel/metrics/nr-proxying-serializer.js
```diff
@@ -0,0 +1,61 @@
...
+      // 1. We can't use `serialized.toBase64()` because that isn't avilable
+      // untile Node.js >= 25.
+      // 2. Using `new DataView(serialized)` to get the byte length would
+      // still be necessary because `UInt8Array` doesn't have a byte length
+      // field.
+      // So it's a bit cheaper to use `Buffer`, but still expensive if we
+      // don't need to log the information.
+      const buffer = Buffer.from(serialized)
+      this.#logger.audit(
+        {
```
bizob2828, line 42: i know you called out that the transport lacks the URI but you could pass in the appropriate configurations here to construct that URL + license key right?

lib/util/logger.js
```diff
@@ -199,6 +200,38 @@ for (const _level of Object.keys(LEVELS)) {
...
+   * We consider trace level logging equivalent to audit logging. But audit
+   * logging can happen at _any_ log level greater than or equal to 'info'
+   * level.
+   *
+   * @param {*} args The same arguments that would be passed to any level
+   * named logging function.
+   *
+   * @returns {*} The result of the logging function or `undefined`.
+   */
+  loggingFunctions.audit = function audit(...args) {
```
bizob2828, line 215: it's prob best to add a follow up PR that uses this method in `collector/remote-method.js`

lib/otel/metrics/nr-exporter.js
```diff
@@ -0,0 +1,44 @@
...
+})
+const NRProxyingDelegate = require('./nr-proxying-delegate.js')
+const NRProxyingSerializer = require('./nr-proxying-serializer.js')
+
+/**
+ * Provides the functionality of upstream's `OTLPMetricExporter`, but wraps
+ * the internal objects so that we can write audit logs.
+ */
+class NROTLPMetricExporter extends OTLPMetricExporterBase {
+  constructor(config, logger = defaultLogger) {
```
bizob2828:
```suggestion
  constructor(config = {}, logger = defaultLogger) {
```

lib/otel/metrics/nr-exporter.js
```diff
@@ -0,0 +1,44 @@
...
+    const convertedOptions = convertLegacyHttpOptions(
+      config || {},
```
bizob2828:
```suggestion
      config
```

test/unit/lib/otel/metrics/nr-proxying-serializer.test.js
```diff
@@ -0,0 +1,96 @@
...
+  ctx.nr.meterProvider = new MeterProvider({
+    readers: [ctx.nr.reader]
+  })
+})
+
+test.afterEach(async (ctx) => {
+  await ctx.nr.meterProvider.shutdown()
+})
+
+test('logs when serializing', async (t) => {
```
bizob2828, line 49: if you're going to keep the auditLogging checks in the serializer, you should have tests that do not have audit logging enabled and verify no writes occur to the logger

bizob2828 (changes requested):

lib/otel/metrics/nr-proxying-serializer.js
jsumners-nr: Because encoding the data for the log is expensive. We can't pass the raw data to the logger and have it do the encoding.

lib/otel/metrics/nr-proxying-serializer.js
jsumners-nr: As noted in the other identical review and the code comment below it.

bizob2828 (approved):

PR #4019, opened 2026-05-21
test: Removed reliance on `glob` in lieu of `fs.glob`

bizob2828:
## Description
In #4008 I merged it but @jsumners-nr had a good comment about this package is now obsolete. Since we support 22+, I removed `glob` in lieu of `fs.glob`


jsumners-nr (dismissed):
Fabulous!

jsumners-nr (approved):

PR #4022, opened 2026-05-21
chore: Updated `nock` to latest

bizob2828:
## Description

This was a confusing one. I had to employ the help of claude with it.
Basically nock drastically changed and because of keep-alives and 
new requirements on the mock responses, `test/lib/nock-utils.js` was
added to handle interactions with our collector.

Also the interceptors of nock changed some of our http instrumentation tests.

### DIFF
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -4956,7 +4956,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 
 ### nock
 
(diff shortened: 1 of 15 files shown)


jsumners-nr (dismissed):
👍 

-----

I used to rely on `nock` and thought it was one of Node's best ecosystem modules. But I've grown to prefer other methods, mainly because of things like the caveats noted throughout these changes. I have come to prefer starting a local server to serve out the desired mock requests and judiciously patching `dns` to redirect requests when necessary (https://github.com/newrelic/node-newrelic/blob/0ced30446a9be2711c96c0afa6453e6b87c014ad/test/integration/core/dns.test.js#L30-L70)

Maybe we can come up with some simple abstraction over that which would remove our reliance on `nock` in the future. 🤷‍♂️

bizob2828:
I had to rebase to resolve conflicts

jsumners-nr (approved):

Issue #3288, opened 2025-08-11
Refactor `mongodb` to subscribe to events emitted


amychisholm03:
## Description
Update instrumentation to create a subscriber for mongodb and a instrumentator for orchestrion.

## Acceptance Criteria
mongodb instrumentation still functions as expected after traditional monkey patching and shim helpers are used.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-469985

PR #4023, opened 2026-05-21
test: Refactor `mongodb-esm` tests to test `>=4.1.4`

amychisholm03:
## Description

This is a precursor for solving issue #3288. Our `mongodb-esm` tests only tested `>=4.1.4 <5` and even as a smoke test, they weren't really useful because we weren't testing if 5.x, 6.x, and 7.x worked with esm. The tests also have to be overhauled quite a bit to support testing versions 4.x - 7.x; they are basically exact ports of the existing CJS tests in `test/versioned/mongodb`.

## How to Test

```
npm run versioned mongodb-esm
```

## Related Issues

N/A


test/versioned/mongodb-esm/collection-common.mjs
```diff
@@ -0,0 +1,255 @@
...
+
+import test from 'node:test'
+import assert from 'node:assert'
+import { createRequire } from 'node:module'
+
+import helper from '../../lib/agent_helper.js'
+import common from '../mongodb/common.js'
+
+const require = createRequire(import.meta.url)
+const mongoPackage = require('mongodb/package.json')
```
jsumners-nr: Why are we doing this? Can we use https://github.com/newrelic/node-newrelic/blob/150476a61cd272a29ed7c03366e879f6eb009298/test/lib/get-package-version.js?

test/versioned/mongodb-esm/collection-misc.test.mjs
```diff
@@ -3,216 +3,136 @@
...
+    data.sort((a, b) => a._id - b._id)
+    assert.deepStrictEqual(data, expectedData)
+
+    verify(null, [`${STATEMENT_PREFIX}/mapReduce`], ['mapReduce'], { strict: false })
+
+    /* eslint-disable */
+    function map() {
+      emit(this.mod10, this.i)
+    }
+    /* eslint-enable */
```
jsumners-nr: Why is linting disabled here?

test/versioned/mongodb-esm/collection-common.mjs
amychisholm03: It was just copy-pasted from the CJS tests, I can change both

test/versioned/mongodb-esm/collection-misc.test.mjs
amychisholm03: Again copy-pasted from existing CJS tests, looks like it is throwing `'emit' is not defined (no-undef)`

jsumners-nr: We should be able to:

```js
/* eslint-disable -- a reason here */
```

But, also, `// eslint-disable-next-line no-undef -- a reason here`.

test/versioned/mongodb-esm/collection-common.mjs
```diff
@@ -0,0 +1,255 @@
...
+const DB_NAME = common.ESM.DB_NAME
+const STATEMENT_PREFIX = common.ESM.STATEMENT_PREFIX
+
+let METRIC_HOST_NAME = null
+let METRIC_HOST_PORT = null
+
+import { fileURLToPath } from 'url'
+import { dirname } from 'path'
+const __filename = fileURLToPath(import.meta.url)
+const pkgVersion = getPackageVersion({ pkgName: 'mongodb', baseDir: dirname(__filename) })
```
jsumners-nr:
```suggestion
const pkgVersion = getPackageVersion({
  pkgName: 'mongodb',
  baseDir: dirname(import.meta.filename)
})
```

jsumners-nr (changes requested):

jsumners-nr (approved):

PR #4034, opened 2026-05-26
chore: Deprecate `instrumentLoadedModule` API

### LINKED ISSUE #4031: Deprecate `instrumentLoadedModule` API
### Description
Deprecate the public `newrelic.instrumentLoadedModule(moduleName, module)` API in favor of preload-only loading (`node -r newrelic` / `NODE_OPTIONS='-r newrelic'`).

### Acceptance Criteria
As a **Node.js agent user**, I **want a clear deprecation warning and doc guidance when I call `instrumentLoadedModule`** so that **I can migrate to `-r newrelic` before the next major removes the API**.

- [ ] `instrumentLoadedModule` emits a one-shot deprecation log on first call.
- [ ] JSDoc carries `@deprecated` with a pointer to the preload docs.
- [ ] Change README, API reference, and any "loaded after my app" guidance to recommend `-r newrelic` only.
- [ ] `NEWS.md` / changelog entry calls out the deprecation and intended removal version. Will be done next minor release.
- [ ] Existing tests in `test/unit/api/api-instrument-loaded-module.test.js` and `test/integration/api/instrument-loaded.test.js` still pass; a new test asserts the deprecation warning fires exactly once.

amychisholm03:
## Description

Deprecates `newrelic.instrumentLoadedModule` effective next release. We will completely remove it next major release (v15).

## How to Test

```
npm run unit
```

## Related Issues

Closes #4031

### DIFF
--- a/api.js
+++ b/api.js
@@ -1493,6 +1493,10 @@ API.prototype.instrumentMessages = function instrumentMessages(moduleName, onReq
  * Applies an instrumentation to an already loaded CommonJs module.
  *
  * Note: This function will not work for ESM packages.
+ *
+ * @deprecated Scheduled for removal in v15. Load the agent via
+ * `node -r newrelic` (or `NODE_OPTIONS='-r newrelic'`), so the agent is active
+ * before user modules resolve.
  * @example
  *
  * // oh no, express was loaded before newrelic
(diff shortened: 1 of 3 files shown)


api.js
```diff
@@ -1511,6 +1515,12 @@ API.prototype.instrumentMessages = function instrumentMessages(moduleName, onReq
  * @returns {boolean} Whether or not the module was successfully instrumented
  */
 API.prototype.instrumentLoadedModule = function instrumentLoadedModule(moduleName, module) {
+  logger.warn(
+    'newrelic.instrumentLoadedModule is deprecated and will be removed in v15. ' +
+      'Load the agent via `node -r newrelic` (or NODE_OPTIONS=\'-r newrelic\'), ' +
+      'so it is active before user modules resolve.'
+  )
```
jsumners-nr: This isn't going to issue a warning where people will see it. It's good to write it out to our agent logs for debugging purposes, but we need to make it known to the user in a visible manner via `process.emitWarning`.

In my opinion, we should use [`process-warning`](https://www.npmjs.com/package/process-warning)'s `createDeprecation` method.

jsumners-nr (changes requested):

api.js
amychisholm03: I'll use `process.emitWarning` because `process-warning` is a new dep.

jsumners-nr (approved):
Note: as written, the process warning will emit for every invocation of the method. The suggested module allows for defining `{ unlimited: false }` for a single issuance.

amychisholm03:
> Note: as written, the process warning will emit for every invocation of the method. The suggested module allows for defining `{ unlimited: false }` for a single issuance.

For this specific API, emitting for every invocation is probably the right call because if a user is using this API for a variety of different modules.

PR #4036, opened 2026-05-27
chore: Added baseline instructions for Claude bot

jsumners-nr:
If we are going to keep using this tool, we should add some common baseline instructions for it. This PR enables the following prompts to work correctly:

- `How would you run a new unit test?`
- `run tests for the fastify instrumentation`
- `run the fastify "add hook" tests`

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -1,0 +1,30 @@
+# newrelic instrumentation agent
+
+`newrelic` is an observability instrumentation agent. It hooks and traces
+third party modules.
+
+## Conventions
+- Code style is enforced using eslint and should be written according to its configuration
+- Tests are written using the `node:test` module
+
+## Important
+- Unit tests live in `test/unit/`
+- When verifying new unit tests, run the test directly as `node --test <test_file>`
+- Agent integration tests live in `test/integration/` and verify agent
+specific functionality
+- Instrumentation of third party modules is verified by "versioned tests"
+- Versioned tests live in `test/versioned/`
+- Docker services should be running for versioned tests to succeed
+- After running a specific versioned test suite at least once, individual
+test files in the suite may be run directly as
+`node --test test/versioned/<suite>/<test_file>`
+
+## Commands
+npm run services:start # start Docker services
+npm run services:stop # stop Docker services
+npm run versioned:major # run all versioned tests
+npm run versioned:major <suite_name> # run specific versioned test suite
+npm run unit # run all unit tests
+npm run lint # verify code style
+npm run lint:verbose # show all code style errors
+npm run lint:fix # automatically fix incorrect code style
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,8 @@ bin/.env
 
 # Local developer resources:
 local/
+# .gitignore - Claude Code personal files
+# per https://computingforgeeks.com/claude-code-dot-claude-directory-guide/
+CLAUDE.local.md
+.claude/settings.local.json
+.claude/agent-memory-local/


.claude/CLAUDE.md
```diff
@@ -0,0 +1,30 @@
+# newrelic instrumentation agent
```
amychisholm03, line 1:
```suggestion
# New Relic Node.js Application Performance Monitoring agent
```

.claude/CLAUDE.md
```diff
@@ -0,0 +1,30 @@
...
+
+`newrelic` is an observability instrumentation agent. It hooks and traces
+third party modules.
+
+## Conventions
+- Code style is enforced using eslint and should be written according to its configuration
```
amychisholm03, line 7:
```suggestion
- Code style is enforced using eslint and should be written according to its configuration. Style can be checked with `npm run lint:fix`.
```

.claude/CLAUDE.md
```diff
@@ -0,0 +1,30 @@
...
+- Tests are written using the `node:test` module
+
+## Important
```
amychisholm03, line 10:
```suggestion
## Testing
```

.claude/CLAUDE.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Commands
+npm run services:start # start Docker services
+npm run services:stop # stop Docker services
+npm run versioned:major # run all versioned tests
+npm run versioned:major <suite_name> # run specific versioned test suite
+npm run unit # run all unit tests
+npm run lint # verify code style
+npm run lint:verbose # show all code style errors
+npm run lint:fix # automatically fix incorrect code style
```
amychisholm03, line 30: I'd move the lint commands up and just have this as testing commands. Having sections for linting and then seperately for testing is probably ideal.

.claude/CLAUDE.md
```diff
@@ -0,0 +1,30 @@
...
+- Agent integration tests live in `test/integration/` and verify agent
+specific functionality
+- Instrumentation of third party modules is verified by "versioned tests"
+- Versioned tests live in `test/versioned/`
+- Docker services should be running for versioned tests to succeed
+- After running a specific versioned test suite at least once, individual
+test files in the suite may be run directly as
+`node --test test/versioned/<suite>/<test_file>`
+
+## Commands
```
amychisholm03, line 22:
```suggestion
### Commands
```

amychisholm03 (changes requested):
Just some nits for a bit better structure. I think we should have separate sections for linting and testing descriptions and commands. I also think a `Repo layout` and `Don't touch` section would be helpful

```
## Repo layout
- `index.js` — entry point that bootstraps the agent
- `api.js` — public API surface exposed to user applications
- `esm-loader.mjs` — ESM loader hook (separate from the CJS path)
- `lib/` — agent source
- `lib/instrumentation/` — legacy `Shim`-based instrumentation
  (monkey-patching); generally one file per third party module
- `lib/subscribers/` — newer subscriber-based instrumentation built on
  Node's `diagnostics_channel` `TracingChannel` via
  `@apm-js-collab/tracing-hooks` and `orchestrion-js`. New
  instrumentations should be written here; existing `Shim`-based ones
  are being migrated. See `lib/subscribers/README.md` for the pattern.
- `test/unit/` — unit tests for agent internals
- `test/integration/` — agent-level integration tests
- `test/versioned/` — versioned tests; one directory per instrumented
  module, each with its own `package.json` declaring the version ranges
  to test against
- `compatibility.md` — generated; source of truth for supported module
  versions
- `CONTRIBUTING.md` — commit and PR conventions

## Generated files (do not hand-edit)
- `third_party_manifest.json`
- `changelog.json`
- `ai-support.json`
- `compatibility.md`
```

.claude/CLAUDE.md
jsumners-nr: The current title has done all it needs to do with less verbosity (necessary for this tool).

.claude/CLAUDE.md
jsumners-nr: Not necessary. Covered below.

.claude/CLAUDE.md
jsumners-nr: Not according to the documentation I followed. The keyword "important" is imperative here.

.claude/CLAUDE.md
jsumners-nr: It is intentionally its own sub-heading. They describe project specific commands, not testing specific.

jsumners-nr:
The structure in this document is for a bot, not humans. Everything is just enough to get the machine to do the right things.

amychisholm03:
> The structure in this document is for a bot, not humans. Everything is just enough to get the machine to do the right things.

Sure, the structure doesn't matter too much, but I do think adding the `Repo layout` section would be helpful so it doesn't have to do as many taxing `grep` calls.

jsumners-nr:
Honestly, I don't know how I would verify those sections are doing anything. I was able to verify that the content of the PR does as intended, and I'd like to start small. We can expand further later (for any definition of "later").

amychisholm03 (approved):
Worth expanding on in future PRs but fine for now.

PR #4037, opened 2026-05-27
refactor: Migrate `mongodb` instrumentation to subscriber-based

### LINKED ISSUE #3288: Refactor `mongodb` to subscribe to events emitted

amychisholm03:
## Description

Migrates the `mongodb` instrumentation from `Shim`-based to subscriber-based.

- Removed all `Shim`-based instrumentation in `lib/instrumentation/mongodb`
- **Added** `lib/subscribers/mongodb/`, which hooks **5 class constructors** (Collection, Db, AbstractCursor, BulkOperationBase, MongoClient) and wraps the relevant prototype methods once at runtime, resolving host/db/collection from the live object at call time. `constants.js` lists the methods to wrap per class; missing methods are skipped at wrap time, so version-specific methods need no special-casing.
- Segment/metric behavior matches the legacy shim, so the existing versioned tests ,pass unchanged. Only `test/unit/shimmer.test.js` was updated, for the new channel names.

### With Scripts
```
npm run versioned mongodb
npm run versioned mongodb-esm
npm run unit
```

### Via Example App

Use the [`graphql-koa-dataloader` app](https://github.com/newrelic/newrelic-node-examples/tree/graphql-koa-dataloader-update/graphql-koa-dataloader) with `NEW_RELIC_DATASTORE=Mongo` in your `.env`.

Comparing the instrumentation styles, the transactions and MongoDB segments are identical:
**Subscriber-based**
- `Koa/POST//query/<anonymous>/user` - https://staging.onenr.io/0oR80Vr3gRG
- `Koa/POST//query/<anonymous>/allUsers` - https://staging.onenr.io/0PwJper7pj7

**`Shim`-based**
- `Koa/POST//query/<anonymous>/user` - https://staging.onenr.io/02wdvoXb7jE
- `Koa/POST//query/<anonymous>/allUsers` - https://staging.onenr.io/0VjYDamkvR0

## Related Issues

Closes #3288


lib/subscribers/mongodb/config.js
```diff
@@ -0,0 +1,144 @@
...
+    {
+      channelName: 'nr_execute',
+      module: { name: 'mongodb', versionRange: '>=4.1.4 <5.0.0', filePath: 'lib/bulk/common.js' },
+      functionQuery: { className: 'BulkOperationBase', methodName: 'execute', kind: 'Sync' }
+    }
+  ]
+}
+
+// ---------------------------------------------------------------------------
+// Export
```
bizob2828: What's the point of this comment?

lib/subscribers/mongodb/config.js
```diff
@@ -0,0 +1,144 @@
...
+// `createIndex` on Db: channel nr_db_createIndex (regex strips db_ prefix → createIndex)
+const dbCreateIndex = dbOperationConfig('createIndex', { channelName: 'nr_db_createIndex' })
+// `indexInformation` on Db: same collision with Collection.indexInformation
+const dbIndexInformation = dbOperationConfig('indexInformation', { channelName: 'nr_db_indexInformation' })
+
+// ---------------------------------------------------------------------------
+// BulkOperationBase.execute
+// ---------------------------------------------------------------------------
+
+// In v4 execute(options, callback) is declared without `async`; v5+ uses `async execute()`.
```
bizob2828: Kind async doesn't mean it has the async keyword. It just means it returns a promise

lib/subscribers/mongodb/config.js
```diff
@@ -0,0 +1,144 @@
...
+    }
+  ]
+}
+
+// ---------------------------------------------------------------------------
+// Db — special cases that don't fit the standard operation config
+// ---------------------------------------------------------------------------
+
+// `createIndex` on Db: channel nr_db_createIndex (regex strips db_ prefix → createIndex)
+const dbCreateIndex = dbOperationConfig('createIndex', { channelName: 'nr_db_createIndex' })
```
bizob2828: I don't understand these comments

lib/subscribers/mongodb/config.js
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// `createIndex` exists on both Collection (query/statement) and Db (operation).
+// They use different subscriber files so they need different channel names.
+const collectionCreateIndex = collectionQueryConfig('createIndex', { channelName: 'nr_createIndex' })
+
+// `indexInformation` exists on both Collection and Db — same collision pattern.
+const collectionIndexInformation = collectionQueryConfig('indexInformation', { channelName: 'nr_indexInformation' })
+
+// Collection.aggregate is a sync cursor factory (returns a cursor, no `async`
```
bizob2828: async keyword doesn't mean kind async. If a function returns a promise the kind is async

lib/subscribers/mongodb/config.js
```diff
@@ -0,0 +1,144 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const { COLLECTION_QUERY_OPS, CURSOR_OPS, DB_OPS, MONGO_VERSION_RANGE } = require('./constants')
+const { collectionQueryConfig, cursorMethodConfig, dbOperationConfig } = require('./config-builder')
+
+// ---------------------------------------------------------------------------
+// Collection — special cases that don't fit the standard query config
```
bizob2828: What's this standard query config that's mentioned in here a lot?

lib/subscribers/mongodb/query.js
```diff
@@ -0,0 +1,51 @@
...
+
+'use strict'
+
+const DbQuerySubscriber = require('../db-query')
+const { MONGODB } = require('../../metrics/names')
+const { getParameters, findCollectionName, operationFromChannel } = require('./utils')
+
+/**
+ * Subscriber for async MongoDB query operations on Collection and Cursor objects.
+ * Covers COLLECTION_OPS (Promise-returning) and CURSOR_OPS.
```
bizob2828:
```suggestion
 * Covers COLLECTION_QUERY_OPS (Promise-returning) and CURSOR_OPS.
```

lib/subscribers/mongodb/query.js
```diff
@@ -0,0 +1,51 @@
...
+class MongoQuerySubscriber extends DbQuerySubscriber {
+  constructor({ agent, logger, channelName, packageName = 'mongodb' }) {
+    super({ agent, logger, channelName, packageName, system: MONGODB.PREFIX })
+    this.events = ['asyncEnd']
+    // Prevent internal MongoDB operations (e.g. cursor.next inside findOne) from
+    // creating nested child segments under this one.
+    this.opaque = true
+  }
+
+  handler(data, ctx) {
```
bizob2828: this isn't handling `callback` from the legacy shim. looking at their changelog they removed callbacks in 5.0.0, or is that the callback in 4.x just wrapped by the promise?

lib/subscribers/mongodb/query.js
```diff
@@ -0,0 +1,51 @@
...
+    const operation = operationFromChannel(this.channelName)
+    this.queryString = operation
+    this._mongoObject = data.self
+
+    const params = getParameters(data.self, this.system)
+    // Collection.rename is always routed to the admin database.
+    if (operation === 'rename') {
+      params.database_name = 'admin'
```
bizob2828: I don't see this logic in the old code

lib/subscribers/mongodb/query.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+    this.parameters = params
+
+    return super.handler(data, ctx)
+  }
+
+  parseQuery(queryString) {
```
bizob2828: we have a query parser for mongo in `lib/db/query-parsers/mongodb.js` does it not work?

lib/subscribers/mongodb/config.js
```diff
@@ -0,0 +1,178 @@
...
+  instrumentations: [
+    {
+      channelName: 'nr_execute',
+      module: { name: 'mongodb', versionRange: '>=5.0.0', filePath: 'lib/bulk/common.js' },
+      functionQuery: { className: 'BulkOperationBase', methodName: 'execute', kind: 'Async' }
+    },
+    {
+      channelName: 'nr_execute',
+      module: { name: 'mongodb', versionRange: '>=4.1.4 <5.0.0', filePath: 'lib/bulk/common.js' },
+      functionQuery: { className: 'BulkOperationBase', methodName: 'execute', kind: 'Sync' }
```
bizob2828: this returns a promise. you don't need different configurations here. remember `async` kind doesn't mean async keyword. it means it returns a promise

lib/subscribers/mongodb/config.js
```diff
@@ -0,0 +1,178 @@
...
+// ---------------------------------------------------------------------------
+// Db — special cases that don't fit the standard operation config
+// ---------------------------------------------------------------------------
+
+// `createIndex` on Db: channel nr_db_createIndex (regex strips db_ prefix → createIndex)
+const dbCreateIndex = dbOperationConfig('createIndex', { channelName: 'nr_db_createIndex' })
+// `indexInformation` on Db: same collision with Collection.indexInformation
+const dbIndexInformation = dbOperationConfig('indexInformation', { channelName: 'nr_db_indexInformation' })
+
+// `collection` on Db: sync factory returning a Collection. Channel
```
bizob2828: say what now?

lib/subscribers/mongodb/config-builder.js
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * @param {object} params - Instrumentation target descriptor.
+ * @param {string} params.channelName - Diagnostics channel name.
+ * @param {string} params.filePath - Module-relative file to instrument.
+ * @param {string} params.className - Class name (used for the pre-v7 stanza).
+ * @param {string} params.methodName - Method to instrument.
+ * @param {string} [params.kind] - 'Async' (default) or 'Sync'.
+ * @returns {Array<object>} The instrumentation stanzas.
+ */
+function versionSplitInstrumentations({ channelName, filePath, className, methodName, kind = 'Async' }) {
```
bizob2828: why is kind an option? It's always async

lib/subscribers/mongodb/operation.js
```diff
@@ -0,0 +1,38 @@
...
+    this.events = ['asyncEnd']
+    this.opaque = true
+  }
+
+  handler(data, ctx) {
+    const operation = operationFromChannel(this.channelName)
+    this.operation = operation
+    const params = getParameters(data.self, this.system)
+    // Db.renameCollection is always routed to the admin database.
+    if (operation === 'renameCollection') {
```
bizob2828: I don't see this in the legacy instrumentation. why is this here or can it not be pulled out of `getParameters`?

lib/shimmer.js
```diff
@@ -391,7 +391,9 @@ const shimmer = (module.exports = {
     for (const subscriberConfigList of Object.values(subscriptions)) {
       for (const subscriberConfig of subscriberConfigList) {
         const Subscriber = require(`./subscribers/${subscriberConfig.path}`)
-        const subscriber = new Subscriber({ agent, logger })
+        // Forward channelName so subscribers reused across multiple channels (e.g. mongodb) can bind to the correct one.
+        const channelName = subscriberConfig.instrumentations[0]?.channelName
+        const subscriber = new Subscriber({ agent, logger, channelName })
```
bizob2828, line 396: you should add a unit test for this in `test/unit/shimmer.test.js`

lib/subscribers/mongodb/utils.js
```diff
@@ -0,0 +1,116 @@
...
+ */
+function bulkCollection(bulk) {
+  return bulk?.collection ?? bulk?.s?.collection ?? null
+}
+
+// Property paths that may resolve to a MongoClient on a given MongoDB object.
+// Covers MongoClient, Collection, Db, AbstractCursor, and the v4.1-v4.3 cursor
+// variant that holds a Topology in place of a client (Topology exposes the same
+// `s.options.hosts` shape).
+const CLIENT_PATHS = [
```
bizob2828: why is this a series of functions and not just a series of case or if/else if statements? creating functions to extract the client every time is going to slow things down

lib/subscribers/mongodb/utils.js
```diff
@@ -0,0 +1,116 @@
...
+ * @param {object} obj - A MongoDB internal object.
+ * @returns {object|null} The MongoClient (or Topology), or null if not found.
+ */
+function findClient(obj) {
+  if (!obj) return null
+  for (const get of CLIENT_PATHS) {
+    const candidate = get(obj)
+    if (candidate?.s?.options?.hosts) return candidate
+  }
+  const coll = bulkCollection(obj)
```
bizob2828: is this needed outside of the `MongoBulkSubscriber`? it's being called directly in there and findClient isn't called in that file at all

test/versioned/disabled-instrumentation/disabled-ioredis.test.js
```diff
@@ -22,21 +21,23 @@ test('Disabled ioredis scenarios', async (t) => {
       }
     })
     const Redis = require('ioredis')
-    const mongodb = require('mongodb')
-    const mongo = await mongoCommon.connect({ mongodb })
```
bizob2828, line 26: did this not work? seems to be close to what's not verbose in this file

bizob2828 (changes requested):
We saved this library for one of the last because we knew it was complex. With that being said I'm concerned about a lot of the utilities being inefficient and a lot of confusing comments/complexity in code.  It's also worth noting this instrumentation is significantly slower than the former.  I ran a full suite of versioned tests on this branch vs main:

This branch:
```sh
mongodb(52)
===============================================================
 ✓ mongodb (70) 8m
===============================================================
Versions executed

Folder: mongodb
	 * mongodb(10): 4.1.4, 4.6.0, 4.12.1, 5.0.1, 5.6.0, 6.1.0, 6.7.0, 6.13.1, 6.19.0, 7.2.0
===============================================================
PASS

real	8m18.265s
user	8m17.978s
sys	0m28.549s
```

main:
```sh
mongodb(52)
===============================================================
 ✓ mongodb (70) 2m
===============================================================
Versions executed

Folder: mongodb
	 * mongodb(10): 4.1.4, 4.6.0, 4.12.1, 5.0.1, 5.6.0, 6.1.0, 6.7.0, 6.13.1, 6.19.0, 7.2.0
===============================================================
PASS

real	2m12.783s
user	1m50.678s
sys	0m26.239s
```

This warrants investigation into what's slowing down the new instrumentation by 4x.

lib/subscribers/mongodb/config.js
amychisholm03: Claude helped me with the config / config-builder "optimization" to try to make it more readable, but I did not do a good job of going through the comments. I'll re-look at this.

lib/subscribers/mongodb/query.js
amychisholm03: It wasn't laid out as simply as this, but the old Shim code did have this behavior.

It was able to set `database_name: admin` without explicitly checking for it because it enabled `monitorCommands` and listened for `commandStarted`. This captured each command's db name from `evnt.databaseName`, and since `renameCollection` always runs against `admin`, that segment got the correct `admin` name automatically.

For us, we don't have the listener anymore, so we have to check for `rename` specifically.

lib/subscribers/mongodb/query.js
amychisholm03: Yeah, it's not really correct anymore because the `this` object differs slightly, but similar logic is in `findCollectionName` in `utils.js`

lib/subscribers/mongodb/config.js
amychisholm03: It was just referring to configs that are generated by these builder methods: collectionQueryConfig, cursorMethodConfig, dbOperationConfig

lib/subscribers/mongodb/operation.js
amychisholm03: Same reason as my comment on `query.js`.

bizob2828:
Superseded by #4059

PR #4040, opened 2026-05-28
feat: Added support for native Next.js otel instrumentation via the Hybrid Agent

### LINKED ISSUE #4027: Next.js hybrid agent support
### Description
Next.js provides native otel instrumentation.  We have difficulty keeping up with Next.js and we'd like to deprecate our own instrumentation and leverage the otel instrumentation.  I took a look and their otel instrumentation is not compliant with the version of the semantic conventions they state they support.  I have [PRd](https://github.com/vercel/next.js/pull/92007) the project but they are not responsive.  Instead, we opted to update our transformation rules to handle Next.js.

### Acceptance Criteria
Native Next.js otel instrumentation properly synthesizes server spans and collects the appropriate attributes and metrics so that we can announce deprecation of our Next.js instrumentation.

bizob2828:
## Description

This PR adds a transformation rule to the hybrid agent that is specifically for Next.js. 
I'm debating on adding versioned tests but we have said not to do so for otel instrumentation.

For customers to use Next.js otel support it's very easy:

In their newrelic.js
```
{
opentelemetry: {
    enabled: true
},
instrumentation: {
    next: {
        enabled: false
    },
    http: {
        enabled: false
    },
    // if you're doing native `fetch` calls
    // you must disable undici instrumentation
    // as Next.js has wrapped `fetch` and create client spans
    // if you do not disable undici instrumentation you will have 
    // duplicate client spans in your traces.
    undici: {
        enabled: false
    }
}
```

I also noticed when testing that all traces had `Truncated/ROOT` segments. This is because we are never touching the root in the hybrid
agent when we create a transaction. I added a `transaction.trace.root.touch()` to fix this.

Regarding how to announce deprecation of our Next.js instrumentation. My thought is for customers using 16+ of Next.js, just enable the hybrid agent.
We only run our versioned tests on Next 14 and 15. We can also update our examples to use the hybrid agent once this ships and update our docs to tell
Next.js customers to use hybrid agent. We can also put in the release notes that Next.js instrumentation is deprecated and will be removed in a future release.

What I still want to dig into is will this work when deploying to Vercel.  We had issues loading our agent and wanted to look into allowing our agent to be bundled.

## Related Issues
Closes #4027


lib/otel/traces/transformation-rules/101-OtelHttpServerNextjs1_20.json
```diff
@@ -0,0 +1,45 @@
...
+    }
+  ],
+  "transaction": {
+    "type": "web",
+    "name": {
+      "verb": "http.method",
+      "path": "http.route"
+    },
+    "url": {
+      "template": "http://localhost${http.target}"
```
bizob2828, line 42: seed a fully qualified URL as next js only provided `http.target` which is the URI + query params, this is just used [here](https://github.com/newrelic/node-newrelic/blob/31b9c8416116c8b04293f53335540729e934c768/lib/otel/traces/span-processor.js#L299) and the pathname is only used

jsumners-nr (approved):
Fabulous! A rebase on main will fix codecov.

Regarding, support. Did we ever claim actual support for Next.js? Or has it always been experimental/best effort? If the later, I don't think we need a major version to fully remove our old way. This will get customers better, direct, support. 

If we want to make exceptions for versioned tests, that seems fair. But we would really just be verifying that upstream does what they advertise they do.

PR #4041, opened 2026-05-29
chore: Added proxy config support to OTEL metrics

### LINKED ISSUE #3998: Add proxy support for OTEL metrics
Spec PR 795 adds the requirement that our OTEL metrics feature honor the agent's configured proxy settings. To accomplish this, we must update the configuration for the metrics exporter at:

https://github.com/newrelic/node-newrelic/blob/b710bacff2231c8038b792ac4c7a570991566835/lib/otel/metrics/index.js#L99-L105

According to a conversation in the [CNCF Slack](https://cloud-native.slack.com/archives/C01NL1GRPQR/p1779110828543569), the appropriate way to do this is to provide a factory function under the `httpAgentOptions` field for the configuration object. The function should return a [`http.Agent`](https://nodejs.org/docs/latest/api/http.html#class-httpagent) (or the `https` variant) compliant instance. This factory function accepts a protocol string so that the correct agent type can be constructed. It is important that the factory function import the appropriate modules lazily, so that the imports do not interfere with instrumentation.

A basic outline for the function looks like:

```js
module.exports = function httpAgentFactory(protocol) {
  const Constructor = protocol === 'http'
    : require('http').Agent
    ? require('https').Agent
  return new Constructor(/* stuff */)
}
```

I recommend that we provide a `defaultAgentFactory` to the `SetupMetrics` constructor so that we can easily unit test the feature.

https://github.com/newrelic/node-newrelic/blob/b710bacff2231c8038b792ac4c7a570991566835/lib/otel/metrics/index.js#L22

Otherwise, or additionally, we will need to verify the feature as we do with our core implementation:

https://github.com/newrelic/node-newrelic/blob/b710bacff2231c8038b792ac4c7a570991566835/test/smoke/proxy-api-connection-ssl.test.js#L20-L64

Note: the actual implementation will need to utilize the `https-proxy-agent` module as we do for our core agent:

https://github.com/newrelic/node-newrelic/blob/b710bacff2231c8038b792ac4c7a570991566835/lib/collector/http-agents.js#L69-L71

jsumners-nr:
This PR resolves #3998.

### DIFF
--- a/lib/otel/metrics/generate-proxy-agent-factory.js
+++ b/lib/otel/metrics/generate-proxy-agent-factory.js
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const defaultLogger = require('../../logger').child({
+  component: 'opentelemetry-metrics-proxy-factory'
+})
+const { buildProxyUrl, proxyAgent } = require('#agentlib/collector/http-agents.js')
+
+/**
+ * Builds a factory function compliant with OTEL's requirements for supplying
+ * an `http.Agent` instance.
+ *
+ * @param {object} params Function parameters.
+ * @param {AgentConfig} params.agentConfig Agent configuration instance.
+ * @param {object} params.logger Agent logger instance.
+ *
+ * @returns {Function} Factory function for OTEL.
+ */
+module.exports = function generateProxyAgentFactory({
+  agentConfig,
+  logger = defaultLogger
+}) {
+  return function httpAgentFactory(protocol) {
+    if (protocol.toLowerCase() === 'http') {
+      // Our collector does not support HTTP. But we return a default
+      // implementation here because it's not clear what will come from the
+      // OTEL library and what we need to pass back in order to skip this
+      // option.
+      //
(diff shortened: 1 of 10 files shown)


test/integration/otel/metrics.test.js
jsumners-nr, line 1: I migrated common code, creating a server and parsing the protobuf data, out into a re-usable file. It wasn't necessary before, but now that we have multiple tests doing roughly the same thing, it makes sense.

lib/otel/metrics/generate-proxy-agent-factory.js
```diff
@@ -0,0 +1,51 @@
...
+ * @param {object} params.logger Agent logger instance.
+ *
+ * @returns {Function} Factory function for OTEL.
+ */
+module.exports = function generateProxyAgentFactory({
+  agentConfig,
+  logger = defaultLogger
+}) {
+  return function httpAgentFactory(protocol) {
+    if (protocol.toLowerCase() === 'http') {
```
bizob2828: I'm not sure I understand this. the host we're sending to is `https`. Our proxy supports providing a proxy url with `http`. you can see that in `test/unit/collector/api-connect.test.js`.  I would think this logic goes away and just has the else bit where it uses `lib/collector/http-agents.js` to create a proxy

bizob2828 (changes requested):

bizob2828 (approved):

Issue #817, opened 2021-07-22
Update versioned test package.json engine fields to drop Node 10


carlo-808:
Update versioned test package.json engine fields to drop Node 10, set to >=12

example: https://github.com/newrelic/node-newrelic/blob/main/test/versioned/mongodb/package.json#L8

PR #4044, opened 2026-06-01
feat(kafka): add MessageBroker/Kafka/Cluster/{id}/Produce|Consume/{topic} metrics

shashank-reddy-nr:
## Why

Customers running **self-hosted Kafka** can't see producer → topic → consumer topology in New Relic - there's no relationship linking a producer APM service, a Kafka topic, and a consumer service.

The blocker is that today's instrumentation keys off broker address (`MessageBroker/Kafka/Nodes/{broker_address}/...`). That's unique by luck in managed Kafka (MSK, Confluent Cloud), but self-hosted clusters routinely reuse identical broker addresses across VPCs (e.g. `kafka-broker-1:9092` in both prod and staging), so address can't reliably join a service to a specific cluster's topic.

The fix is Kafka's native **cluster UUID** ([KIP-78](https://cwiki.apache.org/confluence/display/KAFKA/KIP-78%3A+Cluster+Id)): stable, globally unique (even for self-hosted), and already present in the broker metadata response. Emitting it in the metric name gives the APM metric and the existing Kafka Topic entity (`kafka.cluster.id` tag, from OTel infra / nri-kafka) a shared key, so a relationship rule can join producer → topic → consumer across services.

Purely additive — no existing metrics, distributed-tracing behavior, or instrumentation is changed or removed.

## Summary

Adds two new agent-level timeslice metrics for Kafka produce/consume:

- `MessageBroker/Kafka/Cluster/{clusterId}/Produce/{topic}`
- `MessageBroker/Kafka/Cluster/{clusterId}/Consume/{topic}`

These complement the existing `Message/Kafka/Topic/Named/{topic}/Received/*` metrics by adding the cluster UUID dimension, enabling cluster-level traffic segmentation in NRDB.

## How it works

**Cluster ID capture (optimized)** — kafkajs builds a brand-new internal `Cluster` object on every single `.producer()`/`.consumer()` call — it isn't reachable from the top-level `Kafka` client. Two new subscribers (`producer-cluster-capture.js`, `consumer-cluster-capture.js`) hook kafkajs's internal factory modules (`src/producer/index.js`, `src/consumer/index.js`) via this repo's existing declarative instrumentation config, capturing that `Cluster` reference the moment kafkajs itself creates it and stashing it on the returned producer/consumer instance via a new `kafkaCluster` Symbol.

This replaces the original approach of resolving the cluster ID via a separate KafkaJS admin client (`describeCluster()`) with its own TTL cache and in-flight-dedup map. The new read (`utils/read-cluster-id.js` → `cluster.brokerPool.metadata.clusterId`) is synchronous, in-memory, and best-effort — no extra connection, no cache, no TTL needed, since kafkajs keeps this field current on its own as part of normal operation. This is the same free-read approach used by upstream OpenTelemetry's kafkajs instrumentation. On a producer's very first send (before kafkajs has fetched any metadata), the cluster id may simply be unavailable yet — expected, and it resolves on the next send.

**Metric recording** - Uses `agentMetrics.getOrCreateMetric(...).incrementCallCount()` so metrics aggregate at agent level, not per-transaction. Produce metrics fire both inside and outside active transactions. Consume metrics fire from `record-data-metrics.js` and from the `eachBatch` wrapper.

**Metric name format + why no aggregation rework was needed** — per [agent-specs #817](https://source.datanerd.us/agents/agent-specs/pull/817) review, the metric name now matches the `Nodes` shape (verb before topic, not after): `Cluster/{id}/Produce/{topic}`, not `Cluster/{id}/Topic/{topic}/Produce`. Separately, the spec no longer requires per-message emission — these metrics exist to power entity relationships, not to report exact throughput, so counts just need to be aggregated over the same interval as `Nodes`. No further code change was needed for that: `getOrCreateMetric(...).incrementCallCount()` already merges into the `MetricAggregator`'s in-memory table and only flushes once per 60s harvest — the same mechanism the existing `Nodes` metric already uses. Calling it once per message only costs a cheap in-memory counter bump; the data New Relic receives was already one aggregated count per harvest interval, not one per message.

## How to Test

**Unit tests**

```
npm test test/unit/subscribers/kafkajs/client-constructor.test.js
npm test test/unit/subscribers/kafkajs/producer-cluster-capture.test.js
npm test test/unit/subscribers/kafkajs/consumer-cluster-capture.test.js
npm test test/unit/subscribers/kafkajs/utils/read-cluster-id.test.js
npm test test/unit/subscribers/kafkajs/utils/record-cluster-produce-metrics.test.js
```

**Manual end-to-end**

1. Enable both feature flags — kafkajs instrumentation as a whole, and the new cluster metrics specifically (each defaults to off): `NEW_RELIC_FEATURE_FLAG_KAFKAJS_INSTRUMENTATION=true NEW_RELIC_FEATURE_FLAG_KAFKA_CLUSTER_METRICS=true`
2. Run a producer and consumer against any Kafka cluster.
3. In New Relic, confirm metric delivery:

```nrql
FROM Metric SELECT uniques(metricTimesliceName)
WHERE metricTimesliceName LIKE 'MessageBroker/Kafka/Cluster/%'
SINCE 10 minutes ago
```

You should see `MessageBroker/Kafka/Cluster/{uuid}/Produce/{topic}` and `…/Consume/{topic}` entries carrying the real Kafka cluster UUID.

**Validated end-to-end** against a self-hosted Kafka cluster (SASL/PLAIN, two-broker setup). All three topics' Produce metrics appeared with the correct cluster UUID (pre-rename; format updated since — see note above).

## Bug fixes

`recordClusterProduceMetrics` ran before kafkajs's own argument validation and
indexed into `data.messages`/`topicMessage.messages` unconditionally. A
malformed `send()`/`sendBatch()` call (missing `messages`) threw synchronously
out of the instrumented call site instead of failing the way kafkajs itself
would (via its own promise rejection). Added guards so a missing/malformed
`messages` array is skipped rather than throwing; added corresponding tests.

## Related Issues

Jira: [NR-572216](https://new-relic.atlassian.net/browse/NR-572216)

[NR-572216]: https://new-relic.atlassian.net/browse/NR-572216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bizob2828:
@shashank-reddy-nr do you mind sharing the context of this PR.  I see it's adding a bunch of attributes and metrics but what is driving this?

shashank-reddy-nr:
Hi @bizob2828, I made code changes to add a couple of new Kafka metrics to establish a relationship between our self-hosted Kafka and the APM. Here is the [work request](https://new-relic.atlassian.net/browse/NR-572216) I raised for the APM team to review the changes and design, which has the complete context and explains why this is needed.

jsumners-nr:
Please update the PR description here with details on why this PR is needed and how this PR solves those needs.

lib/subscribers/kafkajs/client-constructor.js
jsumners-nr, line 1: Why are the changes in this file not part of the exported class? Are the two caches really necessary? What is the expected size of these caches in a typical, or extreme, deployment scenario?

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -78,13 +188,34 @@ module.exports = class ConstructorSubscriber extends Subscriber {
     const { arguments: args, self: client } = data
     client[kafkaCtx] = { brokers: args[0].brokers ?? ['none'] }
 
+    _fetchAndCacheClusterId(client, client[kafkaCtx].brokers).then((id) => {
+      if (id) {
+        // Update the shared client context. Producers share this object by reference so they
+        // see the update immediately. Consumers spread-copy the context at creation time, so
```
jsumners-nr: Where is this "spread-copy" happening? Consumers should not have access to our internal symbol and context object.

lib/subscribers/kafkajs/utils/record-data-metrics.js
```diff
@@ -45,4 +49,54 @@ module.exports = function recordDataMetrics({ data, kafkaCtx, tx }) {
...
+      kafkaCtx.groupId
+    )
+  }
+  const rawNr = data?.message?.headers?.newrelic
+  if (rawNr) {
+    try {
+      const encoded = Buffer.isBuffer(rawNr) ? rawNr.toString() : String(rawNr)
+      const payload = JSON.parse(Buffer.from(encoded, 'base64').toString())
+      const accountId = payload?.d?.ac
+      const appId = payload?.d?.ap
```
jsumners-nr: I don't understand. Where is this coming from? It says "rawNr" which I assume means "New Relic." Do we not have control over this data? If so, why are we not storing it in an easier, less expensive to read, manner?

test/unit/transaction/trace/segment.test.js
```diff
@@ -242,8 +242,12 @@ test('TraceSegment', async (t) => {
     sinon.stub(segment.timer, 'softEnd').returns(true)
     sinon.stub(segment.timer, 'endsAfter').returns(true)
 
-    // Make root duration calculation predictable
+    // Make root duration calculation predictable.
+    // Set hrDuration to 0 so getDurationInMillis() returns 0 — this ensures
+    // _computeTotalTime's condition (computedDuration > currentDuration) holds and
+    // the root gets updated to exactly 4ms (1001+3-1000).
     root.timer.start = 1000
+    root.timer.hrDuration = [0, 0]
```
jsumners-nr: Where did this change come from? How are the Kafaka subscriber changes related to the timing of segments?

test/versioned/kafkajs/kafka.test.js
```diff
@@ -443,3 +443,77 @@ test('consume batch inside of a transaction', async (t) => {
 
   await plan.completed
 })
+
+// ---------------------------------------------------------------------------
+// Cluster-ID metric and header tests
+// ---------------------------------------------------------------------------
```
jsumners-nr: Completely useless comment.

bizob2828: yolo'd AI slop

shashank-reddy-nr: Sorry @jsumners-nr  I'm new to this repo and used Claude to help implement the idea, which introduced several things that had no business being in this PR. I will cleanup all manually and test again them properly.

test/unit/transaction/trace/segment.test.js
shashank-reddy-nr: It didn't belong here at all - it was an unrelated change that crept in from a local debugging session. Reverted in the latest commit. Sorry for the confusion.

lib/subscribers/kafkajs/utils/record-data-metrics.js
shashank-reddy-nr: You're right - that block should never have been there. Apologies for the noise.

lib/subscribers/kafkajs/client-constructor.js
shashank-reddy-nr: Hi @jsumners-nr, I'm new to this repo and leaned heavily on AI assistance to implement this, so I apologize upfront for anything that doesn't follow conventions here.

The two caches serve different lifecycles:

- `cluster-id-cache.js` is a module-level map (brokers string → cluster UUID). It's intentionally process-global because the cluster UUID is the same for every producer/consumer connecting to the same Kafka cluster - it makes no sense to refetch it per client instance.
- `_clusterIdByInstance` doesn't exist here - the only per-instance storage is via the existing kafkaCtx symbol already set on each client/consumer.

In a typical deployment you'd have 1–3 distinct Kafka clusters, so the cache holds 1–3 entries. In an extreme multi-tenant scenario (many distinct broker strings) it could grow unbounded - that's a fair concern. Happy to add a size cap or TTL if you'd prefer. 

Open to moving the fetch logic inside the class if that's the convention - I'll follow your guidance. Can you please guide me?

lib/feature_flags.js
```diff
@@ -12,7 +12,7 @@ exports.prerelease = {
   internal_test_only: false,
   reverse_naming_rules: false,
   unresolved_promise_cleanup: true,
-  kafkajs_instrumentation: false,
+  kafkajs_instrumentation: true,
```
jsumners-nr: This should not be changed in this PR. We can open a new issue to discuss if this is ready to be enabled by default, but that's separate from this work.

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -335,12 +384,21 @@ module.exports = class ConstructorSubscriber extends Subscriber {
...
+          topic: batch.topic,
           producer: false
         })
+        // Emit one cluster Consume metric per message in the batch.
+        const clusterId = instance[kafkaCtx]?.clusterId ?? getClusterIdFromCache(instance[kafkaCtx]?.brokers)
+        if (clusterId) {
+          const m = self.agent.metrics.getOrCreateMetric(
+            `MessageBroker/Kafka/Cluster/${clusterId}/Topic/${batch.topic}/Consume`
+          )
+          for (let i = 0; i < batch.messages.length; i++) m.incrementCallCount()
```
jsumners-nr: Why are we using a loop instead of incrementing the count by the value of the length?

https://github.com/newrelic/node-newrelic/blob/185bb66c6b8c8ddc26c934adccc28bf47e9ddfd9/lib/stats/index.js#L77-L82

lib/subscribers/kafkajs/client-constructor.js
jsumners-nr: There will only be one instance of `ConstructorSubscriber` for the duration of an application's lifecycle. It can manage any required caches.

If it is possible for the cache to grow infinitely, then a size limited LRU cache should be used. We do not currently include such cache functionality. So if this is really a necessary feature, we will need to implement `lib/lru-map-cache.js`.

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -14,6 +14,33 @@ const { kafkaCtx } = require('#agentlib/symbols.js')
...
+ * Records MessageBroker/Kafka/Cluster cluster-level produce metrics.
+ * For send() records one metric for the single topic.
+ * For sendBatch() records one metric per distinct topic in the batch.
+ *
+ * @param {object} metrics The agent metrics aggregator.
+ * @param {string} clusterId Kafka cluster UUID.
+ * @param {boolean} batch Whether this is a sendBatch call.
+ * @param {object} data The send/sendBatch arguments object.
+ */
+function recordClusterProduceMetrics(metrics, clusterId, batch, data) {
```
jsumners-nr: Why isn't this function defined a file within the `utils/` directory?

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -76,15 +103,28 @@ module.exports = class ConstructorSubscriber extends Subscriber {
   end(data, ctx) {
     const self = this
     const { arguments: args, self: client } = data
-    client[kafkaCtx] = { brokers: args[0].brokers ?? ['none'] }
+    const rawBrokers = args[0]?.brokers
+    client[kafkaCtx] = { brokers: typeof rawBrokers === 'function' ? [] : (rawBrokers ?? ['none']) }
+
+    _fetchAndCacheClusterId(client, client[kafkaCtx].brokers).then((id) => {
```
jsumners-nr: Why does this function have a leading `_`? Typically, in JavaScript, the convention is that things with a leading `_` are to indicate they are private to the containing entity. This convention is really not necessary any longer, and it's not clear to me why it would have ever been needed for this function.

Also, why is it promise based? The `end` function is a synchronous function. By using a promise here, we are introducing activity happening outside of the context in which `end` is invoked.

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -76,15 +103,28 @@ module.exports = class ConstructorSubscriber extends Subscriber {
   end(data, ctx) {
     const self = this
     const { arguments: args, self: client } = data
-    client[kafkaCtx] = { brokers: args[0].brokers ?? ['none'] }
+    const rawBrokers = args[0]?.brokers
+    client[kafkaCtx] = { brokers: typeof rawBrokers === 'function' ? [] : (rawBrokers ?? ['none']) }
```
jsumners-nr: What is this change doing? I don't see `rawBrokers` documented anywhere on https://kafka.js.org/docs/configuration as being an option for the client constructor.

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -76,15 +103,28 @@ module.exports = class ConstructorSubscriber extends Subscriber {
...
     const origConsumer = client.consumer
     client.consumer = function nrConsumer(...args) {
       const consumer = origConsumer.apply(client, args)
-      consumer[kafkaCtx] = client[kafkaCtx]
+      // Capture clientId at consumer creation so it is available on every
+      // transaction without waiting for the async REQUEST event.
+      consumer[kafkaCtx] = {
+        ...client[kafkaCtx],
+        clientId: args[0]?.clientId ?? client[kafkaCtx]?.clientId ?? null
+      }
```
jsumners-nr: Please remove the copy operation. It's cheaper and cleaner to:

```js
consumer[kafkaCtx] = client[kafkaCtx]
consumer[kafkaCtx].clientId = one ? or : the_other
```

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -76,15 +103,28 @@ module.exports = class ConstructorSubscriber extends Subscriber {
...
 
       consumer.on(consumer.events.REQUEST, function nrListener(data) {
-        consumer[kafkaCtx].clientId = data?.payload?.clientId
+        // REQUEST event may refine clientId from the live connection payload
+        if (data?.payload?.clientId) {
+          consumer[kafkaCtx].clientId = data.payload.clientId
+        }
```
jsumners-nr: This pattern is what the nullish coalescing assignment operator is for:

```js
consumer[kafkaCtx] ??= data?.payload?.clientId
```

lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -174,6 +214,10 @@ module.exports = class ConstructorSubscriber extends Subscriber {
           'Not recording consumer function %s, not in a transaction',
           methodName
         )
+        if (instance[kafkaCtx].clusterId) {
+          const [data] = args
+          recordClusterProduceMetrics(self.agent.metrics, instance[kafkaCtx].clusterId, batch, data)
+        }
```
jsumners-nr: We are not in a transaction here. What is the purpose of recording metrics in this case?

jsumners-nr (commented):
A few more comments subsequent to the revised changes.

Is there a spec PR number to review that defines what these changes are trying to accomplish?

lib/subscribers/kafkajs/client-constructor.js
shashank-reddy-nr: Addressed in the latest commit (`3b6e628`): the two maps are now `#clusterIdCache` and `#clusterIdInFlight` private fields on `ConstructorSubscriber` itself.

On LRU vs hard cap: I opted for a simpler `MAX_CACHE_SIZE = 128` constant rather than a full LRU implementation. The rationale is that each cache entry represents one *distinct broker list* (i.e. one Kafka cluster). In practice, a single Node.js process connects to a small number of clusters — typically 1–3, rarely more than 10. 128 is already generous for any realistic deployment. If the cap is ever hit, new broker-sets simply do not get cached (we silently skip the `cache.set`) but still fetch the cluster ID on first call and return it — so the metric still fires, we just do not avoid the re-fetch for that broker-set.

If you would prefer a true LRU, I am happy to implement `lib/lru-map-cache.js` — just let me know and I will add it as part of this PR.

lib/subscribers/kafkajs/client-constructor.js
shashank-reddy-nr: These use `self.agent.metrics` — the agent-level metric aggregator — not the per-transaction `tx.metrics`. The distinction matters: Kafka producers frequently run outside of any active web/background transaction (fire-and-forget background producers, CLI scripts, etc.). Using the global aggregator means the cluster-level produce count is always captured regardless of transaction context.

The consume side already does the same thing for the same reason: `agentMetrics.getOrCreateMetric(...)` in `record-data-metrics.js:50`. This PR follows the same pattern for the produce side.

lib/subscribers/kafkajs/client-constructor.js
shashank-reddy-nr: Both addressed in the latest commit:

- The leading `_` is gone.
- The function is now in `lib/subscribers/kafkajs/utils/cluster-id-cache.js`.

On the async shape: `AdminClient.connect()` and `AdminClient.describeCluster()` are inherently async network operations, so there is no synchronous alternative. We fire-and-forget from `end()` and set `client[kafkaCtx].clusterId` when the promise resolves. The first few sends before the cluster ID arrives will not record the cluster metric, but that is intentional best-effort behaviour — the metric fires for every send after the ID is populated.

lib/feature_flags.js
shashank-reddy-nr: This was touched for testing, I reverted this change.

lib/subscribers/kafkajs/client-constructor.js
shashank-reddy-nr: Fixed: `record-cluster-produce-metrics.js` now calls `.incrementCallCount(data.messages.length)` and `.incrementCallCount(topicMessage.messages.length)` directly — no loop.

lib/subscribers/kafkajs/client-constructor.js
shashank-reddy-nr: Moved to `lib/subscribers/kafkajs/utils/cluster-id-cache.js`. It exports two pure functions (`getClusterIdFromCache`, `fetchAndCacheClusterId`) that accept the cache maps as parameters so they remain independently testable.

lib/subscribers/kafkajs/client-constructor.js
shashank-reddy-nr: Removed. The line now reads:
```js
const brokers = args[0]?.brokers
client[kafkaCtx] = { brokers: typeof brokers === 'function' ? [] : (brokers ?? ['none']) }
```
This handles both array brokers and the function-based brokers form that kafkajs supports (if brokers is a function we store `[]` and skip the admin fetch, since we cannot sort/key by a function).

lib/subscribers/kafkajs/client-constructor.js
shashank-reddy-nr: Fixed — shared reference assignment now:
```js
consumer[kafkaCtx] = client[kafkaCtx]
consumer[kafkaCtx].clientId = args[0]?.clientId ?? client[kafkaCtx]?.clientId ?? null
```
This also means `clusterId` populated asynchronously on `client[kafkaCtx]` is visible on `consumer[kafkaCtx]` immediately.

lib/subscribers/kafkajs/client-constructor.js
shashank-reddy-nr: Fixed — updated to `consumer[kafkaCtx].clientId ??= data?.payload?.clientId`.

shashank-reddy-nr:
> ments subsequent to the revised changes.
> 
> Is there a spec PR number to review that defines what these changes are trying to accomplish?

@jsumners-nr I just raised a PR for agent kafka spec update and I will get it approved from APM agent board and let you know to review this PR post approval. For reference here is the [#817](https://source.datanerd.us/agents/agent-specs/pull/817).

lib/feature_flags.js
```diff
@@ -13,6 +13,7 @@ exports.prerelease = {
   reverse_naming_rules: false,
   unresolved_promise_cleanup: true,
   kafkajs_instrumentation: false,
+  kafka_cluster_metrics: false,
```
bizob2828, line 16: For future eyes. The spec states `be gated behind an opt-in configuration `kafka.metrics.cluster.metrics.enabled` (disabled by default)`. This is a feature flag, not configuration. it should live in `lib/config/default.js`

bizob2828:
Thanks for this PR @shashank-reddy-nr, but we plan on addressing this in our own PR. We will use this to help drive the new PR though

PR #4049, opened 2026-06-05
feat: Add `span_id` and `trace_id` labels to `CpuProfiler` sample data

### LINKED ISSUE #3964: Add trace/span id into CPU profiler
For the CPU profiling data, we need to inject the trace and span IDs.

We need to spike on how this should be implemented. How do we inject the IDs (alter the cpu profiling samples or send something along side the cpu profiling data)?

amychisholm03:
## Description

This PR tags each `pprof` sample with the active code's `span_id` and `trace_id`, attached as two separate `pprof` labels: `"span_id": <span id>` and `"trace_id": <trace id>`.

We capture the active ids by setting `withContexts: true` in `pprof` and hooking the agent's ALS instance. We wrap the instance rather than the ALS prototype, so other libraries aren't affected.

How those ids get carried per sample depends on the runtime:
- **AsyncContextFrame active (Node 24+):** `pprof` carries the context per async-context-frame via CPED. We hook `enterWith` and write the active ids straight into the current frame.
- **Older Node:** `pprof` captures a context holder by reference, so we wrap the ALS instance's `run` and reset the holder's `current` to the active ids on each context switch.

## How to Test

```
npm run integration
```

## Related Issues

Closes #3964

### DIFF
--- a/lib/profiling/profilers/cpu.js
+++ b/lib/profiling/profilers/cpu.js
@@ -4,13 +4,34 @@
  */
 
 'use strict'
+const { AsyncLocalStorage } = require('async_hooks')
 const BaseProfiler = require('./base')
+
+/**
+ * True when V8's AsyncContextFrame is active, detected by whether
+ * `AsyncLocalStorage.run` delegates to `enterWith`. Feature-detected rather than
+ * version-gated so it tracks `--experimental-async-context-frame` (Node 22/23)
+ * and `--no-async-context-frame` (Node 24+).
+ */
+const ASYNC_CONTEXT_FRAME_ACTIVE = (() => {
+  let active = false
+  const als = new AsyncLocalStorage()
+  als.enterWith = () => {
+    active = true
+  }
+  als.run(1, () => {})
+  als.disable()
+  return active
+})()
 
 /**
  * Produces a wall time stack profile via `pprof`.
  * The native sampler fires at a fixed rate against whatever JS is currently
- * on the stack, and the active context is tracked by hooking the agent's
- * `AsyncLocalStorage` (`this.#store`) instance.
+ * on the stack, and the active transaction's trace/span ids are attached to
+ * each sample by hooking the agent's `AsyncLocalStorage` (`this.#store`).
+ *
+ * With AsyncContextFrame active, pprof's `useCPED` carries the context per
+ * async-context-frame; otherwise a single holder is swapped per sample.
  *
  * The wall-clock (or just wall) time for a function measures the time elapsed
  * between entering and exiting a function. Wall time includes all wait time,
@@ -22,26 +43,33 @@ class CpuProfiler extends BaseProfiler {
   #durationMillis
   #intervalMicros = (1e3 / 99) * 1000 // samples at 99hz(99 times per second)
   #kSampleCount
-  // The "context" reference handed to the pprof time profiler. The native
-  // sampler captures whatever `current` holds at the moment a sample is taken,
-  // so we keep it pointed at the active transaction's trace/span ids.
+  #useCPED
+  // (non-CPED) pprof captures this holder by reference at sample time, so we
+  // keep `current` pointed at the active transaction's trace/span ids.
   #context = this.#defaultContext()
   #state
   #lastSampleCount = 0
-  // The agent's `AsyncLocalStorage` instance we use to get the active context.
+  // The agent's `AsyncLocalStorage` instance.
   #store
-  // Caches the resolved transaction name so `getFullName()` (which re-runs the
-  // name normalizer while a transaction is in-flight) isn't called on every
-  // context change.
+  // Caches the resolved transaction name; `getFullName()` re-runs the name
+  // normalizer while a transaction is in-flight, so avoid it per context change.
   #nameCache = { transaction: null, name: '' }
+  // Saved ALS methods, restored by `#unhookContext` on stop.
+  #origRun
+  #origEnterWith
+  // (non-CPED) async hook that refreshes context on async resumption.
+  #asyncHook
 
-  constructor({ logger, samplingInterval, tracer }) {
+  constructor({ logger, samplingInterval, tracer, useCPED = true }) {
     super({ logger })
     this.#pprof = require('@datadog/pprof')
     this.#kSampleCount = this.#pprof.time.constants.kSampleCount
     this.#tracer = tracer
     this.#store = tracer._contextManager._asyncLocalStorage
     this.#durationMillis = samplingInterval
+    // CPED requires AsyncContextFrame. A caller may force the holder strategy
+    // with useCPED: false, but can't force CPED on a runtime that lacks ACF.
+    this.#useCPED = useCPED && ASYNC_CONTEXT_FRAME_ACTIVE
   }
 
   start() {
@@ -55,11 +83,19 @@ class CpuProfiler extends BaseProfiler {
       durationMillis: this.#durationMillis,
       intervalMicros: this.#intervalMicros,
       // needed for trace_id + span_id label
-      withContexts: true
+      withContexts: true,
+      // carry the sample context per async-context-frame when ACF is active
+      useCPED: this.#useCPED
     })
-    this.#state = this.#pprof.time.getState()
-    this.#resetContext()
-    this.#hookContext()
+
+    if (this.#useCPED) {
+      this.#hookEnterWith()
+    } else {
+      this.#state = this.#pprof.time.getState()
+      this.#resetContext()
+      this.#hookRun()
+      this.#hookAsyncResume()
+    }
   }
 
   stop() {
@@ -68,21 +104,27 @@ class CpuProfiler extends BaseProfiler {
       return
     }
 
+    // Restore the wrapped ALS methods before stopping.
+    this.#unhookContext()
     this.#pprof.time.stop(false)
   }
 
   async collect() {
+    if (!this.#useCPED) {
+      // Flush the active span before sampling restarts.
+      this.#updateContext()
+    }
     const profile = this.#pprof.time.stop(true, this.#linkContext)
-    // `stop(true)` restarts sampling, so start a fresh context for the next
-    // collection cycle.
-    this.#resetContext()
+    if (!this.#useCPED) {
+      // `stop(true)` restarts sampling, so start a fresh holder.
+      this.#resetContext()
+    }
     return this.#pprof.encode(profile)
   }
 
   /**
-   * A fresh pprof context holder. Returns a new object on every call: pprof
-   * captures the holder by reference at sample time, so each swap needs its own
-   * object or already-captured samples would pick up later mutations.
+   * A fresh holder on every call: pprof captures it by reference, so each swap
+   * needs its own object or earlier samples would pick up later mutations.
    *
    * @returns {{ current: object }} a new, empty context holder
    */
@@ -91,8 +133,8 @@ class CpuProfiler extends BaseProfiler {
   }
 
   /**
-   * Points the pprof time profiler at a fresh `#context` and resyncs the
-   * sample counter.
+   * (non-CPED) Points the pprof time profiler at a fresh `#context` and resyncs
+   * the sample counter.
    */
   #resetContext() {
     this.#context = this.#defaultContext()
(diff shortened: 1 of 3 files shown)


lib/profiling/profilers/cpu.js
```diff
@@ -39,9 +72,111 @@ class CpuProfiler extends BaseProfiler {
...
+   */
+  #updateContext = () => {
+    const sampleCount = this.#state?.[this.#kSampleCount]
+    if (sampleCount !== this.#lastSampleCount) {
+      this.#lastSampleCount = sampleCount
+      this.#context = { current: {} }
+      this.#pprof.time.setContext(this.#context)
+    }
+
+    const ctx = this.#tracer.getContext()
```
bizob2828: we have a method on the agent that retrieves active trace/span `agent.getLinkingMetadata(true)`

lib/profiling/profilers/cpu.js
```diff
@@ -39,9 +72,111 @@ class CpuProfiler extends BaseProfiler {
...
+    this.#resetContext()
     return this.#pprof.encode(profile)
   }
+
+  /**
+   * Points the pprof time profiler at a fresh `#context` and resyncs the
+   * sample counter.
+   */
+  #resetContext() {
+    this.#context = { current: {} }
```
bizob2828: i'd keep the default context in some constant on class as it's being replicated in a few places within this file

lib/profiling/profilers/cpu.js
```diff
@@ -39,9 +72,111 @@ class CpuProfiler extends BaseProfiler {
...
+   *
+   * @param {Transaction} transaction the active transaction
+   * @returns {string} the label key
+   */
+  #spanKey(transaction) {
+    if (transaction !== this.#nameCache.transaction || !this.#nameCache.name) {
+      this.#nameCache = { transaction, name: transaction.getFullName() || '' }
+    }
+
+    const prefix = `span:${transaction.id}`
```
bizob2828: i don't understand this. this is not a span id

lib/profiling/profilers/cpu.js
```diff
@@ -39,9 +72,111 @@ class CpuProfiler extends BaseProfiler {
...
+  }
+
+  /**
+   * Wraps `run` on the agent's single `AsyncLocalStorage` so the pprof holder
+   * is refreshed on every context switch: on `run` entry (the agent propagates
+   * all transaction context through this one instance) and on `run` exit (so a
+   * sibling doesn't inherit a returned child's span). It wraps the instance,
+   * not the prototype, leaves other libraries' ALS untouched.
+   *
+   * @todo On Node 24+ (AsyncContextFrame is on by default), pprof's `useCPED: true`
```
bizob2828: I don't feel like this is a todo, it should be done. by using useCPED it drastically reduces the instrumentation.

test/integration/profiling/cpu.test.js
```diff
@@ -0,0 +1,179 @@
...
+const logger = { trace() {}, debug() {}, error() {} }
+
+/**
+ * Burns CPU synchronously for at least `ms` milliseconds so the time profiler
+ * has a chance to take samples while a context is active.
+ *
+ * @param {number} ms minimum time to burn
+ * @returns {number} an accumulated value, to keep the loop from being optimized away
+ */
+function burnCpu(ms) {
```
bizob2828, line 25: i think these tests are fine to assert trace/span id but I'd like to see a test, if possible that does many diff actions and the appropriate spans are being set. I don't think we have proof that this feature is working without it

lib/profiling/profilers/cpu.js
```diff
@@ -25,8 +53,13 @@ class CpuProfiler extends BaseProfiler {
...
+      intervalMicros: this.#intervalMicros,
+      // needed for trace_id + span_id label
+      withContexts: true
     })
+    this.#state = this.#pprof.time.getState()
+    this.#resetContext()
+    this.#hookContext()
   }
 
   stop() {
```
bizob2828, line 98: the async local instrumentation will run even if the profiler is stopped. you need some way to undo that code. this is a common use case as profiling is enabled/disabled via server side config

bizob2828 (changes requested):

lib/profiling/profilers/cpu.js
amychisholm03: `span` is just the prefix, I'm just following the format the Go agent is using: `'span:<first 16 of trace_id>:<transaction name>':  'span_id=…,trace_id=…'`. The `<transaction.id>:<transaction name>` is just the random unique identifier we decided on after `span` (which is the only part that the backend cares about).

bizob2828: I guess it's difficult to understand at the moment because we lack a specification

lib/profiling/profilers/cpu.js
```diff
@@ -4,13 +4,34 @@
...
+const { AsyncLocalStorage } = require('async_hooks')
 const BaseProfiler = require('./base')
 
+/**
+ * True when V8's AsyncContextFrame is active, detected by whether
+ * `AsyncLocalStorage.run` delegates to `enterWith`. Feature-detected rather than
+ * version-gated so it tracks `--experimental-async-context-frame` (Node 22/23)
+ * and `--no-async-context-frame` (Node 24+).
+ */
+const ASYNC_CONTEXT_FRAME_ACTIVE = (() => {
```
bizob2828: I don't think we have to be this thorough. we should just check `process.version >= 24`. If folks are using `--experimental-async-context-frame` in 22 so be it.

lib/profiling/profilers/cpu.js
```diff
@@ -4,16 +4,72 @@
...
   #pprof
+  #tracer
   #durationMillis
   #intervalMicros = (1e3 / 99) * 1000 // samples at 99hz(99 times per second)
-  constructor({ logger, samplingInterval }) {
+  #kSampleCount
+  #useCPED
+  // (non-CPED) pprof captures this holder by reference at sample time, so we
+  // keep `current` pointed at the active transaction's trace/span ids.
+  #context = this.#defaultContext()
```
bizob2828:
```suggestion
  DEFAULT_CONTEXT = { current: {} }
  #context = this.DEFAULT_CONTEXT
```

no need to call a function to set a constant

lib/profiling/profilers/cpu.js
```diff
@@ -35,13 +104,189 @@ class CpuProfiler extends BaseProfiler {
...
+
+  /**
+   * Refreshes the pprof sample context with the active span. With CPED the label
+   * is written straight into the current async-context-frame; otherwise a fresh
+   * holder is swapped in once a sample has landed (so earlier samples keep their
+   * ids) before reassigning its `current`.
+   */
+  #updateContext = () => {
+    if (this.#useCPED) {
+      this.#pprof.time.setContext(this.#buildLabel())
```
bizob2828: this will be quite expensive to do string concatenating many times a second. you should just store the object and then build the strings at collection time in #linkContext and do the label formatting there.

lib/profiling/profilers/cpu.js
```diff
@@ -35,13 +104,189 @@ class CpuProfiler extends BaseProfiler {
...
+   * transaction name (`span:<traceId[0:16]>:<name>`) to match other language
+   * agents' implementation.
+   *
+   * The backend only cares about the `span:` prefix. Anything after that is just
+   * a unique identifier format that other language agents have agreed upon.
+   *
+   * @param {Transaction} transaction the active transaction
+   * @returns {string} the label key
+   */
+  #spanKey(transaction) {
```
bizob2828: i know this was proposed but i don't understand it. we have to store more in the context for this and it doesn't provide any value. we should be storing `trace_id` and `span_id` as two labels where applicable for every given sample. I ran a test and this behaves as expected. there's no reason to have unique keys they are just references to some internal string table. unique values will be associated as expected

bizob2828 (changes requested):
I have a few comments. Have you done perf testing without this PR and with against a real app to ensure the overhead isn't a deal breaker? I also want us to push for 2 labels and not 1 with some unique key and then special formatting for the value. the intent of labels is to provide metadata for a given profile sample, the keys do not have to be unique.

lib/profiling/profilers/cpu.js
amychisholm03: What about for `--no-async-context-frame` 24+?

lib/profiling/profilers/cpu.js
amychisholm03: Consensus from DACI was that we'll proceed with the `span_id` and `trace_id` as two separate labels.

lib/profiling/profilers/cpu.js
```diff
@@ -5,15 +5,61 @@
...
+  constructor({ logger, samplingInterval, tracer, useCPED = true }) {
     super({ logger })
     this.#pprof = require('@datadog/pprof')
+    this.#kSampleCount = this.#pprof.time.constants.kSampleCount
+    this.#tracer = tracer
+    this.#store = tracer._contextManager._asyncLocalStorage
     this.#durationMillis = samplingInterval
+    // CPED requires AsyncContextFrame which is on by default with node >=24
+    // @todo should we account for `--no-async-context-frame`?
+    this.#useCPED = useCPED && semver.satisfies(process.versions.node, '>=24.0.0')
```
bizob2828: TIL. i've always used `process.version`

bizob2828: I get you're using `useCPED` for testing but wouldn't these just be covered with node 22 and then when we drop node 22 we no longer need this option?

lib/profiling/profilers/cpu.js
```diff
@@ -35,13 +94,147 @@ class CpuProfiler extends BaseProfiler {
...
+  }
+
+  /**
+   * Refreshes the pprof sample context with the active span. With CPED, the label
+   * is written straight into the current async-context-frame; otherwise a fresh
+   * holder is swapped in once a sample has landed (so earlier samples keep their
+   * ids) before reassigning its `current`.
+   */
+  #updateContext = () => {
+    const { 'trace.id': traceId, 'span.id': spanId } = this.#tracer.agent.getLinkingMetadata(true)
```
bizob2828: my understanding of CPED is that it's a native propagating of context.  So we really on have to set the context for a given active segment once.  this is reaching into the context manager and getting/setting the context every time, even for segments that it already retrieved, thus kind of defeating the purpose of CPED.  This may be overkill but here's an idea to optimize when to reached into our context manager and set the context on the pprof time package:

```
diff --git a/lib/profiling/profilers/cpu.js b/lib/profiling/profilers/cpu.js
index 9de3a2bec..68135e513 100644
--- a/lib/profiling/profilers/cpu.js
+++ b/lib/profiling/profilers/cpu.js
@@ -7,6 +7,18 @@
 const BaseProfiler = require('./base')
 const semver = require('semver')
 
+/**
+ * (CPED) Caches the per-sample context object on the active segment so repeated
+ * activations of the same span yield the same object identity, letting us skip
+ * redundant `setContext` calls (each of which allocates a native holder).
+ */
+const PROFILING_CONTEXT = Symbol('NewRelicProfilingContext')
+/**
+ * (CPED) Shared context installed when no span is active, so idle stretches
+ * don't churn native holders and don't inherit a prior span's ids.
+ */
+const EMPTY_CONTEXT = Object.freeze({})
+
 /**
  * Produces a wall time stack profile via `pprof`.
  * The native sampler fires at a fixed rate against whatever JS is currently on
@@ -140,24 +152,49 @@ class CpuProfiler extends BaseProfiler {
    * ids) before reassigning its `current`.
    */
   #updateContext = () => {
-    const { 'trace.id': traceId, 'span.id': spanId } = this.#tracer.agent.getLinkingMetadata(true)
-    const spanContext = { traceId, spanId }
-
     if (this.#useCPED) {
-      this.#pprof.time.setContext(spanContext)
+      this.#updateCpedContext()
       return
     }
 
+    const { 'trace.id': traceId, 'span.id': spanId } = this.#tracer.agent.getLinkingMetadata(true)
+    const spanContext = { traceId, spanId }
     const sampleCount = this.#state?.[this.#kSampleCount]
     if (sampleCount !== this.#lastSampleCount || this.#context === this.DEFAULT_CONTEXT) {
       this.#lastSampleCount = sampleCount
-      this.#context = spanContext ? { current: spanContext } : this.DEFAULT_CONTEXT
+      this.#context = { current: spanContext }
       this.#pprof.time.setContext(this.#context)
     } else {
       this.#context.current = spanContext
     }
   }
 
+  /**
+   * (CPED) Installs the active span's context into the current async-context-frame.
+   * The context object is cached on the segment, so re-entering the same span
+   * across async resumptions returns the same identity; when that object is
+   * already installed we skip `setContext`, which would otherwise allocate a fresh
+   * native context holder on every context switch.
+   */
+  #updateCpedContext() {
+    const segment = this.#tracer.getSegment()
+    let spanContext
+    if (segment) {
+      spanContext = segment[PROFILING_CONTEXT]
+      if (spanContext === undefined) {
+        const { 'trace.id': traceId, 'span.id': spanId } = this.#tracer.agent.getLinkingMetadata(true)
+        spanContext = { traceId, spanId }
+        segment[PROFILING_CONTEXT] = spanContext
+      }
+    } else {
+      spanContext = EMPTY_CONTEXT
+    }
+
+    if (this.#pprof.time.getContext() !== spanContext) {
+      this.#pprof.time.setContext(spanContext)
+    }
+  }
+
```

bizob2828 (changes requested):
A couple of comments, getting close though

lib/profiling/profilers/cpu.js
amychisholm03: True, though I was also trying to mock out behavior for `--no-async-context-frame`; if that is set then `useCPED` would be false and in node >=24, CPED wouldn't be used.

lib/profiling/profilers/cpu.js
amychisholm03: Yeah, this is an optimization that we could look into later, but I don't really see "getting/setting the context every time, even for segments that it already retrieved" being that big of an issue on performance. I had Claude run a benchmark test and it noted only 2% increase with the current approach.

bizob2828: it's not really an optimization, it's adhering to the design of CPED.  The gain with CPED is you only have to set the context once for a given async context frame as it doesn't change. Right now you're getting our context and setting it every time which is essentially the legacy model we have in here.

bizob2828 (approved):

PR #4051, opened 2026-06-08
chore: Added supportability metrics to OTEL metrics export

### LINKED ISSUE #4050: Add supportability metrics to OTEL metrics exporter
Spec PR 795 added requirements around retrying delivery of OTEL based metrics delivery and recording New Relic supportability metrics accordingly. The way we have implemented the OTEL based metrics feature gives us the retrying portion, but we need to update our interception of the `export` method to record the supportability metrics:

https://github.com/newrelic/node-newrelic/blob/6c04b8ab780006ff3e15406a56e9efafa79b1509/lib/otel/metrics/nr-proxying-delegate.js#L26-L31

1. If `result.code` is a success code, record `Supportability/Metrics/${agent}/OpenTelemetryBridge/export/success`
2. If `result.code` is a failure code, record `Supportability/Metrics/${agent}/OpenTelemetryBridge/export/failure`

We cannot record the `/retry` metric as the spec defines because we do not have insight into the actual retries:

<details>
<summary>Bot findings</summary>

Based on my analysis of the code, the export callback is invoked only after all retries have been attempted, not after each individual request.

  Here's how it works:

  1. RetryingTransport (retrying-transport.ts:42-88): The send() method performs up to 5 retry attempts in a loop when it receives a 'retryable'
  status. It only returns a final ExportResponse after either:
    - The export succeeds (status === 'success')
    - All retry attempts are exhausted
    - The retry timeout would exceed the deadline
    - A non-retryable error occurs
  2. OTLPExportDelegate (otlp-export-delegate.ts:84-131): The resultCallback is invoked in the .then() handler of the promise returned by
  this._transport.send(). Since the retrying transport completes all retry attempts before resolving its promise, the callback is only called once
   with the final result.
  3. Result handling (otlp-export-delegate.ts:111-117): Even if the final result after all retries is still 'retryable' status, the callback is
  invoked with ExportResultCode.FAILED.

  So to directly answer your question: The callback is invoked once, after all retries have been attempted, with the final export result (either
  success or failure).
</details>

A conversation in the [CNCF Slack](https://cloud-native.slack.com/archives/C01NL1GRPQR/p1780677756946239?thread_ts=1780602314.062499&cid=C01NL1GRPQR) confirms this is accurate. Thus, we can only record the success or failure metrics.

jsumners-nr:
This PR resolves #4050.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -241,32 +241,32 @@ jobs:
       - name: Download artifacts
         uses: actions/download-artifact@v8
       - name: Post Unit Test Coverage
-        # v6.0.1
-        uses: codecov/codecov-action@e79a6962e0d4c0c17b229090214935d2e33f8354
+        # v7.0.0
+        uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f
         with:
           fail_ci_if_error: true
           token: ${{ secrets.CODECOV_TOKEN }}
           directory: unit-tests-${{ matrix.node-version }}
           flags: unit-tests-${{ matrix.node-version }}
       - name: Post Integration CJS Test Coverage
-        # v6.0.1
-        uses: codecov/codecov-action@e79a6962e0d4c0c17b229090214935d2e33f8354
+        # v7.0.0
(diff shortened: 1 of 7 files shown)


bizob2828 (approved):
LGTM

PR #4059, opened 2026-06-09
chore: Converted mongodb instrumentation to subscriber based

### LINKED ISSUE #3288: Refactor `mongodb` to subscribe to events emitted

jsumners-nr:
This PR resolves #3288.

----

- [x] Review https://github.com/newrelic/node-newrelic/blob/main/lib/instrumentation/mongodb/v4-mongo.js to determine which classes were originally instrumented. This PR addresses the same classes via instrumenting the base classes and add at least one test to assert this.
- [x] Replicate the `commandStarted` listener in https://github.com/newrelic/node-newrelic/blob/main/lib/instrumentation/mongodb/v4-mongo.js
- [x] Review https://github.com/newrelic/node-newrelic/blob/main/lib/instrumentation/mongodb/common.js for explicitly instrumented methods that are not listed in the `constants.js`
    * https://github.com/newrelic/node-newrelic/blob/6c04b8ab780006ff3e15406a56e9efafa79b1509/lib/instrumentation/mongodb/common.js#L26 - the `each` method was available prior to v4. Between v4 and v7 it has been replaced with `forEach`. But that has been slated for removal post v7 in favor of `for ... await of`.
    * https://github.com/newrelic/node-newrelic/blob/6c04b8ab780006ff3e15406a56e9efafa79b1509/lib/instrumentation/mongodb/common.js#L27 - `pipe` was removed in v4 in favor of native streams (via `cursor.stream()`).
    * https://github.com/newrelic/node-newrelic/blob/6c04b8ab780006ff3e15406a56e9efafa79b1509/lib/instrumentation/mongodb/common.js#L78-L82 - I cannot find any current or previous reference to `Db.prototype.connect`.

-----

`npm run versioned mongodb` on this branch:

```
Finding all versions for a package
mongodb(53)
===============================================================
 ✓ mongodb (90) 5m
===============================================================
Versions executed

Folder: mongodb
         * mongodb(10): 4.1.4, 4.6.0, 4.12.1, 5.0.1, 5.6.0, 6.2.0, 6.8.2, 6.14.2, 6.20.0, 7.3.0
===============================================================
PASS

real    4m40.788s
user    4m18.294s
sys     0m50.182s
```

On `main`:

```
Finding all versions for a package
mongodb(53)
===============================================================
 ✓ mongodb (70) 3m
===============================================================
Versions executed

Folder: mongodb
         * mongodb(10): 4.1.4, 4.6.0, 4.12.1, 5.0.1, 5.6.0, 6.2.0, 6.8.2, 6.14.2, 6.20.0, 7.3.0
===============================================================
PASS

real    3m8.515s
user    2m38.790s
sys     0m42.052s
```

A few tests have been added in this PR to cover some existing gaps.

-----

`npm run versioned:major mongodb` on this branch:

```
Finding all versions for a package
mongodb(4)
===============================================================
 ✓ mongodb (36) 2m
===============================================================
Versions executed

Folder: mongodb
         * mongodb(4): 4.17.2, 5.9.2, 6.21.0, 7.3.0
===============================================================
PASS

real    2m0.250s
user    1m46.087s
sys     0m20.653s
```

On `main`:

```
Finding all versions for a package
mongodb(4)
===============================================================
 ✓ mongodb (28) 1m
===============================================================
Versions executed

Folder: mongodb
         * mongodb(4): 4.17.2, 5.9.2, 6.21.0, 7.3.0
===============================================================
PASS

real    1m11.027s
user    1m4.293s
sys     0m16.939s
```


lib/subscribers/mongodb/cursor.js
```diff
@@ -0,0 +1,128 @@
...
+  constructor({ agent, logger }) {
+    super({
+      agent,
+      logger,
+      channelName: 'nr_cursor',
+      packageName: 'mongodb',
+      system: MONGODB_METRIC_CONSTANTS.PREFIX
+    })
+
+    this.events = ['end']
```
bizob2828: add `this.opaque = true` to avoid creating child mongo segments to keep parity with legacy instrumentation

lib/subscribers/mongodb/db.js
```diff
@@ -0,0 +1,126 @@
...
+  constructor({ agent, logger }) {
+    super({
+      agent,
+      logger,
+      channelName: 'nr_db',
+      packageName: 'mongodb',
+      system: MONGODB_METRIC_CONSTANTS.PREFIX
+    })
+
+    this.events = ['end']
```
bizob2828: this is missing `this.opaque = true` based on the old instrumentation, but weirdly we don't have assertions in the db related methods to catch this. I'm wondering if assertions should be added for this.

lib/subscribers/mongodb/db.js
```diff
@@ -0,0 +1,126 @@
...
+      }
+
+      self.logger.debug('Recording function %s', methodName)
+      const segmentAttributes = {}
+      if (ADMIN_COMMANDS.includes(methodName) === true) {
+        segmentAttributes.database_name = 'admin'
+      }
+      ctx = self.createSegment({
+        name: `${DB_METRIC_CONSTANTS.OPERATION}/${self.system}/${methodName}`,
+        recorder: genericRecorder,
```
bizob2828: the previous instrumentation used the database-operation metrics recorder, not generic. we should add these metric assertions in the versioned tests to ensure we don't regress going forward as well

lib/subscribers/mongodb/bulk.js
```diff
@@ -0,0 +1,135 @@
...
+      //
+      // We have to capture these values in a closure scope because multiple
+      // cursor methods may fire in one transaction. For example, when using
+      // the `.toArray` method the `.next` method will be invoked at least once
+      // to get the documents to add to the resulting array. If we stored this
+      // information on our object instance, the values would get overwritten
+      // by the nested operations.
+      const capturedOperation = instance.constructor.name === 'OrderedBulkOperation'
+        ? 'orderedBulk'
+        : 'unorderedBulk'
```
bizob2828:
```suggestion
      const capturedOperation = instance.isOrdered ? 'orderedBulk' : 'unorderedBulk'
```

This is what was done before and is safer than keying of a class name that could change

lib/subscribers/mongodb/cursor.js
```diff
@@ -0,0 +1,134 @@
...
+        continue
+      }
+
+      this.#wrapMethod(dbClient, method, collection)
+    }
+
+    return ctx
+  }
+
+  #wrapMethod(instance, methodName, collection) {
```
bizob2828: there's quite a lot of redundant code in these wrapMethods amongst all the classes. I'm wondering if you create some mongo class for wrapMethod and handle bespoke behavior in some override/super call

bizob2828 (changes requested):
overall looks good, a few comments about some missing bits and abstraction suggestions

lib/subscribers/mongodb/client.js
```diff
@@ -0,0 +1,65 @@
...
+      if (Object.hasOwn(event, 'connectionId') === false) {
+        return
+      }
+
+      const address = event.address
+      const lastColon = address.lastIndexOf(':')
+      let host = address.slice(0, lastColon)
+      const port = address.slice(lastColon + 1)
+
+      if (['127.0.0.1', '::1', '[::1]'].includes(host)) {
```
bizob2828: i think you can remove this and just assign host and then update `utils/get-host-details.js` to do:

```
const LOCALHOST_ALIASES = ['127.0.0.1', '::1', '[::1]']
if (LOCALHOST_ALIASES.includes(host)) {
    host = 'localhost'
  }
```
at the end of the util because right now only client calls are getting the normalized host to `localhost`

lib/subscribers/mongodb/cursor.js
```diff
@@ -0,0 +1,76 @@
...
+} = require('#agentlib/metrics/names.js')
+
+/**
+ * This list is a list of all methods we are interested in instrumenting,
+ * regardless of client version. As we iterate through the list we will perform
+ * a presence check for each method.
+ *
+ * @type {string[]}
+ */
+const CURSOR_METHODS = [
```
bizob2828, line 23: i just noticed that `pipe` is no longer being instrumented and we don't have tests for it but it looks like this was leftover in the legacy instrumentation from pre v4, so n/m

lib/subscribers/mongodb/wrapper.js
```diff
@@ -0,0 +1,109 @@
...
+const DbSubscriber = require('../db.js')
+const databaseRecorder = require('#agentlib/metrics/recorders/database.js')
+
+/**
+ * This class provides a baseline for the MongoDB instrumentations that
+ * share a lot of common code when wrapping methods, but require some slight
+ * variations in the collected data. Anything extending this class is
+ * expected to leverage the {@link #wrapDatabaseMethod} method.
+ */
+module.exports = class WrapperSubscriber extends DbSubscriber {
```
bizob2828, line 17: nice work, this cuts down on a lot of redundancy

bizob2828 (changes requested):
overall this looks great. one suggestion on moving the remapping of host to the util to ensure all host values get it

lib/subscribers/mongodb/client.js
jsumners-nr: Good idea.

amychisholm03 (dismissed):

bizob2828 (approved):

PR #4061, opened 2026-06-19
Add getBrowserAgentLoader() API and hasToRemoveLoaderScript option to getBrowserTimingHeader()

Andreyco:
**Background**

The existing `getBrowserTimingHeader()` returns a single `<script>` block combining two conceptually distinct things:
- the NREUM config blob — transaction-specific metadata that is unique per request and must not be cached.
- the agent loader script — a static JS payload (js_agent_loader) that is identical across all requests for a given agent version and is safe to cache aggressively.

Bundling them together forces both to be delivered through the same path, which is either uncacheable (correct for config, wasteful for the loader) or cached (correct for the loader, incorrect for config).

**What this changes**

**New method:** `api.getBrowserAgentLoader(options)`

Returns only the static loader script, optionally wrapped in a `<script>` tag (with nonce support). Because the loader never changes between requests, it can be served through a CDN or long-lived cache, or inlined once into a layout template.

**New option:** `getBrowserTimingHeader({ hasToRemoveLoaderScript: true })`

Returns only the per-request `NREUM` config block — no loader body. This is the piece that must be injected dynamically on each response (e.g. via a reverse proxy inserting it into <head>, or a server-side template). **Change is opt in and non breaking.**

**Internal refactor:** `_wrapScriptTag(content, options)`

The `<script>` wrapping logic (plain / nonce / unwrapped) was duplicated between _generateRUMHeader and the new method. Extracted into a shared private helper.

**Why this matters**

This split enables deployment patterns that weren't previously possible with a single APM-connected Node.js agent:
- cached loader delivery — serve getBrowserAgentLoader() from a caching reverse proxy or CDN; invalidate only on agent upgrades.
- lightweight per-request injection — inject only the small config blob (hasToRemoveLoaderScript: true) via a proxy or template on each request, keeping the response path thin.
- Separation of concerns

**The existing getBrowserTimingHeader() behaviour is unchanged — callers that don't opt in to either new option get the same combined output as before.**

### DIFF
--- a/api.js
+++ b/api.js
@@ -632,25 +632,34 @@ function _gracefail(errorCode, quiet) {
  * @private
  * @param {object} options Configuration options for RUM
  * @param {string} [options.nonce] Nonce to inject into `<script>` header.
+ * @param {boolean} [options.hasToRemoveLoaderScript] Used to strip loader script. Renders only RUM header.
  * @param {boolean} [options.hasToRemoveScriptWrapper] Used to import agent script without `<script>` tag wrapper.
  * @param {string} metadata Stringified representation of rumHash metadata
  * @param {string} loader Agent Loader script
  * @returns {string} fully formed RUM header
  */
-function _generateRUMHeader(options = {}, metadata, loader) {
-  const formatArgs = []
-
+/**
+ * Wraps `content` in a `<script>` tag, respecting nonce and hasToRemoveScriptWrapper options.
+ *
+ * @param {string} content
+ * @param {object} options
+ * @param {string} [options.nonce]
+ * @param {boolean} [options.hasToRemoveScriptWrapper]
+ * @returns {string}
+ */
+function _wrapScriptTag(content, options = {}) {
   if (options.hasToRemoveScriptWrapper) {
-    formatArgs.push(RUM_STUB)
-  } else if (options.nonce) {
-    formatArgs.push(RUM_STUB_SHELL_WITH_NONCE_PARAM, `nonce="${options.nonce}"`)
-  } else {
-    formatArgs.push(RUM_STUB_SHELL)
+    return content
   }
+  if (options.nonce) {
+    return `<script type='text/javascript' nonce="${options.nonce}">${content}</script>`
+  }
+  return `<script type='text/javascript'>${content}</script>`
+}
 
-  formatArgs.push(metadata, loader)
-
-  return util.format(...formatArgs)
+function _generateRUMHeader(options = {}, metadata, loader) {
+  const body = util.format(RUM_STUB, metadata, options.hasToRemoveLoaderScript ? '' : loader)
+  return _wrapScriptTag(body, options)
 }
 
 /**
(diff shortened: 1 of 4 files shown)


metal-messiah:
Hi @Andreyco, I'm from the Browser Agent team over at [https://github.com/newrelic/newrelic-browser-agent](https://github.com/newrelic/newrelic-browser-agent)

Thank you for putting this together and providing such a thorough breakdown of the background and benefits. This is a very thoughtful contribution!

Internally, we have discussed the option of configuring the agent to modify the existing API's string contents to return two scripts instead of one, one that wraps the config portion and another that points remotely to our already-served CDN assets (e.g., `js-agent.newrelic.com`) for allowing the browser to automatically cache the loader portion of the code. 

Naively, it would be going from something like:
```html
<script type="text/javascript">
  window.NREUM||(window.NREUM={});
  NREUM.info={
    beacon:"bam.nr-data.net",
    errorBeacon:"bam.nr-data.net",
    licenseKey:"NRJS-xxxxxxxxxxxxxx",
    applicationID:"12345678",
    transactionName:"Z1YAZREEV0AHAUQLXFkbWBBbS0IDWAsXGFlfDw==",
    queueTime:12,
    applicationTime:42,
    atts:"S0VREAtDTR4=",
    // ...etc...
  };

  !function(){"use strict";var e=window.NREUM||{}; ... /* static minified JS agent code here */ }()
</script>
```

to something more like:
```html
<script type="text/javascript">
  window.NREUM||(window.NREUM={});
  NREUM.info={
    beacon:"bam.nr-data.net",
    errorBeacon:"bam.nr-data.net",
    licenseKey:"NRJS-xxxxxxxxxxxxxx",
    applicationID:"12345678",
    transactionName:"Z1YAZREEV0AHAUQLXFkbWBBbS0IDWAsXGFlfDw==", // Live dynamic data
    queueTime:12,                                              // Live dynamic data
    applicationTime:42,                                        // Live dynamic data
    atts:"S0VREAtDTR4=",
    agent:""
  };
</script>
<script src="https://js-agent.newrelic.com/nr-loader-spa-1.317.0.min.js"></script>
```

We'd love to get feedback from you on this topic. Would moving on an approach like this satisfy your request? We are currently driving these discussions internally through the browser agent team.  If the Node Agent team decides to close this issue, please open an issue on the browser agent repo and we can continue discussions over there!

Thanks!

Andreyco:
@metal-messiah hello there, thanks for responding, much appreciated.

Looking at the proposed output, where original script is split into two separate `script` tags would do the job!

```
<script src="https://js-agent.newrelic.com/nr-loader-spa-1.317.0.min.js"></script>
```

Above part is exactly what I believe we're all missing. I see `https://js-agent.newrelic.com/nr-loader-spa-{version}.min.js` has high enough cache time and get's invalidated on version change. 

The snippet you provided is exactly what I intended to do - when implemented "natively" we get it without heavy lifting 👍 
Is it possible to share any ballpark ETA?

bizob2828:
I'm going to close this PR as @metal-messiah laid out a possible future solution. Please continue to have this conversation though or move it to https://github.com/newrelic/newrelic-browser-agent

PR #4070, opened 2026-06-25
fix: Fixed Azure functions crashing when they include response headers

### LINKED ISSUE #4062: Azure Functions subscriber crashes with "Cannot read properties of null (reading 'addSpanAttribute')" on any HTTP response with collected headers
## Description

The `@azure/functions` subscriber instrumentation (introduced in #3915, shipped in v13.20.0) throws a `TypeError` while finalizing an HTTP-trigger transaction, which propagates out of the worker and turns **every** invocation into an HTTP 500 with an empty body.

```
TypeError: Cannot read properties of null (reading 'addSpanAttribute')
    at _collectHeaders        (node_modules/newrelic/lib/header-attributes.js:137:15)
    at collectResponseHeaders (node_modules/newrelic/lib/header-attributes.js:159:3)
    at Transaction.finalizeWeb (node_modules/newrelic/lib/transaction/index.js:449:22)
    at HttpHandler.finalizeTransaction (node_modules/newrelic/lib/subscribers/azure-functions/http-handler.js:41:17)
    at HttpHandler.handle             (node_modules/newrelic/lib/subscribers/azure-functions/azure-handler-base.js:47:17)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
```

Root cause:

- `AzureHandler.handle()` runs the user handler inside the transaction context via `tracer.runInContext`/`bindFunction` (scoped `AsyncLocalStorage.run`), but then `await`s it and calls `finalizeTransaction()` **outside** that `run` scope. So at finalize time `tracer.getContext()` returns a fresh empty `Context` and `tracer.getSegment()` returns `null`.
- `Transaction.finalizeWeb()` is itself inconsistent: it uses `this.baseSegment.addSpanAttribute(...)` for `http.statusCode` (works, `baseSegment` is set), but response-header collection delegates to `header-attributes._collectHeaders`, which uses `tracer.getSegment()` (null) and then calls `segment.addSpanAttribute(...)` with **no null guard** → crash.
- Trigger condition is simply that the response contains a header in the collected set (`content-type`, `content-length`, etc.). Because virtually every HTTP response has `content-type`/`content-length`, this reproduces for essentially **all** HTTP-trigger functions — it is not specific to any framework (we hit it with Apollo Server, but a trivial handler reproduces it too).

## Expected Behavior

The agent should record the transaction without throwing. An error inside agent instrumentation must never bubble out of the function and cause the invocation to fail (500). Response header attributes should be attached to the transaction's base segment regardless of whether an AsyncLocalStorage context is active at finalize time.

Host-side log (Azure Functions, scrubbed):

```
Microsoft.Azure.WebJobs.Host.FunctionInvocationException : Exception while executing function: Functions.<fn>
 ---> Microsoft.Azure.WebJobs.Script.Workers.Rpc.RpcException :
 Result: Failure
 Exception: Cannot read properties of null (reading 'addSpanAttribute')
 Stack: TypeError: Cannot read properties of null (reading 'addSpanAttribute')
    at _collectHeaders (/home/site/wwwroot/node_modules/newrelic/lib/header-attributes.js:137:15)
    at Object.collectResponseHeaders (/home/site/wwwroot/node_modules/newrelic/lib/header-attributes.js:159:3)
    at Transaction.finalizeWeb (/home/site/wwwroot/node_modules/newrelic/lib/transaction/index.js:449:22)
    at HttpHandler.finalizeTransaction (/home/site/wwwroot/node_modules/newrelic/lib/subscribers/azure-functions/http-handler.js:41:17)
    at HttpHandler.handle (/home/site/wwwroot/node_modules/newrelic/lib/subscribers/azure-functions/azure-handler-base.js:47:17)
```

There is no application-level error log — the throw happens inside the agent's finalize step, after the user handler has already returned its response.

## Steps to Reproduce

1. Create an Azure Functions app using the Node.js v4 programming model (`@azure/functions` >= 4.7.0), deployed to a real Azure Function App (so `WEBSITE_OWNER_NAME` / `WEBSITE_RESOURCE_GROUP` / `WEBSITE_SITE_NAME` are present and the subscriber initializes).
2. Load the agent (e.g. app setting `NODE_OPTIONS=-r newrelic`, with `NEW_RELIC_APP_NAME` / `NEW_RELIC_LICENSE_KEY` set).
3. Register a trivial HTTP function that returns a response with a header:

   ```js
   const { app } = require('@azure/functions');

   app.http('repro', {
     methods: ['GET'],
     authLevel: 'anonymous',
     handler: async () => {
       return { status: 200, headers: { 'content-type': 'text/plain' }, body: 'ok' };
     },
   });
   ```

4. `curl` the endpoint → HTTP 500 with empty body; the stack above appears in the host logs.

(Setting no headers at all avoids the crash, since `_collectHeaders` returns early when there are no collected headers — but any normal response, including the host's default `content-type`, triggers it.)

## Your Environment

* newrelic (Node agent) version: reproduced on **13.20.0** and **14.1.0** (latest)
* `@azure/functions`: 4.x (v4 programming model)
* Node version: 20.x (Azure Functions Linux runtime)
* Azure Functions host version: 4.x
* OS: Linux (Azure Functions)

## Additional context

- Introduced by #3915 ("Updated `@azure/functions` instrumentation to a subscriber type", v13.20.0). #3811 later touched the same area (`bindFunction` → `runInContext`) but did not fix this.
- `lib/header-attributes.js` has not changed since 2025-10, so the missing null guard is still present on `main` and in the latest release.
- Suggested fix (either or both):
  - Add a null guard in `header-attributes._collectHeaders` before `segment.addSpanAttribute(...)`.
  - Have `header-attributes._collectHeaders` / `Transaction.finalizeWeb` use `transaction.baseSegment` for span attributes (as `finalizeWeb` already does for `http.statusCode`) instead of the context-dependent `tracer.getSegment()`, and/or run `finalizeTransaction` within the transaction context.
- Workaround currently in use: disable the subscriber via `newrelic.js` (`instrumentation: { '@azure/functions': { enabled: false } }`) and create the web transaction manually with `newrelic.startWebTransaction(...)` (the manual path does not go through `finalizeWeb`, so it is unaffected).

jsumners-nr:
When an Azure function includes headers in the response object, our instrumentation was crashing at:

https://github.com/newrelic/node-newrelic/blob/0ae24b865d61b0035cffdafbcfc6543cd7dce27d/lib/header-attributes.js#L137

This is because the `segment` was unavailable at:

https://github.com/newrelic/node-newrelic/blob/0ae24b865d61b0035cffdafbcfc6543cd7dce27d/lib/header-attributes.js#L120

The root cause is that our transaction finalization was not happening within the right context.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -182,7 +182,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [22.x, 24.x]
+        node-version: [22.22.3, 24.16.0]
 
     steps:
       - uses: actions/checkout@v6
--- a/lib/subscribers/azure-functions/azure-handler-base.js
+++ b/lib/subscribers/azure-functions/azure-handler-base.js
@@ -42,9 +42,22 @@ module.exports = class AzureHandler {
     const newCtx = this.createTransaction({ handlerArgs, ctx })
     const transaction = newCtx?.transaction
     this.addFaasAttributes(transaction, context)
-    const result = await this.runHandlerInContext({ originalHandler, ctx: newCtx, thisArg, handlerArgs })
-    this.handleColdStart(transaction)
-    return this.finalizeTransaction({ result, transaction })
+
+    // We need to wrap up all of our operations into a function that we
+    // can execute within the same context. If we only run the original handler
+    // in the context, then when we try to finalize the transaction we will
+    // not have access to the segment.
+    const wrappedHandler = async (...args) => {
+      const result = await originalHandler.apply(thisArg, args)
+      this.handleColdStart(transaction)
+      return this.finalizeTransaction({ result, transaction })
+    }
+    return await this.runHandlerInContext({
+      originalHandler: wrappedHandler,
+      ctx: newCtx,
+      thisArg,
+      handlerArgs
+    })
   }
 
   runHandlerInContext({ originalHandler, ctx, thisArg, handlerArgs }) {
--- a/test/versioned/azure-functions/http.test.js
+++ b/test/versioned/azure-functions/http.test.js
@@ -387,3 +387,40 @@ test('ends transaction on stream close', async (t) => {
     })
   })
 })
+
+test('collects response headers', async (t) => {
+  bootstrapModule({ t })
+  const { agent, mockApi } = t.nr
+  agent.config.allow_all_headers = true
+
+  const handler = async function () {
+    const response = new AzureFunctionHttpResponse()
+    response.body = 'ok'
+    response.status = 200
+    response.headers = {
+      'content-type': 'application/json',
+      'x-custom-header': 'custom-value'
+    }
+    return response
+  }
+  const options = { handler }
+
+  const txFinished = once(agent, 'transactionFinished')
+  mockApi.app.get('a-test', options)
+  const wrappedHandler = global.azure.handlers.at(-1)
+  const response = await mockApi.httpRequest('get', wrappedHandler)
+  assert.equal(response.body, 'ok')
+
+  const [tx] = await txFinished
+  assert.ok(tx)
+
+  // Verify response headers were collected as transaction attributes
+  const attributes = tx.trace.attributes.get(DESTS.TRANS_COMMON)
+  assert.equal(attributes['response.headers.contentType'], 'application/json')
+  assert.equal(attributes['response.headers.xCustomHeader'], 'custom-value')
+
+  // Verify response headers were also added to the base segment as span attributes
+  const spanAttributes = tx.baseSegment.attributes.get(DESTS.SPAN_EVENT)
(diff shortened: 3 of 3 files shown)


.github/workflows/ci-workflow.yml
```diff
@@ -182,7 +182,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [22.x, 24.x]
```
jsumners-nr, line 185: This is to work around an issue in the latest LTS releases:

+ https://jdstaerk.substack.com/p/nodejs-security-fix-silently-broke
+ https://github.com/nodejs/node/commit/bc0b53813e
+ https://github.com/nodejs/node/issues/63989
+ https://github.com/nodejs/node/pull/64004

We should be able to restore this in a week or so after new versions ship.

amychisholm03 (approved):

PR #4084, opened 2026-06-30
feat: Added Node.js 26 support

### LINKED ISSUE #4081: Add Node 26 support
### Description
In preparation for Node 26 being LTS in October 2026 we have to do a few things to support this version:

 * Include new versions of bundled deps we manage: `@newrelic/native-metrics`, `@newrelic/fn-inspect`
 * Update the check for unsupported versions: https://github.com/newrelic/node-newrelic/blob/7eaed22f92af2da5db31232b60b5bf35f0ca3e02/index.js#L72-L80
 * Add Node 26 to CI jobs, where applicable

jsumners-nr:
This PR resolves #4081.

### DIFF
--- a/package.json
+++ b/package.json
@@ -225,7 +225,7 @@
   "optionalDependencies": {
     "@datadog/pprof": "^5.13.3",
     "@newrelic/fn-inspect": "^5.0.0",
-    "@newrelic/native-metrics": "^13.0.0",
+    "@newrelic/native-metrics": "^13.1.0",
     "@prisma/prisma-fmt-wasm": "^4.17.0-16.27eb2449f178cd9fe1a4b892d732cc4795f75085"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -223,7 +223,7 @@
     "winston-transport": "^4.5.0"
   },
   "optionalDependencies": {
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -225,7 +225,7 @@
   "optionalDependencies": {
     "@datadog/pprof": "^5.13.3",
     "@newrelic/fn-inspect": "^5.0.0",
```
bizob2828: do you want to update this to `^5.1.0` or shall I? That's the minimum version that supports node 26

jsumners-nr: I will once I come back to this when the dependencies can be installed.

package.json
```diff
@@ -224,8 +224,8 @@
   },
   "optionalDependencies": {
     "@datadog/pprof": "^5.13.3",
```
bizob2828: should we also update this to to [5.14.2](https://github.com/DataDog/pprof-nodejs/pull/331) as that's where node 26 support was added?

bizob2828 (changes requested):
one small suggestion for pprof min version. Also please keep in mind we have to address [this](https://github.com/newrelic/newrelic-agent-init-container/issues/261) before releasing the minor

bizob2828 (approved):

PR #4086, opened 2026-07-01
chore: Added OTEL diagnostics log forwarding to OTEL metrics

### LINKED ISSUE #4076: Enable diagnostics logging for OTEL metrics
Spec PR 814 updates the OTEL metrics instrumentation requirements to include capturing diagnostics logging generated by the OTEL libraries.

https://github.com/newrelic/node-newrelic/blob/0520c354df0ffcefc31e4581641741f03d7e525b/lib/otel/metrics/index.js#L24-L27

We need to update the above to look at the agent's configured logging level. When it is set to `debug` or `trace` we need to provide a "logger" instance to the OTEL diagnostics system that will capture these logs and forward them to our regular agent logger at the debug level.

jsumners-nr:
This PR resolves #4076.

### DIFF
--- a/lib/otel/metrics/index.js
+++ b/lib/otel/metrics/index.js
@@ -38,6 +38,15 @@ class SetupMetrics extends SetupSignal {
       exportTimeout = 10_000
     }
 
+    // Our OTEL metrics spec dictates that we forward OTEL diagnostics logging
+    // to our agent logger. It does not specify what levels should be targeted.
+    // We opt to do this only when debug logging has been enabled.
+    if (logger.debugEnabled() === true || logger.traceEnabled() === true) {
+      const DiagLogger = require('./diag-logger.js')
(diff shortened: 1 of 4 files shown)


lib/otel/metrics/index.js
```diff
@@ -38,6 +38,15 @@ class SetupMetrics extends SetupSignal {
       exportTimeout = 10_000
     }
 
+    // Our OTEL metrics spec dictates that we forward OTEL diagnostics logging
+    // to our agent logger. It does not specify what levels should be targeted.
+    // We opt to do this only when debug logging has been enabled.
+    if (logger.debugEnabled() === true || logger.traceEnabled() === true) {
```
bizob2828, line 44: so in setup we have `createOtelLogger` which gets set for all otel related stuff, i'm not sure I understand why this is needed

Issue #4083, opened 2026-06-30
Add source map support to `CpuProfiler` so flamegraph frames map to original source


amychisholm03:
### Description
The goal is to make continuous-profiler flamegraph frames resolve back to the **original authored source** (TypeScript, bundled/minified JS) instead of the compiled output that actually runs — so a hot frame links to the right line in the user's GitHub repo.

Today the profiler already records function name + file + line for every frame, but those are positions in the **compiled/running** JS. The agent never passes a `sourceMapper` to `pprof`, so for any transpiled or bundled app the frames point at generated `.js`, not the source the user wrote.

- Build a `pprof.SourceMapper` once and pass it into the CPU profiler (`time.start({ sourceMapper })`) and the heap profiler's serialize path.
- Config to enable it and specify which directories to scan for `.map` files.
- Lifecycle handling: build once at profiler start (it's async), keep it for the whole session, dispose on stop.

### Acceptance Criteria
- As a Node developer profiling a TypeScript/bundled app, I want flamegraph frames to resolve to my original source files and lines so that I can find the hot code path without mentally reversing the compile step.
- A `SourceMapper` is wired into the CPU profiler. _Heap profiler support will be a separate issue_.
- When `.map` files are present, captured stacks report original-source file/line; when they're absent, it falls back to today's compiled file/line with no errors.
- The mapper is built **once** at profiler init (not per collect), and its build cost is measured and acceptable.

### Design Consideration / Limitations
- Q: Will source code mapping always occur when profiling is turned on? Or should it be a separate config option? 
  - A: We will add a new config option called `profiling.source_mapping.enabled` with `boolean` type and `false` by default.
- Q: `searchDirs` will default to the application root directory. Do we need a config option for users to specify additional/alternate directories (e.g. when maps are emitted outside the app root)?

### Additional context
Please [see initial work request](https://new-relic.atlassian.net/browse/NR-551253).

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-583521

PR #4090, opened 2026-07-01
feat: Add `profiling.source_mapping.enabled` config option

amychisholm03:
## Description

Preliminary work for issue #4083. This PR adds a new config option, `profiling.source_mapping.enabled`, which is set to `false` by default. When we do the work for #4083, customers who enabled Continuous Profiling may not care about or need source mapping. (Source mapping in our case refers to mapping pre-compiled TS/JS source code; we already map compiled JS.) Thus, we need a config option that will allow customers to toggle this behavior on and off. This PR just adds the config option, but no logic because that will be completed in #4083.

## How to Test

```
npm run unit
```

## Related Issues

Prereq to #4083


lib/config/default.js
```diff
@@ -1274,6 +1274,14 @@ defaultConfig.definition = () => {
       duration: {
         formatter: int,
         default: 0
+      },
+      /**
+       * When set to `true`, resolves profiler frames to their original
+       * source files/lines using source maps, instead of the compiled output.
+       */
+      source_mapping_enabled: {
```
bizob2828: we typically follow `source_mapping.enabled` so this should be nested

bizob2828 (changes requested):
You'll also want to add this to the agent-specs

lib/config/default.js
amychisholm03, line 1: Not sure why the auto-formatter in VS code did this

bizob2828 (approved):

PR #4091, opened 2026-07-02
chore: Removed spanId in outgoing tracestate header

### LINKED ISSUE #4075: Stop sending `spanId` in tracestate header
Spec PR 804 updates the requirements around building the outbound tracestate header to dictate that we must not include the `spanId`.

The following must be updated to set the `spanId` field to an empty string for all cases:

https://github.com/newrelic/node-newrelic/blob/0520c354df0ffcefc31e4581641741f03d7e525b/lib/w3c/tracestate.js#L157-L159

jsumners-nr:
This PR resolves #4075.

### DIFF
--- a/test/integration/distributed-tracing/trace-context-cross-agent.test.js
+++ b/test/integration/distributed-tracing/trace-context-cross-agent.test.js
@@ -44,6 +44,7 @@ async function runTestCase(testCase, parentTest) {
       'web_transaction',
       'comment',
       'transaction_events_enabled',
+      'exclude_newrelic_header',
       // Sampling
       'distributed_tracing_enabled',
       'full_granularity_enabled',
--- a/test/lib/cross_agent_tests/distributed_tracing/trace_context.json
+++ b/test/lib/cross_agent_tests/distributed_tracing/trace_context.json
@@ -14,20 +14,20 @@
     "force_adaptive_sampled": false,
     "expected_priority_between": [ 0, 1 ],
     "inbound_headers": [
-    {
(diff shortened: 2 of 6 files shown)


test/lib/cross_agent_tests/distributed_tracing/trace_context.json
jsumners-nr, line 1: Updated from `cross_agent_tests` PR 173.

bizob2828: interesting we missed a few other syncs. i see changing to `partial_granularity_ratio": 0.00000001,`  as well as a test for excluding NR header. thanks for syncing!

test/integration/distributed-tracing/trace-context-cross-agent.test.js
```diff
@@ -44,6 +44,7 @@ async function runTestCase(testCase, parentTest) {
       'web_transaction',
       'comment',
       'transaction_events_enabled',
+      'exclude_newrelic_header',
```
bizob2828, line 47: These cross agent tests seem broken now. Only a few have `exclude_newrelic_header` set to false. But our default is true. However most of the cross agent tests don't specify `exclude_newrelic_header` but they assert they're present 🤔

bizob2828 (approved):
nice work. just a comment on the cross agent tests but this is unrelated to the scope of this ticket. it's just that we were out of sync and the tests operate on the assumption we don't exclude new relic header by default but then there are a few tests that explicitly set `exclude_newrelic_header` to false, which is a no-op

PR #4094, opened 2026-07-06
chore: Fixed package version resolution on Node 26

jsumners-nr:
This solves our newly broken CI on Node 26. When running `npm run versioned:major undici` the issue does not present itself. But when running `npm run versioned undici`, we trigger [this prophecy](https://github.com/nodejs/node/issues/61149). So we have to be a bit more creative with our stack inspection to resolve the _user_ installed version of `undici`.


bizob2828 (approved):
verified undici versioned tests work on 22/24/26

PR #4095, opened 2026-07-06
feat: Add source mapping to CPU (wall time) profiler

### LINKED ISSUE #4083: Add source map support to `CpuProfiler` so flamegraph frames map to original source

amychisholm03:
## Description

This PR adds source code mapping* support for our current `CpuProfiler`. When enabled via the `profiling.source_mapping.enabled` config option, a `pprof.SourceMapper` is built once during agent startup by scanning the application root (`process.cwd()`) for source maps. It's cached on the `ProfilingManager` and passed into the `CpuProfiler`'s constructor when the profilers are registered. It builds it once at startup, rather than on each profiling cycle, to keep the profiling start/stop path synchronous and avoids re-scanning on every collection. Caching it on the manager also lets the same `SourceMapper` be shared with the `HeapProfiler` when we add heap source mapping later.

If the `SourceMapper` fails to build (or no source maps are found), the profiler falls back to its original behavior and reports the compiled file/line/code information.

*Source code mapping refers to mapping the profiling sample to the _pre-compiled_ code e.g. TypeScript, pre-compiled JavaScript. The goal is that users will be able to more easily correlate the node in the flamegraph with the function information (file location, code line number, etc.) in their written code instead of the compiled version.

## How to Test

```
npm run integration
npm run unit

# or more specifically
node --test test/integration/profiling/source-mapping.test.js
node --test test/unit/aggregators/profiling-aggregator.test.js
node --test test/unit/lib/profiling/profilers/cpu.test.js
node --test test/unit/lib/profiling/index.test.js
```

With the example app: https://github.com/newrelic/newrelic-node-examples/pull/378

## Related Issues

Closes #4083


lib/aggregators/profiling-aggregator.js
```diff
@@ -42,8 +42,8 @@ class ProfilingAggregator extends BaseAggregator {
    * as we want to collect profiling data for `cpu` and/or `heap`
    * and send the gzipped binary encoded data for each profiler
    */
-  start() {
-    this.profilingManager.register()
+  async start() {
```
bizob2828: the aggregators are not awaited in `lib/harvester.js` nor is the harvester. I'm not sure how this would be behave if you update that code

lib/profiling/index.js
```diff
@@ -18,15 +18,51 @@ class ProfilingManager {
...
+  }
+
+  /**
+   * Builds the `pprof` SourceMapper by scanning the app root (`process.cwd()`) for
+   * `.map` files. Logs the build cost; returns `null` on failure so frames fall
+   * back to compiled file/line.
+   *
+   * @returns {Promise<object|null>} the SourceMapper, or `null` on failure
+   */
+  async #buildSourceMapper() {
```
bizob2828: you might have to do this somewhere in the agent and pass in the source mapper based on the comments i pointed out in the profiling aggregator

test/unit/agent/agent.test.js
```diff
@@ -265,6 +265,9 @@ test('#harvesters.start should start  and stop all aggregators', (t) => {
   })
 
   agent.harvester.start()
+  // The profiling aggregator starts asynchronously (it builds the CPU profiler's
+  // SourceMapper), so let its start settle before asserting the timers are armed.
+  await new Promise((resolve) => setImmediate(resolve))
```
bizob2828: this points out how this could be problematic in a real word setting

lib/profiling/index.js
```diff
@@ -18,15 +18,51 @@ class ProfilingManager {
...
     if (this.config.include.includes('heap') && !this.profilers.has('HeapProfiler')) {
       const { HeapProfiler } = require('./profilers')
+      // @todo pass in sourceMapper if enabled
       this.profilers.set('HeapProfiler', new HeapProfiler({ logger: this.logger }))
     }
 
     if (this.config.include.includes('cpu') && !this.profilers.has('CpuProfiler')) {
       const { CpuProfiler } = require('./profilers')
-      this.profilers.set('CpuProfiler', new CpuProfiler({ logger: this.logger, samplingInterval: this.samplingInterval, tracer: this.tracer }))
+      const sourceMapper = this.config.source_mapping.enabled ? await this.#buildSourceMapper() : null
```
bizob2828: you have to keep in mind profiling gets dynamically started/stopped. you wouldn't want to do this every time the profiler starts but just once per agent running instance, all the more reason to put it elsewhere and pass it in.  if source_mapping is enabled regardless of if profiling is enabled, you'd prob want to generate this source mapper

bizob2828 (changes requested):
I don't think we could land this in the current state based on my comments

lib/profiling/index.js
amychisholm03: Okay, I'll do some more digging into where the best spot to build the `SourceMapper` would be, so we don't have the `async` "roll-up" issue. But the overall idea of giving the `SourceMapper` instance to `ProfilingManager` and then into the constructors of `CpuProfiler` and `HeapProfiler` is a decent way to do it, right?

lib/agent.js
```diff
@@ -452,9 +456,13 @@ Agent.prototype.onConnect = function onConnect(shouldImmediatelyHarvest, callbac
     this.metrics.getOrCreateMetric(NAMES.FEATURES.CERTIFICATES).incrementCallCount()
   }
 
-  this._scheduleHarvests(shouldImmediatelyHarvest, callback)
-
-  this.setState('started')
+  // Harvests start the profilers, which consume the source mapper.
```
bizob2828: this still reads funny. what about doing this?

```
diff --git a/lib/agent.js b/lib/agent.js
index d07db0ed8..4fc8c9e4e 100644
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -401,7 +401,7 @@ Agent.prototype.start = function start(callback) {
   // start/stop cycles; awaited in `onConnect` before harvests start the profilers.
   this.sourceMapperReady = this.profilingData.initSourceMapper()
 
-  this.collector.connect(function onStartConnect(error, response) {
+  this.collector.connect(async function onStartConnect(error, response) {
     if (error || response.shouldShutdownRun()) {
       agent.healthReporter.setStatus(HealthReporter.STATUS_CONNECT_ERROR)
       agent.setState('errored')
@@ -412,6 +412,9 @@ Agent.prototype.start = function start(callback) {
 
     if (agent.collector.isConnected()) {
       const config = response.payload
+      // await for source mapper before starting the harvester
+      // which registers the profiling aggregator which could take the sourceMapper
+      await agent.sourceMapperReady
 
       const shouldImmediatelyHarvest = !agent.config.no_immediate_harvest
       agent.onConnect(shouldImmediatelyHarvest, () => {
@@ -456,13 +459,8 @@ Agent.prototype.onConnect = function onConnect(shouldImmediatelyHarvest, callbac
     this.metrics.getOrCreateMetric(NAMES.FEATURES.CERTIFICATES).incrementCallCount()
   }
 
-  // Harvests start the profilers, which consume the source mapper.
-  // Wait for its build before scheduling, so the first profiling
-  // session resolves frames to their original source.
-  Promise.resolve(this.sourceMapperReady).then(() => {
-    this._scheduleHarvests(shouldImmediatelyHarvest, callback)
-    this.setState('started')
-  })
+  this._scheduleHarvests(shouldImmediatelyHarvest, callback)
+  this.setState('started')
 }
```

bizob2828: either way you should be adding tests that have profiling source mapping enabled/disabled and verify it's waiting for that to finish in the agent unit tests

bizob2828 (changes requested):
tried using https://github.com/newrelic/newrelic-node-examples/pull/378 but i don't see any samples that have user code in it. it's all either the agent or core node code. either way take a look at my suggestions

lib/agent.js
amychisholm03: I initially had an async `onStartConnect` but was worried that was going to cause issues. I can revert it back and I'll add the tests.

Regarding the example app, it can take a while to see the `fib` nodes. What you should be looking for in `*.ts` instead of `*.js`.

test/unit/agent/agent.test.js
```diff
@@ -516,6 +516,55 @@ test('when starting', async (t) => {
...
+      agent.collector.isConnected = () => true
+      callback(null, CollectorResponse.success(null, { agent_run_id: RUN_ID }))
+    }
+
+    let started = false
+    agent.start(() => {
+      started = true
+    })
+
+    await new Promise((resolve) => setImmediate(resolve))
```
bizob2828: what are these for?

test/unit/lib/profiling/profilers/cpu.test.js
```diff
@@ -0,0 +1,75 @@
...
+  sandbox.stub(pprof.time, 'start').callsFake(() => {
+    started = true
+  })
+  sandbox.stub(pprof.time, 'stop').callsFake(() => {
+    started = false
+  })
+  sandbox.stub(pprof.time, 'getState').returns({})
+  sandbox.stub(pprof.time, 'setContext')
+
+  ctx.nr = { sandbox, logger, tracer: createTracer(), profiler: null }
```
bizob2828: why not just use the mocked agent in `test/lib/agent_helper`?

bizob2828 (changes requested):
a few tests suggestions/questions but the code looks great!

bizob2828 (approved):

PR #4096, opened 2026-07-06
fix: Updated tracer, middleware wrapper, and wrapPromise to properly run promise context propagating outside current context to avoid leaking objects on promises that are never fulfilled

### LINKED ISSUE #4092: Regression bracketed to 13.16.0→13.17.0: old agent reaches steady state under sustained load, new agent never does
## Environment
- `newrelic`: reproduced on both `13.20.0` and `14.1.0`; `12.25.0` does not reproduce
- Node.js: `24.15`
- App: Nuxt 4 (SSR), running on AWS Fargate, moderate-to-high concurrent request volume
- Related to #3870, but distinct — see below

## Summary

We isolated a memory leak to a regression somewhere between `newrelic@12.25.0` and `13.20.0`. It's not just a magnitude difference — under the same nightly high-traffic window on the same staging environment, the old agent settles into a stable plateau while the new agent shows no sign of stopping. We have this replicated across 9 separate nights and one 48-hour continuous-container control. We're posting this as a separate issue from #3870 because our data doesn't support that issue's `AsyncContextFrame`/`enterWith()` mechanism (details below), while still reproducing the same class of symptom.

## The core finding: old agent plateaus, new agent climbs all night

Every night, staging sees a step increase in memory around 8pm ET (an environmental effect, same magnitude with the agent on or off — not part of this report) followed by roughly 10 hours of elevated request volume until ~6am ET. Segmenting memory by hour instead of averaging a whole-night slope shows a clear qualitative split during that window:

- **`newrelic@12.25.0`:** settles into a flat plateau (residual drift ~0.4–1.5 MB/h) for the rest of the night.
- **`newrelic@13.20.0` / `14.1.0`:** never settles — climbs continuously and monotonically for the full ~10h window, shows no sign of plateauing (r² ≈ 0.99+), until traffic drops off in the morning.

### Replication matrix

All slopes: 01:30–10:00 UTC, fleet average across 5 instances, RSS via `apm.service.memory.physical`. Night 7's disabled agent reports no APM data, so that night used host-level `SystemSample.memoryUsedBytes` instead.

| Night | Config | Slope | r² | Result |
|---|---|---|---|---|
| Night 1 | `14.1.0` stock | +46.9 MB/h | 0.998 | leaks |
| Night 2 | `12.25.0` | −0.2 MB/h | 0.03 | flat |
| Night 3 | `12.25.0` | +0.4 MB/h | 0.06 | flat |
| Night 4 | `14.1.0` + `--no-async-context-frame` | +45.0 MB/h | 0.997 | leaks |
| Night 5 | `14.1.0` + undici instrumentation disabled | +47.4 MB/h | 0.997 | leaks |
| Night 6 | `14.1.0` stock | +45.4 MB/h | 0.999 | leaks |
| Night 7 | agent fully disabled (`NEW_RELIC_ENABLED=false`) | — | — | flat |
| Night 8 | `13.20.0` stock | +44.7 MB/h | 0.989 | leaks |
| Night 9 | `13.20.0` + undici instrumentation disabled | +45.2 MB/h | 0.997 | leaks |

Every "new agent, enabled" night: 44.7–47.4 MB/h, r² ≥ 0.989 — a tight, consistent slope regardless of which instrumentation flag is toggled. Every "old agent or disabled" night: flat. These nights are interleaved with our config changes over a two-week span, not clustered, ruling out a simple day-of-week or calendar traffic effect.

### 48-hour continuous-container control

The table above uses different containers each night (this environment recycles fairly often). To confirm the `12.25.0` plateau is genuine steady state and not an artifact of containers restarting before they'd otherwise start climbing, we found a 48-hour window on `12.25.0` where the same 5 instances ran continuously with no restart. Across two complete overnight cycles on the *same* containers: memory returned to the same range both nights (plateau ~0.345–0.348 GB night 1 vs. ~0.337–0.347 GB night 2; trough ~0.259 GB vs. ~0.279 GB) — zero net drift over 48 hours. This is a real, repeating, stable oscillation, not a leak masked by container turnover.

## What this rules out

- **Not v14-specific.** `13.20.0` (last pre-v14 release) leaks at the same magnitude as `14.1.0`.
- **Not undici instrumentation.** Disabling it did not reduce the slope on either `14.1.0` or `13.20.0`.
- **Not the `AsyncContextFrame` backend specifically.** Running with `--no-async-context-frame` on Node 24 did not flatten the leak, despite this being the exact condition under which #3870's mechanism would apply.
- **Not native/buffer retention.** Heap-used growth tracks RSS growth closely over the same windows (ratio ~0.9–1.25 vs. ~0.02–0.10 on the flat `12.25.0` nights), indicating JS object retention.

## What we can't yet confirm

We don't have a confirmed root-cause line or object graph — only that the regression happened somewhere in the `12.25.0`→`13.20.0` range, and that fully disabling the agent is the only thing that has ever restored steady-state behavior. We looked at whether `Context` objects (`_transaction`/`_segment`) failing to null after `transaction.end()` could explain this, since that pattern is real in the current source — but the mechanism that would make that retention matter (`AsyncContextFrame` propagation) is the same one we directly ruled out above, so we're not asserting it as the cause.

We understand a maintainer attempted a minimal repro (Express + axios) on #3870 and could not reproduce a leak. Our data suggests this needs sustained, moderately high concurrent transaction volume under real SSR-style request patterns to manifest — we don't yet know the minimum conditions required.

## Config for reference

Non-default agent config (everything else is default):

```
NEW_RELIC_APP_NAME=Frontend-staging
NEW_RELIC_APPLICATION_LOGGING_ENABLED=true
NEW_RELIC_APPLICATION_LOGGING_FORWARDING_ENABLED=true
NEW_RELIC_APPLICATION_LOGGING_LOCAL_DECORATING_ENABLED=true
NEW_RELIC_BROWSER_MONITOR_ENABLE=true
```

We've pinned production to `12.25.0` as a workaround in the meantime.

---

If there are specific config toggles, debug settings, or heap-snapshot captures that would help narrow this down, we can run them in staging and turn results around within a day or two.

bizob2828:
## Description

We had a report in #4092 of memory leaks since 13.20.0. We're still trying to confirm that is the version introduced, but there was a pattern in 3 different places where we were attaching logic to a promise chain and actioning on NR primitives from context manager.  The theory is a promise could never be fulfilled, thus leaving references to segments, or transactions that can never be GCd, and create a memory leak.

A lot of the analysis, and fixes were done by claude with refactoring from me to keep a consistent API in our context manager.

I think we should ask the issue author to test this PR to ensure leaks are no longer present before merging.

## Related Issues
Closes #4092

### DIFF
--- a/lib/context-manager/async-local-context-manager.js
+++ b/lib/context-manager/async-local-context-manager.js
@@ -56,6 +56,17 @@ class AsyncLocalContextManager {
   runInContext(context, callback, cbThis, args = []) {
     return this._asyncLocalStorage.run(context, Reflect.apply, callback, cbThis, args)
   }
+
+  /**
+   * Run a function outside the current context. The primary use case is to ensure
+   * objects are not being held and not able to GC if the function never finishes.
+   *
+   * @param {Function} callback to run outside the current async context
+   * @returns {*} Returns the value returned by the callback function.
+   */
+  runOutOfContext(callback) {
+    return this._asyncLocalStorage.exit(callback)
+  }
 }
 
 module.exports = AsyncLocalContextManager
--- a/lib/subscribers/middleware-wrapper.js
+++ b/lib/subscribers/middleware-wrapper.js
@@ -128,14 +128,25 @@ class MiddlewareWrapper {
       try {
         const result = self.agent.tracer.runInContext({ handler, context: newCtx, thisArg: this, args, full: true })
         if (result?.then) {
-          return result.then(function onThen(val) {
-            self.maybeHandleError(txInfo)
+          // Hold `txInfo` (which stores the transaction object) weakly and run the propgation
+          // outside the current context.  This is to prevent a middleware promise that
+          // settles late or is never resolved to not retain the transaction (see issue #4092), and cause a possible leak.
+          // If `txInfo` has been collected by the time the promise settles there
+          // is no longer a transaction to record an error against.
+          const txInfoRef = new WeakRef(txInfo)
+          return self.agent.tracer._contextManager.runOutOfContext(() => result.then(function onThen(val) {
+            const info = txInfoRef.deref()
+            if (info) {
+              self.maybeHandleError(info)
+            }
             return val
-          },
-          function onCatch(err) {
-            self.maybeHandleError(txInfo, err)
+          }, function onCatch(err) {
+            const info = txInfoRef.deref()
+            if (info) {
+              self.maybeHandleError(info, err)
+            }
             throw err
-          })
+          }))
         }
         return result
       } catch (err) {
--- a/lib/subscribers/utils.js
+++ b/lib/subscribers/utils.js
@@ -4,6 +4,7 @@
  */
 
 'use strict'
+
 /**
  * Temporary fix as `tracePromise` wraps the promise in a native one.
  * Certain 3rd party library functions are wrapped in a `traceSync` call,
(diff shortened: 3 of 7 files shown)


bizob2828:
Moving to draft until we can confirm this actually addresses leaks

bizob2828:
I'm going to close this until we have evidence this is an issue with customers

PR #4098, opened 2026-07-07
chore: Attempted to fix memcached CI issue again

jsumners-nr:
This is another attempt to resolve our woes with the memcached test in GHA. Personally, I find the explanation in the comments confusing, but I'm willing to give it a go. 

-----

Previous attempt (PR #4093): Diagnosed the flake as sub-millisecond measurement noise and added a minDelta = 2ms absolute floor to a symmetric ratio check (diff <= 2ms || ratio <= 20%).

Why it failed again: The new CI failure was segment duration (1.62ms) vs actual duration (4.56ms) — a diff of 2.94ms, which blows past both the 2ms floor and the 20% ratio. My 2ms floor was just a bigger number on the same broken symmetric check; the underlying gap is unbounded on a loaded runner, so no fixed floor would have held.

The actual root cause

The two measurements aren't symmetric noise around a true value. Both are taken from segment.timer.hrstart, but:

- segmentDuration is frozen when the memcached op completes (timer.end()).
- actualTime is sampled strictly later — after the callback resolves its promise and the await continuation resumes.

So actualDuration structurally always contains the segment window plus post-completion event-loop scheduling delay. On a contended CI runner that delay grew to ~3ms, making actualDuration >> segmentDuration. That direction is pure scheduling noise and can never indicate a duration bug — yet the symmetric check treated it as failure.

The fix

Made the comparison asymmetric. actualDuration > segmentDuration is now always tolerated (it's the noise-only direction). The ratio/floor gate applies only to the reverse direction — segmentDuration overshooting the observed wall-clock window — which is the only direction that can actually reveal a mis-measured segment.

### DIFF
--- a/test/lib/custom-assertions/assert-segment-duration.js
+++ b/test/lib/custom-assertions/assert-segment-duration.js
@@ -16,6 +16,20 @@
  * runners the diagnostics channel and subscriber overhead between that clock
  * and `hrstart` is large enough to make any fixed threshold unreliable.
  *
+ * The comparison is intentionally **asymmetric**. The segment's duration is
+ * frozen when the instrumented operation completes (`timer.end()`/`touch()`),
+ * whereas `actualTime` is sampled strictly later — after the callback resolves
+ * its promise and the awaiting continuation resumes. That ordering means
+ * `actualTime` structurally contains the segment window plus however long the
+ * event loop took to schedule the continuation. So `actualDuration >
+ * segmentDuration` is expected, and its magnitude is post-completion
+ * scheduling noise (unbounded on a loaded runner), never evidence of a
+ * duration bug — a symmetric percentage gate on that direction is what made
+ * this assertion flaky. We therefore forgive `actualDuration > segmentDuration`
+ * freely and only enforce the ratio in the direction that can actually reveal
+ * a mis-measured segment: `segmentDuration` running meaningfully longer than
+ * the real wall-clock window it was observed within.
+ *
  * Capture `actualTime` as the very first thing after the awaited operation —
  * before any assertions, event lookups, or other work — to keep the
  * measurement window as small as possible:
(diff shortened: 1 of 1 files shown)


test/lib/custom-assertions/assert-segment-duration.js
```diff
@@ -16,6 +16,20 @@
...
+ * `actualTime` structurally contains the segment window plus however long the
+ * event loop took to schedule the continuation. So `actualDuration >
+ * segmentDuration` is expected, and its magnitude is post-completion
+ * scheduling noise (unbounded on a loaded runner), never evidence of a
+ * duration bug — a symmetric percentage gate on that direction is what made
+ * this assertion flaky. We therefore forgive `actualDuration > segmentDuration`
+ * freely and only enforce the ratio in the direction that can actually reveal
+ * a mis-measured segment: `segmentDuration` running meaningfully longer than
+ * the real wall-clock window it was observed within.
+ *
```
amychisholm03, line 32: I feel like Claude makes comments unnecessarily long. This could be summarized a bit.

jsumners-nr: Sure. Would you like to offer a suggestion change?

amychisholm03, line 32:
```suggestion
 *
 * Asymmetric by design: `actualTime` is sampled after the continuation resumes,
 * so it carries event-loop scheduling delay on top of the segment window.
 * `actualDuration` > `segmentDuration` is expected noise, not a bug. Only the
 * reverse (`segmentDuration` exceeding its observed window) is a real error.
 *
```

jsumners-nr: I prefer the current comment. The suggestion is too sparse for me. I think that if something deserves a comment, the comment should be thorough and easy (relatively) to understand when read again months later. The suggestion is assuming too much intimate knowledge of the problem. The current comment explains why the algorithm is designed the way it is.

amychisholm03: Sure we can keep it as-is.

amychisholm03 (approved):

PR #4101, opened 2026-07-07
chore: Shard versioned tests across CI runners

### LINKED ISSUE #4104: Improve continuous integration run time
Currently, our continuous integration (CI) for every pull request takes around 20 minutes to run. Given that our CI is routinely flaky (due to various factors both internal and external), this means we waste a lot of time waiting on CI to run. We should be able to refactor our CI configuration such that the work completes much faster.

jsumners-nr:
This PR resolves #4104.

This significantly reduces the amount of time needed to run our versioned tests in CI. The only other thing I can think of that we could do would be to add a pre-tests step that installs the agent dependencies and caches them for all of the various shards to re-use. But I think this complicates the workflow enough, and the dependencies step is short compared to everything else. So I am omitting it for now.

The big problem is all of the Docker services we rely upon. I tried loading them up in to groups at the top (i.e. group all Docker dependent services, then alphabetize, and then shard), but that resulted in test timeouts. It seems best to leave them in a somewhat random configuration.

-----

## Motivation

Measured per-step timings showed the **entire CI critical path was a single job leg: `versioned`**. Everything else (lint, ci, unit, integration) finished 10+ minutes earlier and did not gate completion.

Breakdown of the slowest PR versioned leg (~17m24s):
- checkout + setup Node: ~10s
- `npm install`: 33s
- `docker compose up --wait` (all 12 services): ~2m14s
- **Run Versioned Tests: ~14m20s** ← the long pole

The three `node-version` legs already ran in parallel, so the slowest leg alone set the finish, bottlenecked by `JOBS=4` on a 2-vCPU runner.

## What this does

1. **Shards the versioned suites across the matrix.** A new `versioned-setup` job runs `bin/versioned-runner/shard-plan.js`, which lists the suites under `test/versioned/` alphabetically and splits them into contiguous chunks of at most `SHARD_SIZE` (5) suites — 57 suites → **12 shards**. The `versioned` job gains a `shard` matrix dimension (3 node-versions × 12 shards = **36 legs**), each running its subset via a new `VERSIONED_DIRS` env var. Contiguous alphabetical chunks keep adjacent suites together and balance far better than round-robin (which interleaved the sorted list and produced a straggler shard).

2. **Starts only the docker services each shard needs.** Each service-using suite declares a top-level `dockerServices` array in its `package.json` (~20 suites; absent means none). `shard-plan.js` computes the per-shard service union and emits a `servicemap`; the `versioned` job starts only those services via a new arg-aware `bin/start-services.sh`, and **skips the Docker step entirely** for service-free shards (5 of 12). An unknown service name fails the setup job. `npm run services` with no args still starts everything (local dev unchanged).

3. **Fixes latent service-readiness flakes exposed by scoping.** Starting all 12 services previously blocked ~2m14s on the slow ones, which incidentally gave gate-less services time to become ready. With per-shard startup that grace period is gone, so:
   - Added healthchecks to `rmq`, `zookeeper`, `memcached`, and `pg` (they had none, so `--wait` returned before they accepted connections). `rmq` uses `check_port_connectivity` to gate on the AMQP listener; `pg` (9.2, predates `pg_isready`) probes with `psql`.
   - Set `RABBITMQ_ERLANG_COOKIE` so RabbitMQ skips the on-disk `.erlang.cookie` whose permission race intermittently crashed boot with `eacces`.

Coverage: each shard uploads `versioned-tests-${node}-shard-${i}`; the `codecov` job builds a per-node file list across shards and posts them under the unchanged `versioned-tests-${node}` flag (Codecov merges them), preserving `fail_ci_if_error: true`.

## Results (measured)

End-to-end CI time on a clean single-attempt run ([run 28892475586](https://github.com/newrelic/node-newrelic/actions/runs/28892475586), 50/50 jobs green): **~5m20s**, down from **13–18m** — roughly a **65–70% reduction**.

| Stage | End-to-end |
|---|---|
| Baseline (no sharding) | 13–18m |
| Sharding (contiguous chunks, 12 shards) | ~7m42s |
| + per-shard services, healthchecks, rmq cookie fix | **~5m20s** |

The 36 legs land between ~47s (`winston`/`winston-esm` — now skips Docker) and ~4m30s (`memcached mongodb mongodb-esm mysql mysql2` — DB-heavy, still starts its services).

## Commits

- `chore: Shard versioned tests across CI runners` — setup job, shard matrix, `VERSIONED_DIRS`, sharded coverage artifacts
- `chore: Shard versioned suites into contiguous alphabetical chunks` — `SHARD_SIZE`-based contiguous chunking (replaces round-robin) for balance
- `chore: Start only the docker services each versioned shard needs` — `dockerServices` field, `servicemap`, `bin/start-services.sh`, conditional Docker step
- `fix: Add healthchecks to rmq, zookeeper, memcached, and pg` — so `--wait` gates on real readiness
- `fix: Set RabbitMQ Erlang cookie via environment to avoid boot crash` — removes the `.erlang.cookie` `eacces` race

## Why directory-based sharding

- The runner already accepts multiple positional suite directories; only the shell wrapper needed a change.
- `--strict` only flags files within a listed suite dir, not wholly-omitted suites — verified — so a shard subset is safe. `shard-plan.js` self-verifies full coverage to guard against a silent drop.
- `--pattern` is a fragile substring match (`pg` matches `pg-esm`), so sharding by directory is the robust choice.

## Notes

- 36 legs vs the previous 3 — more runner-minutes, but the goal was wall-clock; runner cost was accepted.
- Under-declaring a suite's services would fail loudly on its shard (no blanket "start everything" fallback). The `dockerServices` mapping is evidence-backed from each suite's real connections.
- Out of scope: a pre-existing `amqplib/promises.test.js` wall-clock timing assertion can flake under CPU contention (did not fire on the final run); tracked separately.

### DIFF
--- a/bin/run-versioned-tests.sh
+++ b/bin/run-versioned-tests.sh
@@ -38,7 +38,16 @@ fi
 
 set -f
 directories=()
-if [[ "$1" != '' ]];
+if [[ -n "${VERSIONED_DIRS}" ]];
+then
+  # VERSIONED_DIRS is a space separated list of suite subdir names, used by CI
+  # to run a shard (subset) of the versioned suites. Word splitting is safe here
+  # because `set -f` is enabled above and suite names are simple identifiers.
+  for d in ${VERSIONED_DIRS};
+  do
+    directories+=( "test/versioned/${d}" )
+  done
+elif [[ "$1" != '' ]];
 then
   directories=(
     "test/versioned/${1}"
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -18,6 +18,7 @@ specific functionality
 - After running a specific versioned test suite at least once, individual
 test files in the suite may be run directly as
 `node --test test/versioned/<suite>/<test_file>`
+- When opening Pull Requests, always open them in draft mode
 
 ## Commands
 npm run services:start # start Docker services
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -175,9 +175,31 @@ jobs:
           name: integration-tests-esm-${{ matrix.node-version }}
           path: ./coverage/integration-esm/lcov.info
 
+  # Partition the versioned test suites across shards so the matrix below can
+  # run them on separate runners in parallel. Emits a shard list and a
+  # shard-index -> suite-dirs map consumed by the `versioned` job matrix.
+  versioned-setup:
+    needs:
+      - should_run
+    if: github.event_name == 'workflow_dispatch' ||
+      (needs.should_run.outputs.javascript_changed == 'true' ||
+      needs.should_run.outputs.deps_changed == 'true')
+    runs-on: ubuntu-latest
+    outputs:
+      shards: ${{ steps.plan.outputs.shards }}
+      dirmap: ${{ steps.plan.outputs.dirmap }}
+    steps:
+      - uses: actions/checkout@v6
+      - name: Plan versioned shards
+        id: plan
+        run: node ./bin/versioned-runner/shard-plan.js
+        env:
+          SHARD_COUNT: ${{ github.ref == 'refs/heads/main' && 8 || 4 }}
+
   versioned:
     needs:
       - should_run
+      - versioned-setup
     if: github.event_name == 'workflow_dispatch' ||
       (needs.should_run.outputs.javascript_changed == 'true' ||
       needs.should_run.outputs.deps_changed == 'true')
@@ -187,6 +209,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version: [22.x, 24.x, 26.x]
+        shard: ${{ fromJSON(needs.versioned-setup.outputs.shards) }}
 
     steps:
       - uses: actions/checkout@v6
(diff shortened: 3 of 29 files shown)


bizob2828 (commented):
I will review more thoroughly but a few things that came to mind:
 * might be good to group some of the flakey suites into its own shard or have some way to dynamically move suites into the problem shard(s)
 * now that we've spread the load horizontally, I'm not sure the 16 core runner is needed on main.
 * since this is fast, thoughts on running full versioned runs on PRs?

jsumners-nr:
> * might be good to group some of the flakey suites into its own shard or have some way to dynamically move suites into the problem shard(s)

I think we should work to resolve the flakes so that we don't have to come up with some categorization system for them. #4098 should resolve our biggest issue in this regard (maybe, possibly, 🤞). But I think I also addressed this above:

> The big problem is all of the Docker services we rely upon. I tried loading them up in to groups at the top (i.e. group all Docker dependent services, then alphabetize, and then shard), but that resulted in test timeouts. It seems best to leave them in a somewhat random configuration.

-----

> * now that we've spread the load horizontally, I'm not sure the 16 core runner is needed on main.

That may be true. I'd like to see how things play out for a while, though.

-----

> * since this is fast, thoughts on running full versioned runs on PRs?

Same as above. I'd like to see how well this works for a bit before we get ambitious.

bizob2828:
> > * now that we've spread the load horizontally, I'm not sure the 16 core runner is needed on main.
> 
> That may be true. I'd like to see how things play out for a while, though.

The issue is we have a 16 core runner that will be now used to run shards of suites 5 at a time. It is probably too big for that and we also have to pay for it. I guess we can see how it does on first run but we should fast follow removing the NR_LOADER logic as I don't think it's needed any longer in the new workflow

bin/run-versioned-tests.sh
```diff
@@ -38,7 +38,16 @@ fi
 
 set -f
 directories=()
-if [[ "$1" != '' ]];
+if [[ -n "${VERSIONED_DIRS}" ]];
```
bizob2828, line 41: this is nice but i'm wondering if we just update the runner to allow for:

```sh
npm run versioned:major pino fastify express
```

as an example that would run all 3 suites instead of doing

```sh
VERSIONED_DIRS="pino fastify express" npm run versioned:major
```

jsumners-nr: Good idea. I'd like to do it in a follow up, though, so we can focus on whatever that ends up looking like.

jsumners-nr: I created issue https://github.com/newrelic/node-newrelic/issues/4105 to track this work.

bizob2828 (dismissed):
CI is failing but this is a great start at speeding up versioned tests

bizob2828 (approved):

PR #4108, opened 2026-07-08
fix: Updated `winston` instrumenting to assign `handleExceptions` to the first NrTransport to avoid registering duplicate `uncaughtException` handlers

bizob2828:
## Description

In #4092 a customer was reporting linear growth.  Although they haven't found the root cause, they pointed to a possible leak in our agent.
We do not handle where a customer is registering more than 1 winston logger instance and our `NrTransport` is assigning an uncaughtException handler 
n times.  We should only be creating one `NrTransport` for an agent instance to handle and forward uncaught exceptions.  This PR addresses that and adds
tests to ensure it is functioning as expected.

## How to Test

```sh
npm run versioned winston
```

### DIFF
--- a/lib/subscribers/winston/configure.js
+++ b/lib/subscribers/winston/configure.js
@@ -17,6 +17,9 @@ module.exports = class WinstonConfigure extends ApplicationLogsSubscriber {
   constructor({ agent, logger }) {
     super({ agent, logger, channelName: 'nr_configure', packageName: 'winston' })
     this.events = ['end']
+    // we only want to assign `handleExceptions` to the first NrTransport
+    // customers could be constructing multiple logger instances
+    this.handlingExceptions = false
   }
 
   handler(data, ctx) {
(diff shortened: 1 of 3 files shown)


lib/subscribers/winston/configure.js
```diff
@@ -17,6 +17,9 @@ module.exports = class WinstonConfigure extends ApplicationLogsSubscriber {
   constructor({ agent, logger }) {
```
jsumners-nr, line 21: We should add something like:

```js
* Handles `NrTransport` injection ...
*
* @property {boolean} handlingExceptions Indicates if this instrumentation has
* already registered a global exception handler or not. We only need a single
* exception handler per process.
*/
```

lib/subscribers/winston/nr-winston-transport.js
```diff
@@ -17,9 +17,6 @@ const { truncate } = require('../../util/application-logging')
  */
 class NrTransport extends TransportStream {
```
jsumners-nr, line 24: We should add a `@property` doc line for `handleExceptions` since it still gets set to `true` eventually.

jsumners-nr (changes requested):
Looks good. But I'd like to see a bit more documentation around the utilized properties. This should help future readers understand what is expected.

jsumners-nr (approved):

PR #4112, opened 2026-07-08
test: Added `samples` to prisma and next to cut down on versions under test on packages we recommend using hybrid ageninstrumentation

bizob2828:
## Description

Now that we have sharded versioned tests, there are 2 suites that take much longer than any others:
 nextjs, and prisma.  Prisma takes upwards of 20 mins on node 22 and next takes around 7. Since we recommend our customers rely on native otel instrumentation with hybrid agent, we should reduce the test matrix. I have applied a samples flag to all 3: prisma < 7 sample at 2 because there are only 2 major versions we support, 7+ sample at 5. Nextjs sample at 2 because we only support 2 major versions.

### DIFF
--- a/test/versioned/nextjs/package.json
+++ b/test/versioned/nextjs/package.json
@@ -9,7 +9,10 @@
         "node": ">=22"
       },
       "dependencies": {
-        "next": ">=14.0.0 <16"
+        "next": {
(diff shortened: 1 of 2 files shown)


jsumners-nr (approved):
Fabulous!

PR #4113, opened 2026-07-09
test: Removed timing race from common.request unit tests

### LINKED ISSUE #4106: Resolve flake in unit tests
In https://github.com/newrelic/node-newrelic/actions/runs/28961699457/job/85934629110?pr=4101 we can see a flaky test that errors with "socket hang up." We have been seeing this exact error for months, but have assumed it was just the GitHub Actions environment messing up network connections. However, having the bot investigate the issue has illustrated that we may be able to solve it ourselves:

```
What failed

The unit (22.x) job — not a versioned/docker shard. It's unrelated to the sharding, Docker, or timeout work on this PR. 3 unit tests failed, all in test/unit/utilization/common.test.js, under the common.request block (5099 passed, 3 failed).

Root cause: these are timing-sensitive flakes

The tests exercise HTTP request timeout behavior against a mocked endpoint set up with a hard-coded delay and tight millisecond timeouts:

nock('http://fakedomain').persist().get('/timeout').delay(150).reply(200, 'woohoo')
...
// "should not timeout when request succeeds"
timeout: 200, path: '/timeout'        // 200ms timeout vs a 150ms delayed reply
setTimeout(verifyInvocations, 250)     // verify after 250ms

The test asserts that a request with a 200ms timeout succeeds against a reply delayed 150ms — only a 50ms margin. The failure shows socket hang up and ifError got unwanted exception, meaning the request timed out when it should have succeeded. On a loaded 2-CPU CI runner, the event loop stalls enough that a 150ms-delayed response doesn't land within the 200ms window, so the socket is aborted. The companion error — generated asynchronous activity after the test ended → 0 == 1 — is the same race: the callback fires (or doesn't) outside the expected window and trips an assertion after end().

These are classic wall-clock timing flakes: the pass/fail depends on whether the machine can service a ~50ms margin, which isn't guaranteed under CI contention.

Is it related to our PR?

No. This is a pre-existing unit test with narrow timing margins; the branch even carries a recent Addressed long standing CI flakes commit, so this family is known. Nothing in the sharding/Docker/timeout changes touches lib/utilization/common.js or these tests. A re-run would very likely pass.

What could resolve it (options, no changes made)

The fix belongs in the test's timing assumptions, not in our CI infra:

1. Widen the margins. The reliable fix: increase the success test's request timeout (e.g. 200ms → 500ms+) and/or reduce the mock delay, so the pass condition isn't a ~50ms race. Same idea for the setTimeout(verifyInvocations, 250) guard.
2. Use fake timers instead of real ones. Drive the delay/timeout with sinon fake timers (or node:test mock timers) so the behavior is deterministic and independent of machine load. The test comment even acknowledges this is hard to do synchronously today — but it's the robust path.
3. Decouple from wall-clock entirely. Assert on the request/timeout logic via injected clock or by stubbing the transport, rather than measuring real elapsed time.
4. Cheapest stopgap: re-run the job — as timing flakes, they usually pass on retry — but that doesn't fix the underlying fragility.

Recommendation: option 1 (widen margins) is the smallest durable fix; option 2 is the proper one. But note this is a separate, pre-existing flake from the sharding work — I'd treat it as its own change/issue rather than folding it into this PR, unless you want it bundled.
```

jsumners-nr:
This PR resolves #4106.

This one was a bit tricky. The flakiness is 100% on the slow GHA system. But the "success" test is rather flaky by design, and I'm not quite sure it provides us with much utility. Getting the bot to not mess up the fix was a chore.

-----

## Problem

The two `common.request` tests in `test/unit/utilization/common.test.js` raced a fixed 150ms `nock` `.delay()` against the request timeout with ~50ms margins:

- **success test**: `timeout: 200` vs `delay(150)` — under CI load the slow response lands after the timeout, the socket aborts, and the test fails with `socket hang up` / `ifError got unwanted exception`.
- **timeout test**: `timeout: 100` vs `delay(150)`, plus a real-clock `setTimeout(..., 200)` observation poll.

These margins are reliable on a developer machine but too tight on a contended runner like GitHub Actions, producing the long-standing intermittent `socket hang up` flake (NR-588003).

## Intent being preserved

The success test is **not** merely "a request succeeds." Its purpose is that a **slow** response — slower than a baseline but still within the timeout — is accepted by the client and does **not** cause the socket timeout to fire and invoke the callback a second time (which would look like a spurious retry). The narrow original timings were deliberately inducing that slow-but-within-timeout condition. The timeout test's purpose is the converse: on an actual timeout, the callback is invoked exactly once with `ECONNRESET`. Both tests issue the request, then wait an observation window and verify the callback fired exactly once.

## Fix

Keep the behaviors and structure; remove only the fragile margins by making each path **decisively** on one side of its timeout, and widening the windows well beyond event-loop jitter:

- **Success path** (`/success`): responds slowly — `delay(500)` — but with a `2000ms` request timeout and a `1000ms` observation window (past the 500ms response). Still exercises "accept a slow response without a spurious timeout/retry," but with ~1.5s of headroom instead of ~50ms.
- **Timeout path** (`/timeout`): `delay(10000)` against a `100ms` timeout, so the socket timeout always fires first and yields `ECONNRESET`. After the abort, `nock.abortPendingRequests()` cancels the still-pending delayed response so its timer doesn't keep the event loop alive for the full delay.
- Both tests retain the original **issue request → wait an observation window → assert callback invoked exactly once** structure. The windows are no longer timing-sensitive because the success/timeout outcome is now decided by the interceptor delays, not by racing the window.

Fake timers were considered but rejected: `nock`'s `.delay()` and the socket `timeout` both use real timers deep in the http/socket machinery — exactly what the old code comment ("sinon and http don't quite seem to work well enough") already found. Making the delays decisive removes the race without fighting that.

## Verification

Test-only change; `lib/utilization/common.js` is untouched and both branches (slow-success acceptance + `ECONNRESET` abort) are still exercised. Ran the file repeatedly, including **25 iterations under full CPU saturation** (the condition that triggers the flake): 0 failures.

### DIFF
--- a/test/unit/utilization/common.test.js
+++ b/test/unit/utilization/common.test.js
@@ -84,13 +84,17 @@ test('Utilization Common Components', async function (t) {
   await t.test('common.request', async (t) => {
     t.before(() => {
       nock.disableNetConnect()
-      // We need separate paths for success and failure so that we do not
-      // have to rely on fragile millisecond based timings. While timings of
-      // 100 - 150ms works reliably on individual developer systems, they are
-      // too tight on contendend systems like GitHub Actions.
+      // We use separate paths, each with a delay that is decisively on one side
+      // of its request timeout, so the outcome does not depend on fragile
+      // millisecond margins. The original single interceptor used delay(150)
+      // against timeouts of 150-200ms, which is reliable on a developer machine
+      // but flaky on a contended runner like GitHub Actions.
 
-      // `/success` replies immediately so the success path never times out
-      nock('http://fakedomain').persist().get('/success').reply(200, 'woohoo')
+      // `/success` responds slowly (500ms) but well within its 2000ms request
+      // timeout. This still exercises the "slow response" behavior the success
+      // test cares about -- the client must accept a slow response and must not
+      // treat it as a timeout -- without racing the timeout.
+      nock('http://fakedomain').persist().get('/success').delay(500).reply(200, 'woohoo')
 
       // `/timeout` delays far longer than any request timeout so the socket
       // timeout always fires first.
@@ -115,13 +119,15 @@ test('Utilization Common Components', async function (t) {
     await t.test('should not timeout when request succeeds', (ctx, end) => {
       const agent = ctx.nr.agent
       let invocationCount = 0
+      // The response is slow (500ms) but comfortably within the 2000ms request
+      // timeout. The client must accept the slow response and must not let the
+      // socket timeout fire and invoke the callback a second time (which would
+      // look like a spurious retry).
       common.request(
         {
           method: 'GET',
           host: 'fakedomain',
-          // We set a high timeout here in order to give the request time to
-          // complete on contended systems like GitHub Actions.
-          timeout: 1000,
+          timeout: 2000,
           path: '/success'
         },
         agent,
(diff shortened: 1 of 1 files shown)


bizob2828 (approved):
LGTM

PR #4114, opened 2026-07-09
fix: Pin CI npm to 11 so native versioned suites build on Node 22

jsumners-nr:
## Problem

`versioned (22.x, 9)` intermittently fails in the `pg`/`pg-esm` suites with `Could not locate the bindings file` for `libpq`. The `force-native`/`native` tests exercise the native postgres client (`pg-native` → `libpq`), whose `addon.node` is built by an install script.

Root cause (confirmed from the install log): the CI `Upgrade npm` step (Node 22 only) upgraded npm to `npm@latest`, which is now **npm 12**. npm 12 added an **`allowScripts` allow-list gate** separate from `ignore-scripts`:

```
npm warn install-scripts 1 package had install scripts blocked because they are not covered by allowScripts:
npm warn install-scripts   libpq@1.11.0 (install: node-gyp rebuild)
```

So even with the suite's `ignore-scripts=false` opt-in, `libpq`'s build is blocked and `addon.node` is never produced. It only affected the 22.x leg because only that leg upgrades npm; 24.x/26.x (bundled older npm) and local (npm 10) have no such gate.

There is **no viable way to allow the script through the runner's install**: `--allow-scripts`/`NPM_CONFIG_ALLOW_SCRIPTS` are rejected in project-scoped installs (`EALLOWSCRIPTS`), the suite `package.json` is never read by the runner's `npm install <pkgs> --no-save`, and a suite-local `.npmrc` is never read either.

## Fix

Pin the `Upgrade npm` step to **`npm@11`**. npm 11 has no `allowScripts` gate, so `libpq` builds via the existing `ignore-scripts=false` opt-in (verified end-to-end on Node 22: `npm run versioned:major pg-esm` passes). This is the minimal change and matches the npm behavior on the other Node versions and locally.

## Verification

- Reproduced the npm 12 `allowScripts` block and confirmed npm 11 builds `libpq` with only `ignore-scripts=false`.
- `npm run versioned:major pg-esm` passes on Node 22 locally.

Note: a temporary no-op comment in `index.js` is included to make the `should_run` filter run the full battery on this branch; it will be reverted before merge.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -119,7 +119,7 @@ jobs:
       # This step can be removed when we drop support for Node 22
       - name: Upgrade npm
         if: matrix.node-version == '22.x'
-        run: npm install -g npm@~11.10.0 && npm install -g npm@latest
+        run: npm install -g npm@11
       - name: Install Dependencies
         run: npm install
       - name: Run Unit Tests
@@ -156,7 +156,7 @@ jobs:
       # This step can be removed when we drop support for Node 22
       - name: Upgrade npm
         if: matrix.node-version == '22.x'
-        run: npm install -g npm@~11.10.0 && npm install -g npm@latest
+        run: npm install -g npm@11
       - name: Install Dependencies
         run: npm install
       - name: Run Integration Tests
@@ -223,7 +223,7 @@ jobs:
       # This step can be removed when we drop support for Node 22
       - name: Upgrade npm
         if: matrix.node-version == '22.x'
-        run: npm install -g npm@~11.10.0 && npm install -g npm@latest
+        run: npm install -g npm@11
       - name: Install Dependencies
         run: npm install
       # Start only the docker services this shard's suites need. Shards whose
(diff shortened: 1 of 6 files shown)


bizob2828 (commented):
you'll also have to change `versioned-security-agent.yml` as well

.github/workflows/ci-workflow.yml
```diff
@@ -119,7 +119,7 @@ jobs:
       # This step can be removed when we drop support for Node 22
       - name: Upgrade npm
         if: matrix.node-version == '22.x'
-        run: npm install -g npm@~11.10.0 && npm install -g npm@latest
+        run: npm install -g npm@11
```
bizob2828, line 122: fwiw this prob isn't needed

.github/workflows/ci-workflow.yml
```diff
@@ -156,7 +156,7 @@ jobs:
...
```
bizob2828, line 159: fwiw this prob isn't needed

bizob2828 (approved):
this is fine for now but most of these aren't needed as our agent doesn't need to build native modules

PR #4117, opened 2026-07-09
chore: Updated test runner to accept suites as positional parameters

### LINKED ISSUE #4105: Improve test runner to accept multiple suites
Currently, we are able to run specific versioned tests via `npm run versioned <suite_name>`. As part of #4101, we gained the ability to specify multiple suites via an environment variable. Instead, we'd like to be able to do `npm run versioned <one> <two> ... <n>`. Solving this should remove the `VERSIONED_DIRS` environment variable and improve our runner to accept them as arguments.

jsumners-nr:
This PR resolves #4105.

https://github.com/newrelic/node-newrelic/actions/runs/29034218785 shows the changes work as expected. Shards still get created and run with subsets of the overall versioned tests suites.

-----

## What

Run a subset of versioned suites by passing suite names as **positional arguments**:

```
npm run versioned:major amqplib pino winston
```

This replaces the `VERSIONED_DIRS` environment variable that previously carried the shard's suite list.

## Why

Positional arguments are the natural CLI form and match how a single suite was already invoked (`npm run versioned:major <suite>`). The env var was an extra, less discoverable mechanism.

## Changes

- **`bin/run-versioned-tests.sh`**: build the `directories` list from all positional arguments (`"$@"`) when any are given; otherwise run the whole tree. Previously only `$1` was honored, so `npm run versioned:major amqplib pino winston` would have silently dropped all but the first suite — that's now fixed, and the separate single-`$1` branch is subsumed by the loop.
- **`.github/workflows/ci-workflow.yml`**: the `versioned` job passes the shard's suite list as arguments (`npm run versioned <dirs>`) instead of via `VERSIONED_DIRS`.

## Verification

- `npm run versioned:major amqplib pino winston` → all three suites reach the runner; single-suite and no-arg (whole tree) invocations still work.
- npm forwards trailing args through the full `versioned:major` → `versioned` → script chain (verified).
- No remaining `VERSIONED_DIRS` references; every shard's suite list is non-empty (whole-tree fallback cannot be accidentally triggered); scripts unit tests pass.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -233,15 +233,15 @@ jobs:
         run: ./bin/start-services.sh $SERVICES
         env:
           SERVICES: ${{ fromJSON(needs.versioned-setup.outputs.servicemap)[matrix.shard] }}
+      # The suite names for this shard are passed as positional arguments so the
+      # runner only executes this shard's subset of the versioned suites.
       - name: Run Versioned Tests
-        run: TEST_CHILD_TIMEOUT=600000 npm run versioned
+        run: TEST_CHILD_TIMEOUT=600000 npm run versioned ${{ fromJSON(needs.versioned-setup.outputs.dirmap)[matrix.shard] }}
         env:
           VERSIONED_MODE: ${{ github.ref == 'refs/heads/main' && '--minor' || '--major' }}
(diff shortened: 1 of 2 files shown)


amychisholm03 (approved):

bizob2828:
nice. verified this works locally

PR #4118, opened 2026-07-09
test: Fixed flaky amqplib async-work duration assertions

jsumners-nr:
In #4093 we attempted to fix flakiness in the `amqplib` tests. We only partially succeeded. This PR attempt to fully resolve the problem.

-----

## Problem

The amqplib versioned suite flakes intermittently (e.g. on the 22.x shard) with:

```
not ok - consume with async message handler ...
error: 'transaction should account for async work'
  test/versioned/amqplib/promises.test.js:453
```

The "consume with async message handler" tests (in both `promises.test.js` and `callback.test.js`) asserted:

```js
assert.ok(tx.trace.getDurationInMillis() >= PROMISE_WAIT, 'transaction should account for async work')
assert.ok(tx.baseSegment.getDurationInMillis() >= PROMISE_WAIT, 'base segment should account for async work')
```

`PROMISE_WAIT` is a fixed 100ms floor. The segment's duration is frozen when the instrumented operation completes, so on a loaded CI runner it can be measured as just under 100ms, failing the assertion. This is post-completion scheduling noise, not a real defect.

## Why PR #4093 didn't fix it

[#4093](https://github.com/newrelic/node-newrelic/pull/4093) diagnosed exactly this class of flake and added the `assertSegmentDuration` helper (`test/lib/custom-assertions/assert-segment-duration.js`), whose asymmetric comparison tolerates the scheduling-noise direction. But it applied the helper only to a **unit** test and separately changed `callback.test.js`'s `agent.on`→`agent.once` (a different, listener-leak flake). The amqplib suites' **duration assertions** were never migrated, so the fixed-floor pattern remained in both `promises.test.js` and `callback.test.js`.

## Fix

Migrate the remaining async-work duration assertions in both files to `assertSegmentDuration`, capturing `actualTime = process.hrtime(tx.baseSegment.timer.hrstart)` and asserting the segment duration against the observed wall-clock rather than a fixed floor.

## Verification

Test-only. Ran each file the way the versioned runner does (one file per process): 0 failures across many iterations, including under CPU contention (half- and full-core load), where the fixed-floor version reproduced the flake.

### DIFF
--- a/test/versioned/amqplib/callback.test.js
+++ b/test/versioned/amqplib/callback.test.js
@@ -15,6 +15,7 @@ const promiseResolvers = require('../../lib/promise-resolvers')
 const getPackageVersion = require('../../lib/get-package-version')
 
 const { assertPackageMetrics } = require('../../lib/custom-assertions')
+const assertSegmentDuration = require('../../lib/custom-assertions/assert-segment-duration')
 const PROMISE_WAIT = 100
 
 const version = getPackageVersion({ pkgName: 'amqplib', baseDir: __dirname })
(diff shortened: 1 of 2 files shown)


bizob2828 (approved):
LGTM, hopefully this stabilizes this suite

PR #4120, opened 2026-07-09
test: Added more scenarios to winston for constructing loggers

bizob2828:
## Description

Although we haven't 100% confirmed yet from customer the theory is that the `NrTransport` leak, fixed in #4108 was never in play for the customer in `newrelic@<13.17.0` because their logger instances weren't getting instrumented.  When we migrated the instrumentation to a subscriber based model, we're instrumenting the underlying `DerivedLogger`, so now their logger constructions were getting instrumented and leaking because of the bug mentioned before.

The theory is that they were constructing loggers like:

```js
 const logger = new winston.Logger({
      transports: [new winston.transports.Stream({ level: 'info', stream: jsonStream })]
    })
    logger.log('info', 'test')
```

which weren't getting instrumented until 13.17.0.  This PR adds more test coverage around 2 other methods of constructing loggers.  If you run these same tests in 13.16.0, `should forward logs from a logger created via new winston.Logger` fails. 

See #4092 for more context of the entire conversation leading to the fix in 14.2.1 and this PR

### DIFF
--- a/test/versioned/winston/winston.test.js
+++ b/test/versioned/winston/winston.test.js
@@ -408,6 +408,56 @@ test('log forwarding enabled', async (t) => {
     assert.equal(added, 1, 'should add exactly one uncaughtException listener for 25 loggers')
   })
 
+  await t.test('should forward logs from a logger created via `new winston.Logger`', (t, end) => {
+    const { agent, winston } = t.nr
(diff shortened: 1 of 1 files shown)


jsumners-nr (approved):
Such a wacky logger.

PR #4121, opened 2026-07-09
test: Added skip for winston test for ranges that do not support it

bizob2828:
## Description

In #4120 I added more winston tests. Turns out the one `new winston.Logger` is invalid for a range between 3.0.0-3.9.x. This Pr updates that check.

### DIFF
--- a/test/versioned/winston/winston.test.js
+++ b/test/versioned/winston/winston.test.js
@@ -10,9 +10,9 @@ const assert = require('node:assert')
 const semver = require('semver')
 const { Writable } = require('node:stream')
 
-const { version: pkgVersion } = require('winston/package')
-
 const helper = require('../../lib/agent_helper')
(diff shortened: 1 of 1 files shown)


test/versioned/winston/winston.test.js
```diff
@@ -7,8 +7,11 @@
 
 const test = require('node:test')
 const assert = require('node:assert')
+const semver = require('semver')
 const { Writable } = require('node:stream')
 
+const { version: pkgVersion } = require('winston/package')
```
jsumners-nr: Should https://github.com/newrelic/node-newrelic/blob/f6123bc939afc0122443fda3eab7d9e842dcbfef/test/lib/get-package-version.js be used instead?

bizob2828: yea prob, ill fix

jsumners-nr (approved):

PR #4122, opened 2026-07-09
feat: Add source mapping support to `HeapProfiler`

### LINKED ISSUE #4102: Add source mapping support to `HeapProfiler`
### Description
[PR #4095](https://github.com/newrelic/node-newrelic/pull/4095) added source mapping support for the `CpuProfiler` (wall-time profling). Now, we will need to add this functionality to the `HeapProfiler`. This should be relatively simple because the `SourceMapper` is built in the `ProfilingManager`'s `register` method and we just need to pass the object into the `HeapProfiler` constructor.

### Acceptance Criteria

- [ ] When `heap` profiling is enabled and `profiling.source_mapping.enabled = true`, raw (pre-compiled) source code appears in the flamegraph nodes in the UI

amychisholm03:
## Description

This PR adds source mapping support for our `HeapProfiler`. This is done by passing in our shared `sourceMapper` instance into `pprof.heap.profile()`, which is called in `HeapProfiler.collect()`. Coordinating unit and integrations tests have been added.

### Test Files
```zsh
node --test test/integration/profiling/source-mapping.test.js
node --test test/unit/lib/profiling/index.test.js
node --test test/unit/lib/profiling/profilers/base.test.js
node --test test/unit/lib/profiling/profilers/heap.test.js
```

### With the Example App
Example app: https://github.com/newrelic/newrelic-node-examples/pull/378

With `heap` profiling enabled, and selecting "Objects" in the dropdown on the Profiling UI, you can see `*.ts` files listed instead of `*.js` (expected behavior with source mapping).
image

## Related Issues

Closes #4102


lib/profiling/profilers/heap.js
```diff
@@ -30,7 +30,7 @@ class HeapProfiler extends BaseProfiler {
   }
 
   async collect() {
-    const profile = this.#pprof.heap.profile()
+    const profile = this.#pprof.heap.profile(undefined, this.sourceMapper)
```
amychisholm03, line 33: This first parameter is [`ignoreSamplePath`](https://github.com/DataDog/pprof-nodejs/blob/2e3ad636041332bb943339b1ac5c47a183508df2/ts/src/heap-profiler.ts#L86) which we do not need, so it is always `undefined`.

bizob2828, line 33: might be good to call out that sourceMapper is passed at collection vs start like cpu profiler

bizob2828 (approved):
LGTM

PR #4123, opened 2026-07-09
feat: Added support for `ai_monitoring.enabled`,`ai_monitoring.streaming.enabled`, and `ai_monitoring.record_content.enabled` to be assigned via server-side config

### LINKED ISSUE #4077: Add server side config support to AIM
Spec PR 808 adds a new requirement that all AI monitoring (AIM) settings must be configurable from server side settings. We must evaluate how we handle AIM settings in our agent and make any necessary changes to support the expanded server side configuration outlined in the spec PR. We already support enabling and disabling of AIM from server side.

bizob2828:
## Description

This PR adds support for ai_monitoring configuration via SSC.  The one caveat worth noting is that `ai_monitoring.enabled` takes precedence over `collect_ai` if the value of `ai_monitoring.enabled` is dff than the compiled config.  Per spec:

> The [connect response](https://source.datanerd.us/agents/agent-specs/blob/main/Connect-LEGACY.md#connect-reply-fields) will include a flag collect_ai that returns a boolean value indicating whether AIM was disabled. This value mirrors what the agent has set locally. Additional ai_monitoring.enabled, ai_monitoring.streaming.enabled, and ai_monitoring.record_content.enabled settings are included in the connect response as well. If ai_monitoring.enabled is present in the connect response, this value MUST take precedence to override the ai_monitoring.enabled value and fallback on using the value from collect_ai. For example, if collect_ai is False, and ai_monitoring.enabled locally is True, the agent would ultimately disable AI Monitoring. If both collect_ai is False and ai_monitoring.enabled is True in the connect response, ai_monitoring.enabled would ultimately enable AI Monitoring. If HSM is enabled, the appropriate AI monitoing settings SHOULD be set appropriately in the connect response. As a safeguard however, agents SHOULD also enforce HSM mode on AI settings they recieve from the connect response.

To support this I had to re-tool all the ai monitoring subscribers to all register regardless of if `ai_monitoring.enabled` was true at startup.  I added tests to verify this works and I also split out that check into `this.aiEnabled` and included in the `this.streamingEnabled` check.

**Note**: I verified all scenarios of toggle ai_monitoring config on/off in UI. 

## Related Issues
Closes #4077

### DIFF
--- a/lib/subscribers/ai-monitoring/base.js
+++ b/lib/subscribers/ai-monitoring/base.js
@@ -26,8 +26,14 @@ class AiMonitoringSubscriber extends Subscriber {
     this.trackingPrefix = trackingPrefix
   }
 
-  get enabled() {
-    return super.enabled && this.agent.config.ai_monitoring.enabled === true
+  /**
+   * Checks if AI monitoring is enabled. The subscriber itself is registered
+   * regardless of this value (gated only by `instrumentation.<lib>.enabled`),
+   * but AI monitoring behavior (segments, events, metrics) is gated on this.
+   * @returns {boolean} if AI monitoring is enabled
+   */
+  get aiEnabled() {
+    return this.agent.config.ai_monitoring.enabled === true
   }
 
   set name(name) {
@@ -96,7 +102,7 @@ class AiMonitoringSubscriber extends Subscriber {
    * @returns {Context} either new context or existing
    */
   handler(data, ctx) {
-    if (!this.enabled) {
+    if (!this.aiEnabled) {
       this.logger.debug('`ai_monitoring.enabled` is set to false, not creating segment.')
       return ctx
     }
--- a/lib/subscribers/ai-monitoring/base.js
+++ b/lib/subscribers/ai-monitoring/base.js
@@ -36,6 +36,15 @@ class AiMonitoringSubscriber extends Subscriber {
     return this.agent.config.ai_monitoring.enabled === true
   }
 
+  /**
+   * Checks if AI monitoring streaming is enabled.  It is predicated on AI monitoring being enabled
+   * as if it isn't, no instrumentation should occur
+   * @returns {boolean} if we should be instrumenting streams
+   */
+  get streamingEnabled() {
+    return this.aiEnabled && this.agent.config.ai_monitoring.streaming.enabled === true
+  }
+
   set name(name) {
     if (!name) {
       throw new Error('subscriber must pass in `name`')
--- a/lib/subscribers/ai-monitoring/chat.js
+++ b/lib/subscribers/ai-monitoring/chat.js
@@ -24,7 +24,7 @@ class AiMonitoringChatSubscriber extends AiMonitoringSubscriber {
   }
 
   get streamingEnabled() {
-    return this.agent.config.ai_monitoring.streaming.enabled === true
+    return this.aiEnabled && this.agent.config.ai_monitoring.streaming.enabled === true
   }
 
   /**
@@ -86,7 +86,7 @@ class AiMonitoringChatSubscriber extends AiMonitoringSubscriber {
    * @param {Array} params.tags used only for langchain events at the moment
    */
   recordChatCompletionEvents({ ctx, request, response, err, timeOfFirstToken, metadata = {}, tags = [] }) {
-    if (!this.enabled) {
+    if (!this.aiEnabled) {
       this.logger.debug('config.ai_monitoring.enabled is set to false, not creating chat completion events.')
       return
     }
--- a/lib/subscribers/ai-monitoring/chat.js
+++ b/lib/subscribers/ai-monitoring/chat.js
@@ -23,10 +23,6 @@ class AiMonitoringChatSubscriber extends AiMonitoringSubscriber {
     super({ agent, logger, packageName, channelName, name, trackingPrefix })
   }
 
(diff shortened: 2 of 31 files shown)


lib/subscribers/google-genai/generate-content-stream.js
amychisholm03, line 1: This was probably an oversight in the initial Google Gen AI instrumentation but `google-genai/generate-content.js` and `google-genai/embed-content.js` don't have any `this.aiEnabled` checks before creating `Llm` events.

lib/subscribers/ai-monitoring/base.js
```diff
@@ -26,8 +26,14 @@ class AiMonitoringSubscriber extends Subscriber {
...
 
-  get enabled() {
-    return super.enabled && this.agent.config.ai_monitoring.enabled === true
+  /**
+   * Checks if AI monitoring is enabled. The subscriber itself is registered
+   * regardless of this value (gated only by `instrumentation.<lib>.enabled`),
+   * but AI monitoring behavior (segments, events, metrics) is gated on this.
+   * @returns {boolean} if AI monitoring is enabled
+   */
+  get aiEnabled() {
```
amychisholm03, line 35: We could also do a `aiStreamingEnabled` too because that's SSC as well?

lib/subscribers/ai-monitoring/chat.js
```diff
@@ -24,7 +24,7 @@ class AiMonitoringChatSubscriber extends AiMonitoringSubscriber {
   }
 
   get streamingEnabled() {
-    return this.agent.config.ai_monitoring.streaming.enabled === true
+    return this.aiEnabled && this.agent.config.ai_monitoring.streaming.enabled === true
```
amychisholm03: Oop never mind here it is

amychisholm03: Though I'd argue this should be moved to `ai-monitoring/base.js` so `langgraph/graph-stream.js` can use it without repeating it.

lib/subscribers/langgraph/graph-stream.js
```diff
@@ -22,7 +22,7 @@ class LangGraphStreamSubscriber extends AiMonitoringSubscriber {
   }
 
   get streamingEnabled() {
```
amychisholm03: Should be defined by `AiMonitoringSubscriber` imo

lib/subscribers/ai-monitoring/chat.js
jsumners-nr: Can streaming be enabled independently of `this.aiEnabled`? I doubt it. But regardless, it is probably better to have:

```js
get streamingEnabled() {
	return this.aiStreamingEnabled
}
```

```js
get aiStreamingEnabled () {
	return this.aiEnabled && <config lookup>
}
```

jsumners-nr (commented):
Everything is looking reasonable to me. We can just clean up the streaming detection.

lib/subscribers/ai-monitoring/base.js
bizob2828: I'm not sure I understand. that's being evaluated at execution through `this.streamingEnabled`

lib/subscribers/ai-monitoring/chat.js
bizob2828: sure i mean didn't you do this refactor 😄

bizob2828: no it can't. meaning if ai monitoring is disabled no instrumentation occurs. that's why i'm checking both in `streamingEnabled` now because previously if ai monitoring was disabled, the subscriber was never registered but we have to register them all now because it could be enabled via SSC

lib/subscribers/google-genai/generate-content-stream.js
bizob2828: ok i'll take a look and try to make this more abstract if possible

lib/subscribers/ai-monitoring/chat.js
amychisholm03: Yeah, I guess I didn't realize it at that time. It makes sense that only `AiMonitoringChatSubscriber`s could be streaming, but `langgraph/graph-stream.js`, because of its agentic nature, differed too much from `AiMonitoringChatSubscriber` and had to just extend from just `AiMonitoringSubscriber`.

lib/subscribers/ai-monitoring/base.js
amychisholm03: This was before I saw `streamingEnabled()` in `ai-monitoring/chat.js`. I was thinking something like James' suggestion.

lib/subscribers/google-genai/generate-content-stream.js
bizob2828: so `recordEmbedding` is gated by `this.aiEnabled` so the google-genai/embed-content.js is fine

bizob2828: recordChatCompletionEvents is also gated by `this.aiEnabled`.  I'll take a look at all of these to ensure we rely on the abstractions in subscribers and not do duplicate checks. I think the only place we need to check is if we also instrument streams

jsumners-nr (dismissed):

amychisholm03 (dismissed):

jsumners-nr (approved):

PR #4128, opened 2026-07-13
feat: Added `otlp_resource_attributes` to all outgoing opentelemetry metrics calls

### LINKED ISSUE #4103: Assign `otlp_resource_attributes` on every otel metric
### Description
This [spec](https://source.datanerd.us/agents/agent-specs/pull/821) update, wants us to store the `otlp_resource_attibutes` from the connect service and attach to every outgoing otel metrics transmission.

We will need a mechanism to store these to be accessed later by the SetupMetrics class [here](https://github.com/newrelic/node-newrelic/blob/058fde6d64e3606b82875380ba51f9bc44a4b994/lib/otel/metrics/index.js#L118)

Perhaps we just attach to the config object as `config.otlp_resource_attributes`?

bizob2828:
## Description

This PR adds logic to assign `config.otlp_resource_attributes` from connect response. All the attributes in this object should be posted to otel metrics calls.  

**Note**: `entity.guid` comes in from `otlp_resource_attributes` so we do not have to merge that with the incoming `otlp_resource_attributes`.

## How to Test

```sh
npm run unit
```

## Related Issues
Closes #4103

### DIFF
--- a/lib/otel/metrics/index.js
+++ b/lib/otel/metrics/index.js
@@ -115,8 +115,7 @@ class SetupMetrics extends SetupSignal {
           { agent, logger }
         )
 
-        const resourceAttrs = config.otlp_resource_attributes
-        const resource = resourceFromAttributes(resourceAttrs)
+        const resource = resourceFromAttributes({ 'entity.guid': config.entity_guid, ...config.otlp_resource_attributes })
         // Assigning the resource after having received the `entity.guid` from
         // the server is a key detail of this implementation. Unfortunately,
         // we don't have real public access to the object that retains the
--- a/lib/otel/metrics/index.js
+++ b/lib/otel/metrics/index.js
@@ -115,7 +115,8 @@ class SetupMetrics extends SetupSignal {
           { agent, logger }
         )
 
-        const resource = resourceFromAttributes({ 'entity.guid': config.entity_guid })
+        const resourceAttrs = config.otlp_resource_attributes
+        const resource = resourceFromAttributes(resourceAttrs)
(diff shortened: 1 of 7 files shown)


lib/otel/metrics/index.js
```diff
@@ -115,7 +115,8 @@ class SetupMetrics extends SetupSignal {
           { agent, logger }
         )
 
-        const resource = resourceFromAttributes({ 'entity.guid': config.entity_guid })
+        const resourceAttrs = config.otlp_resource_attributes
+        const resource = resourceFromAttributes(resourceAttrs)
```
jsumners-nr: Is this fully deployed? As in, if we ship this today, will everyone already using the OTEL metrics feature continue working without issue?

bizob2828: I think so but confirming that now

jasonjkeller: I think it should be, but I'm not 100% sure if that is true for the Node-specific fix that was added last week. If you're able to test Node in staging and prod that should probably be a good indication.

bizob2828: yes i confirmed this has been 100% deployed

jsumners-nr (dismissed):
Looks good to me.

jsumners-nr (approved):

sonewman:
how do we use this? is this a property in `newrelic.cjs` ? i am mostly interested in having `appName` and other base attributes added. but i can't see from the code where this is happening

jsumners-nr:
This happens automatically via internal mechanisms. Please open an issue that fully describes your use case and desired outcome.

PR #4130, opened 2026-07-14
fix: Fixed OTEL crash when adding exceptions to spans

### LINKED ISSUE #4129: OTel bridge: FakeSpan implements only spanContext(); Next.js 16 span.recordException() throws uncaughtException
### Description

When the agent runs in OpenTelemetry-bridge mode (`NEW_RELIC_OPENTELEMETRY_ENABLED=true`) alongside a framework that emits its own OTel spans and then operates on the **active** span, the bridge's `FakeSpan` throws, because `FakeSpan` implements only `spanContext()`.

Concretely, with Next.js 16 (which instruments App Router rendering with native OpenTelemetry spans), during **error rendering** Next calls `span.recordException(err)` on the active span. The active span is a New Relic `FakeSpan`, which has no `recordException`, so:

```
TypeError: e.recordException is not a function
```

This surfaces as an **unhandledRejection / uncaughtException** (it happens in the streaming/OTel layer, outside the React error boundary), which crashes the in-flight RSC stream and, under load, crash-loops the Node process.

`lib/otel/fake-span.js` on `main` (and in the latest release, `14.2.1`) still implements only `spanContext()`:

```js
module.exports = class FakeSpan {
  constructor(segment, transaction) { /* ... */ }
  spanContext() { /* ... */ }
}
```

But the OpenTelemetry JS [`Span` interface](https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/trace/span.ts) declares `setAttribute`, `setAttributes`, `addEvent`, `addLink`, `addLinks`, `setStatus`, `updateName`, `recordException`, `end`, and `isRecording` as required members. Any consumer that treats the active span as a real OTel `Span` (per the spec) will throw against `FakeSpan`.

### Steps to Reproduce

1. Node app instrumented with the New Relic Node agent, OTel bridge enabled (`NEW_RELIC_OPENTELEMETRY_ENABLED=true`).
2. Framework that emits OTel spans and calls Span methods beyond `spanContext()` on the active span — e.g. Next.js 16 App Router, which calls `span.recordException(err)` during error rendering.
3. Trigger an error render (e.g. a route that throws). Observe `TypeError: recordException is not a function` as an uncaughtException.

### Expected Behavior

`FakeSpan` should be a drop-in for an OTel `Span`: methods it does not meaningfully support should be safe no-ops (returning `this` where the interface is chainable, `false` for `isRecording()`), so consumers calling the standard Span API don't crash. Recording still flows through New Relic segments; these calls just shouldn't throw.

### Your Environment

- `newrelic` 14.1.0 (also verified `fake-span.js` unchanged through 14.2.1 and `main`)
- Node.js 24
- Next.js 16 (App Router), `NEW_RELIC_OPENTELEMETRY_ENABLED=true`

### Additional context

Workaround we're running: augment `FakeSpan.prototype` with no-op implementations of the remaining `Span` interface methods. Happy to open a PR implementing this in `lib/otel/fake-span.js` if that's a welcome direction.

jsumners-nr:
This PR resolves #4129.

### DIFF
--- a/lib/otel/fake-span.js
+++ b/lib/otel/fake-span.js
@@ -10,12 +10,25 @@
  * within otel bridge. We have to create fake spans for new relic
  * segments.  The only thing needed is a method for `spanContext`
  * which should return the spanId(segment id) and traceId(transaction trace id).
- * We hardcode traceFlags to 1.
+ * We hard code traceFlags to 1.
+ *
+ * @see https://open-telemetry.github.io/opentelemetry-js/interfaces/_opentelemetry_api._opentelemetry_api.Span.html
  */
 module.exports = class FakeSpan {
+  #segment
+  #transaction
+
   constructor(segment, transaction) {
-    this.segmentId = segment.id
-    this.traceId = transaction.traceId
+    this.#segment = segment
+    this.#transaction = transaction
+  }
+
+  get segmentId() {
+    return this.#segment.id
+  }
+
+  get traceId() {
+    return this.#transaction.traceId
   }
 
   spanContext() {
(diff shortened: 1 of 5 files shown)


lib/otel/fake-span.js
```diff
@@ -25,4 +38,103 @@ module.exports = class FakeSpan {
...
+   * @returns {FakeSpan}
+   */
+  setAttributes(attributes) {
+    for (const [k, v] of Object.entries(attributes)) {
+      this.setAttribute(k, v)
+    }
+    return this
+  }
+
+  addEvent(name, _attrsOrTime, _time) {
```
jsumners-nr, line 74: We do not support OTEL events due to spec instability.

lib/otel/fake-span.js
```diff
@@ -10,12 +10,25 @@
...
- * We hardcode traceFlags to 1.
+ * We hard code traceFlags to 1.
+ *
+ * @see https://open-telemetry.github.io/opentelemetry-js/interfaces/_opentelemetry_api._opentelemetry_api.Span.html
  */
 module.exports = class FakeSpan {
+  #segment
+  #transaction
+
   constructor(segment, transaction) {
```
bizob2828, line 25: I'm still not sure we understand this issue completely.  Either way, these methods should  be pure stubs.

bizob2828: Also, this should be reverted so we don't store the entire segment/tx with this fake span

lib/otel/fake-span.js
```diff
@@ -25,4 +38,103 @@ module.exports = class FakeSpan {
...
+   * @returns {FakeSpan}
+   */
+  addLinks(links) {
+    for (const link of links) {
+      this.addLink(link)
+    }
+    return this
+  }
+
+  setStatus(otelSpanStatus) {
```
bizob2828, line 112: i'm unclear why you wouldn't set the status, that's used in the otel bridge

lib/otel/fake-span.js
```diff
@@ -25,4 +38,103 @@ module.exports = class FakeSpan {
...
+    this.#segment.logger.warn(
+      'setStatus is not implemented. Not setting status: %s.',
+      otelSpanStatus?.code
+    )
+    return this
+  }
+
+  updateName(name) {
```
bizob2828, line 120: i'm unclear why you wouldn't set the name

lib/otel/fake-span.js
```diff
@@ -25,4 +38,103 @@ module.exports = class FakeSpan {
...
+    this.#segment.logger.warn(
+      'updateName is not implemented. Not setting name: %s.',
+      name
+    )
+    return this
+  }
+
+  end(_time) {
```
bizob2828, line 128: i'm unclear why you wouldn't end the span

bizob2828 (commented):
a few questions about unimplemented methods

lib/otel/fake-span.js
jsumners-nr: All of the methods I left as stubs are methods I think don't make sense in the context of our bridge. I think we manage those things internally. I can be convinced that `setStatus` should actually do something, but I'm not sure the others need to be updated. We take care of naming segments and we take care of ending them.

bizob2828 (approved):

PR #4133, opened 2026-07-14
chore: Changed smoke-test schedule with our bot user

newrelic-node-agent-team:
The current cron for smoke test is tied to @michaelgoin and we want it to be invoked on behalf of our bot user


.github/workflows/smoke-test-workflow.yml
```diff
@@ -13,7 +13,7 @@ on:
     # Run once a day at 9AM PDT (16 UTC) on week days (1-5).
```
bizob2828:
```suggestion
    # Run once a day at 8AM PDT (15 UTC) on week days (1-5).
```

bizob2828 (approved):

Issue #4126, opened 2026-07-13
Enable collection of OTEL metrics from AWS Lambda


jsumners-nr:
Currently, the collection of metrics from the OTEL metrics API relies on capturing information from the "connect lifecycle" before metrics can be shipped. In AWS Lambda, we do not have a connect lifecycle. In Lambda, we have two methods of collecting data:

1. Through parsing CloudWatch logs
2. Through our [Lambda extension](https://github.com/newrelic/newrelic-lambda-extension)

See "Lambda.md" in our specs for details on how the second mode is implemented.

The key for implementation is: when we serialize the transaction/trace data, the metrics data must be added as a new `otlp_payload` field with the value being base64 encoded protobuf data flushed from the exporter.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-589733

PR #4140, opened 2026-07-15
chore: Attached serverless mode config to agent instance

jsumners-nr:
While conducting research for #4126 I determined that we are referencing `config.serverless_mode.enabled` in all of the places that we need to branch on that setting. I think it makes sense for the agent instance to indicate if it is operating in serverless mode or not. So this PR updates the agent to expose the information directly, and then updates all of the usage locations to reference the agent instead of the parsed config.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -186,7 +186,13 @@ function Agent(config) {
   this.environment = require('./environment')
   this.version = config.version
 
-  if (config.serverless_mode.enabled) {
+  Object.defineProperty(this, 'serverlessMode', {
+    get () {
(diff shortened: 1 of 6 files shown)


lib/agent.js
bizob2828: missed updating this

bizob2828 (changes requested):
missed one place to update

lib/agent.js
jsumners-nr: Strange. My IDE refuses to find that one line.

bizob2828 (approved):

PR #4146, opened 2026-07-15
feat: Collect on-CPU time alongside wall time in `CpuProfiler`

### LINKED ISSUE #4048: Add `cpu` to wall time profiling samples
### Description
Our wall-clock (wall) time profiler, `CpuProfiler`, currently sends `wall` (in nanoseconds) and `sample` (hit count). Product has also asked us to include `cpu` (in nanoseconds) as well. Thankfully, `pprof` handles this for us. All we need to do is:

```js
  this.#pprof.time.start({
      durationMillis: this.#durationMillis,
      intervalMicros: this.#intervalMicros,
      // add the following
      collectCpuTime: true
    })
```

### Acceptance Criteria

- [ ] Add `collectCpuTime: true`
- [ ] Create tests that assert that `cpu` is being included in the sample value and that it is reasonable (e.g. `cpu` must always be less than `wall` by definition)
- [ ] Consider using `pprof-format` to deserialize the payload and make sure `cpu` is still there

- [ ] Should `collectCpuTime` be configurable or is it being always `true` okay?

amychisholm03:
## Description

Our `CpuProfiler` is a wall-clock profiler; previously each pprof example carried only `wall:nanoseconds` and `sample:count` values. This PR adds `collectCpuTime: true` to pprof's `time.start`, adding a measured `cpu:nanoseconds` value type to every sample alongside `wall`, so profiles now report actual on-CPU time in addition to wall time.

Previous: 
```
sampleType = [
  { type: 'sample', unit: 'count' },
  { type: 'wall',   unit: 'nanoseconds' }
]
```
Now: 
```
sampleType = [
  { type: 'sample', unit: 'count' },
  { type: 'wall',   unit: 'nanoseconds' },
  { type: 'cpu',    unit: 'nanoseconds' }
]
```

## How to Test

```
node --test test/unit/lib/profiling/profilers/cpu.test.js
node --test test/integration/profiling/cpu.test.js
```

Appears in the UX correctly:
image

image

## Related Issues

Closes #4048

### DIFF
--- a/lib/profiling/profilers/cpu.js
+++ b/lib/profiling/profilers/cpu.js
@@ -83,6 +83,10 @@ class CpuProfiler extends BaseProfiler {
       withContexts: true,
       // carry the sample context per async-context-frame when ACF is active
       useCPED: this.#useCPED,
+      // add a measured on-CPU (`cpu`) nanoseconds value to each sample
+      // alongside `wall`. cpu is the on-CPU time the native profiler
+      // measures per context, so cpu <= wall by definition.
+      collectCpuTime: true,
       sourceMapper: this.sourceMapper
     })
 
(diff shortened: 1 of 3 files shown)


bizob2828 (approved):
LGTM

PR #4147, opened 2026-07-17
feat: Add internal OTEL API to agent

### LINKED ISSUE #4143: Add internal OTEL API
As we build new features and functionality in to our OTEL bridge, it will be necessary to have an internal API that we can use to work with the OTEL objects. To facilitate this, we need to build an API object in the OTEL setup function:

https://github.com/newrelic/node-newrelic/blob/05b072f8b5c7b4ec18b36c0aa4267a86f5966b9c/lib/otel/setup.js#L16

That new object should look like:

```js
{
  logs,
  metrics,
  traces
}
```

Where each field value is the signal wrapper instance. For example, we may add a `flushMetrics` method to the class defined in `lib/otel/metrics/index.js`.

To make this API accessible across the agent, `lib/agent.js` should be updated with:

```js
const kOtel = Symbol('otel.api')
Object.defineProperty(Agent.prototype, 'otel', {
  get () { return this[kOtel] },
  set (val) { this[kOtel] = val }
})
```

So that `setupOtel` can `agent.otel = {/* api object */}`.

User's of the API would need to invoke things as `agent.?otel.?metrics.flushMetrics()`. We have the option of disabling all or some of the OTEL signals. As such, the API may or may not be defined. Another possibility would be to add a stub API to attach to the agent. But this seems overkill because 1. we'd have to maintain parity on it as we update the actual implementations and 2. this is an internal only API.

jsumners-nr:
Closes #4143

## Summary

Adds an internal `agent.otel` API so that features built on top of the OTEL bridge can interact with the OTEL signal objects.

- `lib/agent.js`: adds an `otel` getter/setter to `Agent.prototype`, backed by a private symbol.
- `lib/otel/setup.js`: `setupOtel` now builds an API object of the shape `{ logs, metrics, traces }`, where each field is the corresponding signal wrapper instance (`SetupLogs` / `SetupMetrics` / `SetupTraces`), and assigns it to `agent.otel`. Signals that are disabled stay `undefined`. When OTEL is disabled entirely, `agent.otel` is never set.

Because signals may be individually disabled (or OTEL disabled entirely), the API may be partially or fully undefined. Consumers use optional chaining, e.g. `agent.otel?.metrics?.flushMetrics()`. As discussed in the issue, no stub API is provided since this is an internal-only API and a stub would require maintaining parity with the real implementations.

## Testing

- Added unit tests in `test/unit/lib/otel/setup.test.js` covering all-enabled, partially-enabled, and OTEL-disabled cases.
- Added a unit test in `test/unit/agent/agent.test.js` for the `otel` getter/setter.
- `npm run lint` passes.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -349,6 +349,16 @@ Object.defineProperty(Agent.prototype, Symbol.toStringTag, {
   value: 'Agent'
 })
 
+const kOtel = Symbol('otel.api')
+Object.defineProperty(Agent.prototype, 'otel', {
+  get() {
+    return this[kOtel]
+  },
+  set(val) {
+    this[kOtel] = val
+  }
+})
+
 /**
  * The agent is meant to only exist once per application, but the singleton is
  * managed by index.js. An agent will be created even if the agent's disabled by
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -169,7 +169,10 @@ const DEFAULT_HARVEST_INTERVAL_MS = 60000
  * available. Don't try to recover here, because without configuration the
  * agent can't be brought up to a useful state.
  *
- * @param {Config} config Agent configuration object
+ * @param {AgentConfig} config Agent configuration object
+ *
+ * @class
+ * @alias Agent
  */
 function Agent(config) {
   EventEmitter.call(this)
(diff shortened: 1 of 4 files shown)


bizob2828 (approved):
Looks fine just struggling to understand its utility

PR #4151, opened 2026-07-20
chore: Translate OTel exception into timed event

### LINKED ISSUE #4138: Translate otel exception into timed event
In #4130 we added support for `fakeSpan.recordException` as a stubbed out method:

https://github.com/newrelic/node-newrelic/blob/3c1a2b3a8b9ee2157c4bc774873a48ed6079334e/lib/otel/fake-span.js#L138-L143

We overlooked the fact that we have support for timed events:

https://github.com/newrelic/node-newrelic/blob/3c1a2b3a8b9ee2157c4bc774873a48ed6079334e/lib/transaction/trace/segment.js#L316-L323

We can improve the `recordException` and `addEvent` methods to use the timed event API. The link to upstream's reference implementation should be a good model to follow.

amychisholm03:
## Description

This PR implements the `addEvent` and `recordException` methods into the `FakeSpan` (`lib/otel/fake-span.js`), which were previously just stubs. 

- `addEvent(name,attributes,time)` now builds a `TimedEvent` from the given parameters and pushes to the actual `TraceSegment` via `segment.addTimedEvent()`.
- `recordException(exception,time)` extracts the exception attibutes from a string or `Error`, mirroring [what OTel does](https://github.com/open-telemetry/opentelemetry-js/blob/9b05f66/packages/sdk-trace/src/Span.ts#L427-L451) and records it as an `'exception'` event via `FakeSpan.addEvent`.

## How to Test
```
node --test test/unit/lib/otel/fake-span.test.js
```
## Related Issues

Closes #4138

### DIFF
--- a/lib/otel/fake-span.js
+++ b/lib/otel/fake-span.js
@@ -5,6 +5,9 @@
 
 'use strict'
 
+const TimedEvent = require('../spans/timed-event')
+const { EXCEPTION_MESSAGE, EXCEPTION_STACKTRACE, EXCEPTION_TYPE } = require('./traces/constants')
+
 /**
  * In order to be able to return the appropriate span context
  * within otel bridge. We have to create fake spans for new relic
@@ -71,11 +74,31 @@ module.exports = class FakeSpan {
     return this
   }
 
-  addEvent(name, _attrsOrTime, _time) {
-    this.#segment.logger.warn(
-      'addEvent is not implemented. Not adding event: %s.',
-      name
-    )
+  /**
+   * Adds a new timed event to the backing segment.
+   *
+   * This replicates what the original actually does.
+   * See:
+   * https://github.com/open-telemetry/opentelemetry-js/blob/9b05f66/packages/sdk-trace/src/Span.ts#L389-L410.
+   *
+   * @param {string} name The name of the event.
+   * @param {object} [attributesOrStartTime] The attributes to associate with
+   * the event.
+   * @param {number[]|number|Date} [timeStamp] The start time of the event.
+   *
+   * @returns {FakeSpan}
+   */
+  addEvent(name, attributesOrStartTime, timeStamp) {
+    const timedEvent = new TimedEvent({
+      event: {
+        name,
+        attributes: attributesOrStartTime ?? {},
+        time: timeStamp
+      },
+      spanContext: this.spanContext()
+    })
+    this.#segment.addTimedEvent(timedEvent)
+
     return this
   }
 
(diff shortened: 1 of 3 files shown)


lib/otel/fake-span.js
```diff
@@ -71,11 +74,31 @@ module.exports = class FakeSpan {
...
+   * https://github.com/open-telemetry/opentelemetry-js/blob/9b05f66/packages/sdk-trace/src/Span.ts#L389-L410.
+   *
+   * @param {string} name The name of the event.
+   * @param {object} [attributesOrStartTime] The attributes to associate with
+   * the event.
+   * @param {number[]|number|Date} [timeStamp] The start time of the event.
+   *
+   * @returns {FakeSpan}
+   */
+  addEvent(name, attributesOrStartTime, timeStamp) {
```
jsumners-nr, line 112: You should add a test that does:

```js
span.addEvent('my-exception', 1784564100000)
```

amychisholm03: `1784564100000` as `attributesOrStartTime` right, not `timeStamp`?

jsumners-nr: Exactly as written: `addEvent(string, number)`.

test/unit/lib/otel/fake-span.test.js
```diff
@@ -67,30 +67,67 @@ test('should add attributes to the segment', () => {
...
+  const tx = { traceId: 'traceId' }
+  const span = new FakeSpan(segment, tx)
+
+  const someTime = 1784564100000
+  span.addEvent('my-exception', someTime)
+  assert.equal(segment.timedEvents.length, 1)
+
+  const [intrinsics, , agentAttrs] = segment.timedEvents[0].toJSON()
+  assert.equal(intrinsics.name, 'my-exception')
+  assert.notEqual(intrinsics.timestamp, someTime, `${someTime} is not treated as a timestamp`)
```
jsumners-nr: These should be equal. This is why I suggested the test. The implementation of the method is not interpreting the arguments list correctly.

jsumners-nr (changes requested):

test/unit/lib/otel/fake-span.test.js
amychisholm03: Ah I was following [what OTel did](https://github.com/open-telemetry/opentelemetry-js/blob/9b05f668ee7ab884a44b04b504e0baaff6c6d2b2/packages/sdk-trace/src/Span.ts#L201-L204) but I didn't make the last two parameters optional

jsumners-nr: See https://github.com/open-telemetry/opentelemetry-js/blob/9b05f668ee7ab884a44b04b504e0baaff6c6d2b2/packages/sdk-trace/src/Span.ts#L227-L232

jsumners-nr (approved):

PR #4155, opened 2026-07-21
fix: Resolved mongodb host details at operation time to support `mongodb+srv://` clients

### LINKED ISSUE #4154: mongodb subscribers crash with `TypeError: Cannot read properties of undefined (reading 'host')` when using `mongodb+srv://` URIs before connect

## Description

The v14 `diagnostics_channel`-based mongodb subscribers crash the application when a
`Db` or `Collection` handle is created on a `MongoClient` configured with a
`mongodb+srv://` connection string before `client.connect()` has resolved.

[Every branch of `lib/subscribers/mongodb/utils/get-host-details.js`](https://github.com/newrelic/node-newrelic/blob/36daebe14ad7b7426456d7bee529087389511459/lib/subscribers/mongodb/utils/get-host-details.js#L46-L74) reads
`options.hosts[0].host` without guarding against an empty array. For SRV URIs the
mongodb driver leaves `options.hosts` as `[]` at parse time and only populates it
after DNS SRV resolution completes during `connect()`. The subscribers fire on
`Db`/`Collection` **construction** (via the patched constructors publishing to the
tracing channel), so any app that creates handles before connecting — which the
driver explicitly supports via lazy connect — throws.

This is not caught by the existing versioned tests because `test/versioned/mongodb/common.js`
only builds plain `mongodb://host:port` connection strings, where `hosts` is populated
synchronously at parse time.

Note the failure mode is particularly severe: because the subscriber throws inside a
`diagnostics_channel` publish, the error is re-raised as an **uncaught exception** —
the `client.db()` / `db.collection()` calls themselves return normally and the process
crashes a tick later, making the source hard to trace back from application code.

## Expected Behavior

`getHostDetails` should degrade gracefully (omit host/port attribution) when
`options.hosts` is empty, instead of throwing from inside `new Db` / `new Collection`.

```
TypeError: Cannot read properties of undefined (reading 'host')
    at getHostDetails (node_modules/newrelic/lib/subscribers/mongodb/utils/get-host-details.js:47:41)
    at DbClassSubscriber.end (node_modules/newrelic/lib/subscribers/mongodb/db.js:70:21)
    at Channel.publish (node:diagnostics_channel:156:9)
    at node_modules/mongodb/lib/db.js:91:29
    at run (node:diagnostics_channel:168:14)
    at AsyncLocalStorage.run (node:internal/async_local_storage/async_context_frame:63:14)
    at node:diagnostics_channel:100:18
    at Channel.runStores (node:diagnostics_channel:177:12)
    at new Db (node_modules/mongodb/lib/db.js:80:34)
    at MongoClient.db (node_modules/mongodb/src/mongo_client.ts:846:16)
```

The equivalent crash occurs via `CollectionSubscriber.end` → `new Collection` when
calling `db.collection(...)` ([get-host-details.js:52](https://github.com/newrelic/node-newrelic/blob/36daebe14ad7b7426456d7bee529087389511459/lib/subscribers/mongodb/utils/get-host-details.js#L52)).

## Steps to Reproduce

Verified against `newrelic@14.3.1` + `mongodb@7.5.0` on Node 24.18.0:

```js
require('newrelic')
const { MongoClient } = require('mongodb')

// Any mongodb+srv:// URI — DNS does not even need to resolve;
// the crash happens before any I/O.
const client = new MongoClient('mongodb+srv://cluster0.example.mongodb.net/mydb')

// client.options.hosts is [] here — the TypeError is raised from the
// DbClassSubscriber during `new Db` and surfaces as an uncaught exception.
const db = client.db()
// db.collection('users') crashes identically via CollectionSubscriber.
```

Works fine with a plain `mongodb://host:27017` URI (hosts populated at parse time),
and with newrelic 13.x (shimmer instrumentation reads host details at operation time,
not construction time).

## Your Environment

* newrelic: 14.3.1
* mongodb driver: 7.5.0
* Node version: 24.18.0
* Operating System: Linux (containerized, k8s)

## Additional context

Workarounds we validated:
* Downgrading to newrelic 13.x.
* `NEW_RELIC_INSTRUMENTATION_MONGODB_ENABLED=false` also avoids the crash on 14.3.1,
  at the cost of losing all mongodb instrumentation.
* Awaiting `client.connect()` before creating any `Db`/`Collection` handle avoids the
  window, but breaks the long-standing lazy-connect usage pattern the driver supports.

jsumners-nr:
## Description

Fixes #4154.

The v14 `diagnostics_channel`-based mongodb subscribers crash the application when a `Db` or `Collection` handle is created on a `MongoClient` configured with a `mongodb+srv://` connection string before `client.connect()` has resolved.

For SRV URIs the driver leaves `options.hosts` as `[]` at parse time and only populates it after DNS SRV resolution completes during `connect()`. The subscribers fire on `Db`/`Collection` **construction**, so `getHostDetails` read `options.hosts[0].host` against an empty array and threw:

```
TypeError: Cannot read properties of undefined (reading 'host')
    at getHostDetails (lib/subscribers/mongodb/utils/get-host-details.js:47:41)
    at DbClassSubscriber.end (lib/subscribers/mongodb/db.js:70:21)
    at Channel.publish (node:diagnostics_channel:156:9)
    ...
    at new Db (mongodb/lib/db.js:80:34)
    at MongoClient.db (mongodb/src/mongo_client.ts:846:16)
```

Because the throw happens inside a `diagnostics_channel` publish, it surfaces as an **uncaught exception** and crashes the process a tick after `client.db()` / `db.collection()` return normally.

## Changes

- **`get-host-details.js`** — Guarded every `hosts[0]` access with optional chaining so it degrades gracefully (host/port omitted) instead of throwing when `options.hosts` is empty, matching the issue's expected behavior.
- **`wrapper.js` + `db.js` / `collection.js` / `bulk.js` / `cursor.js`** — Resolve host details **at operation time** via a new `getParameters` callback in `wrapDatabaseMethod`, rather than caching them once at handle construction. By the time an operation runs, `connect()` has populated `hosts`, so segments once again carry the resolved server name and port for SRV clients (not just avoiding the crash).

## Testing

Added `test/versioned/mongodb/srv-connect.test.js`. Following the DNS-patching pattern in `test/integration/core/dns.test.js`, the SRV hostname resolves to the local Docker MongoDB container, so the tests exercise the real connect path against a live server:

1. `client.db()` before connect does not throw.
2. `db.collection()` before connect does not throw.
3. Operations over an SRV connection are instrumented **and the segment carries the SRV-resolved server name/port**.
4. An operation issued before SRV resolution degrades gracefully — the `getHostDetails` guard is exercised directly (segment recorded with host/port omitted, driver's own error surfaces, no `TypeError`).

Verified each test fails against the unpatched code and passes with the fix, and the full mongodb versioned suite passes across all four driver versions (4.17.2, 5.9.2, 6.21.0, 7.5.0). `npm run lint` is clean.

### DIFF
--- a/lib/subscribers/mongodb/bulk.js
+++ b/lib/subscribers/mongodb/bulk.js
@@ -41,13 +41,6 @@ module.exports = class BulkSubscriber extends Wrapper {
     const { self: bulkOperation, arguments: args } = data
     const [collection] = args
     const collectionName = collection.collectionName
-    const details = getHostDetails(collection)
-    this.parameters = {
-      host: details.host,
-      port_path_or_id: details.port_path_or_id,
-      database_name: details.database_name,
-      product: this.system
-    }
 
     for (const method of CURSOR_METHODS) {
       if (typeof bulkOperation[method] !== 'function') {
@@ -58,6 +51,17 @@ module.exports = class BulkSubscriber extends Wrapper {
         getSegmentName: () => {
           const operation = bulkOperation.isOrdered ? 'orderedBulk' : 'unorderedBulk'
           return `${DB_METRIC_CONSTANTS.STATEMENT}/${this.system}/${collectionName}/${operation}/batch`
+        },
+        // Resolve host details when the operation runs, not when the bulk
+        // operation is constructed. See `WrapperSubscriber#wrapDatabaseMethod`.
+        getParameters: () => {
+          const details = getHostDetails(collection)
+          return {
+            host: details.host,
+            port_path_or_id: details.port_path_or_id,
+            database_name: details.database_name,
+            product: this.system
+          }
         },
         getRecorderContext: () => {
           return {
--- a/lib/subscribers/mongodb/collection.js
+++ b/lib/subscribers/mongodb/collection.js
@@ -78,13 +78,6 @@ module.exports = class CollectionSubscriber extends Wrapper {
   end(data, ctx) {
     const { self: dbClient, arguments: args } = data
     const [dbInstance, collectionName] = args
-    const details = getHostDetails(dbInstance)
-    this.parameters = {
-      host: details.host,
-      port_path_or_id: details.port_path_or_id,
-      database_name: details.database_name,
-      product: this.system
-    }
 
     for (const method of CURSOR_METHODS) {
       if (typeof dbClient[method] !== 'function') {
@@ -93,6 +86,18 @@ module.exports = class CollectionSubscriber extends Wrapper {
 
       this.wrapDatabaseMethod(dbClient, method, {
         getSegmentName: (method) => `${DB_METRIC_CONSTANTS.STATEMENT}/${this.system}/${collectionName}/${method}`,
+        // Resolve host details when the operation runs, not when the
+        // `Collection` handle is constructed. See
+        // `WrapperSubscriber#wrapDatabaseMethod`.
+        getParameters: () => {
+          const details = getHostDetails(dbInstance)
+          return {
+            host: details.host,
+            port_path_or_id: details.port_path_or_id,
+            database_name: details.database_name,
+            product: this.system
+          }
+        },
         getRecorderContext: (method) => {
           return {
             operation: method,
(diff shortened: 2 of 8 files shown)


jsumners-nr:
The "uncovered" lines are covered by `npm run versioned` instead of `npm run versioned:major`.

bizob2828 (approved):

PR #4161, opened 2026-07-21
chore: Added workflows to generate a config schema and version before agent release

mvicknr:
## Description

Adds a script to automatically generate and update a configuration schema for the agent. This PR has the initial schema generated from the script.

The script will automatically run on local pre-commit and on pushes (except main). It will determine if there are config changes and if so, it will update the schema.O n local, the dev should include these changes in their commit. In the PR build, if the changes aren't already made in the schema, it will make the changes and add them to the PR.

The schema version will be updated during the release process.

I split the work into commits so you can review in chunks

## How to Test

Unit tests provided. GH workflows run locally with act

## Related Issues

N/A


.fleetControl/schemas/config.json
```diff
@@ -0,0 +1,1556 @@
...
+      "type": "string",
+      "default": "",
+      "description": "Proxy user name when required."
+    },
+    "proxy_pass": {
+      "type": "string",
+      "default": "",
+      "description": "Proxy password when required."
+    },
+    "trusted_account_key": {
```
mvicknr, line 69: `trusted_account_key`, `primary_application_id`, and `account_id` are marked as "Serverless DT config defaults" in the [default.js](https://github.com/newrelic/node-newrelic/blob/main/lib/config/default.js#L104-L113) file but these items don't have a description here in the schema because the comment is for all 3 items. Can we modify the default.js file to have a comment for each config item so it is clear in the schema?

.fleetControl/schemas/config.json
```diff
@@ -0,0 +1,1556 @@
...
+              "description": "Prefix of attributes to include in browser monitoring. Allows * as wildcard at end."
+            }
+          },
+          "additionalProperties": true
+        },
+        "enable": {
+          "type": "boolean",
+          "default": true,
+          "description": "Enable browser monitoring header generation. This does not auto-instrument, rather it enables the agent to generate headers. The newrelic module can generate the appropriate <script> header, but you must inject the header yourself, or use a module that does so. Usage: var newrelic = require('newrelic'); router.get('/', function (req, res) { var header = newrelic.getBrowserTimingHeader(); res.write(header) // write the rest of the page }); This generates the <script>...</script> header necessary for Browser Monitoring This script must be manually injected into your templates, as high as possible in the header, but _after_ any X-UA-COMPATIBLE HTTP-EQUIV meta tags. Otherwise you may hurt IE! This method must be called _during_ a transaction, and must be called every time you want to generate the headers. Do *not* reuse the headers between users, or even between requests."
```
mvicknr: The code here in the description won't have formatting so could this be updated with @example in the default.js file? If so, the generator could omit anything with that tag. There may be a few more we will also want to do this for.

.github/workflows/agent-config-schema-bump.yml
```diff
@@ -0,0 +1,103 @@
...
+  contents: read
+
+# Release-prep workflow: opens a PR that bumps the `version:` field in
+# .fleetControl/configurationDefinitions.yml, based on the cumulative schema
+# diff since the previous release tag.
+#
+# Triggered manually before a release tag is cut, so the bumped version can
+# be merged to main and included in the release.
+#
+# Release ordering:
```
mvicknr: Is there somewhere I can call this from that would make it more automatic?

package.json
```diff
@@ -252,6 +256,7 @@
     "express": "*",
     "git-raw-commits": "^2.0.11",
     "got": "^15.0.5",
+    "js-yaml": "^4.3.0",
```
bizob2828: do you need 4.x? This package is on 5.2.1

mvicknr: fixed

bizob2828 (approved):
we're really hands off on this. I trust @mvicknr that she verified that this will fit her needs. There's a merge conflict but LGTM

package.json
```diff
@@ -178,6 +180,7 @@
     "test": "npm run integration && npm run unit",
     "third-party-updates": "oss third-party manifest --includeOptDeps && oss third-party notices --includeOptDeps && git add THIRD_PARTY_NOTICES.md third_party_manifest.json",
     "unit": "rm -f newrelic_agent.log && time c8 -o ./coverage/unit borp --timeout 180000 --reporter ./test/lib/test-reporter.mjs 'test/unit/**/*.test.{js,mjs}'",
+    "unit:config-schema": "time c8 -o ./coverage/config-schema-unit borp --reporter ./test/lib/test-reporter.mjs '.fleetControl/schemaGeneration/tests/**/*.test.js'",
```
bizob2828, line 183: do you want to run these in CI or is it just something you manually run if any edits occur in the future?

bin/update-config-schema.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+#
+# Checks all staged files for changes to the agent's config definition or
+# the schema generator itself. If either changed, regenerates the agent
+# config schema and stages the result.
+
+STAGED_FILES=$(git diff-index --cached --name-only HEAD)
+
+SHOULD_REGENERATE=0
```
bizob2828: @mvicknr we have `lint-staged` which allows you to run something if a file changes. you could add it to `https://github.com/newrelic/node-newrelic/blob/main/.lintstagedrc.json` and run what you need

.fleetControl/schemas/config.json
```diff
@@ -0,0 +1,1559 @@
...
+            }
+          },
+          "additionalProperties": true,
+          "description": "Toggles the capturing of Llm events when using streaming based methods in AIM supported libraries(i.e.- openai, AWS bedrock, langchain)"
+        }
+      },
+      "additionalProperties": true,
+      "description": "When enabled, instrumentation of supported AI libraries will be in effect."
+    },
+    "instrumentation": {
```
bizob2828, line 1453: this property is dynamic but doesn't enumerate all the available options. I'm not sure if that's an issue or not

.github/workflows/agent-config-schema-bump.yml
bizob2828: we have a other workflows that run as a post release, like the agent-metadata https://github.com/newrelic/node-newrelic/blob/e9787624ec4a2ce2823fec8652bfab02274cb94a/.github/workflows/agent-metadata.yml#L26

.github/workflows/agent-config-schema.yml
```diff
@@ -0,0 +1,61 @@
...
+# Regenerates .fleetControl/schemas/config.json whenever a push touches the
+# agent's config definition or the generator itself, then commits the
+# result back to the pushed branch.
+#
+# The paths filter deliberately excludes .fleetControl/schemas/** (the
+# generated file) — the commit step below only touches that path, so its
+# own push does not re-trigger this workflow.
+#
+# Skips main; this only runs on feature branches.
+on:
```
bizob2828, line 9: so am i reading this correct that this could live on an existing PR branch if config changes? If so, you may want to remove the pre-commit hook as it's redundant and this is more reliable because some people may not have registered the pre-commit hook

bizob2828 (changes requested):
actually i decided to dig more into this. a few comments/suggestions

.github/workflows/agent-config-schema-bump.yml
mvicknr: The issue is that the schema version needs to be updated before the release tag is applied

package.json
mvicknr: Updated

bin/update-config-schema.sh
mvicknr: updated

package.json
```diff
@@ -160,8 +160,10 @@
   "scripts": {
```
bizob2828, line 160: this has to be rebased because it contains older versions of packages as well as packages we removed

.github/workflows/agent-config-schema-bump.yml
```diff
@@ -0,0 +1,103 @@
...
+#   1. Run this workflow (workflow_dispatch).
+#   2. Wait for the PR to open (or for the workflow to report no bump needed).
+#   3. Review and merge the bump PR if one was opened.
+#   4. Cut the release from the post-merge main.
+#
+# Bootstrap: if the resolved ref predates the schema (or its
+# configurationDefinitions.yml doesn't have a version yet), the script exits
+# cleanly with a bootstrap message and no PR is opened.
+on:
```
bizob2828: the git tag for a release isn't created until after we release so I don't think this will work

.github/workflows/agent-config-schema.yml
```diff
@@ -0,0 +1,61 @@
...
+
+# Regenerates .fleetControl/schemas/config.json whenever a push touches the
+# agent's config definition or the generator itself, then commits the
+# result back to the pushed branch.
+#
+# The paths filter deliberately excludes .fleetControl/schemas/** (the
+# generated file) — the commit step below only touches that path, so its
+# own push does not re-trigger this workflow.
+#
+# Skips main; this only runs on feature branches.
```
bizob2828: will this work on forks? all of our PRs are opened on forks

.lintstagedrc.json
```diff
@@ -1,5 +1,9 @@
 {
   "**/*.{cjs,mjs,js}": [
     "eslint --fix"
+  ],
+  "{lib/config/default.js,lib/config/samplers.js,.fleetControl/schemaGeneration/**}": [
```
bizob2828: I may have confused things in my past round of reviews but the guarantee that this runs as a pre-commit hook cannot be relied on. Since we also have a workflow to do the same thing which opens a separate PR it seems like that's more reliable

.fleetControl/schemaGeneration/generate-schema.js
```diff
@@ -0,0 +1,525 @@
...
+    for (const prefix of prefixes) {
+      commentIndex.set(`${prefix}.${key}`, entry)
+    }
+  }
+}
+
+// Every input defaults to the real thing, so tests can pass synthetic ones instead.
+function generateSchema({
+  definition = defaultConfig.definition(),
+  defaultConfigSourceText = fs.readFileSync(DEFAULT_CONFIG_SOURCE, 'utf8'),
```
bizob2828: I think instead of reading the file you require it and call the `.definition` function which compiles any dynamic pieces of the default configuration

.github/workflows/agent-config-schema.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Regenerate schema
+        run: node .fleetControl/schemaGeneration/generate-schema.js
+
+      - name: Commit and push regenerated schema
+        run: |
+          if [ -z "$(git status --porcelain .fleetControl/schemas/config.json)" ]; then
+            echo "No schema changes — nothing to commit."
+            exit 0
+          fi
+          git config user.name 'github-actions[bot]'
```
bizob2828: we should prob use the `node-agent-bot` user for this.

bizob2828 (commented):
a few comments from discussion in google meet

.lintstagedrc.json
mvicknr: I'll remove the lintstaged change. A note: the workflow should update your same PR (not a new one) with the updated schema.

.fleetControl/schemaGeneration/generate-schema.js
mvicknr: the default configuration values were being purposefully dropped (gotta love Claude) so I fixed that. The `.definition` is used for everything except the descriptions of the fields which is why we are parsing the file.

.fleetControl/schemas/config.json
mvicknr: updated

.github/workflows/agent-config-schema.yml
mvicknr: updated

.fleetControl/schemas/config.json
mvicknr: updated

.fleetControl/schemas/config.json
mvicknr: updated

.github/workflows/agent-config-schema-bump.yml
```diff
@@ -0,0 +1,105 @@
...
+  contents: read
+
+# Release-prep workflow: opens a PR that bumps the `version:` field in
+# .fleetControl/configurationDefinitions.yml, based on the cumulative schema
+# diff since the previous release tag.
+#
+# Triggered manually before a release tag is cut, so the bumped version can
+# be merged to main and included in the release.
+#
+# Release ordering:
```
bizob2828: I thought we discussed running this as part of `Prepare Release` so it updates the release prep PR and we won't have to remember to review a separate PR?

.github/workflows/agent-config-schema.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Use Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: lts/*
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Regenerate schema
+        run: node .fleetControl/schemaGeneration/generate-schema.js
```
bizob2828: You already added a script in package.json. I would update this
```suggestion
        run: npm run generate:config-schema
```

.github/workflows/prep-release.yml
```diff
@@ -20,6 +20,12 @@ on:
         description: Whether or not it should generate a changelog.json (should only be for node-newrelic)
         type: boolean
         required: false
+    outputs:
```
bizob2828: this is the reusable workflow. I wouldn't put that here. I would put all the logic into `prepare-release` as other repos which call this will not need to run the config schema update

.github/workflows/prepare-release.yml
```diff
@@ -24,3 +24,39 @@ jobs:
...
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: lts/*
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Compute config schema bump
+        run: node .fleetControl/schemaGeneration/bump-schema-version.js --write
```
bizob2828: i would add this to package.json as a script and just call it as `npm run <script-name>`

.github/workflows/agent-config-schema-bump.yml
bizob2828: oh wait i think you did update `Prepare Release`, can't this just be removed?

bizob2828 (changes requested):
a few suggestions. getting close though, nice work!

.github/workflows/agent-config-schema.yml
bizob2828: we like to have changelog items in past. I can't suggest but can you change this to `chore: Regenerated agent config schema` for both PR title and commit message?

bizob2828 (dismissed):
very small tweak but don't sweat it. i can handle this later.  I'm approving but I have to see from team when to land this because we know we may have issues.

bizob2828 (approved):

PR #4163, opened 2026-07-22
fix: Avoid unconditional @opentelemetry/core require on agent startup

### LINKED ISSUE #4162: Slim Node.js Lambda layer/image fails with "Cannot find module '@opentelemetry/core'" since v14.3.2
## Description

Since `newrelic` v14.3.2, the agent crashes at startup with
`Error: Cannot find module '@opentelemetry/core'` on the **slim**
Node.js Lambda layer / container image, which ships without any
`@opentelemetry/*` packages.

Cause: #4151 added an unconditional `require('@opentelemetry/core')`
in the require chain `tracer/index.js → ... → lib/spans/timed-event.js →
lib/otel/normalize-timestamp.js`, which now runs on every agent startup —
not just when the OTel bridge is used. The slim build strips
`@opentelemetry/*`, so it now crashes unconditionally.

## Expected Behavior

Slim layer/image should start without requiring any `@opentelemetry/*`
package when the OTel bridge isn't in use.

## Steps to Reproduce

1. Use `newrelic-lambda-layers-nodejs-slim` (or the slim layer) built
   against `newrelic@14.3.2`+.
2. Invoke the Lambda function.
3. `Error: Cannot find module '@opentelemetry/core'` at init.

Or: install `newrelic@14.3.2`, delete `node_modules/newrelic/node_modules/@opentelemetry`, then `require('newrelic')`.

## Your Environment

* Node: 22.x (Lambda `nodejs22.x`)
* `newrelic`: 14.3.2 (last known-good slim build: 14.3.1)
* Slim Lambda layer / `newrelic-lambda-layers-nodejs-slim` image

## Additional context

Slim build strips OTel here: https://github.com/newrelic/newrelic-lambda-layers/blob/master/nodejs/publish-layers.sh
Regression introduced by #4151.

jsumners-nr:
## Description

Fixes #4162.

Since v14.3.2 (#4151), the agent crashes at startup with `Cannot find module '@opentelemetry/core'` on the **slim** Node.js Lambda layer/image, which ships without any `@opentelemetry/*` packages.

### Root cause

#4151 wired `lib/spans/timed-event.js` into the **always-on** agent startup require chain:

```
lib/transaction/tracer/index.js
  → lib/context-manager/async-local-context-manager.js
    → lib/otel/context.js
      → lib/otel/fake-span.js
        → lib/spans/timed-event.js
          → lib/otel/normalize-timestamp.js
            → require('@opentelemetry/core')   ← unconditional
```

`lib/otel/normalize-timestamp.js` did a top-level `require('@opentelemetry/core')` (for `isTimeInputHrTime`), so simply requiring the agent now loaded that package on **every** startup — not just when the OTel bridge is enabled. The slim build strips `@opentelemetry/*`, so it crashed unconditionally.

### Fix

Replace the single used export (`isTimeInputHrTime`) with a local copy of the upstream implementation in `lib/otel/is-otel-hr-time.js`, shared by `lib/otel/is-time-input.js` and `lib/otel/normalize-timestamp.js`. Nothing on the always-on path now requires an `@opentelemetry/*` package. The OTel-bridge paths (traces/logs/metrics via `lib/otel/setup.js`) are unchanged — they still use `@opentelemetry/core` and only run when the bridge is enabled and the package is present.

### Tests

Adds `test/unit/lib/otel/normalize-timestamp-no-otel-require.test.js`, which runs in child processes (for a pristine module cache) and asserts:

1. Requiring `normalize-timestamp.js` does not load `@opentelemetry/core`.
2. The always-on tracer require chain does not load any `@opentelemetry/*` package.
3. `normalize-timestamp.js` produces correct output even when every `@opentelemetry/*` resolution throws `MODULE_NOT_FOUND` (slim-build simulation).

Verified the test fails without the fix and passes with it.

### DIFF
--- a/lib/otel/fake-span.js
+++ b/lib/otel/fake-span.js
@@ -5,28 +5,13 @@
 
 'use strict'
 
-const TimedEvent = require('../spans/timed-event')
-const { EXCEPTION_MESSAGE, EXCEPTION_STACKTRACE, EXCEPTION_TYPE } = require('./traces/constants')
-
-/**
- * Checks whether a value looks like a point in time (a `Date`, an epoch
- * number, or an OTEL hrtime `[seconds, nanoseconds]` tuple) rather than a
- * set of event attributes.
- *
- * @param {*} value The value to check.
- *
- * @returns {boolean} True if the value should be treated as a time.
- */
-function isTimeInput(value) {
-  return (
-    typeof value === 'number' ||
-    value instanceof Date ||
-    (Array.isArray(value) &&
-      value.length === 2 &&
-      typeof value[0] === 'number' &&
-      typeof value[1] === 'number')
-  )
-}
+const isTimeInput = require('./is-time-input.js')
+const TimedEvent = require('../spans/timed-event.js')
+const {
+  EXCEPTION_MESSAGE,
+  EXCEPTION_STACKTRACE,
+  EXCEPTION_TYPE
+} = require('./traces/constants.js')
 
 /**
  * In order to be able to return the appropriate span context
--- a/lib/otel/is-otel-hr-time.js
+++ b/lib/otel/is-otel-hr-time.js
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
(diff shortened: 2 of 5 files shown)


bizob2828 (approved):
we prob need better ways to guard against this. I won't block this but we should think about how to improve this slim build use case

PR #4166, opened 2026-07-22
feat: Collect OTEL metrics in AWS Lambda

### LINKED ISSUE #4126: Enable collection of OTEL metrics from AWS Lambda

jsumners-nr:
See https://github.com/newrelic/newrelic-node-examples/pull/379.

We should probably just ship this as "prep for an upcoming feature" instead of keeping a stale branch around. The agent is going to do the right thing, but the data won't get shipped until the extension is updated (possibly in a future quarter).

-----

## Summary

Adds collection of OpenTelemetry metrics in the AWS Lambda (serverless) environment. In serverless mode the periodic exporter is effectively disabled and metrics are instead flushed on demand during the agent's synchronous harvest: `SetupMetrics.flushToString()` collects and serializes the metrics to a base64 OTLP protobuf payload, which `ServerlessCollector` attaches to the flushed payload as `otlp_payload`.

Closes #4126.

## Changes

- `SetupMetrics`: in serverless mode, use a never-firing export interval, finalize the exporter eagerly (no `started`-event wait), and expose `flushToString()`.
- `NRProxyingSerializer` / `NROTLPMetricExporter`: cache the last serialization (`lastSerialization` / `serializedData`) so the harvest can retrieve the encoded payload after an export; `awsLambdaMode` forces caching independent of audit logging.
- `ServerlessCollector.flushPayloadSync`: pull OTEL metrics via `agent.otel.metrics.flushToString()` when present and include them as `otlp_payload`; made internal flush helpers private.
- `Agent`: internal `otel` API accessor (attached by `lib/otel/setup.js`).

## Testing

- Fixed an audit-logging regression found while writing coverage: `NRProxyingSerializer` compared the `auditEnabled` function reference to `true` instead of calling it.
- Repaired existing tests broken by the refactor (private flush helpers; serverless-mode branch in the metrics index mock).
- Added coverage for the serialization cache/purge + `awsLambdaMode`, the `serializedData` getter, the serverless no-`started`-listener path, `flushToString` (in its own file — it registers a process-global meter provider and reaches the network, nock-stubbed), and `ServerlessCollector` attaching `otlp_payload`.
- Affected unit suites: 172 tests passing; lint clean.


lib/collector/serverless.js
jsumners-nr, line 1: This object has a few methods it only uses internally. I found it difficult to determine which methods actually needed to be exposed for the agent to work correctly. So I updated the internal only methods to be private. This makes it clear which methods are meant to be the public API and which are meant to be private.

test/unit/collector/serverless.test.js
jsumners-nr, line 1: As a consequence of making the API clearer, tests needed to be updated such that they monkey patch the underlying filesystem methods instead of replacing methods directly on the collector.

test/unit/collector/serverless.test.js
```diff
@@ -125,25 +125,76 @@ test('ServerlessCollector API', async (t) => {
...
+      assert.equal(outData[0], 1)
+      assert.equal(outData[1], 'NR_LAMBDA_MONITORING')
+      const decoded = JSON.parse(zlib.gunzipSync(Buffer.from(outData[2], 'base64')))
+      assert.notEqual(decoded.metadata, undefined)
+      assert.deepStrictEqual(decoded.data, testPayload)
       assert.equal(Object.keys(api.payload).length, 0)
-      assert.equal(flushed, true)
+    })
+
+    await t.test('includes OTEL metrics as otlp_payload when the metrics API is present', async (t) => {
```
jsumners-nr, line 168: This is a new test that verifies the collector attaches the serialized protobuf data correctly.

lib/collector/serverless.js
```diff
@@ -147,14 +146,29 @@ class ServerlessCollector {
       return
     }
 
+    if (this._agent.otel?.metrics?.flushToString) {
+      // Unfortunately, the OTEL API is designed in such a way that we cannot
+      // be fully synchronous here. That API is _both_ callback and promise
+      // based. So we have to finalize our export out-of-band when the metrics
+      // API is available (i.e. OTEL + OTEL Metrics have been enabled).
+      this._agent.otel.metrics.flushToString().then((serializedMetrics) => {
```
bizob2828, line 154: what happens if this fails? There's no catch handler so the agent may produce an unhandledRejection

lib/collector/serverless.js
```diff
@@ -147,14 +146,29 @@ class ServerlessCollector {
       return
     }
 
+    if (this._agent.otel?.metrics?.flushToString) {
```
bizob2828, line 149: Feels weird to be checking if there's a property on `otel.metrics`. What if there's some idea of enabled flags for all these otel apis?

lib/otel/metrics/index.js
```diff
@@ -88,13 +114,24 @@ class SetupMetrics extends SetupSignal {
...
+      // the `agent.on('started')` event. Which means that we can't finalize the
+      // metrics client configuration until after the `started` event.
+      logger.debug('Waiting for agent connect to finish bootstrapping OTEL metrics.')
+      agent.on('started', postReady)
+    } else {
+      // In this case, we simply need to assign a capturing exporter to our
+      // proxy so that we can manually collect metrics when the agent performs
+      // a harvest during serverless mode execution.
+      logger.debug('Finalizing OTEL metrics in serverless mode.')
+      proxyExporter.exporter = new NRCapturingExporter({ logger })
```
bizob2828, line 130: since we do not do a round trip with collector it's missing the entity_guid and any other `otlp_resource_attibutes`. Is the extension aware they will need to decorate metrics with those attributes?

lib/collector/serverless.js
bizob2828: Also are we certain this is actually going to work? this happens right before the lambda shutdown, feels like there's a chance of data loss because the promise may not have resolved

lib/otel/metrics/index.js
```diff
@@ -135,6 +175,22 @@ class SetupMetrics extends SetupSignal {
...
+  /**
+   * Initiates a manual metrics collection and returns the resulting protobuf
+   * array as a Base64 encoded string.
+   *
+   * @returns {Promise<string>} The encoded metrics data.
+   */
+  async flushToString() {
+    const { resourceMetrics: collectedMetrics } = await this.#metricReader.collect()
+    await new Promise((resolve) => {
+      this.#metricExporter.exporter.export(collectedMetrics, () => {
```
bizob2828, line 188: can you provide a comment explaining this. so we collect the metrics, then call export but return what was last serialized?  What happens if the export fails, could it return what was last collected on warm lambda invocations?

lib/otel/metrics/index.js
```diff
@@ -16,16 +16,40 @@ const {
 const { proxySettingsPresent } = require('#agentlib/collector/http-agents.js')
 const defaultLogger = require('../../logger').child({ component: 'opentelemetry-metrics' })
 const NROTLPMetricExporter = require('./nr-exporter.js')
+const NRCapturingExporter = require('./nr-capturing-exporter.js')
 const SetupSignal = require('../setup-signal.js')
 const ProxyingExporter = require('./proxying-exporter.js')
 const generateProxyAgentFactory = require('./generate-proxy-agent-factory.js')
 
+// See https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout#maximum_delay_value
+const MAXIMUM_TIMEOUT_TIME = ((2 ^ 32) / 2) - 1
```
bizob2828, line 25: do you have to provide an export interval? Doesn't this just create a timer that will never resolve and thus hang the process?

bizob2828 (changes requested):
A few comments/questions. I assumed you verified this with real lambdas?

lib/collector/serverless.js
jsumners-nr: Example please?

lib/collector/serverless.js
jsumners-nr: It's not really possible for it to fail.

The promise comes from the `.collect` method on the metrics reader object. That method is just compiling a JSON object of all the data that has been buffered. 

The callback comes from the `.export` method. Since we supply our own exporter that only deals with the in-memory data, there's nothing for the callback to fail on.

Regarding shutdown, I'm open to suggestions. The OTEL API requires the codes as I have written it. Our code expects fully synchronous operations. So I can't await the promise but I have to use the promise. 🤷‍♂️

lib/otel/metrics/index.js
jsumners-nr: Yes, we have to provide an interval. Otherwise we have to write our own metrics reader. The only actual implementation upstream provides is the periodic reader (https://open-telemetry.github.io/opentelemetry-js/classes/_opentelemetry_sdk-node.metrics.PeriodicExportingMetricReader.html).

The timer will not keep the process alive:

https://github.com/open-telemetry/opentelemetry-js/blob/3f9756ab3fb15aec3ac420f255cdbe70d3d280a8/packages/sdk-metrics/src/export/PeriodicExportingMetricReader.ts#L246-L257

lib/otel/metrics/index.js
jsumners-nr: That's a question for the team implementing the extension. To my knowledge, it already does this.

lib/otel/metrics/index.js
jsumners-nr: What suggestion do you have for the jsdoc block? I think it already provides the explanation.

There isn't any failure case possibility here that I am aware of. As explained above, all of this is happening on in-memory data. The only failure case I know of for `.export` is it not being able to ship the collected data to a remote HTTP endpoint. We are avoiding that completely.

jsumners-nr:
> A few comments/questions. I assumed you verified this with real lambdas?

I verified this with AWS's Docker based Lambda environment.

bizob2828 (approved):
We talked, I'm ok with this code, if there is an issue we can always address later. especially since this isn't 100% wired up with the lambda extension and this feature isn't GA

PR #4173, opened 2026-07-23
fix: Updated `openai` instrumentation to cache the results of `responses.parse` chain on openai custom promises

### LINKED ISSUE #4171: OpenAI 6.48.0 responses.parse() consumes an already-read response body when New Relic 14.3.3 is enabled
## Description

When the New Relic Node.js agent is enabled, OpenAI requests fail with:

```text
TypeError: Body is unusable: Body has already been read
```

This occurs with both:

- The `openai` package used directly.
- OpenAI clients instrumented through `@posthog/ai/openai`.

The failing application path uses `client.responses.parse(...)`, which ultimately calls `Response.json()` in OpenAI SDK 6.48.0. Undici reports that the response body was already consumed before the SDK can parse it.

The failure does not occur when the New Relic agent is disabled.

`NEW_RELIC_AI_MONITORING_ENABLED=false` and `NEW_RELIC_DISTRIBUTED_TRACING_ENABLED=false` are set, but the error still occurs.

The issue reproduces both inside the Docker container built from `node:22.22-slim` and when the application runs directly on the macOS host. It is therefore not container-specific.

## Expected Behavior

OpenAI requests, including `responses.parse()`, should work normally with the New Relic agent enabled.

Disabling AI Monitoring should prevent AI-specific monitoring behavior from affecting OpenAI SDK request/response handling. Disabling distributed tracing should not be required for OpenAI response parsing to work.

Configured:

```text
NEW_RELIC_AI_MONITORING_ENABLED=false
NEW_RELIC_DISTRIBUTED_TRACING_ENABLED=false
```

Observed error:

```text
TypeError: Body is unusable: Body has already been read
    at consumeBody (node:internal/deps/undici/undici:5854:15)
    at _Response.json (node:internal/deps/undici/undici:5807:18)
    at openai/src/internal/parse.ts:66:35
    at APIPromise.defaultParseResponse (openai/src/internal/parse.ts:72:5)
    at APIPromise.parseResponse (openai/src/core/api-promise.ts:40:41)
    at APIPromise.parseResponse (openai/src/core/api-promise.ts:40:41)
```

Relevant observations:

- The duplicate `APIPromise.parseResponse` frames may indicate that the response parsing path is invoked twice. This is a suspicion, not a confirmed root cause.
- `openai@6.48.0` is explicitly listed as supported by New Relic agent 14.3.3’s compatibility report.
- The agent source documents that `ai_monitoring.enabled=false` disables AI events/segments but keeps the OpenAI subscriber registered. It is gated by `instrumentation.openai.enabled`, not `ai_monitoring.enabled`.
- `distributed_tracing.enabled=false` only prevents distributed-trace header insertion; it does not disable OpenAI instrumentation.
- This appears related to, but is not the same failure as, fixed issue [#3379](https://github.com/newrelic/node-newrelic/issues/3379), where OpenAI `parse()` broke because the instrumented call did not preserve `APIPromise`.

Potential isolation test not yet run:

```text
NEW_RELIC_INSTRUMENTATION_OPENAI_ENABLED=false
```

If this resolves the failure, it would isolate the problem to New Relic’s OpenAI subscriber rather than generic Undici instrumentation.

## Steps to Reproduce

1. Run a Node.js application with `newrelic@14.3.3` loaded before application imports.
2. Set:

   ```text
   NEW_RELIC_AI_MONITORING_ENABLED=false
   NEW_RELIC_DISTRIBUTED_TRACING_ENABLED=false
   ```

3. Create either:
   - A standard `OpenAI` client from `openai`, or
   - An `OpenAI` client from `@posthog/ai/openai`.
4. Call `client.responses.parse(...)` and await its result.
5. Observe `TypeError: Body is unusable: Body has already been read` from `Response.json()`.
6. Disable the New Relic agent and repeat the same request.
7. The request succeeds with the agent disabled.

No standalone minimal reproduction repository is available yet. The production call uses a structured `responses.parse()` request and does not include sensitive request data in this report.

## Your Environment

- New Relic Node.js agent: `14.3.3`
- OpenAI SDK: `6.48.0`
- PostHog AI wrapper: `@posthog/ai@8.4.0`
- PostHog Node SDK: `posthog-node@5.46.0`
- Node.js: `22.22.3`
- Application container base image: `node:22.22-slim`
- Runtime: Docker container and direct host execution
- Development host: macOS `15.7.7`
- Framework: NestJS `11.1.19`
- Package manager: pnpm `10.34.5`

## Additional context

The OpenAI SDK’s `APIPromise` deliberately caches its parsed response so its response body is consumed once. With New Relic enabled, Undici rejects the SDK’s first observed `response.json()` call because another layer has already disturbed or locked the body stream.

New Relic’s current OpenAI subscriber instruments:

- `Responses.create` for OpenAI `>=4.87.0`
- `OpenAI.makeRequest` for OpenAI `>=4.0.0`

The affected SDK version falls within both supported ranges.

## Workaround

Disabling the affected New Relic library instrumentation resolves the issue while keeping the New Relic agent enabled:

```text
NEW_RELIC_INSTRUMENTATION_OPENAI_ENABLED=false
NEW_RELIC_INSTRUMENTATION_GOOGLE-GENAI_ENABLED=false
```

The process/container must be restarted after changing these settings.

With these settings applied, OpenAI requests complete successfully both when using the OpenAI SDK directly and through the PostHog OpenAI wrapper.

## Documentation and Configuration Feedback

I could not find documentation for the per-library settings required to disable OpenAI or Google GenAI instrumentation.

The documented `NEW_RELIC_AI_MONITORING_ENABLED=false` setting strongly implies that New Relic will stop instrumenting AI SDK calls. In practice, it only suppresses AI Monitoring events and segments while the OpenAI and Google GenAI subscribers remain registered and continue to alter the request execution path.

This behavior is unexpected, difficult to diagnose, and caused production-impacting OpenAI failures. I am severely frustrated that the documented disable setting did not disable the instrumentation responsible for the issue, while the effective per-library settings were undocumented.

Please document the distinction clearly and document these settings:

```text
NEW_RELIC_INSTRUMENTATION_OPENAI_ENABLED=false
NEW_RELIC_INSTRUMENTATION_GOOGLE-GENAI_ENABLED=false
```

bizob2828:
## Description

When our agent instruments `Responses.create`/ `Completions.create`, it side-chains onto the returned `APIPromise` by calling `.then()`. This triggers `APIPromise.parse()`, which creates an internal parsedPromise chain that reads the response body via `response.json()`. When the caller then invokes `responses.parse()`/`completions.parse()`, that method calls `_thenUnwrap()` on the same APIPromise to create a second promise — and that second promise also independently calls `response.json()` on the already-consumed body, producing `TypeError: Body is unusable: Body has already been read.`

This PR adds instrumentation for `APIPromise._thenUnwrap()` that intercepts the moment a second promise is chained off an already-in-flight parse, and prevents it from re-reading the response body by caching the result.  openai custom promises are mess, fun fun fun. Turns out our tests were asserting this actual bug due to a misunderstanding of how `respones.parse`/`completions.parse` works.  I also fixed the tests to assert that when calling those methods you actually get the responses expected and not an error.

## How to Test

```sh
npm run versioned openai
```

## Related Issues
Closes #4171

### DIFF
--- a/lib/subscribers/openai/legacy-api-promise.js
+++ b/lib/subscribers/openai/legacy-api-promise.js
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+const OpenAIApiPromiseSubscriber = require('./api-promise')
+
+/**
+ * Overrides the behavior of wrapping `_thenUnwrap` for openai < 5
+ */
+class OpenAILegacyApiPromiseSubscriber extends OpenAIApiPromiseSubscriber {
+  constructor({ agent, logger }) {
+    super({ agent, logger, channelName: 'nr_legacyThenUnwrap' })
+  }
+
+  /**
+   * In openai <5 parse() captures parseResponse by reference at setup time.
+   * Wrap `parsedPromise` and cache the first resolved parsedResponse to avoid
+   * it getting called multiple times
+   *
+   * @param {object} data from tracing channel
+   */
+  cacheWrappedPromise(data) {
+    const original = data.self
+    const newPromise = data.result
+
+    const cached = original.parsedPromise
+    newPromise.parsedPromise = cached.then(async function wrappedParsedPromise(baseData) {
+      const props = await original.responsePromise
+      const savedParseResponse = original.parseResponse
+      // assign to original parseResponse but it is an async function
+      // however baseData was already resolved so this just injects it
+      // into the call path of _thenUnwrap
+      original.parseResponse = () => Promise.resolve(baseData)
+      try {
+        return await newPromise.parseResponse(props)
+      } finally {
+        original.parseResponse = savedParseResponse
+      }
+    })
+  }
+}
+
+module.exports = OpenAILegacyApiPromiseSubscriber
--- a/lib/subscribers/openai/api-promise.js
+++ b/lib/subscribers/openai/api-promise.js
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+const Subscriber = require('../base')
+
+/**
+ * Instruments APIPromise._thenUnwrap to prevent double response body consumption.
+ *
+ * When NR's end handler for nr_responses calls wrapPromise(), it calls
+ * apiPromise.then() which triggers APIPromise.parse(), setting parsedPromise
+ * (a chain: responsePromise → defaultParseResponse → response.json()).
+ *
+ * responses.parse() then calls _thenUnwrap() on that same APIPromise to create
+ * a second APIPromise whose parseResponse closure calls the original's parseResponse
+ * field directly, bypassing the parsedPromise cache. Both chains subscribe to the
+ * same responsePromise and both call response.json(), causing "Body is unusable".
+ */
+class OpenAIApiPromiseSubscriber extends Subscriber {
+  constructor({ agent, logger, channelName = 'nr_thenUnwrap' }) {
+    super({ agent, logger, packageName: 'openai', channelName })
+    this.events = ['end']
+    this.requireActiveTx = false
+  }
+
+  /**
+   * Not all calls to _thenUnwrap have to be instrumented. This check verifies
+   * if the properties exist that need wrapping on all openai versions
+   * @param {object} data from tracing channel
+   * @returns {boolean} if the _thenUnwrap should be instrumented
+   */
+  skipWrapping(data) {
+    return data?.self?.parsedPromise === undefined || data.result === undefined
+  }
+
+  /**
+   * parse() uses a closure that reads parseResponse via a live property lookup at call time, so caching it works.
+   * @param {object} data from tracing channel
+   */
+  cacheWrappedPromise(data) {
+    const original = data.self
+    const originalParseResponse = original.parseResponse
+    let cachedResult = null
+    original.parseResponse = function wrappedParseResponse(client, props) {
+      if (!cachedResult) {
+        cachedResult = originalParseResponse(client, props)
+      }
+      return cachedResult
+    }
+  }
+
+  end(data) {
+    if (this.skipWrapping(data)) {
+      return
+    }
+
+    this.cacheWrappedPromise(data)
+  }
+}
+
+module.exports = OpenAIApiPromiseSubscriber
--- a/lib/subscribers/openai/config.js
+++ b/lib/subscribers/openai/config.js
@@ -68,6 +68,34 @@ module.exports = {
       ]
     },
     {
+      path: './openai/legacy-api-promise.js',
+      instrumentations: [
+        {
+          channelName: 'nr_legacyThenUnwrap',
+          module: { name: 'openai', versionRange: '>=4.87.0 <5.0.0', filePath: 'core.js' },
+          functionQuery: {
+            className: 'APIPromise',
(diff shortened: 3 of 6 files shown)


lib/subscribers/openai/legacy-api-promise.js
```diff
@@ -0,0 +1,45 @@
...
+      const props = await original.responsePromise
+      const savedParseResponse = original.parseResponse
+      // assign to original parseResponse but it is an async function
+      // however baseData was already resolved so this just injects it
+      // into the call path of _thenUnwrap
+      original.parseResponse = () => Promise.resolve(baseData)
+      try {
+        return await newPromise.parseResponse(props)
+      } finally {
+        original.parseResponse = savedParseResponse
```
jsumners-nr, line 39: I'm confused about these `parseResponse` assignments. It looks like `savedParseResponse` will overwrite the `() => Promise.resolve(baseData)` before that function has had an opportunity to be invoked. Which seems like it would sporadically not return the `baseData` back upstream. Should the original method be restored after the `baseData` returning function has completed? e.g.:

```js
original.parseResponse = () => {
	return Promise.resolve(baseData).finally(() => {
		original.parseResponse = savedParseResponse
	})
}
```

jsumners-nr (approved):
If I am misunderstanding how things work, I don't see any blockers. But I do have a comment for clarification.

lib/subscribers/openai/legacy-api-promise.js
bizob2828: So this is saving off the original `parseResponse` on the instance `data.self`.  It's redefining it as a function that immediately returns with baseData as it's already been resolved. Then re-assigns the original `parseResponse` back on the instance `data.self`. 

So in openai < 5 it does this 

```
    _thenUnwrap(transform) {
        return new APIPromise(this.responsePromise, async (props) => _addRequestID(transform(await this.parseResponse(props), props), props.response));
    }
```

and in 5+ it does:

```
    _thenUnwrap(transform) {
        return new APIPromise(tslib_1.__classPrivateFieldGet(this, _APIPromise_client, "f"), this.responsePromise, async (client, props) => (0, parse_1.addRequestID)(transform(await this.parseResponse(client, props), props), props.response));
    }
```

bizob2828: I asked claude to explain it better because my brain is done for the day:

api-promise.js (openai >= 5): parse() reads parseResponse via a live property lookup at call time — it accesses this.parseResponse each time it runs. So patching the property on data.self is sufficient: when parse() eventually executes, it picks up the wrapped version that memoizes the result.

legacy-api-promise.js (openai < 5): parse() captures parseResponse by reference in a closure at setup time (when _thenUnwrap is called). Patching data.self.parseResponse afterward is too late — the closure already holds a direct reference to the original function and will never see the patch.

So the legacy subscriber has to intercept parsedPromise (the promise chain) itself instead. By wrapping that chain, it can temporarily swap in a parseResponse that just returns the already-resolved baseData, call through to newPromise.parseResponse, then restore the original — effectively hijacking the data path without needing the closure to pick up a property change.

In short: openai >= 5 does a live this.parseResponse lookup (patchable), openai < 5 closes over the function reference at construction time (not patchable after the fact).

jsumners-nr: > So this is saving off the original `parseResponse` on the instance `data.self`. It's redefining it as a function that immediately returns with baseData as it's already been resolved. Then re-assigns the original `parseResponse` back on the instance `data.self`.

Yeah, but I'm saying it looks like the execution is:

1. patch with a function that returns the value
2. patch the original function back to the object

When my guess is it should be:

1. patch with a function that returns the value
2. stuff happens and the patched function is invoked
3. patch the original function back to the object

PR #4182, opened 2026-07-29
ci: Added a poll npm workflow to ensure new package version exists before executing any dependent workflows

### LINKED ISSUE #4180: Add poll before invoking post release jobs
### Description
npm is rolling out some changes to [harden against supply chain attacks](https://github.blog/changelog/2026-07-28-npm-publish-time-malware-scanning-and-dual-use-metadata/).  We have a few workflows that run after we publish and then install the latest version of the agent:
 * https://github.com/newrelic/node-newrelic/blob/main/.github/workflows/release-lambda-init-containers.yml
 * https://github.com/newrelic/node-newrelic/blob/main/.github/workflows/azure-site-extension.yml

We should introduce some workflow after `Create Release` that polls npm to ensure the new version is in registry before calling the above workflows

bizob2828:
## Description

This PR introduces a polling workflow after create release. This is necessary for the azure site extension and lambda and init container workflows because they run `npm i newrelic@<new-version>`. This new workflow verifies the new version of `newrelic` exists before exiting. It will poll every 30 s for n attempts. The default is so but in our repo action variables i set it to 30. So this will wait 15 minutes, for now, before failing.  I can't seem to get a straight answer on time to publish now but have a question out if `npm view newrelic@<version>` is the right mechanism to ensure the new version exists on registry.

As a sanity I modified the script locally to change the version to a legit one after 3 checks:

```sh
VERSION="14.5.0"
MAX_ATTEMPTS="20"
INTERVAL=30

echo "Polling npm for newrelic@${VERSION}..."

for i in $(seq 1 $MAX_ATTEMPTS); do
  NPM_VERSION=$(npm view "newrelic@${VERSION}" --prefer-online version 2>/dev/null || echo "")
  echo "Attempt ${i}/${MAX_ATTEMPTS}: registry returned '${NPM_VERSION}'"
  if [ "$i" = "2" ]; then
    VERSION="14.3.4"
  fi

  if [ "${NPM_VERSION}" = "${VERSION}" ]; then
    echo "newrelic@${VERSION} is available on npm."
    exit 0
  fi

  if [ "$i" -lt "$MAX_ATTEMPTS" ]; then
    echo "Not yet available, sleeping ${INTERVAL}s..."
    sleep $INTERVAL
  fi
done

echo "ERROR: newrelic@${VERSION} did not appear on npm after $((MAX_ATTEMPTS * INTERVAL))s."
exit 1
```

and the output was:

```
Polling npm for newrelic@14.5.0...
Attempt 1/20: registry returned ''
Not yet available, sleeping 30s...
Attempt 2/20: registry returned ''
Not yet available, sleeping 30s...
Attempt 3/20: registry returned ''
newrelic@14.3.4 is available on npm.
```

## Related Issues
Closes #4180

### DIFF
--- a/.github/workflows/azure-site-extension.yml
+++ b/.github/workflows/azure-site-extension.yml
@@ -6,7 +6,7 @@ permissions:
 on:
   workflow_dispatch:
   workflow_run:
-    workflows: ["Create Release"]
+    workflows: ["Poll NPM for Release Availability"]
     types:
       - completed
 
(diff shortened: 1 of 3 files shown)


.github/workflows/azure-site-extension.yml
```diff
@@ -45,62 +45,31 @@ jobs:
...
-          echo "Done with delayed check. Published version: $env:npm_agent_version Local version: ${{env.AGENT_VERSION}}"
-          echo "NPM_AGENT_VERSION=$env:npm_agent_version" | Out-File -FilePath $env:GITHUB_ENV -Append
-
-      - name: Has the new agent been published?
-        if: ${{ env.NPM_AGENT_VERSION != env.AGENT_VERSION }}
-        run: |
-          echo "Published agent version (${{env.NPM_AGENT_VERSION }}) is behind local agent version (${{env.AGENT_VERSION}}); exiting."
-          exit 1;
+          $version = node -p "require('./package.json').version"
+          echo "version=$version" >> $env:GITHUB_OUTPUT
```
jsumners-nr, line 52: This could probably be simplified:

```
jq -r '.version' package.json >> $env:GITHUB_OUTPUT
```

jsumners-nr (approved):

.github/workflows/azure-site-extension.yml
bizob2828: good point

bizob2828: I'm going to keep because you have to name the variable so it'd be something like `jq -r '"version=\(.version)"' ./package.json >> $GITHUB_OUTPUT`

jsumners-nr: Odd. 👍

bizob2828: we have so many ways we get the version:

```
agent-metadata.yml
11:        description: 'Version number (needs to match package.json)'
51:            VERSION=$(node -p "require('./package.json').version")
62:      # Our 'version' in package.json does not have a 'v'. This reusable workflow checks out a tag which is prefixed with a 'v', so we'll add it here.

azure-site-extension.yml
48:      - name: Get agent version from package.json
51:          $version = node -p "require('./package.json').version"

poll-npm-availability.yml
23:      - name: Get agent version from package.json
25:        run: echo "version=$(node -p "require('./package.json').version")" >> $GITHUB_OUTPUT

post-release.yml
48:        run: echo "latest_tag=$(cat package.json | jq .version)" >> $GITHUB_OUTPUT
70:        run: echo "latest_tag=$(cat package.json | jq .version)" >> $GITHUB_OUTPUT

release-lambda-init-containers.yml
25:        run: echo "latest_tag=$(cat package.json | jq .version)" >> $GITHUB_OUTPUT
```

bizob2828: maybe i'll do a follow up to normalize on this

PR #4184, opened 2026-07-30
fix: Ensuring addAttribute only increases the count for new Attributes

jmjones88:
## Description

This fixes an issue in the new relic agent where `addAttribute`  always assumed the attribute wasn't present when called and increased the attribute count. This led to reaching the attribute limit faster then expected/

## How to Test

Tested with unit tests and with a running application to ensure addAttribute now checks for updates

### DIFF
--- a/lib/attributes.js
+++ b/lib/attributes.js
@@ -118,7 +118,8 @@ class Attributes {
    * @param {boolean} [truncateExempt] - Flag marking value exempt from truncation
    */
   addAttribute(destinations, key, value, truncateExempt = false) {
-    if (this.attributeCount + 1 > this.limit) {
+    const isNewAttribute = !this.has(key)
+    if (isNewAttribute && this.attributeCount + 1 > this.limit) {
       return logger.debug(
         `Maximum number of custom attributes have been added.
         Dropping attribute ${key} with ${value} type.`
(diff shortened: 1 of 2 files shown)


jsumners-nr:
I'm having a difficult time understanding the need for this change. Can you please describe, or provide a minimal reproduction of, the scenario that leads to needing this change?

jmjones88:
The way that this was discovered @jsumners-nr is I was calling `addCustomAttribute` with the same attribute and different values. Every time that happens, `attributeCount` is increased, leading to future custom attributes being silently dropped (since the New Relic agent already thinks it's sent 64 custom attributes).

I added a unit test to simulate that simply updating a value shouldn't be counted as a new attribute.

jsumners-nr:
Am I understanding this issue correctly? You're expecting the following:

```js
addAttribute(0x01, 'foo', 'foo') // count = 1, "foo = foo"
addAttribute(0x01, 'bar', 'bar') // count = 2, "foo = foo";"bar = bar"
addAttribute(0x01, 'foo', 'baz') // count = 2, "foo = baz";"bar = bar"
```

jmjones88:
Yes exactly. That way you can update the current value of attributes without adding to the amount of user attributes per transaction.

If this is intentional, then I'm happy to close this. It was just a little unexpected that even sending the same value for an attribute increases the attribute count too.

jsumners-nr:
> Yes exactly. That way you can update the current value of attributes without adding to the amount of user attributes per transaction.
> 
> If this is intentional, then I'm happy to close this. It was just a little unexpected that even sending the same value for an attribute increases the attribute count too.

Being transparent: we've just been trying to understand the intention of this API internally. Standby for a resolution.

jsumners-nr (approved):
Thank you for contributing this fix.

jsumners-nr:
Please give release https://github.com/newrelic/node-newrelic/releases/tag/v14.3.6 a try.

PR #4194, opened 2026-08-03
chore: Updated how we reference agent version in the site extension file name

bizob2828:
## Description

In #4182 we got rid of `AGENT_VERSION` as an env var. Not all the places that references that env var were updated to the new github variable.


jsumners-nr (approved):
Very easy miss.

PR #4196, opened 2026-08-03
fix: Correctly parse `subquery` collection names in SQL; assert previously-skipped cross agent test cases

### LINKED ISSUE #2521: Update CAT tests in `sql.test.js`

amychisholm03:
## Description

This PR solves the issue described in #2521. In `test/unit/db/query-parsers/sql.test.js`, we had a couple of `t.todo`s that skipped asserting certain behavior in the CAT subtests: `("subquery")` and `db.table` specifically. 

Before this PR, the parser returned a unhelpful collection name (i.e. `("SELECT")`) for `FROM`-subqueries instead of a real table. Added detection in `lib/db/query-parsers/sql.js` so it now returns `"(subquery)"` per the [cross-agent tests](https://source.datanerd.us/agents/cross_agent_tests/blob/2e25086365918708a9cb4a1dd47d3d30941dd961/sql_parsing.json#L21-L22), matching what other language agents already do.

However, for `db.table`, we cannot strip the schema/database prefix (e.g. `"database"` in `"database.foobar"`) because that would conflict with existing metric names. So, I changed the `todo` to assert the behavior we currently produce.

## How to Test

```
npm run unit
```

## Related Issues

Closes #2521

### DIFF
--- a/lib/db/query-parsers/sql.js
+++ b/lib/db/query-parsers/sql.js
@@ -23,6 +23,10 @@ const OPERATIONS = [
   new StatementMatcher('delete', /^[^\S]*?delete[^\S]+?from[^\S]+([^\s\n\r,(;]+)/gim)
 ]
 const COMMENT_PATTERN = /\/\\*.*?\\*\//g
+// Matches a SELECT whose FROM clause is itself a subquery, e.g.
+// `SELECT * FROM (SELECT * FROM foobar)`. The outer statement matchers can't
+// name a real table for these, so they're reported as the literal '(subquery)'.
+const SUBQUERY_PATTERN = /^\s*select\b[\s\S]*?\bfrom[\s\n\r]*\(\s*select\b/i
 /* eslint-enable no-useless-escape, sonarjs/super-linear-regex, sonarjs/duplicates-in-character-class */
 
 // This must be called synchronously after the initial db call for backtraces to
@@ -51,6 +55,15 @@ module.exports = function parseSql(sql) {
   }
 
   sql = sql.replace(COMMENT_PATTERN, '').trim()
+
+  if (SUBQUERY_PATTERN.test(sql)) {
+    return {
+      operation: 'select',
+      database: null,
+      collection: '(subquery)',
+      query: sql
+    }
+  }
 
   let parsedStatement
 
--- a/test/benchmark/db/query-parsers/sql.bench.js
+++ b/test/benchmark/db/query-parsers/sql.bench.js
@@ -31,6 +31,14 @@ const tests = [
     fn: selectStatement
   },
   {
+    name: 'plain-select-statement',
+    fn: plainSelectStatement
+  },
+  {
+    name: 'subquery-select-statement',
+    fn: subquerySelectStatement
+  },
+  {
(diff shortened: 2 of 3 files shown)


lib/db/query-parsers/sql.js
```diff
@@ -23,6 +23,10 @@ const OPERATIONS = [
   new StatementMatcher('delete', /^[^\S]*?delete[^\S]+?from[^\S]+([^\s\n\r,(;]+)/gim)
 ]
 const COMMENT_PATTERN = /\/\\*.*?\\*\//g
+// Matches a SELECT whose FROM clause is itself a subquery, e.g.
+// `SELECT * FROM (SELECT * FROM foobar)`. The outer statement matchers can't
+// name a real table for these, so they're reported as the literal '(subquery)'.
+const SUBQUERY_PATTERN = /^\s*select\b[\s\S]*?\bfrom[\s\n\r]*\(\s*select\b/i
```
bizob2828, line 29: my only concern is the [spec](https://source.datanerd.us/agents/agent-specs/blob/main/SQL-Obfuscation-LEGACY.md) isn't updated to include how to parse subqueries.

amychisholm03: Yeah I'll reach out to other agent teams to see if it is still valid and if so, add it to the spec

jsumners-nr (commented):
Can you provide some numbers for this PR:

1. How does this change the baseline performance?
2. How does this compare to https://github.com/newrelic/node-newrelic/pull/2714

jsumners-nr:
Note: we have a benchmark that would need to be updated to cover new cases -- https://github.com/newrelic/node-newrelic/blob/229ff0d1f1926ae937b8a1c54975c5ee8b1f0d2c/test/benchmark/db/query-parsers/sql.bench.js

amychisholm03:
> Can you provide some numbers for this PR:
> 
> 1. How does this change the baseline performance?
> 2. How does this compare to #2714

Had Claude help me run the benchmarks and this is what it came up with:
**Benchmark comparison: current vs. this fix vs. #2714**

Ran all three `sql.js` implementations against the same 6 queries (`hrtime`, 2M iters × 5 trials, median ns/op):

| query | current | this PR | #2714 |
|---|---|---|---|
| insert w/ comment | 220 | 231 (+5%) | 790 (3.6x) |
| CTE select | 91 | 100 (+10%) | 487 (5.3x) |
| plain select | 147 | 180 (+22%) | 1232 (8.4x) |
| subquery select | 155 | 50 (**-68%**) | 687 (4.4x) |
| update | 150 | 158 (+5%) | 842 (5.6x) |
| delete | 170 | 179 (+5%) | 717 (4.2x) |

This fix adds a few ns on most queries and makes subqueries *faster* (short-circuits before the matcher loop). #2714's procedural rewrite is 3.6–8.4x slower across the board (matches the regression its own author reported), and it still wouldn't pass the `(subquery)` CAT assertions this PR fixes — it resolves the real inner table name instead of the literal `"(subquery)"` the spec expects.

jsumners-nr (approved):

PR #4203, opened 2026-08-04
test: Add benchmarks for lib/attributes.js methods

jsumners-nr:
## Description

Adds benchmark coverage for every method in `lib/attributes.js` and extends the benchmark suite to support setup that is excluded from the measured work.

### `test/benchmark/lib/attributes.bench.js`
Benchmarks each `Attributes` method following the pattern in `test/benchmark/db/query-parsers/sql.bench.js` (flat `tests` array of `{ name, fn }`, a loop calling `suite.add`, then `suite.run()`):

- `constructor`, `isValidLength`, `_set`, `get`, `has`, `reset`, `addAttribute`, `addAttributes`, `hasValidDestination`

Because the `Attributes` constructor reads config via `Config.getInstance()`, the suite loads an agent through the framework's `agent: { config }` option. Per-run setup (creating / pre-populating instances) is done in `before` so it is not part of the timed function.

### `test/lib/benchmark.js`
The `before` function's return value is now captured and awaited, so `before` may be synchronous or an `async function`. The returned value is passed to the test `fn` as a context object (second argument), for both the transaction and non-transaction paths. This is backward-compatible — no existing benchmark `fn` reads a second positional argument, and `before`'s return value was previously discarded.

## Testing
- `node test/benchmark/lib/attributes.bench.js` emits stats for all nine methods.
- `npm run lint` passes (0 errors).

### DIFF
--- a/test/benchmark/lib/attributes.bench.js
+++ b/test/benchmark/lib/attributes.bench.js
@@ -1,0 +1,129 @@
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+const benchmark = require('#testlib/benchmark.js')
+const { Attributes } = require('#agentlib/attributes.js')
+const AttributeFilter = require('#agentlib/config/attribute-filter.js')
+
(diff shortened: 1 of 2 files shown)


bizob2828 (approved):
LGTM

PR #4206, opened 2026-08-05
fix: Improved Apollo server resolver instrumentation performance

jsumners-nr:
We have a customer reporting that they have seen a significant performance impact upon moving from our external Apollo server plugin instrumentation approach to our current subscriber approach. This PR attempts to improve the situation by reducing the scenarios under which expensive work is performed.

-----

## Description

`wrapResolve` in the `@apollo/server` resolve subscriber runs once per resolved GraphQL field, so its per-call cost is multiplied across every field of every query.

Previously it flattened the field path (`flattenToArray` + `reverse().join('.')`) and the resolver args (`flattenArgs`) on **every** call, *before* the check that short-circuits scalar fields. Under the default config (`apollo_server.scalars = false`), a non-top-level scalar field creates no segment — so all of that work was computed and immediately discarded. Scalar leaf fields (`id`, `name`, `title`, ...) dominate real responses (~11:1 vs object fields in a typical nested query), so this was wasted work on the hot path for the majority of resolver invocations.

This PR reorders `wrapResolve` to:
1. Run the cheap scalar / top-level check **first**.
2. Build the segment name (`formattedPath`) only when a segment will actually be created.
3. Flatten resolver args only when they are consumed — i.e. `field_metrics` is enabled, or a segment is kept and the field has args. Arg-less and skipped fields no longer pay for it. `maybeAddAttributes` now guards against a null `flattenedArgs`.

Behavior is unchanged — the resolver segment tree, field attributes, field metrics, and error handling are all identical.

## How to Test

**Behavior (unchanged):** the `apollo-server` versioned suite passes in full against the change — 115/115 across `segments`, `attributes`, `metrics`, `errors`, `transaction-naming`, and `query-obfuscation`. The `attributes` and `metrics` tests specifically exercise the arg-flattening path that was made lazy.

**Performance:** this PR also adds a microbenchmark for the resolve subscriber hot path at `test/benchmark/lib/subscribers/apollo-server/`, covering each field-type case plus the individual helpers. Run with:

```
node ./bin/run-bench.js lib/subscribers/apollo-server
```

Before → after (mean per call, stable across runs):

| case | before | after | change |
|------|-------:|------:|-------:|
| `wrapResolve/scalar-no-args`   | 0.00171ms | 0.00132ms | **−23%** |
| `wrapResolve/scalar-with-args` | 0.00242ms | 0.00171ms | **−29%** |
| `wrapResolve/object-no-args`   | 0.00446ms | ~0.00428ms | −6% |
| `wrapResolve/object-with-args` | 0.00524ms | ~0.00515ms | ~same |

The scalar (no-segment) cases get materially faster; the object (segment-creating) path is unchanged, as expected — its cost is real segment creation.

## Related Issues

N/A

### DIFF
--- a/lib/subscribers/apollo-server/resolve.js
+++ b/lib/subscribers/apollo-server/resolve.js
@@ -57,7 +57,12 @@ module.exports = class ApolloResolveSubscriber extends BaseSubscriber {
     }
 
     if (fieldMetrics) {
-      this.maybeCaptureFieldMetrics({ transaction, info, args: flattenedArgs || {} })
+      this.captureFieldMetrics({
+        transaction,
+        fieldName: info.fieldName,
+        fieldType: info.parentType.toString(),
+        args: flattenedArgs || {}
+      })
     }
 
     if (skipSegment) {
@@ -147,26 +152,6 @@ module.exports = class ApolloResolveSubscriber extends BaseSubscriber {
           }
         }
       }
-    }
-  }
-
-  /**
-   * Captures both field and args of resolvers as metrics.
-   *
-   * This is intended to be used to determine if a field within a graphql schema is still being requested.
-   *
-   * @param {object} params to function
-   * @param {object} params.info info key from resolver context
-   * @param {object} params.transaction active transaction
-   * @param {object} params.args args key from resolver context
-   *
-   */
-  maybeCaptureFieldMetrics({ info, transaction, args }) {
-    const { config } = this
-    if (config.apollo_server.field_metrics) {
-      const fieldName = info.fieldName
-      const fieldType = info.parentType.toString()
-      this.captureFieldMetrics({ transaction, args, fieldName, fieldType })
     }
   }
 
--- a/lib/subscribers/apollo-server/resolve.js
+++ b/lib/subscribers/apollo-server/resolve.js
@@ -37,15 +37,36 @@ module.exports = class ApolloResolveSubscriber extends BaseSubscriber {
     const operationSegment = ctx?.segment
     const transaction = ctx.transaction
     const [, resolverArgs, , info] = args
-    const pathArray = this.flattenToArray(info.path)
-    const formattedPath = pathArray.reverse().join('.')
-    const flattenedArgs = this.flattenArgs({ obj: resolverArgs })
 
-    this.maybeCaptureFieldMetrics({ transaction, info, args: flattenedArgs })
+    // Do the cheap short-circuit check first. When scalar capture is disabled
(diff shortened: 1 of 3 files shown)


lib/subscribers/apollo-server/resolve.js
```diff
@@ -37,16 +37,37 @@ module.exports = class ApolloResolveSubscriber extends BaseSubscriber {
...
+    // them: field_metrics is enabled, or we are keeping the segment and there
+    // are args to attach. Arg-less fields (the common case) skip this.
+    const fieldMetrics = this.config.apollo_server.field_metrics
+    const hasArgs = resolverArgs != null && typeof resolverArgs === 'object'
+    let flattenedArgs = null
+    if (hasArgs && (fieldMetrics || !skipSegment)) {
+      flattenedArgs = this.flattenArgs({ obj: resolverArgs })
+    }
+
+    if (fieldMetrics) {
```
bizob2828, line 59: this check is done within `this.maybeCaptureFieldMetrics`

bizob2828 (commented):
I'm not sure this is the root cause of their issue. if you compare with our [old plugin](https://github.com/newrelic/newrelic-node-apollo-server-plugin/blob/977e93a815387e6e40e23e6337b8932f7b096e7d/lib/hooks.js#L147), it's identical

jsumners-nr:
> I'm not sure this is the root cause of their issue. if you compare with our [old plugin](https://github.com/newrelic/newrelic-node-apollo-server-plugin/blob/977e93a815387e6e40e23e6337b8932f7b096e7d/lib/hooks.js#L147), it's identical

I agree. But it is a win.

lib/subscribers/apollo-server/resolve.js
```diff
@@ -37,16 +37,42 @@ module.exports = class ApolloResolveSubscriber extends BaseSubscriber {
...
+      this.captureFieldMetrics({
+        transaction,
+        fieldName: info.fieldName,
+        fieldType: info.parentType.toString(),
+        args: flattenedArgs || {}
+      })
+    }
 
-    if (!this.config.apollo_server.scalars && !this.isTopLevelField(info) && this.isScalar(info)) {
+    if (skipSegment) {
```
bizob2828, line 68: so i think this may be the root cause of the slowness. I don't think we should be running scalar resolvers in context when we have them skipped

bizob2828 (approved):
nice optimization, left a comment that I _think_ might be the root cause of the slowness reported by customer

PR #4207, opened 2026-08-06
chore: Improved docs and benchmarking for attributes

jsumners-nr:
This PR wraps up my investigation in to our `Attributes` object. Initially, I wanted to see if rewriting the data structure would be a benefit, but the more I studied it the more convinced I became that would be a fruitless endeavor. Indeed, I had the bot consider the idea of a better data structure and it came up with the same ideas I had and struck them down for the same reasons I did.

There was one possible optimization, though. We currently truncate excessively long values on _every_ retrieval. In commit https://github.com/newrelic/node-newrelic/pull/4207/commits/a9eb9619b99c0da97c31cc101936fb80d5a333c8 we cache this operation on the first retrieval so that subsequent ones are quicker. In my opinion, a better approach would be to truncate at store time, but I don't know what the consequences of that would be. I don't think we ever read that full value out for anything, but maybe we do?

I'm also on board with dropping that last commit. It's an edge case optimization unlikely to have effect in the majority of cases.

-----

## Description

Improves the `Attributes` class: documentation, benchmarking, a module-resolution tidy-up, and a truncation-memoization performance optimization for the harvest path.

### Docs & tooling (ea56c608f)
- JSDoc improvements in `lib/attributes.js` and `lib/config/attribute-filter.js`
- `jsdoc-conf.js` and `package.json` adjustments

### `realistic-usage` benchmark (1c48641a9)
Adds a benchmark to `test/benchmark/lib/attributes.bench.js` that exercises the real-world usage pattern:
- **Seeds 60 attributes** into a `limit: 64` instance in the `before` hook, so the seed cost is excluded from the measured window.
- **During measurement**, adds new keys to reach and exceed the 64-attribute maximum (crossing the limit and exercising the drop path), then updates existing keys to exercise the overwrite path.
- **Uses a mix of destination masks** (single- and multi-destination) so both single-bit and multi-bit filtering paths are covered.

### Reduce module resolution cost (b47d02db2)
Minor import adjustments in `lib/spans/*` and `lib/transaction/trace/*` to reduce module-resolution overhead.

### Memoize attribute value truncation (a9eb9619b)
A single `Attributes` instance is harvested once per destination (e.g. `trace.attributes` is read for `TRANS_EVENT`, `TRANS_TRACE`, and `ERROR_EVENT`), and each harvest re-truncated every string value via a byte-length scan plus a binary search for multibyte values.

The value limit is fixed for an instance's lifetime and `_set` replaces the whole record on overwrite, so the truncated value is now **memoized on the record** (`attr.truncated`) and reused across harvests. Overwriting a key transparently invalidates the cache because the record object is replaced.

**Benchmarked** on a 64-attribute instance with long multibyte values (median of 3 runs):
- Repeated-harvest `get`: **~4.8us -> ~2.8us per get (~41% faster)**
- Single-harvest case: unchanged within noise

Adds unit tests for repeated-harvest idempotency and overwrite invalidation, plus a `get-repeated-harvest` benchmark case so CI tracks this path.

## Testing
- `node test/benchmark/lib/attributes.bench.js` runs the full suite and emits stats for every method plus `realistic-usage` and `get-repeated-harvest`.
- `node --test test/unit/attributes.test.js` — 23 pass.
- Consumer suites pass: `test/unit/transaction/trace` (147), `test/unit/spans` (147).
- `npm run lint` passes (0 errors).

### DIFF
--- a/jsdoc-conf.js
+++ b/jsdoc-conf.js
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+// This configuration file is used to generate the full documentation, including
+// private information. The `jsdoc-conf.jsonc` is used as a baseline, as that
+// is the configuration for our public API docs.
+
+const fs = require('node:fs')
+const baselineData = fs.readFileSync('./jsdoc-conf.jsonc').toString('utf-8')
+const config = JSON.parse(
+  baselineData
+    .split('\n')
+    .map((l) => l.trim())
+    .filter((l) => l.startsWith('//') === false)
+    .join('\n')
+)
+
+// We want everything in `lib/` for our internal docs.
+config.source.include.push('lib/')
+
+delete config.opts.template
+delete config.opts.theme_opts
+
+module.exports = config
--- a/lib/attributes.js
+++ b/lib/attributes.js
@@ -10,6 +10,7 @@ const logger = require('./logger').child({ component: 'attributes' })
 const isValidType = require('./util/attribute-types')
 const byteUtils = require('./util/byte-limit')
 const properties = require('./util/properties')
+const { DESTINATIONS } = require('./config/attribute-filter')
 
 const MAXIMUM_CUSTOM_ATTRIBUTES = 64
 const MAXIMUM_ATTR_VALUE_LENGTH = 4_096
@@ -19,6 +20,7 @@ const MAXIMUM_ATTR_VALUE_LENGTH = 4_096
  * @private
  */
 class Attributes {
+  static DESTINATIONS = DESTINATIONS
   static SCOPE_TRANSACTION = 'transaction'
   static SCOPE_SEGMENT = 'segment'
 
@@ -47,7 +49,7 @@ class Attributes {
   /**
    * Checks if a given string is within agent attribute limits.
    *
-   * @param {string} str - Object key name or value
+   * @param {string} str Object key name or value
    */
   isValidLength(str) {
     return typeof str === 'number' || byteUtils.isValidLength(str, 255)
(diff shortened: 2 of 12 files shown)


lib/attributes.js
```diff
@@ -19,6 +20,7 @@ const MAXIMUM_ATTR_VALUE_LENGTH = 4_096
  * @private
  */
 class Attributes {
+  static DESTINATIONS = DESTINATIONS
```
jsumners-nr, line 23: This serves two purposes:

1. Destinations are tightly coupled to this object. So it makes sense to expose them from the object.
2. Enables in-IDE navigation to the declaration through IDE rendered documentation.

package.json
```diff
@@ -161,7 +161,7 @@
     "addhook:precommit": "git config --local core.hooksPath .githooks",
     "bench": "node ./bin/run-bench.js",
     "docker-env": "./bin/docker-env-vars.sh",
-    "docs": "rm -rf ./out && jsdoc -c ./jsdoc-conf.jsonc --private",
+    "docs": "rm -rf ./out && NODE_OPTIONS='--max-old-space-size=12288' jsdoc -c ./jsdoc-conf.js --private",
```
jsumners-nr, line 164: Yeah, it takes that much memory for jsdoc to render the theme (any theme) from the data in our `lib/` directory.

jsdoc-conf.js
```diff
@@ -0,0 +1,28 @@
...
+    .map((l) => l.trim())
+    .filter((l) => l.startsWith('//') === false)
+    .join('\n')
+)
+
+// We want everything in `lib/` for our internal docs.
+config.source.include.push('lib/')
+
+delete config.opts.template
+delete config.opts.theme_opts
```
jsumners-nr, line 26: I originally did this to see if using the standard theme would remove the need to specify the process memory limit. That was a failure, but I like the default theme better for the internal docs (I find it easier to read and navigate) so I left it.

bizob2828 (commented):
It looks like there's identical code in `lib/prioritized-attributes.js` I think you should update that as well right?  Otherwise looks good, this memoization seems to improve perf slightly

jsumners-nr:
I have addressed the feedback. We can ignore Codecov's complaint, because I already have a follow-up PR ready to migrate over here (https://github.com/jsumners-nr/node-newrelic/pull/3) that eliminates the uncovered path. I just can't stack it here because GitHub didn't implement stacking correctly.

bizob2828 (approved):

PR #4209, opened 2026-08-06
fix: Updated `getBrowserTimingHeader` to handle when a transaction may not be finalized and still return the browser agent string without transaction intrinsics

### LINKED ISSUE #4205: No transaction name for Next.js Browser RUM
## Description

When using the new [Hybrid mode](https://docs.newrelic.com/docs/apm/agents/nodejs-agent/extend-your-instrumentation/nextjs-instrumentation/) for Next.js instrumentation, the browser script are never injected, because of `NREUM: browser_monitoring headers need a transaction name`.

## Expected Behavior

The RUM script should be injected even if the transaction name is not present.

See reproduction.

## Steps to Reproduce

Here is a reproduction of the issue + some context that could be useful (found by AI): https://github.com/afoures/newrelic-node-examples/tree/issue-with-next-js-browser-rum/nextjs/nextjs-rum-unnamed-transaction

## Your Environment

* Next.js: 16.3.0
* Node: 22.23.2
* Agent version: 14.3.5

bizob2828:
## Description

This PR fixes behavior when using `newrelic.getBrowserTimingHeader` in a non-traditional server-side html rendering situation.  In full stack frameworks like Next.js, customers want to inject the browser agent and use our api to do so with `newrelic.getBrowserTimingHeader`. The issue is before this PR it required to have an active, finalized transaction.  We did add an option `allowTransactionlessHeader` but it didn't do enough. It simply just checked if there was an active transaction, which was true.  However due to loading over of the script tag in the next.js layout and the completion of the server side Next.js code, it was always returning `<!-- NREUM: (3) -->`, which meant it lacked a transaction name.  Web transactions aren't finalized until the server handlers are done which in this case wasn't true.  In fact in our legacy Next.js instrumentation, this didn't work correctly. It just encoded `WebTransaction/NormalizedUri/*` as the transaction name.

After speaking with folks on the browser agent, the idea of sending `transactionName` in the intrinsics for the browser agent isn't applicable.  The intention of this property is that it would be associated with all browser events. But since Next.js is just building a layout to be used in client and may have interactions from server it'd be Client -> Server. In the legacy server side rendering of javascript it's different it's Server -> Client. I also found we were passing in attributes that have been deprecated since 2014: ttGuid and agentToken, so I removed those.

So the fix is really to not make any of the transaction intrinsics required and just return what we can or in the case of when `allowTransactionlessHeader`, it doesn't even inject any transaction intrinsics into the browser agent.  This doesn't affect the experience and will still load the browser agent and properly capture things like DT between Next.js client and server interactions. You can see that from this trace:

screenshot 2026-08-06 at 5 11 04 PM

Lastly, I plan to amend our spec and call out that transaction intrinsics aren't required. The only required keys on the browser agent string are: `agent, beacon, errorBeacon, licenseKey, and applicationID`.

## Related Issues
Closes #4205

### DIFF
--- a/api.js
+++ b/api.js
@@ -725,6 +725,10 @@ function validateBrowserMonitoring(config, transaction, allowTransactionlessInje
     return { isValidConfig: false, failureIdx: 1 }
   }
 
+  if (transaction && transaction.isIgnored()) {
+    return { isValidConfig: false, failureIdx: 1 }
+  }
+
   return { isValidConfig: true }
 }
 
@@ -754,11 +758,11 @@ API.prototype.getBrowserTimingHeader = function getBrowserTimingHeader(options =
   )
   metric.incrementCallCount()
 
-  const trans = this.agent.getTransaction()
+  const transaction = this.agent.getTransaction()
 
   const { isValidConfig, failureIdx, quietMode } = validateBrowserMonitoring(
     this.agent.config,
-    trans,
+    transaction,
     options.allowTransactionlessInjection
   )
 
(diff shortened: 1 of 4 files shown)


jsumners-nr (approved):
Looks good to me.

PR #4211, opened 2026-08-07
ci: Updated `child_process.exec` calls to `child_process.execFile` in `bin/npm-commands.js`

anupamme:
## Summary
Harden input handling in `bin/npm-commands.js` (flagged by semgrep).

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | javascript.lang.security.detect-child-process.detect-child-process |
| **Severity** | MEDIUM |
| **Scanner** | semgrep |
| **Rule** | `javascript.lang.security.detect-child-process.detect-child-process` |
| **File** | `bin/npm-commands.js:22` |
| **Assessment** | Defensive hardening |

**Description**: Detected calls to child_process from a function argument `command`. This could lead to a command injection if the input is user-controllable. Try to avoid calls to child_process, and if it is needed, ensure user input is correctly sanitised or sandboxed. 

## Threat Model Context

This is a Node.js library - vulnerabilities affect downstream consumers who use this package.

## Changes
- `bin/npm-commands.js`

## Behaviour Preservation
The change is scoped to 1 file on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*This patch removes an exploit primitive — a code pattern that, while not independently exploitable today, could be chained with other weaknesses by automated exploit-development tooling. Proactive removal of such primitives raises the bar against increasingly capable automated attack tools.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/bin/npm-commands.js
+++ b/bin/npm-commands.js
@@ -5,21 +5,23 @@
 
 'use strict'
 
-const { exec } = require('child_process')
+const { execFile } = require('child_process')
 
 async function version(typeOrVersion, shouldCommitAndTag) {
-  let command = `npm version ${typeOrVersion}`
+  const args = ['version', typeOrVersion]
 
-  command += shouldCommitAndTag ? '' : ' --no-git-tag-version'
+  if (!shouldCommitAndTag) {
+    args.push('--no-git-tag-version')
+  }
 
-  await execAsPromise(command)
+  await execAsPromise(args)
 }
 
-function execAsPromise(command) {
+function execAsPromise(args) {
   return new Promise((resolve, reject) => {
-    console.log(`Executing: '${command}'`)
+    console.log(`Executing: 'npm ${args.join(' ')}'`)
 
-    exec(command, (err, stdout) => {
+    execFile('npm', args, (err, stdout) => {
       if (err) {
         return reject(err)
       }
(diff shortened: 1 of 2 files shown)


bizob2828 (changes requested):
@anupamme although these PRs have been valid, this code is purely used internal and not shipped with our software thus the exploits you're patching are not possible. If you want to continue to provide PRs and there are tests missing, please add them accordingly. As of now, we do not have unit tests for `bin/npm-commands.js`. They should go into `bin/test/npm-commands.test.js`

anupamme:
Thanks for the clarification; that makes sense. I agree that bin/npm-commands.js is internal tooling and isn’t shipped to downstream consumers, so I’ll correct the PR description and avoid framing this as a vulnerability affecting the published package.

I’d still like to keep the underlying hardening: switching from exec() to execFile() ensures that the typeOrVersion value is passed as an argument rather than being interpreted by a shell.

I’ll also add the missing tests under bin/test/npm-commands.test.js, including coverage for both shouldCommitAndTag paths and a regression test showing that shell metacharacters in typeOrVersion aren’t executed.

I’ll update the description and push the tests. Thanks for pointing me in the right direction.

bizob2828 (approved):

PR #4212, opened 2026-08-07
refactor: Refactored `child_process` instrumentation to use tracing channel instead of `Shim`

### LINKED ISSUE #4223: Migrate `child_process` to emit events over tracing channel
## Description

## Acceptance Criteria
All methods we previously instrumented with shim on the `child_process` API will be run over tracing channel and subscribers will create the necessary trace segments.

amychisholm03:
## Description

This PR converts the `child_process` `exec`/`execFile` instrumentation from `Shim`-based monkeypatching to utilize Node's native tracing channel API. It creates one `tracingChannel` per method and monkey patches each `exec`/`execFile` function once per process.

## How to Test

```
npm run unit
npm run integration
```

## Related Issues

Closes #4223


lib/instrumentation/core/child_process.js
```diff
@@ -5,33 +5,112 @@
...
 'use strict'
 
-const { RecorderSpec } = require('../../../lib/shim/specs')
+// eslint-disable-next-line n/no-unsupported-features/node-builtins
+const { tracingChannel } = require('node:diagnostics_channel')
+const logger = require('../../logger').child({ component: 'child_process' })
+
+// Create channels based on instrumented methods
+const methods = ['exec', 'execFile']
+const channels = new Map(methods.map((name) => [name, tracingChannel('child_process.' + name)]))
```
jsumners-nr: This is confusing and difficult to read for such a simple action. I'd write:
```suggestion
const channels = {
	exec: tracingChannel('child_process.exec'),
	execFile: tracingChannel('child_process.execFile')
}
```

lib/instrumentation/core/child_process.js
```diff
@@ -5,33 +5,112 @@
...
+// Patched once per process and never unwrapped; each `initialize()` call
+// just re-points the bound store at the current agent.
+let patched = false
+let currentTracer = null
+let boundStore = null
 
 module.exports = initialize
 
-function initialize(agent, childProcess, moduleName, shim) {
+function initialize(agent, childProcess) {
```
jsumners-nr: Since we are refactoring this code already, I'd inject the logger for easier testing:

```js
function initialize(agent, childProcess, { logger = defaultLogger } = {}) {
```

Note: this will require some more refactoring of the remaining code. Looking at it, I think I'd do:

```js
class ChildProcessInstrumentation {
	/* the support methods in your original code */
}

module.exports = function initialize (agent, childProcess, { logger = dl } = {}) {
	if (go_away) return
	const instrumentation = new ChildProcessInstrumentation(/* stuff */)
	return instrumentation
}
```

lib/instrumentation/core/child_process.js
```diff
@@ -5,33 +5,112 @@
...
+    return
+  }
+  patched = true
 
-  makePromisifyCompatible(shim, childProcess)
-}
+  for (const [methodName, channel] of channels) {
+    const original = childProcess[methodName]
+
+    childProcess[methodName] = function wrappedMethod(...args) {
```
jsumners-nr: Add an indicator that the method has been wrapped. See #4216.

lib/instrumentation/core/child_process.js
```diff
@@ -5,33 +5,112 @@
...
+// refactors over to tracing channel.
+function createSegment(tracer, ctx, name) {
+  const segment = tracer.createSegment({ name, parent: ctx?.segment, transaction: ctx?.transaction })
+  if (segment) {
+    segment.start()
   }
+  return segment
+}
+
+function rebindStore(agent) {
```
jsumners-nr: If we refactor this as I have suggested, the class will be decorated with all of the necessary things and so we don't have to pass the agent in here. The function would reference `this.#agent` and `this.#logger`.

jsumners-nr (dismissed):
Looking good, but I think we can improve a few things.

lib/instrumentation/core/child_process.js
```diff
@@ -1,37 +1,119 @@
...
-  if (!childProcess) {
-    shim.logger.debug('Could not find child_process, not instrumenting')
-    return false
+class ChildProcessInstrumentation {
+  /** currently active `ChildProcessInstrumentation` instance */
+  static #active = null
+
+  constructor(agent) {
+    this.tracer = agent.tracer
+    this.store = this.tracer._contextManager._asyncLocalStorage
```
jsumners-nr: Are these used by anything outside of the class?

lib/instrumentation/core/child_process.js
```diff
@@ -1,37 +1,119 @@
...
+  execFile: tracingChannel('child_process.execFile')
+}
 
-function initialize(agent, childProcess, moduleName, shim) {
-  if (!childProcess) {
-    shim.logger.debug('Could not find child_process, not instrumenting')
-    return false
+class ChildProcessInstrumentation {
+  /** currently active `ChildProcessInstrumentation` instance */
+  static #active = null
```
jsumners-nr: I'm having difficulty following this. Why do we need to keep track of "active" instances of this object?

amychisholm03: It was a bandaid fix for the multi-agent scenario (in `test/integration/core/child-process.js` for example), but I will commit a way to teardown this new instrumentation style because it's not on `Shim` and it's not a subscriber.

lib/instrumentation/core/child_process.js
amychisholm03: Nothing reads off of the specific instance of `ChildProcessInstrumentation.tracer` and `ChildProcessInstrumentation.tracer`, but `agent.tracer` and `agent.tracer._contextManager._asyncLocalStorage` is used widely across our codebase.

lib/instrumentation/core/child_process.js
bizob2828: our agent is a singleton so this is never in play for a real application

bizob2828 (changes requested):
I haven't studied this but we should keep the instrumentation separate from the subscribers of instrumentation. I know they are in separate classes but the subscribers should live in `lib/subscribers`. I know this is the first but we should be creating abstractions like the orchestrion and native diagnostics channel based instrumentation so it can be reused for future core node library subscribers

lib/instrumentation/core/child_process.js
amychisholm03: True, but we still need a way to teardown this new `tracingChannel`-only instrumentation, like we do with subscribers and `Shim`.

lib/instrumentation/core/child_process.js
```diff
@@ -1,37 +1,120 @@
...
+
+        const lastArg = args[args.length - 1]
+        const hasCallback = typeof lastArg === 'function'
+        const data = { methodName, callbackName: hasCallback ? lastArg.name || '<anonymous>' : null }
 
-function makePromisifyCompatible(shim, childProcess) {
-  const originalExec = shim.getOriginal(childProcess.exec)
-  for (const symbol of Object.getOwnPropertySymbols(originalExec)) {
-    childProcess.exec[symbol] = originalExec[symbol]
+        return hasCallback
```
bizob2828: if it doesn't have a callback i would think you would want to `tracePromise` or just never check this because the promisified versions just wrap the callback right?

lib/instrumentation/core/child_process.js
```diff
@@ -1,37 +1,120 @@
...
+  }
+}
+
+module.exports = function initialize(agent, childProcess, { logger = defaultLogger } = {}) {
+  if (!childProcess) {
+    logger.debug('Could not find child_process, not instrumenting')
+    return false
+  }
+
+  const instrumentation = new ChildProcessInstrumentation(agent)
```
bizob2828: I don't see a benefit of abstracting this in a class but i guess it's ok

lib/instrumentation/core/child_process.js
```diff
@@ -1,37 +1,120 @@
...
+      })
+    }
   }
+
+  teardown() {
+    for (const [methodName, channel] of Object.entries(channels)) {
+      if (this.originals[methodName]) {
+        this.childProcess[methodName] = this.originals[methodName]
+      }
+      channel.start.unbindStore(this.store)
```
bizob2828: i would think the subscriber would just unsubscribe from these events.  I'm not sure what unbinding the store does. Is this an issue where since these are native, the code is already wrapped with an agent that is no longer active? If so, I think we have to solve this in our test setup/teardown not in instrumentation because our agent is a singleton aside from our tests

bizob2828 (commented):
a few more comments. this is a good start but we probably want to study the abstracted design more as we add more core node.js subscribers and instrumentation that runs in tracing channels

lib/instrumentation/core/child_process.js
bizob2828: or is this a mapping to `execFileSync` as an example?

bizob2828: Either way I think our integration tests need more tests:
 * `execFileSync` `execSync`
 * using `util.promisify` and running `await exec` and `await execFile`

Also comparing the segment times from before and after this PR would be good, not sure if we can reliable assert times but if we can then I would recommend we do that as that's a common issue we've seen with migrating to subscribers and not running the function in the right tracing channel method or touching segments

lib/instrumentation/core/child_process.js
```diff
@@ -1,37 +1,120 @@
...
+  return instrumentation
+}
+
+// `end`/`asyncEnd` are agent-agnostic (operate only on what's already
+// stashed on `data` by the bindStore transforms above), so unlike
+// `start`/`asyncStart` they only ever need to be subscribed once.
+for (const channel of Object.values(channels)) {
+  channel.subscribe({
+    end(data) {
+      data.ctx?.segment?.touch()
```
bizob2828: is this correct? I would think the `asyncEnd` would touch both the segment and callbackSegment, you may want to use the debugger to see how this worked with shim before

lib/instrumentation/core/child_process.js
jsumners-nr: The benefit was to encapsulate the shared resources that either needed to be passed to the various functions or defined at a module scope level (making testing more difficult).

lib/instrumentation/core/child_process.js
bizob2828: ok looking at shim. the `ctx.segment` is getting touched when returning a wrapped callback so this is probably fine but may want to dig a little deeper on your own

lib/instrumentation/core/child_process.js
amychisholm03: We don't currently instrument `execFileSync` or `execSync`, just `execFile` and `exec`, but I can add the `util.promisify` and segment duration timing tests.

amychisholm03: Interestingly enough, the old `Shim` way didn't properly instrument `const asyncExecFile = util.promisify(cp.execFile)`, so I'll see if I can fix it in this refactor.

bizob2828: i would wait to solve new issues in diff PRs

lib/subscribers/core/child_process/index.js
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  /**
+   * Same as the base class's `handleCallback`, but parented off `data.ctx`
+   * (see {@link #handler}) instead of the ambient context.
+   *
+   * @param {TracingChannel} channel The tracing channel whose `asyncStart`
+   * event should be bound.
+   */
+  handleCallback(channel) {
```
bizob2828: i'm not sure i follow why this is being overriden

lib/subscribers/core/child_process/index.js
```diff
@@ -0,0 +1,116 @@
...
+   * `execFile` runs synchronously inside `exec`'s opaque call, its own
+   * `start`/`end` still enter and exit its segment's context around that
+   * window, and that's the context an async completion arriving afterward
+   * resumes in -- not `exec`'s.
+   *
+   * @param {object} data Event data published to the `start` event.
+   * @param {object} ctx The current async context.
+   * @returns {object} The context entered for this call.
+   */
+  handler(data, ctx) {
```
bizob2828: i'm not sure i follow why this is being overriden

lib/subscribers/core/child_process/index.js
```diff
@@ -0,0 +1,116 @@
...
+
+const INSTRUMENTED_METHODS = ['exec', 'execFile']
+
+// `exec`'s raw implementation calls `execFile` internally, forwarding its
+// own already-instrumented callback. Marking `exec` opaque means: while its
+// synchronous body is running (including whatever it calls internally),
+// nested instrumented methods skip their own callback tracking -- exec's own
+// tracking already covers that same completion. Only `exec` needs this;
+// `execFile` doesn't call any other instrumented method.
+const OPAQUE_METHODS = new Set(['exec'])
```
bizob2828: i don't think you need to do this now that we aren't creating callback segments

test/integration/core/child-process.test.js
bizob2828: you're going to have to update all `verifySegments` to pass in `assertCallbacks: false`

bizob2828 (changes requested):

lib/subscribers/core/child_process/index.js
```diff
@@ -0,0 +1,55 @@
...
+    const self = this
+    shimmer.wrapMethod(childProcess, 'child_process', INSTRUMENTED_METHODS, function wrapMethod(original, method) {
+      const channel = tracingChannel(`${self.id}:${method}`)
+
+      function wrappedMethod(...args) {
+        const lastArg = args.at(-1)
+        const hasCallback = typeof lastArg === 'function'
+        const data = { name: `${self.packageName}.${method}`, callbackName: hasCallback ? (lastArg.name || '<anonymous>') : null }
+
+        return hasCallback
```
bizob2828, line 40: i'm not sure i follow this. it's not that it has a callback, it's that it has nested calls right?

lib/subscribers/core/child_process/index.js
```diff
@@ -0,0 +1,55 @@
...
+   */
+  instrument(childProcess) {
+    const self = this
+    shimmer.wrapMethod(childProcess, 'child_process', INSTRUMENTED_METHODS, function wrapMethod(original, method) {
+      const channel = tracingChannel(`${self.id}:${method}`)
+
+      function wrappedMethod(...args) {
+        const lastArg = args.at(-1)
+        const hasCallback = typeof lastArg === 'function'
+        const data = { name: `${self.packageName}.${method}`, callbackName: hasCallback ? (lastArg.name || '<anonymous>') : null }
```
bizob2828: callbackName is no longer used

lib/subscribers/core/child_process/index.js
bizob2828: I think we need to introduce the `internal` property like we do in the core subscriber:

```
diff --git a/lib/subscribers/core/base.js b/lib/subscribers/core/base.js
index 91177a7f3..06d415abf 100644
--- a/lib/subscribers/core/base.js
+++ b/lib/subscribers/core/base.js
@@ -20,9 +20,6 @@ const { tracingChannel } = require('node:diagnostics_channel')
  * produces `'nr:promises:dns'`. Omit to use the short form `'nr:dns'`.
  * @property {string[]} [instrumentedMethods=[]] Names of the module methods
  * to instrument. One tracing channel is created per entry.
- * @property {boolean} [hasCallback=false] Set when the instrumented methods are
- * traced with `traceCallback`. Causes the segment created in `start` to be
- * stashed on the event data so `asyncEnd` can touch that exact segment.
  */
 
 /**
@@ -43,12 +40,12 @@ class BaseCoreSubscriber {
   /**
    * @param {CoreSubscriberParams} params the subscriber constructor params
    */
-  constructor({ agent, logger, packageName, prefix, instrumentedMethods = [], hasCallback = false }) {
+  constructor({ agent, logger, packageName, prefix, instrumentedMethods = [] }) {
     this.agent = agent
     this.config = agent.config
     this.logger = logger.child({ component: `${packageName}-subscriber` })
     this.packageName = packageName
-    this.hasCallback = hasCallback
+    this.internal = false
     this.id = prefix ? `nr:${prefix}:${packageName}` : `nr:${packageName}`
     this.channels = this.buildChannels(instrumentedMethods)
     this.store = agent.tracer.store
@@ -79,6 +76,11 @@ class BaseCoreSubscriber {
     return names.map((name) => tracingChannel(`${this.id}:${name}`))
   }
 
+  shouldCreateSegment(parent) {
+    return !(parent?.opaque ||
+    (this.internal && this.packageName === parent?.shimId))
+  }
+
   /**
    * Creates a new segment, starts its timer, and returns an updated context
    * that has entered the new segment. Returns the original context unchanged
@@ -95,6 +97,12 @@ class BaseCoreSubscriber {
   createSegment({ name, recorder, ctx }) {
     const parent = ctx.segment
 
+    if (this.shouldCreateSegment(parent) === false) {
+      this.logger.trace('Skipping segment creation for %s, %s(parent) is of the same package: %s and incoming segment is marked as internal', name, parent?.name, this.packageName)
+      return ctx
+
+    }
+
     const segment = this.agent.tracer.createSegment({
       name,
       parent,
@@ -103,6 +111,7 @@ class BaseCoreSubscriber {
     })
 
     if (segment) {
+      segment.shimId = this.packageName
       segment.start()
       return ctx.enterSegment({ segment })
     }
@@ -148,12 +157,8 @@ class BaseCoreSubscriber {
    * @param {object} data Event data published to the `asyncEnd` event.
    */
   asyncEnd(data) {
-    if (this.hasCallback) {
-      data?.segment?.touch()
-      return
-    }
-
+    console.log('touching segment', ctx?.segment?.name)
     ctx?.segment?.touch()
   }
 
@@ -195,11 +200,7 @@ class BaseCoreSubscriber {
           return ctx
         }
 
-        const result = this.handler(data, ctx)
-        if (this.hasCallback) {
-          data.segment = result?.segment
-        }
-        return result
+        return this.handler(data, ctx)
       })
     }
   }
diff --git a/lib/subscribers/core/child_process/index.js b/lib/subscribers/core/child_process/index.js
index 86787d30b..05c08a4dc 100644
--- a/lib/subscribers/core/child_process/index.js
+++ b/lib/subscribers/core/child_process/index.js
@@ -14,7 +14,8 @@ const INSTRUMENTED_METHODS = ['exec', 'execFile']
 
 class ChildProcessSubscriber extends BaseCoreSubscriber {
-    super({ agent, logger, packageName: 'child_process', hasCallback: true, instrumentedMethods: INSTRUMENTED_METHODS })
+    super({ agent, logger, packageName: 'child_process', instrumentedMethods: INSTRUMENTED_METHODS })
+    this.internal = true
   }
 
   /**
diff --git a/test/integration/core/child-process.test.js b/test/integration/core/child-process.test.js
index dcf5f83f5..66596448b 100644
--- a/test/integration/core/child-process.test.js
+++ b/test/integration/core/child-process.test.js
@@ -35,7 +35,6 @@ test('exec', function (t, end) {
         agent,
         end,
         name: 'child_process.exec',
-        children: ['child_process.execFile'],
         assertCallbacks: false
       })
     })

```

There's no reason to trace 2 methods when only one is called because the other one is being called within the wrapping method. so when you call `exec` we only create one segment not 2 for `exec` and `execFile`

bizob2828 (changes requested):

lib/subscribers/core/child_process/index.js
amychisholm03: Okay, this makes sense. I was just trying to keep existing behavior with asserting `children: ['child_process.execFile']` in the tests

lib/subscribers/core/child_process/index.js
amychisholm03: I noticed `dns` still has it, so modified it too

bizob2828: welp, i missed removing that. i'll fix

bizob2828 (approved):
LGTM

PR #4216, opened 2026-08-10
fix: Fixed OTEL metrics wrapping meters too many times

### LINKED ISSUE #4213: OTel bridge: `nrGetMeter` re-wraps the cached Meter's `create*` methods on every `getMeter()` call, leaking an unbounded closure chain
## Summary

When `opentelemetry.enabled` is on, the agent patches `provider.getMeter` (in
`lib/otel/metrics/index.js`) to count bridge usage. The patch re-wraps the
returned meter's `create*` methods on **every** call, with no
already-wrapped guard:

```js
const meter = getMeter.apply(provider, args)   // SDK returns the SAME cached Meter per scope
const originals = {}
for (const method of methods) {
  originals[method] = meter[method]            // call 2+: this is the PREVIOUS call's wrapper
  meter[method] = function nrWrappedMethod(...args) {
    agent.metrics.getOrCreateMetric(`…/meter/${method}`).incrementCallCount()
    return originals[method].apply(meter, args)  // closure pins `originals`, which pins the previous wrapper
  }
}
```

Because `@opentelemetry/sdk-metrics` caches and returns the same `Meter`
object per instrumentation scope, from the second call onward
`originals[method]` captures the *previous call's wrapper* rather than the
SDK method. Each new wrapper's closure references its `originals`, so every
`getMeter()` call permanently appends one `{originals + 7 nrWrappedMethod
functions}` block to a chain whose head is a property of the process-lived
Meter. Nothing ever removes a layer. (A later `createCounter()` call would
also recurse through all N layers, incrementing the supportability metric N
times.)

Calling `getMeter()` repeatedly is explicitly sanctioned by the OTel API
("It is unspecified whether or under which conditions the same or different
`Meter` instances are returned" — callers are told acquisition is cheap and
idempotent), so per-use `getMeter()` is a legitimate pattern the bridge
turns into a memory leak.

## Reproduction

```js
// repro.js — run: NEW_RELIC_OPENTELEMETRY_ENABLED=true NEW_RELIC_ENABLED=true \
//   NEW_RELIC_APP_NAME=leak NEW_RELIC_LICENSE_KEY=0000000000000000000000000000000000000000 \
//   NEW_RELIC_NO_CONFIG_FILE=true node --expose-gc repro.js
require('newrelic')
const { metrics } = require('@opentelemetry/api')

setTimeout(() => {
  console.log('patched:', metrics.getMeterProvider().getMeter.name) // "nrGetMeter"
  global.gc()
  const h0 = process.memoryUsage().heapUsed
  for (let i = 0; i < 100_000; i++) metrics.getMeter('leak-demo')
  global.gc()
  const h1 = process.memoryUsage().heapUsed
  console.log(`retained after 100k getMeter calls: ${((h1 - h0) / 1e6).toFixed(1)} MB`)
  process.exit(0)
}, 3000)
```

Observed: **~87 MB retained after 100,000 calls (~873 bytes and ~9 heap
objects per call), surviving forced full GC**. The identical loop without
the agent (plain `@opentelemetry/sdk-metrics` `MeterProvider`) retains
0.0 MB.

## Real-world impact

A production GraphQL gateway acquired the meter once per Redis command
(instruments memoized; only the `getMeter()` lookup was per-call). At
~600 commands/min/pod this leaked ~28 MB/hour/pod (~750 MB/day) of
GC-unreclaimable heap, and the millions of retained closures drove
major-GC pause from ~20 ms on fresh pods to 400–600 ms within a day —
an aged-pod event-loop-lag incident that took several days to trace,
partly because the leak lives below the application in the agent.

## Environment

- `newrelic` 14.3.0 (code unchanged on `main` / 14.3.6 as of 2026-08-08)
- Node 26.6.0
- `@opentelemetry/api` 1.9.x, `@opentelemetry/sdk-metrics` 2.9.0

## Suggested fix

Wrap once per meter — e.g. guard with a `WeakSet<Meter>` before wrapping,
wrap the prototype once instead of the instance per call, or have
`nrGetMeter` cache and return already-wrapped meters. Any of these keeps
the supportability counts while making the patch idempotent.

## Workaround

Memoize the `getMeter()` result application-side and never call it on a hot
path while the bridge is enabled.

jsumners-nr:
This PR resolves #4213 and supersedes #4215.

### DIFF
--- a/lib/otel/metrics/index.js
+++ b/lib/otel/metrics/index.js
@@ -77,6 +77,7 @@ class SetupMetrics extends SetupSignal {
     this.#metricExporter = proxyExporter
     this.#metricReader = reader
 
+    const symDecorated = Symbol('otel.meter.decorated')
     const getMeter = provider.getMeter
     provider.getMeter = function nrGetMeter(...args) {
       agent.metrics
@@ -88,6 +89,7 @@ class SetupMetrics extends SetupSignal {
       const methods = Object.getOwnPropertyNames(proto).filter((name) => name.startsWith('create'))
       const originals = {}
       for (const method of methods) {
+        if (meter[method][symDecorated] === true) continue
         originals[method] = meter[method]
         // As of 2025-06-17:
         // + createGauge
@@ -103,6 +105,7 @@ class SetupMetrics extends SetupSignal {
             .incrementCallCount()
           return originals[method].apply(meter, args)
         }
+        meter[method][symDecorated] = true
       }
 
       return meter
--- a/test/unit/lib/otel/metrics/index.test.js
+++ b/test/unit/lib/otel/metrics/index.test.js
@@ -178,6 +178,42 @@ test('serverless mode does not wait for the started event', (t) => {
   t.assert.ok(bootstrapped, 'should emit otelMetricsBootstrapped from the constructor')
 })
 
+test('getMeter called multiple times for the same scope does not re-wrap create* methods', async (t) => {
+  const { agent } = t.nr
+  const createCounterCounts = {}
+  agent.metrics.getOrCreateMetric = function (name) {
+    createCounterCounts[name] = (createCounterCounts[name] ?? 0) + 1
+    return this
+  }
+
+  // eslint-disable-next-line no-new
+  new SetupMetrics({ agent })
+  const provider = otelApi.metrics.getMeterProvider()
+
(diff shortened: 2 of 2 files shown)


amychisholm03 (approved):

sonewman:
❤️

PR #4224, opened 2026-08-12
refactor: Refactored `dns` instrumentation to subscribe to events emitted

### LINKED ISSUE #4221: Migrate `dns` to emit events over tracing channel
### Description

### Acceptance Criteria
All methods we previously instrumented with shim on the `dns` API will be run over tracing channel and subscribers will create the necessary trace segments.

bizob2828:
## Description
This PR sets the groundwork for migrating core instrumentation to subscriber models.  Every subscriber must extend the BaseCoreSubscriber and implement the instrument method and use shimmer to patch.  We require shimmer because it allows easy unwrapping of methods when the agent unloads.  Also most core instrumentation is very simple. It just create a segment with a name and typically no metric recorder.  

Closes #4221


lib/subscriber-configs.js
```diff
@@ -17,6 +17,7 @@ const subscribers = {
   ...require('./subscribers/bunyan/config'),
   ...require('./subscribers/cassandra-driver/config'),
   ...require('./subscribers/connect/config'),
+  ...require('./subscribers/dns/config'),
```
jsumners-nr: We don't want?: `./subscribers/core/dns/config`?

bizob2828: i guess we _could_ because my next commit will have some sort of abstraction for core library instrumentation so i can live there as well

lib/subscribers/core/base.js
```diff
@@ -0,0 +1,256 @@
...
+  /**
+   * @param {CoreSubscriberParams} params the subscriber constructor params
+   */
+  constructor({ agent, logger, packageName, hasCallback = false, prefix, instrumentedMethods = [] }) {
+    this.agent = agent
+    this.config = agent.config
+    this.logger = logger.child({ component: `${packageName}-subscriber` })
+    this.packageName = packageName
+    this.hasCallback = hasCallback
+    this.id = prefix ? `nr:${prefix}:${packageName}` : `nr:${packageName}`
```
bizob2828, line 52: this is id to conform to the `lib/subscribers/base.js`. we should prob have a generic class that is extended here and the base subscriber so we can ensure we implement all the right methods that are needed to bootstrap and teardown subscribers

test/integration/core/dns-utils.js
```diff
@@ -0,0 +1,80 @@
...
+    setImmediate(() => {
+      cb(null, ['127.0.0.1'])
+    })
+  })
+  sandbox.stub(dns.Resolver.prototype, 'resolve').callsFake((_, cb) => {
+    setImmediate(() => {
+      cb(null, ['127.0.0.1'])
+    })
+  })
+  sandbox.stub(dns.promises, 'resolve').callsFake(async () => Promise.resolve(['127.0.0.1']))
```
bizob2828, line 34: i forgot to clean some of this up but it will be in use when we do #4264

lib/subscribers/core/base.js
```diff
@@ -0,0 +1,256 @@
...
+   * @param {object} data Event data published to the tracing channel's `start` event.
+   * @param {string} data.name Name to use for the created segment.
+   * @param {object} ctx The current async context.
+   * @returns {object} The updated context returned by {@link #createSegment}.
+   */
+  handler(data, ctx) {
+    return this.createSegment({ name: data.name, ctx })
+  }
+
+  /**
```
amychisholm03, line 129: I feel like these default `end`/`asyncEnd`/`handleCallback` definitions will work for most core subscribers, but may touch the wrong segment if a module's instrumented methods call each other. This is the case with `child_process.exec` (it calls `execFile` internally).

If the child subscriber has to override these methods, we'll have to make sure their description is updated to avoid confusion.

bizob2828: Hmm I'd make the case that `exec` should have opaque = true, and maybe we build that in.

amychisholm03: Yeah I can explore that in my PR. And I guess there may be cases where we have to override `handleCallback`/`end`/`asyncEnd` regardless, but this is a fine default.

lib/subscribers/core/base.js
```diff
@@ -0,0 +1,256 @@
...
+    this.id = prefix ? `nr:${prefix}:${packageName}` : `nr:${packageName}`
+    this.channels = this.buildChannels(instrumentedMethods)
+    this.store = agent.tracer.store
+    this.handlers = {
+      asyncEnd: this.asyncEnd.bind(this)
+    }
+
+    if (hasCallback) {
+      this.handlers.end = this.end.bind(this)
+    }
```
amychisholm03, line 61:
```suggestion
    this.handlers = {
      asyncEnd: this.asyncEnd.bind(this),
      end: this.end.bind(this)
    }
```

Shouldn't `end` always be bound? For `traceSync` calls, `asyncEnd` never fires, so `end` is the only thing that closes out the segment's timing.

lib/subscribers/core/base.js
```diff
@@ -0,0 +1,256 @@
...
+'use strict'
+// eslint-disable-next-line n/no-unsupported-features/node-builtins
+const { tracingChannel } = require('node:diagnostics_channel')
+
+/**
+ * The baseline parameters required to construct a core subscriber.
+ *
+ * @typedef {object} CoreSubscriberParams
+ * @property {object} agent A New Relic Node.js agent instance.
+ * @property {object} logger An agent logger instance.
```
jsumners-nr:
```suggestion
 * @property {Agent} agent A New Relic Node.js agent instance.
 * @property {AgentLogger} logger An agent logger instance.
```

lib/subscribers/core/base.js
```diff
@@ -0,0 +1,256 @@
...
+ * `'nr:'` and `packageName` when building the channel id, e.g. `'promises'`
+ * produces `'nr:promises:dns'`. Omit to use the short form `'nr:dns'`.
+ * @property {string[]} [instrumentedMethods=[]] Names of the module methods
+ * to instrument. One tracing channel is created per entry.
+ */
+
+/**
+ * @property {object} agent A New Relic Node.js agent instance.
+ * @property {object} config The agent configuration object.
+ * @property {object} logger A child logger scoped to this subscriber.
```
jsumners-nr:
```suggestion
 * @property {Agent} agent A New Relic Node.js agent instance.
 * @property {AgentConfig} config The agent configuration object.
 * @property {AgentLogger} logger A child logger scoped to this subscriber.
```

lib/subscribers/core/base.js
```diff
@@ -0,0 +1,256 @@
...
+   * Any error thrown by {@link #instrument} is caught and logged as a warning
+   * rather than propagated to the caller.
+   *
+   */
+  enable() {
+    try {
+      this.instrument(require(this.packageName))
+    } catch (error) {
+      this.logger.warn(error, 'Failed to instrument %s', this.packageName)
+    }
```
jsumners-nr, line 172: Is this missing a return?
```suggestion
    } catch (error) {
      this.logger.warn(error, 'Failed to instrument %s', this.packageName)
      return
    }
```

jsumners-nr (commented):
Looks good to me.

lib/subscribers/core/base.js
bizob2828: yea, let me add a test

jsumners-nr (approved):

PR #4227, opened 2026-08-18
refactor: Co-located Attributes and PrioritizedAttributes under lib/attributes/

jsumners-nr:
## Summary

Stacks on #4207.

- Moves `lib/attributes.js` → `lib/attributes/index.js` and `lib/prioritized-attributes.js` → `lib/attributes/prioritized.js`
- `PrioritizedAttributes` now extends `Attributes`, removing duplicated `isValidLength`, `get`, `has`, `reset`, `addAttributes`, `hasValidDestination`, and `makeFilter`
- Map-based storage replaced with `Object.create(null)` to match the base class; `addAttribute` and `_getDroppableAttributeKey` rewritten to use plain-object operations
- `attributeCount` (already tracked by the base class) replaces `Object.keys().length` for capacity checks, avoiding a per-call array allocation
- Both classes now accept a `logger` instance via constructor options (defaulting to the module-level child logger) and store it as a private `#logger` field
- All `require` calls in `lib/attributes/` use explicit file extensions to short-circuit Node's file resolution algorithm
- Unit tests moved to `test/unit/lib/attributes/`
- Benchmarks moved to `test/benchmark/lib/attributes/`, including a new `realistic-usage` case for the prioritized suite

**Benchmarked** against the tip of #4207 (mean of 100k runs):

The `Attributes` (index) suite is neutral to slightly faster (~1–7%) across all methods, within noise.

`PrioritizedAttributes` — plain-object + `attributeCount` is faster than the original Map-based implementation:

| test | before (µs) | after (µs) | delta |
|------|-------------|------------|-------|
| add past maximum, all high priority | 0.48 | 0.21 | -56% |
| add past maximum, all low priority | 0.44 | 0.28 | -37% |
| add past maximum, mixed priority | 0.47 | 0.27 | -43% |

## Test plan

- `node --test test/unit/lib/attributes/index.test.js` — 23 pass
- `node --test test/unit/lib/attributes/prioritized.test.js` — 33 pass
- `npm run lint` — 0 errors


test/unit/lib/attributes/index.test.js
```diff
@@ -8,9 +8,9 @@
 const test = require('node:test')
 const assert = require('node:assert')
 
-const helper = require('../lib/agent_helper')
-const { Attributes } = require('../../lib/attributes')
-const AttributeFilter = require('../../lib/config/attribute-filter')
+const helper = require('../../../lib/agent_helper')
+const { Attributes } = require('../../../../lib/attributes/index')
+const AttributeFilter = require('../../../../lib/config/attribute-filter')
```
amychisholm03:
```suggestion
const helper = require('#testlib/agent_helper.js')
const { Attributes } = require('#agentlib/attributes/index.js')
const AttributeFilter = require('#agentlib/config/attribute-filter.js')
```

test/unit/lib/attributes/prioritized.test.js
```diff
@@ -8,10 +8,10 @@
 const test = require('node:test')
 const assert = require('node:assert')
 
-const helper = require('../lib/agent_helper')
+const helper = require('../../../lib/agent_helper')
 
-const { PrioritizedAttributes, ATTRIBUTE_PRIORITY } = require('../../lib/prioritized-attributes')
-const AttributeFilter = require('../../lib/config/attribute-filter')
+const { PrioritizedAttributes, ATTRIBUTE_PRIORITY } = require('../../../../lib/attributes/prioritized')
+const AttributeFilter = require('../../../../lib/config/attribute-filter')
```
amychisholm03:
```suggestion
const { PrioritizedAttributes, ATTRIBUTE_PRIORITY } = require('#agen tlib/attributes/prioritized.js')
const AttributeFilter = require('#agentlib/config/attribute-filter.js')
```

amychisholm03 (changes requested):
We should try to use the import aliases, especially for 3+ `..`s, for consistency across the attribute tests.

amychisholm03 (approved):

Issue #4226, opened 2026-08-13
Update our default configuration `lib/config/default.js` to be a json schema


bizob2828:
### Description
We've hit limits with `lib/config/default.js` and we've had requests from fleet control(#4161) to publish a json schema that can be consumed for user controlled configuration UIs.  To eliminate extra processing in #4161, we would benefit from defining our default, user controlled configuration in a json schema.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-605793

bizob2828:
I will file another ticket but once we land the json schema we should also automate our docs website configuration to generate this as a post release step like we do for our compatiblity

amychisholm03:
Whether we truly need all of the tests in `test/unit/config/config-defaults.test.js` and `config-env.test.js` should also be investigated.

PR #4228, opened 2026-08-18
fix: Updated `openai` instrumentation to support `>=7.5.0`

jsumners-nr:
## Description

openai@7.5.0 introduced a `responsePromise()` method on the `OpenAI` client that overrides `_thenUnwrap` on each returned `APIPromise` **instance** with an arrow function that closes over the `parse` parameter. This shadows the prototype method that the existing `>=5.0.0` hook instruments via `nr_thenUnwrap`, so the channel never fires for calls going through the instance override.

The double-consumption bug (`response.json()` called twice → `"Body is unusable"`) still applies: NR's `asyncEnd` on `nr_responses` triggers `parse()` on the original promise while `responses.js` calls `_thenUnwrap()`, which creates a new promise chain that also calls `parse()` via the closed-over variable.

Since this is the 3rd different iteration, this PR instruments the `defaultParseResponse` which calls `response.text` or `response.json` on the undici instance.  We now wrap those and cache the first time they are called, thus greatly simplifying this intrumentation across versions.

## How to Test

```sh
npm run versioned openai
```


amychisholm03:
I think there's a typo with the GH issue this PR is closing. I don't think it should be closing https://github.com/newrelic/node-newrelic/issues/4226

jsumners-nr (commented):
Approved.

jsumners-nr (commented):
Nice.

bizob2828 (approved):

PR #4231, opened 2026-08-18
refactor: Removed dead `wrapNoSegment` code from `zlib` instrumentation

jsumners-nr:
## Summary

Removes dead code from the `zlib` core-module instrumentation: the `wrapNoSegment` function and the `zlib.Deflate` prototype-wrapping block.

That code manually rebound zlib stream callbacks (`_transform`, or legacy `write`/`flush`/`end`) to the active segment. It was necessary in the continuation-local-storage (CLS) era, when callbacks crossing zlib's internal C++ boundary lost transaction context. The agent now uses only `AsyncLocalContextManager`, which propagates context across the native boundary automatically, so the manual rebind is redundant.

The live `shim.record(...)` path that produces the `zlib.*` segments/metrics is unchanged.

## Verification

Confirmed empirically that toggling the callback binding on/off produces **byte-identical** behavior across Node **22, 24, and 26** (the full supported `engines: >=22` range), for:

- transaction context in stream `data` / `end` / `flush` / `setImmediate` handlers
- segment parenting of children created inside stream callbacks
- an adversarial promise-based `gzip` → `gunzip` `stream/promises.pipeline`

The legacy `write`/`flush`/`end` branch was also dead, since modern zlib streams always expose `_transform`.

## Tests

Adds three integration tests as regression guards:

- **error path** — a recorded method ends its segment and binds the callback even when the operation errors
- **stream context propagation** — the transaction survives into a zlib stream's own async event handlers
- **stream outside a transaction** — zlib streams work normally with no active transaction

All 17 tests in `test/integration/core/zlib.test.js` pass on Node 22/24/26. `eslint` is clean.

### DIFF
--- a/lib/instrumentation/core/zlib.js
+++ b/lib/instrumentation/core/zlib.js
@@ -15,32 +15,7 @@ const methods = ['deflate', 'deflateRaw', 'gzip', 'gunzip', 'inflate', 'inflateR
 function initialize(agent, zlib, moduleName, shim) {
   shim.record(zlib, methods, recordZLib)
 
-  if (zlib.Deflate && zlib.Deflate.prototype) {
-    const proto = Object.getPrototypeOf(zlib.Deflate.prototype)
-    if (proto._transform) {
-      // streams2
-      shim.wrap(proto, '_transform', wrapNoSegment)
-    } else if (proto.write && proto.flush && proto.end) {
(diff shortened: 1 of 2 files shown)


bizob2828 (approved):
this is def an artifact of our old event loop based context manager. LGTM

PR #4232, opened 2026-08-18
refactor: Include all `.mjs` equivalents in `openai` instrumentation to properly function when using ESM

bizob2828:
## Description

When looking into #4228, I noticed we aren't wrapping the `.mjs` files for openai.  This would lead to a loss of instrumentation when using ESM. This PR updates the openai config to wrap both `.js` and `.mjs`. I also fixed the version ranges for `core.js/mjs` as it only exists in <5.0.0.

### DIFF
--- a/lib/subscribers/openai/config.js
+++ b/lib/subscribers/openai/config.js
@@ -21,7 +21,25 @@ module.exports = {
         },
         {
           channelName: 'nr_completionsCreate',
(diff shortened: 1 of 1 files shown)


jsumners-nr (approved):
How we are not encountering the same startup slowness we saw with mongodb when having so many patches is a mystery, but LGTM.

bizob2828:
> How we are not encountering the same startup slowness we saw with mongodb when having so many patches is a mystery, but LGTM.

There were significantly more for mongo originally

PR #4233, opened 2026-08-18
chore: Refactored `zlib` instrumentation to be subscriber based

### LINKED ISSUE #4222: Migrate `zlib` to emit events over tracing channel
### Description

### Acceptance Criteria
All methods we previously instrumented with shim on the `zlib` API will be run over tracing channel and subscribers will create the necessary trace segments.

jsumners-nr:
This PR resolves #4222.


lib/context-manager/async-local-context-manager.js
```diff
@@ -24,6 +24,16 @@ class AsyncLocalContextManager {
...
 
+  /**
+   * The underlying local storage object.
+   *
+   * @see https://nodejs.org/api/async_context.html#class-asynclocalstorage
+   * @returns {AsyncLocalStorage}
+   */
+  get store() {
+    return this._asyncLocalStorage
+  }
```
jsumners-nr, line 35: I get that accessing `_` prefixed fields within our code base is acceptable and the actual intention of those fields. But I think, even within our own code base, it is better to expose things via public fields. It gives users confidence that they are doing the right thing instead of the anxiety induced by utilizing conventionally private fields.

lib/transaction/tracer/index.js
```diff
@@ -47,6 +47,20 @@ Tracer.prototype.wrapFunctionFirst = wrapFunctionFirst
...
+   *
+   * @see https://nodejs.org/api/async_context.html#class-asynclocalstorage
+   * @returns {AsyncLocalStorage}
+   */
+  asyncStore: {
+    get () {
+      return this._contextManager.store
+    }
+  }
+})
```
jsumners-nr, line 62: Just bubbling up the underlying getter to the interface we typically access the storage object from.

lib/subscribers/subscriber.js
jsumners-nr, line 1: I'm not sure if we really need this. This comes from the work I was experimenting with around `@platformatic/kafka` in October of 2025. My memory of that work is hazy at best. I think my idea was to give a base to inherit from that doesn't come with any implementation detail baggage, and then refactor the existing ones atop it (e.g. `dc-base` and `db`).

lib/subscribers/tracing-channel-subscriber.js
jsumners-nr, line 1: This also comes from my experimentation with `@platformatic/kafka`, but it has been tweaked to actually do stuff. This smaller task helped me narrow in on how these classes should work. In short:

1. `TracingChannelSubscription` describes the event handlers an instrumentation will require. By defining only the event methods the instrumentation needs, we don't need to update something like `this.events`.
2. `TracingChannelSubscriber` utilizes one, or many, instances of `TracingChannelSubscription` to register event listeners and manage segment life cycles.

lib/subscribers/wrap-method.js
jsumners-nr, line 1: I ported the shimmer wrap methods to these simple utilities. It's likely that I missed some obscure detail of the shimmer implementation.

jsumners-nr (commented):
At this time, I have not explored uncovered variations of the `zlib` API. I am opening the PR in its most basic state so that we can discuss the ideas in it before finalizing our approach. Once we do that, I will revisit updating API coverage.

Issue #4230, opened 2026-08-18
[spike] Add support for inference profile model IDs in AWS Bedrock


amychisholm03:
### Description

AWS Bedrock's **Application Inference Profiles** let customers invoke a model via a customer-created ARN (`arn:aws:bedrock:<region>:<account-id>:application-inference-profile/<id>`) instead of a direct foundation-model ID or system-defined cross-region profile ID, typically for cost allocation/tagging. When a customer passes one of these ARNs as `modelId` to `InvokeModelCommand`/`ConverseCommand`, the Node.js agent's Bedrock instrumentation silently fails to capture request/response content, and reports the model as an unreadable ARN instead of a resolved model name.

This is included: fixing vendor/model detection in `lib/llm-events/aws-bedrock/bedrock-command.js` and `bedrock-response.js` so it works for application inference profile ARNs, for both `InvokeModel` and `Converse` code paths.

This is *not* included: changes to how cross-region system-defined inference profiles are detected (e.g. `us.anthropic.claude-3-haiku-...`) — those already work correctly today via the existing `modelId.split('.').slice(-2)` matching, since the vendor/model is embedded in the string.

**Root cause:** `BedrockCommand`'s vendor-detection helpers (`isClaude`, `isClaude3`, `isTitan`, `isCohere`, `isLlama` — `lib/llm-events/aws-bedrock/bedrock-command.js`) do prefix matching against `modelId`, assuming it always looks like `<vendor>.<model>` or `<region>.<vendor>.<model>`. An application inference profile ARN has no `.` characters at all, so every one of these checks returns `false`. That cascades into:
- `BedrockCommand.prompt` returning `[]` (no vendor branch matches) — request content lost.
- `BedrockResponse#extractCompletionsAndId` never populating `#completions` (same reason) — response content lost.
- Zero `LlmChatCompletionMessage` events emitted per call (not just blank content — there's nothing to iterate, so the events themselves never get created).
- `request.model` / `response.model` reported as the raw ARN rather than a resolved model name.

Token usage (`response.usage.prompt_tokens`/`completion_tokens`/`total_tokens`) is *not* affected — it's read from `x-amzn-bedrock-*-token-count` response headers (InvokeModel) or `response.output.usage` (Converse), independent of vendor detection. This is why affected customers see a model entry with token/cost data but no completions content, which is the exact discrepancy reported.

[...]

PR #4235, opened 2026-08-19
refactor: Split `BedrockResponse` and `BedrockCommand` into `Converse` and `InvokeModel` variants

amychisholm03:
## Description

While I was triaging a customer's reported bug with AWS Bedrock, I noticed that the `BedrockResponse` and `BedrockCommand` classes have gotten a bit complex and hard to reason about. So I split `BedrockResponse` into `InvokeModelResponse` and `ConverseResponse` and likewise with `BedrockCommand`. We can deduce which API it is (`InvokeModel` vs. `Converse`) when we wrap the Bedrock middleware because we have access to the initial command name there.

## How to Test

```
node --test test/unit/llm-events/aws-bedrock/*.test.js test/unit/subscribers/aws-sdk/*.test.js
npm run versioned aws-sdk-v3
```

## Related Issues

Discovered while researching https://github.com/newrelic/node-newrelic/issues/4230

### DIFF
--- a/lib/llm-events/aws-bedrock/invoke-model-command.js
+++ b/lib/llm-events/aws-bedrock/invoke-model-command.js
@@ -8,12 +8,13 @@
 const { stringifyClaudeChunkedMessage } = require('./stringify-message')
 
 /**
- * Parses an AWS Bedrock InvokeModel/InvokeModelWithResponseStream command
+ * Parses an AWS Bedrock `InvokeModel`/`InvokeModelWithResponseStream` command
  * instance into a re-usable entity.
  */
 class InvokeModelCommand {
   #input
   #modelId
+  /** The parsed input body. */
   #body
 
   /**
@@ -68,7 +69,7 @@ class InvokeModelCommand {
    * @returns {string} One of `embedding` or `completion`.
    */
   get modelType() {
-    // TODO: this.#input.body.inputText enough to determine if embedding?
+    /// This logic is common to both command types
     return this.#modelId.toLowerCase().includes('embed') ? 'embedding' : 'completion'
   }
 
--- a/lib/llm-events/aws-bedrock/invoke-model-command.js
+++ b/lib/llm-events/aws-bedrock/invoke-model-command.js
@@ -6,15 +6,14 @@
 'use strict'
 
 const { stringifyClaudeChunkedMessage } = require('./stringify-message')
+const ModelCommand = require('./model-command')
 
 /**
  * Parses an AWS Bedrock `InvokeModel`/`InvokeModelWithResponseStream` command
  * instance into a re-usable entity.
  */
-class InvokeModelCommand {
+class InvokeModelCommand extends ModelCommand {
   #input
-  #modelId
-  /** The parsed input body. */
   #body
 
   /**
@@ -22,16 +21,9 @@ class InvokeModelCommand {
    * or `InvokeModelWithResponseStreamCommand` instance.
    */
   constructor(input) {
+    super(input)
     this.#input = input
-    this.#modelId = this.#input.modelId?.toLowerCase() ?? ''
     this.#body = JSON.parse(this.#input.body)
-  }
-
-  /**
-   * @returns {boolean} Always `false`, this command is not from the Converse API.
-   */
-  get isConverse() {
-    return false
   }
 
   /**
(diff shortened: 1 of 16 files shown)


lib/llm-events/aws-bedrock/invoke-model-command.js
```diff
@@ -1,47 +1,37 @@
...
- * Parses an AWS Bedrock command instance into a re-usable entity,
- * unifying logic for both InvokeModel and Converse commands.
+ * Parses an AWS Bedrock `InvokeModel`/`InvokeModelWithResponseStream` command
+ * instance into a re-usable entity.
  */
-class BedrockCommand {
+class InvokeModelCommand {
   #input
   #modelId
+  /** The parsed input body. */
```
jsumners-nr: This comment isn't fully formed JSDoc and really isn't providing any utility.

lib/llm-events/aws-bedrock/invoke-model-command.js
```diff
@@ -1,47 +1,37 @@
...
-      this.#messages = input.messages
-      this.#isConverseCommand = true
-    }
+    this.#body = JSON.parse(this.#input.body)
   }
 
   /**
-   *
-   * @returns {boolean} True if the command is from the Converse API.
+   * @returns {boolean} Always `false`, this command is not from the Converse API.
```
jsumners-nr: So why does this method need to be present?

lib/llm-events/aws-bedrock/invoke-model-command.js
```diff
@@ -85,7 +69,7 @@ class BedrockCommand {
    * @returns {string} One of `embedding` or `completion`.
    */
   get modelType() {
-    // This logic is common to both command types
+    /// This logic is common to both command types
```
jsumners-nr: ?

lib/llm-events/aws-bedrock/invoke-model-command.js
```diff
@@ -85,7 +69,7 @@ class BedrockCommand {
...
     return this.#modelId.toLowerCase().includes('embed') ? 'embedding' : 'completion'
```
jsumners-nr: Presumably I originally wrote this line. We can improve it by returning `InvokeModelCommand.MODEL_TYPE_EMBEDDING` or `MODEL_TYPE_COMPLETION`. This is more beneficial now that multiple objects are implementing the same method and returning the same constants. The root object should have the above statics defined and referenced from the implementations.

lib/subscribers/aws-sdk/middleware/bedrock/utils.js
```diff
@@ -172,8 +173,8 @@ function recordChatCompletionMessages({
  * @param {Logger} params.logger Logger instance
  * @param {TraceSegment} params.segment Current segment
  * @param {Transaction} params.transaction Current transaction
- * @param {BedrockCommand} params.bedrockCommand BedrockCommand object
- * @param {BedrockResponse} params.bedrockResponse BedrockResponse object
+ * @param {InvokeModelCommand|ConverseCommand} params.bedrockCommand BedrockCommand object
+ * @param {InvokeModelResponse|ConverseResponse} params.bedrockResponse BedrockResponse object
```
jsumners-nr: If we need to split up the original implementation to support AWS's current architecture, that is okay. But the split should not make the usage of the items more complicated. Let's pretend AWS adds another variation for doing this stuff, then we'd have to add a third entity in all places that accept these things.

What we should do is define a base object that the data handlers can use:

```js
class ModelCommand {}

class InvokeModelCommand extends ModelCommand {}

class ModelResponse {}

class InvokeModelResponse extends ModelResponse {}
```

Thus:

```
* @param {ModelCommand} params.command
```

lib/llm-events/aws-bedrock/invoke-model-command.js
amychisholm03: Yeah, I thought it was JSDoc because the comment shows up in my IDE when I hover over the field. I'll change/remove the comment.

lib/llm-events/aws-bedrock/invoke-model-command.js
amychisholm03: It's a quick way to determine InvokeModelCommand vs ConverseCommand, but we should just define `toString` and compare that instead.

jsumners-nr (approved):
Looks good to me.

PR #4245, opened 2026-08-24
refactor: Update Claude detection logic in `InvokeModelCommand`

amychisholm03:
## Description

Follow up to PR #4235. Our current Claude detection logic in `InvokeModelCommand` lives across 4 separate booleans and there are no test cases for v4 or v5 Claude models. This PR simplifies the Claude detection logic to be just two booleans; they detect the Claude API called rather than relying on the model version in the model ID. This PR also adds tests for the Claude v4/v5 models via making the tests not relying on the model ID version number, but instead on the API response shape.

* `isClaude` and `isClaudeTextCompletionApi` -> `isClaudePromptApi` (matches v1/v2)
* `isClaude3` and `isClaudeMessagesApi` -> `isClaudeMessagesApi` (matches v3+)

## How to Test

```
node --test test/unit/llm-events/aws-bedrock/invoke-model-command.test.js
node --test test/unit/llm-events/aws-bedrock/invoke-model-response.test.js
node --test test/unit/subscribers/aws-sdk/stream-handler.test.js

npm run versioned aws-sdk-v3
```

## Related Issues

Discovered while researching #4230

### DIFF
--- a/ai-support.json
+++ b/ai-support.json
@@ -98,7 +98,7 @@
     "footnote": "Note: if a model supports streaming, we also instrument the streaming variant.",
     "models": [
       {
-        "name": "Anthropic Claude",
+        "name": "Anthropic Claude (model versions 1-5)",
         "features": [
           {
             "title": "Text",
@@ -128,7 +128,7 @@
         ]
       },
       {
-        "name": "Meta Llama3",
+        "name": "Meta Llama (all versions)",
         "features": [
           {
             "title": "Text",
--- a/ai-support.json
+++ b/ai-support.json
@@ -98,7 +98,7 @@
     "footnote": "Note: if a model supports streaming, we also instrument the streaming variant.",
     "models": [
       {
-        "name": "Anthropic Claude (model versions 1-5)",
+        "name": "Anthropic Claude",
         "features": [
           {
             "title": "Text",
@@ -128,7 +128,7 @@
         ]
       },
       {
-        "name": "Meta Llama (all versions)",
+        "name": "Meta Llama3",
         "features": [
           {
             "title": "Text",
--- a/lib/llm-events/aws-bedrock/invoke-model-command.js
+++ b/lib/llm-events/aws-bedrock/invoke-model-command.js
@@ -61,14 +61,14 @@ class InvokeModelCommand extends ModelCommand {
     }
 
     if (
-      this.isClaudePromptApi(this.#body) ||
+      this.isClaudePromptApi() ||
       this.isCohere() ||
       this.isLlama()
     ) {
       return [{ role: 'user', content: this.#body.prompt }]
     }
 
-    if (this.isClaudeMessagesApi(this.#body)) {
+    if (this.isClaudeMessagesApi()) {
       return normalizeClaudeMessages(this.#body?.messages)
     }
 
--- a/lib/llm-events/aws-bedrock/invoke-model-command.js
+++ b/lib/llm-events/aws-bedrock/invoke-model-command.js
@@ -32,10 +32,10 @@ class InvokeModelCommand extends ModelCommand {
    * @returns {number|undefined}
    */
   get maxTokens() {
-    if (this.isClaude() === true) {
+    if (this.isClaudePromptApi() === true) {
       return this.#body.max_tokens_to_sample
     }
-    if (this.isClaude3() === true || this.isCohere() === true) {
+    if (this.isClaudeMessagesApi() === true || this.isCohere() === true) {
       return this.#body.max_tokens
     }
     if (this.isLlama() === true) {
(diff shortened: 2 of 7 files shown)


ai-support.json
```diff
@@ -98,7 +98,7 @@
     "footnote": "Note: if a model supports streaming, we also instrument the streaming variant.",
     "models": [
       {
-        "name": "Anthropic Claude",
+        "name": "Anthropic Claude (model versions 1-5)",
```
jsumners-nr: I'm not clear why we need to indicate the model versions if we are now able to differentiate the payloads by a common structure instead of the model that generated it.

lib/llm-events/aws-bedrock/invoke-model-command.js
```diff
@@ -93,12 +93,27 @@ class InvokeModelCommand extends ModelCommand {
...
-  isClaude() {
-    return this.modelId.split('.').slice(-2).join('.').startsWith('anthropic.claude-v')
+  /**
+   * Detects if the command used the Claude Text Completions API (v1/v2)
+   * via the presence of the `prompt` field on the body.
+   * @returns {boolean} True if the command used a Claude model and the Claude Text Completions API.
+   */
+  isClaudePromptApi() {
+    const matchesModelId = this.modelId.split('.').slice(-2).join('.').startsWith('anthropic.claude-v')
+    const matchesBodyShape = !!this.#body.anthropic_version && 'prompt' in this.#body
```
jsumners-nr: Please don't use `!!`. It's very difficult to read. It's much easier to reason through `foo != null`.

I also recommend `Object.hasOwn` over `in`. The `in` operator will evaluate all the way up the prototype chain, where as `hasOwn` checks for existence on the actual object instance.

lib/llm-events/aws-bedrock/invoke-model-command.js
```diff
@@ -93,12 +93,27 @@ class InvokeModelCommand extends ModelCommand {
...
-    return this.modelId.split('.').slice(-2).join('.').startsWith('anthropic.claude-3')
+  /**
+   * Detects if the command used the Claude Messages API (v3+)
+   * via the presence of the `messages` field on the body.
+   * @returns {boolean} True if the command used a Claude model and the Claude Messages API.
+   */
+  isClaudeMessagesApi() {
+    const strippedModelId = this.modelId.split('.').slice(-2).join('.')
+    const matchesModelId = strippedModelId.startsWith('anthropic.claude-') && !strippedModelId.startsWith('anthropic.claude-v')
+    const matchesBodyShape = !!this.#body.anthropic_version && 'messages' in this.#body
```
jsumners-nr: Same note as above.

test/unit/llm-events/aws-bedrock/invoke-model-command.test.js
```diff
@@ -34,30 +31,7 @@ const claude35 = {
...
-  }
-}
-
-const regionClaude3 = {
-  modelId: 'us.anthropic.claude-3-haiku-20240307-v1:0',
-  body: {
-    messages: [{ role: 'user', content: 'who are you' }]
-  }
-}
+const regionClaudeMsgsApi = { ...claudeMsgsApi, modelId: `us.${claudeMsgsApi.modelId}` }
```
jsumners-nr, line 34: I don't think these objects are mutated by anything, so we should be safe with this kind of object cloning. If I'm wrong, though, we should utilize `structuredClone()`.

ai-support.json
amychisholm03: My thought process was that, if for some reason, Claude released another API with v6 for example, we wouldn't automatically state that we support all versions.

amychisholm03: Though, I guess I could just state "Text Completions and Messages API".

jsumners-nr: But with this note we will have to keep track of what is released and update our doc. I think the release dates and git history will cover us sufficiently.

test/unit/llm-events/aws-bedrock/invoke-model-command.test.js
amychisholm03: Yeah, they're not being used anywhere else but this file.

ai-support.json
amychisholm03: Yeah, for us, git history would be enough but my concern is customers reading the coordinating docs website and trying to glean what exactly we support within AWS Bedrock <> Claude.

jsumners-nr: I understand. Both approaches suffer the same problem, but the current approach has the benefit of reducing our maintenance burden. If anything, I'd adjust the preamble with some language like "As of August 2026, all Claude models provided by Bedrock are supported."

We'd still have to maintain that language, but it gives us far more leeway than what is currently proposed.

amychisholm03: Yeah, I'll just leave it as-is for now.

jsumners-nr (approved):

PR #4246, opened 2026-08-26
refactor: Refactored the Transaction class and affected other entities

jsumners-nr:
😮‍💨

The `Transaction` class isn't going anywhere. So I decided to take an opportunity and refactor it into modern syntax, privatize the methods that should be private, remove the methods that are no longer needed, and overall trim it up. There were several far reaching tentacles that I ended up dealing with, hence the large number of affected files. Most of those changed files are test suites being updated to match the current code.

I know this is a lot to sort through. People tell me the bot can help, but 🤷‍♂️. Personally, I'd look at individual commits to understand the changes. I tried to be methodical and build this up in sensible chunks.

I think you'll find that there is much less usage of mystery literals.

-----

## What

Refactors the `Transaction` class into an ES class and reshapes its surrounding code: distributed-trace payloads and the transaction/transport enumerations are codified into dedicated classes, the internal payload methods are eliminated or made private, and now-dead public members are removed. Test coverage follows the code.

## Changes

**Transaction becomes an ES class**
- `lib/transaction/index.js` is converted from a prototype-based constructor to `class Transaction`.
- Internal-only helpers are true `#private` methods: `#cleanUnneededReferences`, `#copyNameToActiveSpan`, `#linkExceptionToSegment`, `#partialNameFromUri`, `#runUserNamingRules`.
- The one-line `createPartialTrace` wrapper is removed; callers construct `new PartialTrace(...)` directly.

**Distributed-trace payloads codified into classes** (`lib/transaction/distributed-trace/`)
- `payload.js` — `Payload` / `PayloadData` model the New Relic DT payload structure and enforce the spec (version validation, required data keys, short/long accessor names).
- `incoming-payload.js` — `IncomingPayload#parseAndApply` parses an inbound header (JSON or base64) via `Payload` and applies the data to the transaction.
- `outgoing-payload.js` — `OutgoingPayload` (and its stub) build/serialize the outbound payload via `Payload`.

**Internal payload methods eliminated / privatized**
- `Transaction#_acceptDistributedTracePayload` is removed; `acceptDistributedTraceHeaders` constructs `IncomingPayload` inline.
- `_createDistributedTracePayload` becomes the private `#createDistributedTracePayload`, driven by `insertDistributedTraceHeaders`.
- A test helper, `#testlib/create-dt-payload.js`, builds an outbound payload directly via `OutgoingPayload` and applies it via `IncomingPayload`, so tests no longer depend on internal transaction methods.

**Transaction and transport types codified into classes**
- `lib/transaction/type.js` — a `Type` class statically exporting `WEB`/`BG`/`MESSAGE` plus `values()`/`entries()`/`isValid()`. `Transaction.TYPES` points at it; `transaction-shim.js` uses `Type.isValid()`/`Type.entries()`.
- `lib/transaction/distributed-trace/transport.js` — a `Transport` class in the same shape. It is the source of truth for transport values: `MessageBrokerDescription` derives its `TRANSPORT_TYPE_*` constants from it, `message-shim` uses it directly, and the accept path validates with `Transport.isValid()`.

**Dead public members removed**
- `Transaction.TRANSPORT_TYPES`, `TransactionShim.TRANSPORT_TYPES`, and `TransactionShim.TRANSACTION_TYPES` had no remaining readers and are removed. The per-type static members (`TransactionShim.WEB`/`BG`/`MESSAGE`), still used by instrumentation, are retained.

**Tests & benchmark**
- New unit suites co-located with the code: `type.test.js`, `distributed-trace/transport.test.js`, `distributed-trace/payload.test.js`, `distributed-trace/incoming-payload.test.js`, `distributed-trace/outgoing-payload.test.js`.
- Round-trip DT setup across the affected suites now goes through the shared `#testlib/create-dt-payload.js` helper; create-path assertions are driven through `insertDistributedTraceHeaders`.
- Adds `test/benchmark/transaction/dt-accept.bench.js`, a high-iteration benchmark (10k accepts per sample) for the DT accept path.

## Performance

The class-based DT accept path is faster than the previous `.bind(this)`-per-call helper approach, avoiding repeated bound-function allocation per accept. Measured against `main` on the same machine:

- Inner accept method (parse + apply, in isolation): ~700 ns vs ~1250 ns — roughly 1.8× faster.
- Full public entry point (`acceptDistributedTraceHeaders`, incl. header dispatch + base64 decode): ~950 ns vs ~1350 ns — roughly 1.4× faster.

The current branch is also markedly more stable run-to-run than `main`.

## Testing

- `node --test test/unit/transaction.test.js` → 128/128
- `node --test test/unit/transaction/type.test.js` and `distributed-trace/*.test.js` → passing
- `node --test test/unit/distributed_tracing/dt-newrelic-header.test.js` → 26/26
- `node --test test/unit/shim/transaction-shim.test.js` → 48/48
- `node --test test/unit/shim/message-shim.test.js` → 83/83
- `node --test test/unit/lib/message-broker-description.test.js` → 52/52
- Error / metrics-recorder / db / analytics suites → passing
- `npm run versioned:major otel-bridge` → PASS
- `npm run lint` → clean


test/unit/transaction/distributed-trace/incoming-payload.test.js
```diff
@@ -0,0 +1,316 @@
...
+ * `Transaction#acceptDistributedTraceHeaders` constructs it internally.
+ *
+ * @param {Transaction} txn the transaction under test
+ * @returns {IncomingPayload} the payload handler
+ */
+function makeHandler(txn) {
+  return new IncomingPayload({ agent: txn.agent, logger, transaction: txn })
+}
+
+test('DistributedTracePayload#parseAndApply', async (t) => {
```
jsumners-nr, line 28: These tests were migrated from https://github.com/newrelic/node-newrelic/blob/44abb5d4de48b4a17dd46a3ee5378b0467013dd1/test/unit/transaction.test.js#L738

Which is to say, the same tests are still present, just moved to this location.

lib/transaction/index.js
```diff
@@ -74,1353 +56,1128 @@ const PARTIAL_TYPES = {
...
+    this.ignoreApdex = false
+    // flag used to create partial trace when transaction ends
+    this.partialType = null
+    // partial trace is created when transaction ends and if transaction is partialType
+    this.partialTrace = null
+
+    agent.emit('transactionStarted', this)
+  }
+
+  get traceId() {
```
bizob2828, line 156: not following this setter/getter, what's the value of putting this on a symbol? this also used to be assigned lazily, not at construction of transaction

lib/transaction/distributed-trace/payload.js
```diff
@@ -0,0 +1,205 @@
...
+  #trustKey
+  #type
+
+  constructor(input) {
+    if (Object.prototype.toString.call(input) === '[object PayloadData]') {
+      return input
+    }
+
+    for (const k of REQUIRED_DT_KEYS) {
+      if (Object.hasOwn(input, k) === false) {
```
bizob2828, line 44: this used to check if the value wasn't null, this also throws vs before it returned a boolean

lib/subscribers/message-consumer.js
```diff
@@ -19,7 +19,7 @@ const tools = require('#agentlib/subscribers/message-consumer-tools.js')
  * @property {string} system canonical mapping of system(i.e. - Kafka, RabbitMq, SNS, SQS)
  * @property {string} type destinationType: Exchange, Queue, Topic
  * @property {number} callback if consumer is callback based, indicates index of callback
- * @property {string} transport identifier of the transport(see Transaction.TRANSPORT_TYPES)
+ * @property {string} transport identifier of the transport(see the Transport class in lib/transaction/distributed-trace/transport.js)
```
bizob2828, line 22: this also has to be updated in `lib/subscribers/message-consumer-tools.js`

lib/shim/transaction-shim.js
```diff
@@ -37,31 +35,18 @@ util.inherits(TransactionShim, Shim)
...
-/**
- * Enumeration of possible transaction transport types used for distributed tracing.
- *
- * This enumeration is also exposed on the `TransactionShim` class.
- *
- * @readonly
- * @memberof TransactionShim.prototype
- * @enum {string}
- */
-Shim.defineProperty(TransactionShim, 'TRANSPORT_TYPES', Transaction.TRANSPORT_TYPES)
```
bizob2828, line 62: this is public and could break some customers setup

lib/transaction/index.js
```diff
@@ -74,1353 +56,1128 @@ const PARTIAL_TYPES = {
...
-  // If a namestate stack exists, copy route parameters over to the trace.
-  if (!this.nameState.isEmpty() && this.baseSegment && this.agent.config.high_security === false) {
-    this.nameState.forEachParams(forEachRouteParams, this)
-  }
-
-  // Allow the API to explicitly set the ignored status.
-  if (this.forceIgnore !== null) {
-    this.ignore = this.forceIgnore
-  }
+    const transport = Transport.isValid(transportType) ? transportType : Transport.UNKNOWN
```
bizob2828, line 373: this builds a new array everytime, this should be a static check like before.

```
const transport = TRANSPORT_TYPES_SET[transportType] ? transportType : TRANSPORT_TYPES.UNKNOWN
``` 

that set was built once per bootstrapped agent, now it builds a static array every time this method is called

lib/transaction/distributed-trace/payload.js
```diff
@@ -0,0 +1,205 @@
...
+    this.#priority = input.pr
+    this.#sampled = input.sa
+    this.#timestamp = input.ti
+    this.#traceId = input.tr
+    this.#transactionId = input.tx
+    this.#trustKey = input.tk
+    this.#type = input.ty
+  }
+
+  get [Symbol.toStringTag]() { return 'PayloadData' }
```
bizob2828, line 61: how are these setters being used? you're assigning them to the private variables. this class also seems very heavy handed as it's only constructed and serialized. i don't see a time where you would need to reference an individual key. The nested `toJSON` is a little confusing to me. aswell

bizob2828: also not that we would add a new field but if we do, you have to add a `get/set` method and an entry to the `toJSON` method, maybe add a comment if we proceed with this

lib/transaction/distributed-trace/incoming-payload.js
```diff
@@ -0,0 +1,167 @@
...
+  #isDtAlreadyRecorded() {
+    if (this.#transaction.isDistributedTrace !== true) {
+      return false
+    }
+
+    this.#logger.warn(
+      'Already accepted or created a distributed trace payload for transaction %s, ignoring call',
+      this.#transaction.id
+    )
+    const supportabilityMetric = this.#transaction.parentId != null
```
bizob2828: prob fine but this used to just check if it was truthy

bizob2828 (commented):
quite a lot but def a good way to modernize important models in our agent.  I called out a few changes in behavior. to me the riskiest are the traceId getting assigned eagerly vs lazily, the payload key check is no longer `!= null`, and the removal of things that were effectively public

lib/transaction/index.js
jsumners-nr: The purpose of the symbol is to internalize the storage so that folks have to go through the setter to update it. In my opinion, if `class` syntax is the way to do things, then it is better to expose properties in this manner instead of making them directly mutable. But I decided to postpone that for all of the other properties because it would be entirely too much churn. In this specific case, I looked at what the original implementation was doing and simplified it:

https://github.com/newrelic/node-newrelic/blob/b090344f510067efbc650d62bb7b1f62a8ac2537/lib/transaction/index.js#L137-L148

That original code is setting `_traceId` to whatever was passed in or to undefined. Then, when `txn.traceId` is accessed, it determines if a value was set or sets it to a made up hash.  I'm just not following the purpose of that lazy evaluation. In my implementation, we still skip the hash generation if a value is passed in, but we take the compute hit immediately instead of later if one wasn't.

What am I missing in regard to the benefit of the lazy evaluation?

lib/transaction/index.js
jsumners-nr: Fair call out. Commit [2fdf774](https://github.com/newrelic/node-newrelic/pull/4246/commits/2fdf77495b616507cccea81460905374fec1d487) addresses this.

lib/transaction/distributed-trace/payload.js
jsumners-nr: Honestly, I am not a fan of this class. The problem is that we have two very similar payload types: incoming and outgoing. It took a fair bit of noodling to come up with a way to codify their common parts so that we aren't just passing around plain JavaScript objects. All of the getters and setters ended up being needed so that all usage locations didn't necessarily need to be updated. Maybe I got a bit lost in this part.

lib/transaction/index.js
bizob2828: traceId is only passed in for the hybrid agent where trace context exists. regarding impact of lazy vs eager, there are times i had to use our homeboy claude to explain:

```
Why it's often not set at construction time:

- The constructor only receives a traceId argument in one special case — OTel bridge mode, where the id needs to match an existing OTel span's trace id (see the JSDoc at lib/transaction/index.js:82-84). For every normal transaction, traceId is undefined, so _traceId starts as null.
- After that, _traceId only gets populated if the transaction turns out to be part of a distributed trace:
  - acceptTraceContextPayload sets this.traceId = traceparent.traceId when a W3C traceparent header is accepted (lib/transaction/index.js:981).
  - The legacy DT payload path sets this.traceId = data.tr (lib/transaction/index.js:1095).
- Most transactions never receive an inbound DT/traceparent payload at all — they're the root of their own trace, not a continuation of one. For those, nothing ever assigns _traceId.

Why lazy beats eager here:

1. Avoids wasted work. hashes.makeId(32) runs on every single transaction if done eagerly, even for transactions that are ignored, never reported, or whose traceId is never read by any code path. Generating it only on first get means you pay that cost only when something actually needs it.
2. Avoids a generate-then-discard race. DT acceptance (acceptTraceContextPayload/legacy payload handling) happens asynchronously, sometime after the transaction is constructed — e.g., once inbound headers are parsed. If a self-generated id were assigned eagerly at construction, there's a window where that "own" id is live and could be read/reported by other code before the real incoming trace id overwrites it — producing an inconsistent trace id for the same transaction. The lazy getter guarantees: if DT payload acceptance runs before anyone reads traceId, _traceId is already correctly set from the payload and the generator branch never fires; if nothing ever accepts a payload, the getter fabricates one on first access, correctly treating the transaction as trace root.

So the lazy getter isn't just a micro-optimization — it resolves an ordering dependency (DT payload acceptance vs. any code reading traceId) without needing to sequence those two things explicitly.
```

bizob2828: but after reading that, the analysis is kinda wrong. we always do assign a trace id. but i think the lazy creation allows it to be a static string from either otel bridge at construction of transaction, or if the upstream calls had a trace id. so not all transactions generate a random id which is kinda expensive. so in the case where it's part of an upstream trace, we'd be creating a random every time and then stomping on it. that's the main reason we'd want to delay creating the traceId

jsumners-nr: Addressed in [f6a14d6](https://github.com/newrelic/node-newrelic/pull/4246/commits/f6a14d67ab8647391cf6bf9de5a710eda00f5d89).

test/unit/transaction/distributed-trace/transport.test.js
bizob2828, line 25: this changed so test has to be updated

test/unit/transaction/type.test.js
bizob2828, line 39: this changed so test has to be updated

lib/transaction/distributed-trace/payload.js
bizob2828: this needs to change
```suggestion
      if(input[k] != null)
```

lib/subscribers/message-consumer.js
bizob2828: this also has to be addressed

lib/transaction/distributed-trace/payload.js
bizob2828: yea i just don't see how the setters are getting used. you're directly assigning the values to the `#` properties

bizob2828 (changes requested):
a few things missed from original review

lib/transaction/distributed-trace/payload.js
jsumners-nr: The getters and setters are used in places like:

https://github.com/newrelic/node-newrelic/blob/f6a14d67ab8647391cf6bf9de5a710eda00f5d89/lib/transaction/index.js#L235-L241

Notice that `data.id` and `data.tk` are being assigned values.

If we look at the following screenshot, we can see a couple of things that I was hoping to accomplish with this class:

image

1. We have code like `data.tk = 'something'` scattered around and unless you know what `tk` is short for, it's difficult to understand without tracing back through the code.
2. Better IDE integration for these difficult to understand properties. At least enough that you can easily navigate to the object definition and read what "tk" is supposed to map to.

In my opinion, there are two possibilities for improving things in this PR:

1. Update all of the getters and setters to have documentation as shown in the screenshot. This will give people the hints I'm trying to give them with this change.
2. Update all of the usage locations to reference their human readable name, e.g. `.trustKey` over `.tk`. I didn't originally do this because I was trying to limit the production code changes as much as I could.

What are your thoughts? I think deciding on what to do here is the biggest blocker for the PR.

jsumners-nr: The changes I made got rid of all of that verbose nonsense while retaining the developer experience improvement I was going for:

image

Issue #4247, opened 2026-08-27
Support nest.js v12


jsumners-nr:
https://github.com/nestjs/nest-cli/releases/tag/12.0.0

v12.0.0 of `@nestjs/cli` is not supported by our current v11 instrumentation. The module has moved to ESM only. We need to investigate the new code and make changes to our instrumentation accordingly.

Newrelic-Boomi:
https://new-relic.atlassian.net/browse/NR-611061

PR #4248, opened 2026-08-27
chore: Pinned nestjs to unblock CI

jsumners-nr:
Issue #4247 will require more work than I am able to perform right now. This pin will unblock our CI until someone is able to take on the task in that issue.


Ak-x (approved):
Approved

PR #4252, opened 2026-08-31
test: Add `type: module` for `@nestjs/core` tests, including version 12

### LINKED ISSUE #4247: Support nest.js v12

amychisholm03:
## Description

`@nestjs/core` version 12 now [ships as ESM-only modules](https://docs.nestjs.com/migration-guide#esm-packages). With our current test suite that builds an example CJS Nest.js application, version 12 fails. This PR modfies the Nest.js app to be `type:module` that tests versions 10 and above, including 12. Our instrumentation already works for both CJS and ESM, so no instrumentation changes were needed, just tests.

## How to Test

```
npm run versioned nestjs
```

## Related Issues

Closes #4247


bizob2828 (approved):
LGTM. we now have a blindspot for CJS instrumentation but I think that's fine.

PR #4256, opened 2026-08-31
feat: Updated GCP hostname to support more Cloud Run instance types

### LINKED ISSUE #4254: Look for other GCP env vars to identify instance hostname
### Description
If `config.utilization.gcp_cloud_run.use_instance_as_host` is true, we use `K_REVISION` to assign the host. Customers on ruby agent reported this not being enough for worker pools or jobs.  We should now look for: `K_REVISION`, `CLOUD_RUN_REVISION` or `CLOUD_RUN_EXECUTION` to assign host when `config.utilization.gcp_cloud_run.use_instance_as_host` is set to true.

### Acceptance Criteria
A a GCP user that has `config.utilization.gcp_cloud_run.use_instance_as_host` set to true. I want the hostname to be pulled from the appropriate GCP env vars to identify the host of the running service.

### Additional Context
Originates from this [spec update](https://source.datanerd.us/agents/agent-specs/pull/832/files)

amychisholm03:
## Description

Spec 832 requires us to look at other Google Cloud Run environment variables to craft our special GCP hostname override if the coordinating config options are set to `true` (`gcp_use_instance_as_host` and `gcp_cloud_run.include_revision_in_host`). Currently, we only use `K_REVISION` if `include_revision_in_host` is `true`, but that env var only applies to services. **This PR adds support for worker pools** (via `CLOUD_RUN_REVISION`) **and jobs** (via `CLOUD_RUN_EXECUTION`). The order to check these env vars is `K_REVISION`, then `CLOUD_RUN_REVISION`, and then finally `CLOUD_RUN_EXECUTION`.

Also, we were using `K_SERVICE` before we constructed the hostname with the revision to detect if the app is in a GCP environment. However, jobs and workers pools will not have `K_SERVICE` set, they will have `CLOUD_RUN_JOB` and `CLOUD_RUN_WORKER_POOL` set instead, [according to Google's docs](https://docs.cloud.google.com/run/docs/container-contract#env-vars). We cannot get rid of this detection because then this Cloud-Run-specific logic would apply to other GCP services.

Also, in `gcp-info.js`, we needed to modify the boolean to determine if we were in Cloud Run vs another GCP service to include jobs and worker pools.

Follow up:
- [X] ~~Our `utilization.gcp_use_instance_as_host` config option differs from the spec's `utilization.gcp_cloud_run.use_instance_as_host`. What would be the impact on customers if we renamed the config option to be spec-compliant?~~ Solved in PR https://github.com/newrelic/node-newrelic/pull/4258

## How to Test

```
node --test test/unit/collector/facts.test.js
```

## Related Issues

Closes #4254

### DIFF
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -562,8 +562,8 @@ defaultConfig.definition = () => {
        * Deprecated and will be removed in v15 of the agent.
        * Please use `utilization.gcp_cloud_run.use_instance_as_host` instead.
        *
-       * When enabled, it will use GCP metadata id
-       * to set the hostname of the running application.
+       * When enabled, it will use the GCP metadata id to set the hostname
+       * of the running application (Services, Worker Pools, and Jobs).
        */
       gcp_use_instance_as_host: {
         default: true,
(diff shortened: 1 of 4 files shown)


lib/config/default.js
```diff
@@ -559,8 +559,8 @@ defaultConfig.definition = () => {
       },
 
       /**
-       * When enabled, it will use GCP metadata id
-       * to set the hostname of the running application.
+       * When enabled, it will use the GCP metadata id to set the hostname
+       * of the running application (Services, Worker Pools, and Jobs).
        */
       gcp_use_instance_as_host: {
```
bizob2828, line 568: can you file a ticket to duplicate this to the spec'd version detailed [here](https://source.datanerd.us/agents/agent-specs/pull/809). Then we can announce this is deprecated and remove in next major release, and just check both for now

bizob2828 (dismissed):
LGTM, we should hold off on merging until the spec is merged

lib/config/default.js
amychisholm03: Please see https://github.com/newrelic/node-newrelic/pull/4258

bizob2828 (approved):

PR #4258, opened 2026-09-01
chore: Deprecated `utilization.gcp_use_instance_as_host` in favor of `utilization.gcp_cloud_run.use_instance_as_host`

amychisholm03:
## Description

Adhering to Spec 832, the config option `utilization.gcp_use_instance_as_host` needs to be renamed to `utilization.gcp_cloud_run.use_instance_as_host` to match its intended sibling `utilization.gcp_cloud_run.include_revision_in_host`. This PR **deprecates** `gcp_use_instance_as_host` and adds `gcp_cloud_run.use_instance_as_host`; they are intentional duplicates. In the next major version of the agent (v15), we will completely remove `gcp_use_instance_as_host`.

## How to Test

```
node --test test/unit/collector/facts.test.js
```

## Related Issues

N/A, precursor to issue #4254.

### DIFF
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -559,6 +559,9 @@ defaultConfig.definition = () => {
       },
 
       /**
+       * Deprecated and will be removed in v15 of the agent.
+       * Please use `utilization.gcp_cloud_run.use_instance_as_host` instead.
+       *
        * When enabled, it will use GCP metadata id
        * to set the hostname of the running application.
        */
@@ -577,7 +580,15 @@ defaultConfig.definition = () => {
         include_revision_in_host: {
           default: false,
           formatter: boolean
-        }
+        },
+        /**
+         * When enabled, it will use GCP metadata id
+         * to set the hostname of the running application.
+         */
+        use_instance_as_host: {
+          default: true,
+          formatter: boolean
+        },
       }
     },
     transaction_tracer: {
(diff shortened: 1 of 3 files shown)


lib/config/index.js
```diff
@@ -1048,7 +1048,12 @@ function getHostnameSafe(gcpId = null) {
     if (config.heroku.use_dyno_names && process.env.DYNO) {
       _hostname = process.env.DYNO || os.hostname()
       logger.info('Using Heroku dyno name for host name: %s', _hostname)
-    } else if (config.utilization.gcp_use_instance_as_host && process.env.K_SERVICE && gcpId) {
+    } else if (
+      // TODO: `gcp_use_instance_as_host` is deprecated (will be removed
+      // in v15 of the agent) in favor of `gcp_cloud_run.use_instance_as_host`.
+      (config.utilization.gcp_cloud_run.use_instance_as_host ||
```
bizob2828: you prob want to encapsulate these checks into a simple function for now

bizob2828 (changes requested):

lib/config/default.js
```diff
@@ -559,6 +559,9 @@ defaultConfig.definition = () => {
       },
 
       /**
+       * Deprecated and will be removed in v15 of the agent.
```
bizob2828, line 562: you may want to have a `warnOnce` in `lib/config/index.js` when it constructs config.  I'm thinking of some way to associate the old and new and automatically log warnings when it compiles config but not sure the best way atm

bizob2828 (commented):
it looks like one of my comments didn't get submitted in that first round

lib/config/default.js
amychisholm03: Hm would the warning be useful because `use_instance_as_host` is true by default? It's not like the customer is manually enabling it.

bizob2828: I think it's moreso that it's set to false or we just don't care

bizob2828 (approved):

PR #4263, opened 2026-09-03
feat: Added support for `MessageBroker/Kafka/Cluster/{cluster_id}/{Produce|Consume}/{topic_name}` metrics

### LINKED ISSUE #4255: Add kafka cluster produce/consume metrics
### Description
Adds two new count metrics to the existing Kafka metrics the instrumentation produces, but only add when `config.kafka.metrics.cluster.metrics.enabled` is true:

`MessageBroker/Kafka/Cluster/{cluster_id}/Produce/{topic_name}`
`MessageBroker/Kafka/Cluster/{cluster_id}/Consume/{topic_name}`

We also have to add this new configuration.

Why this is necessary: Customers running multiple Kafka clusters frequently reuse the same topic names across clusters, which makes the existing per-topic throughput metrics (MessageBroker/Kafka/Topic/...) ambiguous — there is currently no way to break produce/consume throughput down by cluster in NRQL.

Neither the broker server address nor the topic name is enough to identify a self-hosted Kafka cluster. The existing per-node metrics (MessageBroker/Kafka/Nodes/{server_name}/...) key off the broker address, but in self-hosted deployments different clusters living in separate VPCs/networks routinely share the same broker addresses (private IPs/hostnames such as 10.0.0.10:9092 or kafka-1:9092 are reused across isolated networks) and the same topic names. So (server address + topic) can collide across completely different clusters, and throughput can't be reliably attributed.

The Kafka cluster id (sourced from the Kafka client's existing metadata/admin API) is a globally unique identifier independent of network addressing, so adding it as a dimension disambiguates throughput even when broker addresses and topic names overlap.

The section also specifies the behavioral contract the implementations follow: best-effort (skip silently until the cluster id resolves), one metadata fetch per cluster (deduplicated by bootstrap-servers), retry-with-backoff, and must-not-crash. Capture is additive — it does not replace the Nodes metric, which is emitted unconditionally regardless of cluster id availability.

### Additional context
Related to this [spec update](https://source.datanerd.us/agents/agent-specs/pull/817/files). Also a [PR](https://github.com/newrelic/node-newrelic/pull/4044/changes) was started by another team. we plan to use this as a framework for our solution

amychisholm03:
## Description

Adhering to spec 817 and as an alternative solution to https://github.com/newrelic/node-newrelic/pull/4044, this PR adds support for `MessageBroker/Kafka/Cluster/{cluster_id}/{Produce|Consume}/{topic_name}`
metrics for `kafkajs` instrumentation. The Node agent already supports similar metrics for Kafka "Nodes" (`MessageBroker/Kafka/Nodes/...`) in `lib/subscribers/kafkajs/utils/record-linking-metrics.js`, so `recordLinkingMetrics` was modified to also increment the `Cluster` metrics at the same call frequency as the `Nodes` metrics.

In order to get the `cluster_id`, I had to instrument kafkajs's private `createCluster` factory on the `Client` instance it assigns onto itself), since `cluster_id` is never exposed anywhere on the `Client`, `producer`, or `consumer` objects. Wrapping that factory lets me get a handle on each `Cluster` it constructs — one per `.producer()`/`.consumer()`/`.admin()` call — and from there wrap `brokerPool.refreshMetadata` so that whenever kafkajs performs a metadata refresh on its own (e.g. on `send()`/`subscribe()`), the resolved `clusterId` gets cached on the client's shared `kafkaCtx`. `recordLinkingMetrics` just reads that cached value when it's called — it doesn't resolve anything itself.

This feature is enabled with the `kafka.metrics.cluster.metrics.enabled` config option and is disabled by default.

## How to Test

```
npm run versioned kafkajs
```

## Related Issues

Closes #4255


lib/subscribers/kafkajs/client-constructor.js
```diff
@@ -125,6 +128,44 @@ module.exports = class ConstructorSubscriber extends Subscriber {
...
+    }
+    if (!cluster?.brokerPool || typeof cluster.brokerPool.refreshMetadata !== 'function') {
+      return
+    }
+
+    const kafkaCtxObj = instance[kafkaCtx]
+    const brokerPool = cluster.brokerPool
+    const origRefresh = brokerPool.refreshMetadata.bind(brokerPool)
+    brokerPool.refreshMetadata = async function nrRefreshMetadata(...args) {
+      const result = await origRefresh(...args)
```
bizob2828:
```suggestion
      const result = await origRefresh.apply(this, args)
```

lib/subscribers/kafkajs/cluster-constructor.js
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * Picks up the data returned from the `kafkajs.Cluster` constructor.
+   *
+   * @param {SubscriberHandlerData} data Data from Orchestrion.
+   * @param {SubscriberHandlerContext} ctx Context from Orchestrion.
+   *
+   * @returns {SubscriberHandlerContext}
+   */
+  end(data, ctx) {
+    ClusterConstructorSubscriber.pendingCluster.current = data.self
```
bizob2828: this is not ideal. we have to find something to attach this to other than our own code

lib/subscribers/kafkajs/cluster-constructor.js
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * Picks up the data returned from the `kafkajs.Cluster` constructor.
+   *
+   * @param {SubscriberHandlerData} data Data from Orchestrion.
+   * @param {SubscriberHandlerContext} ctx Context from Orchestrion.
+   *
+   * @returns {SubscriberHandlerContext}
+   */
+  end(data, ctx) {
```
bizob2828: end does not emit ctx

bizob2828 (commented):
Make sure you PR docs website for new config option

bizob2828 (approved):
LGTM

PR #4269, opened 2026-09-08
feat: Added `browser_monitoring.version` configuration option

### LINKED ISSUE #4262: Add `config.browser_monitoring.version`
### Description
Spec 835 - This is a request to allow customers to specify the browser agent version to load. At the moment they always get the latest.  Not only should we allow this to be configurable, default is empty string(defaults to latest), but it should be sent up as part of the settings on connect(this will happen automatically in our agent)

### Acceptance Criteria
Add a new configuration option `browser_monitoring.version`, default to empty string.

amychisholm03:
## Description

Added support for the `browser_monitoring.version` configuration option, which lets customers request a specific browser agent loader version (e.g. `"1.317.0"`) instead of always receiving the latest. Per the collector spec, this value is customer-supplied, defaults to `""` (unset), and must be included in the settings map sent up on connect — the collector is responsible for resolving it and returning a matching loader.

To support the collector's version resolution behavior, this PR also:
- Wires up `browser_monitoring.loader_version` so the resolved version reported back by the collector on connect is actually stored on the config, rather than being silently dropped as an unknown key.
- Adds a local warning, logged when `browser_monitoring.version` is set but the collector does not return a `js_agent_loader`, since that combination usually means the requested version could not be resolved. The warning correctly excludes the two other cases where a missing loader is expected and not an error: `browser_monitoring.loader` set to `"none"`, and browser monitoring disabled (`browser_monitoring.enable` is `false`).

## How to Test

```
node --test test/unit/config/*.test.js
node --test test/unit/agent/agent.test.js
```

## Related Issues
Closes #4262

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -770,7 +770,26 @@ Agent.prototype._enabledChange = function _enabledChange() {
  * always occurs after handling a response from a connect call.
  */
 Agent.prototype._configChange = function _configChange() {
+  this._checkBrowserMonitoringVersion()
   this.collector.reportSettings()
+}
+
+/**
+ * Logs a local warning if a customer-requested browser agent loader version
+ * was not resolved by the collector. A missing loader does not necessarily
+ * mean the requested version was bad, since the loader is also omitted when
+ * `browser_monitoring.loader` is `'none'` or browser monitoring is disabled,
+ * so those cases are excluded.
+ */
+Agent.prototype._checkBrowserMonitoringVersion = function _checkBrowserMonitoringVersion() {
+  const bm = this.config.browser_monitoring
+  if (bm.version && bm.enable && bm.loader !== 'none' && !bm.js_agent_loader) {
+    logger.warn(
+      'browser_monitoring.version was set to "%s" but no js_agent_loader was returned by ' +
+        'the collector; the requested version may not have been resolved.',
+      bm.version
+    )
+  }
 }
 
 Agent.prototype._addIntrinsicAttrsFromTransaction = _addIntrinsicAttrsFromTransaction
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -715,6 +715,7 @@ Agent.prototype.reconfigure = function reconfigure(configuration) {
   }
 
   this.config.onConnect(configuration)
+  this._checkBrowserMonitoringVersion()
 }
 
 /**
@@ -770,7 +771,6 @@ Agent.prototype._enabledChange = function _enabledChange() {
  * always occurs after handling a response from a connect call.
  */
 Agent.prototype._configChange = function _configChange() {
-  this._checkBrowserMonitoringVersion()
   this.collector.reportSettings()
 }
 
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -715,7 +715,6 @@ Agent.prototype.reconfigure = function reconfigure(configuration) {
   }
 
   this.config.onConnect(configuration)
-  this._checkBrowserMonitoringVersion()
 }
 
 /**
@@ -771,6 +770,7 @@ Agent.prototype._enabledChange = function _enabledChange() {
  * always occurs after handling a response from a connect call.
  */
 Agent.prototype._configChange = function _configChange() {
+  this._checkBrowserMonitoringVersion()
   this.collector.reportSettings()
 }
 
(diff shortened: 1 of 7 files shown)


lib/config/default.js
```diff
@@ -816,6 +816,15 @@ defaultConfig.definition = () => {
...
         env: 'NEW_RELIC_BROWSER_MONITOR_DEBUG'
+      },
+      /**
+       * The browser agent loader version to request, e.g. `"1.317.0"`. See the
+       * [browser agent EOL policy](https://docs.newrelic.com/docs/browser/browser-monitoring/getting-started/browser-agent-eol-policy/)
+       * for which versions are currently available and supported.
+       */
+      version: {
+        default: '',
+        env: 'NEW_RELIC_BROWSER_MONITOR_VERSION'
```
bizob2828: this doesn't need specified and it'd be wrong anyways, it's browser_monitoring

lib/config/index.js
```diff
@@ -276,9 +276,31 @@ Config.prototype.onConnect = function onConnect(json, recursion) {
...
 }
 
+/**
+ * Logs a local warning if a customer-requested browser agent loader version
+ * was not resolved by the collector. A missing loader does not necessarily
+ * mean the requested version was bad, since the loader is also omitted when
+ * `browser_monitoring.loader` is `'none'` or browser monitoring is disabled,
+ * so those cases are excluded.
+ */
+Config.prototype._checkBrowserMonitoringVersion = function _checkBrowserMonitoringVersion() {
```
bizob2828: I'm not following what this is doing

lib/config/index.js
```diff
@@ -276,9 +276,31 @@ Config.prototype.onConnect = function onConnect(json, recursion) {
     this._fromServer(json, key)
   }
 
+  if (!recursion) {
+    this._checkBrowserMonitoringVersion()
```
bizob2828: we emit a change event. which is handled in lib/agent.js. you should probably just add code in there to log this warning instead of directly in the onConnect

bizob2828 (changes requested):

lib/config/default.js
amychisholm03: I was just the following the other `browser_monitoring` configs: https://github.com/newrelic/node-newrelic/blob/19b22cfafee6af2cbc0ea46b947b37e9c030de6e/lib/config/default.js#L807-L819 , but those are probably special cases.

bizob2828: looking at our docs, the only ones that are `BROWSER_MONITOR` are debug and enable

lib/config/index.js
amychisholm03: It was mostly just logging the warning: [link to spec](https://source.datanerd.us/agents/agent-specs/blob/main/BAM-Agent-Auto-Instrumentation-PORTED.md#:~:text=Agents%20MAY%20also,is%20false).

lib/agent.js
```diff
@@ -770,9 +770,28 @@ Agent.prototype._enabledChange = function _enabledChange() {
  * always occurs after handling a response from a connect call.
  */
 Agent.prototype._configChange = function _configChange() {
+  this._checkBrowserMonitoringVersion()
```
bizob2828: i was trying to test this and i don't see it working. however i noticed this gets called twice. `config.onConnect` is emitting change but also there's a configuration option emitting change as well which isn't necessary:

```js
    // interpret AI Monitoring account setting.
    // `ai_monitoring.enabled` from the connect response server-side config(SSC) takes precedence over
    // SSC settings come in an `agent_config` object
    // `collect_ai`; collect_ai is only the fallback when it is absent.
    case 'collect_ai':
      if (!params?.agent_config?.['ai_monitoring.enabled']) {
        this._disableOption(params.collect_ai, 'ai_monitoring')
        this.emit('change', this)
      }
      break
```

amychisholm03: Works with `"1.321.0"` but not `"1.314.0"`; this is likely a server-side issue.

lib/agent.js
```diff
@@ -715,6 +715,7 @@ Agent.prototype.reconfigure = function reconfigure(configuration) {
   }
 
   this.config.onConnect(configuration)
+  this._checkBrowserMonitoringVersion()
```
bizob2828: this may happen too early. i'd put it back in `_configChange`

bizob2828 (changes requested):
I had to test with prod because not all versions live on staging.  one small tweak regarding where the warning should be called though

bizob2828 (approved):

PR #4272, opened 2026-09-09
fix: Fixed duplicated router mount path in transaction names for async middleware

jsumners-nr:
## Problem

After upgrading from `13.5.0` → `14.3.x`, transaction names duplicate a router's mount path when an `async` middleware is registered at the same path as a router. For example:

- Expected: `WebTransaction/Expressjs/GET//resources/:resourceId/:itemId/details`
- Broken: `WebTransaction/Expressjs/GET//resources/:resourceId/resources/:resourceId/:itemId/details`

The mount path `/resources/:resourceId` appears twice.

## Root cause

The express instrumentation was rewritten as `diagnostics_channel` subscribers in `13.6.0` (#3424). In the shared `lib/subscribers/middleware-wrapper.js`, `wrapDoneHandler` only wrapped the `next`/`done` callback for **synchronous** middleware. The route/mount path appended by an `async` middleware was therefore never popped from the transaction name state, so a subsequent middleware or router registered at the same path appended the identical prefix a second time.

## Fix

The `next`/`done` callback is now wrapped for async handlers as well, so the path is popped (and errors handled) regardless of whether the handler is sync or async. Context re-entry via `runInContext` remains sync-only, since async handlers already propagate context through their awaited promise.

## Testing

Added a versioned regression test (`test/versioned/express/async-middleware-naming.test.js`) that registers an async middleware and a router at the same param mount path and asserts the transaction name is not duplicated.

With the **broken** code, the test's request to `/resources/abc/123/details` produced:

```
WebTransaction/Expressjs/GET//resources/:resourceId/resources/:resourceId/:itemId/details
```

instead of the expected:

```
WebTransaction/Expressjs/GET//resources/:resourceId/:itemId/details
```

Verified passing after the fix across the shared middleware wrapper's consumers: express (4.22.2, 5.2.1), connect (3.7.0), fastify (3/4/5), koa (2/3), hapi (20/21), and the `middleware-wrapper` unit tests.

### DIFF
--- a/lib/subscribers/middleware-wrapper.js
+++ b/lib/subscribers/middleware-wrapper.js
@@ -189,11 +189,11 @@ class MiddlewareWrapper {
   /**
    * Wraps the done/next handler so that it pops the middleware's route from the
    * transaction name state and handles any error passed to it. When the handler
-   * is synchronous the wrapped done also restores and re-enters the previous
+   * is synchronous the wrapped `done` also restores and re-enters the previous
    * context; async handlers propagate context via their returned promise so
-   * only the path pop / error handling is needed.
+   * only the `path pop` / `error` handling is needed.
    *
-   * The path pop must happen for async handlers too. Otherwise a route/mount
+   * The path pop must happen for async handlers too. Otherwise, a route/mount
    * path appended by an async middleware is never removed, and a subsequent
    * middleware or router registered at the same path duplicates the prefix in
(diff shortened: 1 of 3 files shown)


bizob2828 (approved):
ran all webframework versioned tests, this doesn't seem to impact any other libraries. people to whacky stuff with express so i can't predict if this is going to break anything, but LGTM

PR #4278, opened 2026-09-15
fix: Updated `undici` instrumentation to not attempt to create an external segment when transaction is not active

### LINKED ISSUE #4276: Undici subscriber creates traceparent with undefined span ID when no external segment is created
## Description

The Undici subscriber can attempt to create a `traceparent` header with an undefined span ID when an external segment is not created.

I observed this warning from `newrelic@14.3.4`:

```text
WARN (newrelic/1): Unable to create external segment
Error: received invalid parent id: undefined
    at new Traceparent (.../newrelic/lib/w3c/traceparent.js:35:13)
    at Traceparent.fromSpanContext (.../newrelic/lib/w3c/traceparent.js:98:12)
    at TraceContext.createTraceparent (.../newrelic/lib/transaction/tracecontext.js:48:39)
    at TraceContext.addTraceContextHeaders (.../newrelic/lib/transaction/tracecontext.js:132:30)
    at Transaction.insertDistributedTraceHeaders (.../newrelic/lib/transaction/index.js:922:21)
    at UndiciSubscriber.addDTHeaders (.../newrelic/lib/subscribers/undici/index.js:74:19)
    at UndiciSubscriber.requestCreateHook (.../newrelic/lib/subscribers/undici/index.js:52:12)
```

The failure path appears to be:

1. `requestCreateHook` continues when the context contains either a segment or a transaction.
2. `createExternalSegment` calls `tracer.createSegment`, which can return `null` if there is no parent segment or the transaction is no longer active.
3. `addDTHeaders` still passes `{ spanId: segment?.id }` to `insertDistributedTraceHeaders`.
4. `Traceparent.fromSpanContext` rejects the resulting undefined parent ID.

The warning is caught by the subscriber and the request continues, but the external segment and distributed tracing headers are not created for that request.

The same nullable access is still present on the current `main` branch and in `14.4.0`.

## Expected Behavior

The subscriber should not attempt to construct a span context with an undefined span ID. If no external segment was created, it should either skip header injection or use the transaction's normal trace-context fallback behavior.

No NR Diag output is available. The stack and source path above consistently identify the Undici subscriber as the source of the warning.

## Steps to Reproduce

I do not yet have a standalone reproducer. In a Node service making outbound `fetch` requests, trigger a request while New Relic's async context contains a transaction but no usable parent segment, or a transaction that has ended before the Undici `request:create` hook runs.

The source path can also be exercised by allowing `createExternalSegment` to receive such a context and return without assigning `request[undiciSegment]`, followed by `addDTHeaders` with distributed tracing enabled.

## Your Environment

* New Relic Node agent: 14.3.4
* Node.js: 24.17.0
* Operating system: Alpine Linux 3.23, running in a container

## Additional context

This is different from #2950. That issue concerned a valid but incorrect parent ID. Here the subscriber passes an undefined span ID after segment creation returns `null`.

bizob2828:
## Description

Our undici instrumentation only checks if there is a transaction or segment, or if segment is opaque before skipping creating an external segment. Creating an external segment will fail if the transaction is not active. So this PR just updates the check to ensure transaction is active and current segment is not opaque.

## How to Test

```sh
node test/unit/lib/subscribers/undici.test.js
npm run versioned undici
```

## Related Issues
Closes #4276

**Note**: It looks like this behavior was borrowed from http outbound and replicated in undici and http/2. I will add follow up PRs.  Those do not fail with the same error because in undici we pass in the span context which doesn't check if transaction is active.


lib/subscribers/undici/index.js
```diff
@@ -37,10 +37,11 @@ class UndiciSubscriber extends DcBase {
     const context = agent.tracer.getContext()
     const { segment, transaction } = context
     request[undiciParent] = segment
-    if (!(segment || transaction) || segment?.opaque) {
+    if (!transaction?.isActive() || segment?.opaque) {
       this.logger.trace(
-        'Not capturing data for outbound request (%s) because parent segment opaque (%s)',
+        'Not capturing data for outbound request (%s) because transaction is not active %s or parent segment opaque (%s)',
```
jsumners-nr:
```suggestion
        'Not capturing data for outbound request (%s) because transaction is not active (%s) or parent segment is opaque (%s)',
```

lib/subscribers/undici/index.js
```diff
@@ -37,10 +37,11 @@ class UndiciSubscriber extends DcBase {
...
         request.path,
+        transaction?.isActive(),
         segment?.name
```
jsumners-nr: I'm guessing the final `(%s)` is meant to receive a `true` or `false` string:

```suggestion
        segment?.opaque === true
```

jsumners-nr (dismissed):
Log can be improved, but otherwise LGTM.

jsumners-nr (approved):

PR #4284, opened 2026-09-18
chore: Migrated configuration to JSON Schema

### LINKED ISSUE #4226: Update our default configuration `lib/config/default.js` to be a json schema

jsumners-nr:
I really wanted to put `getOrCreateInstance` and `initialize` in their own modules, and then export a constructed plain object from `lib/config/index.js`. But all of the tests assume the `Config` object has the functions attached as methods. So I compromised with what I think is a cleaner structure for a reduction in blast radius.

I have not reviewed each individual schema file. They are all built from the schema file contributed by Melissa. I first instructed the bot to copy that schema over and build a loader around it. Then I instructed it to break the large file down into the separated components. This should have resulted in all files containing the same data as the original definition (aside from changes needed to complete this task). Our tests bear this assumption out.

Speaking of the schema contributed by Melissa, this work contains a mechanism for replacing that file via refactoring the existing script, or replacing it with a new one, to invoke `schema.renderedSchema` and writing the result out as `JSON.strinify`-ied data to a file. I went ahead and performed this work instead of doing it as a follow up. A follow up would have required keeping the decommissioned `default.js` around (along with everything depending upon it). And the generator would not have been accurate until the follow up work was done. So we could have shipped a bad artifact once this PR is merged.

One consequence of the JSON Schema approach that we need to be mindful of going forward is that when we add a new instrumentation, `lib/config/schemas/instrumentation.js` needs to be updated with a corresponding block. We should follow up with a CI check or something that will keep us consistent here.

-----

## Summary

Makes a JSON Schema the authoritative, complete definition of every agent
configuration option, replacing the bespoke `lib/config/default.js`
`definition()` machinery. The schema (JSON Schema Draft 2020-12) is authored as
per-block JavaScript modules under `lib/config/schemas/`, and the `Config`
object is rebuilt to derive its defaults, environment-variable handling,
coercion, and validation from that schema.

Closes #4226.

## What changed

- **Schema as source of truth.** `lib/config/schemas/root.js` plus per-block
  modules (`attributes.js`, `distributed-tracing.js`, `error-collector.js`, …)
  describe every setting: type, default, constraints, and environment-variable
  bindings. Vendor keywords (`x-newrelic-env-var`, `x-newrelic-coerce`,
  `x-newrelic-internal`, `x-newrelic-sampler`) carry agent-specific metadata.
  See `lib/config/schemas/README.md`.
- **Config construction rebuilt** (`lib/config/index.js`): defaults come from
  `generate-default-config.js`; environment overrides from
  `apply-environment-overrides.js`; validation/coercion from `schema.js` +
  `validate-config.js`. Server-side config application (`onConnect`) and
  host-name resolution were ported onto the new class.
- **ajv type coercion** for basic scalar types (boolean/integer/number) is
  enabled on the validating schema, so a passed config value like
  `high_security: 'true'` is coerced to a real boolean. The more complicated
  coercions (object/objectList/regex/allowList) remain applied to environment
  variables by the existing formatters.
- **Bug fixes surfaced while migrating the unit suites:** serverless-mode
  auto-detection now happens at construction time (not module load); GCP/Heroku
  host-name resolution corrected; `rules.ignore` schema now allows regular
  expressions; environment-variable values are redacted when logged.
- **Unit tests updated** across config, samplers, shims, spans, attributes,
  logger, high security, and feature flags to exercise the new public seams
  (`onConnect`, `{ logger }` constructor option, construction-time high
  security) rather than removed private methods.
- **`lib/config/default.js` retired.** With the schema authoritative, the old
  `definition()` module and the now-dead `build-instrumentation-config.js` (plus
  its test) were removed. The Fleet Control schema generator
  (`.fleetControl/schemaGeneration/generate-schema.js`) was rewritten to build
  `.fleetControl/schemas/config.json` by pruning `schema.js`'s `renderedSchema`
  (dropping `x-newrelic-internal` settings and stripping
  `$id`/`x-newrelic-coerce`/`x-newrelic-sampler`, while keeping
  `x-newrelic-env-var`) instead of walking `default.js`. The
  `agent-config-schema` workflow now triggers on `lib/config/schemas/**`. The
  benchmark's `generate-full-config.js` (also `default.js`-based) was removed;
  its `full-config.json` fixture is unchanged.

## Behavior changes to note

- **Passed-config booleans must be real booleans or the exact strings
  `"true"`/`"false"`.** ajv's coercion is stricter than the legacy
  `formatters.boolean`, which also accepted `"yes"`, `"no"`, `"on"`, `"off"`,
  `"enabled"`, `"disabled"`, `"0"`, `"1"`, etc. Those lenient forms passed
  directly in the config object / `newrelic.js` now fail validation.
  **Environment variables are unaffected** — `NEW_RELIC_*` values still go
  through `formatters.boolean`.
- **Invalid `distributed_tracing.sampler.*.adaptive.sampling_target`** (out of
  `[1,120]` or non-integer) is now rejected by validation rather than silently
  clamped/floored.

## Testing

- `npm run unit` for the affected config and consumer suites is green.
- Each migrated suite was run individually and verified passing.

## Performance

`test/benchmark/lib/config/index.bench.js` compared against the pre-refactor
`default.js` config (the benchmark's introducing commit), same machine and
harness, `runs: 5000`, averaged over 3 runs. Times are ms per parse (median;
lower is better):

| fixture | original | new | change |
|---|---|---|---|
| baseline (suggested `newrelic.js`) | 0.201 | 0.111 | −44.7% |
| complex (object samplers, coerced scalars, env vars) | 0.220 | 0.114 | −48.4% |
| full (every field) | 1.514 | 1.484 | −2.0% |

Realistic configs (baseline and complex — the common load path) parse ~45–48%
faster at the median; the maximal "every field" config is within noise (~2%).
Both versions were measured with a no-op logger to exclude logging I/O. (The
benchmark's logger option was corrected from `{ loggerInstance }` to `{ logger }`
in this PR so it measures what it claims.)

### One-time cold start

The per-parse figures above are steady state: the schema is read from disk and
compiled into an ajv validator **once per process**, then cached, so individual
`Config` builds hit the warm cache. That build is a new cost the pre-refactor
config did not have (it derived defaults from an in-memory `definition()` object
with no file reads or validator compile). Measured with the same methodology
(build the default config from a cold module cache, median of 5000 runs):

| cold start | original | new | change |
|---|---|---|---|
| build defaults from cold | ~0.37 ms | ~44.8 ms | +~44 ms |

This ~44 ms is a **one-time, per-process startup cost** paid on first config
load, not per parse. It does not appear in the per-parse numbers above; the
benchmark's `cold schema build` case measures the new side of it directly.

## Follow-ups

- **Two cross-agent sampler cases are skipped.** In
  `test/integration/samplers/config.test.js`,
  `sampling_target_specified_uses_new_sampler_instance` and
  `no_ratio_falls_back_to_global_sampler` are skipped. Their configs (from the
  vendored cross-agent `sampler_configuration.json`) are malformed under the
  strict JSON schema — a `trace_id_ratio_based` without a `ratio`, and a sampler
  object with an unrecognized key. The cross-agent spec expects these to be
  tolerated and fall back to the global sampler, but schema validation runs
  before the sampler formatters (`buildSamplers`) and rejects them. Needs a
  decision on reconciling strict schema validation with the cross-agent
  fallback contract.
- **One cross-agent trace-context case is skipped** for the same reason. In
  `test/integration/distributed-tracing/trace-context-cross-agent.test.js`,
  `no_headers_root_uses_adaptive_sampler_when_no_ratio` supplies a
  `trace_id_ratio_based` sampler without a `ratio`; the spec expects a fall back
  to the adaptive sampler, but schema validation rejects it first.


lib/config/index.js
```diff
@@ -1,1849 +1,1976 @@
...
+   *
+   * @throws {Error} When the configuration cannot be validated or read from
+   * disk.
+   *
+   * @returns {AgentConfig} Instantiated configuration object.
+   */
+  static initialize(config, { logger = defaultLogger } = {}) {
+    if (config != null) {
+      validateConfig(config)
+      return new (this)(config, { logger })
```
jsumners-nr, line 247: This construct is strange, and likely not necessary, but is used for extensibility. See https://gist.github.com/jsumners/ed8597a20363c395e45b011bb22ac4f1

lib/config/schema.js
```diff
@@ -0,0 +1,265 @@
...
+  // keywords (e.g. x-newrelic-internal); we're validating config data, not the
+  // schema itself. `coerceTypes: true` lets ajv coerce basic scalar types
+  // (boolean/integer/number) in place during validation, so a passed config
+  // value like `high_security: 'true'` becomes `true` rather than failing the
+  // `type: boolean` check. Coercion happens within the same validation walk, so
+  // there is no second parse. The more complicated coercions
+  // (object/objectList/regex/allowList) are not handled here; those remain the
+  // job of the formatters applied to environment variables in
+  // `apply-environment-overrides.js`.
+  const ajv = new Ajv({ allErrors: true, strict: false, coerceTypes: true })
```
jsumners-nr, line 80: `allErrors` is typically flagged as not recommended because it can blow up AJV's parser and/or leak details. In our case, I think we want to present all of the errors. And the typical flagging is really geared toward situations where unknown data is received from unknown sources, i.e. through an HTTP endpoint. We do not have that problem.

lib/config/schemas/agent-control.js
```diff
@@ -0,0 +1,38 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+module.exports = {
+  $schema: 'https://json-schema.org/draft/2020-12/schema',
+  $id: 'agent-control.js',
+  type: 'object',
+  'x-newrelic-internal': true,
```
bizob2828, line 12: i don't see this being used at all

lib/config/schemas/README.md
```diff
@@ -0,0 +1,104 @@
...
+  `additionalProperties`, so per-package settings for packages beyond the
+  enumerated set are permitted but are not individual schema settings.
+- **`NEWRELIC_`-prefixed variables are out of scope.** A few special/test
+  variables use the `NEWRELIC_` (no underscore) prefix and are read directly by
+  their owning modules (e.g. gRPC test metadata, `NEWRELIC_PIPE_PATH`). They are
+  intentionally not configuration settings and are not represented here.
+
+## Adding or changing a setting
+
+1. Edit the relevant block file (or `root.js` for a top-level scalar). Give the
```
bizob2828, line 100: can we require `description`

lib/config/schemas/instrumentation.js
```diff
@@ -0,0 +1,708 @@
+/*
+ * Copyright 2026 New Relic Corporation. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+'use strict'
+
+module.exports = {
```
bizob2828, line 8: this should be dynamic like the past. it can be done since this is gets evaluated at require time

lib/config/index.js
```diff
@@ -1,1849 +1,1974 @@
...
     }
 
-    return filteredLabels
-  }, {})
-}
+    let userConf
+    try {
+      userConf = require(filepath).config
+    } catch (error) {
+      logger.error(error)
```
bizob2828, line 275: is the `ERR_REQUIRE_ESM` no longer applicable? 

```
    if (error.code === 'ERR_REQUIRE_ESM') {
      // Attempted to import newrelic.js or similar from an ESM module, error out early.
      throw error
    }
```

lib/config/validate-config.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  const details = validate.errors
+    .map((error) => {
+      const location = error.instancePath || '<root>'
+      return `  ${location} ${error.message}`
+    })
+    .join('\n')
+
+  throw new Error(`New Relic configuration failed schema validation:\n${details}`)
```
bizob2828, line 39: I get this is the most ideal situation, to throw errors on invalid values but this is technically a breaking change.  Should we just log warnings for now and announced in the next major release we will throw errors and not start? Also I may be wrong but this failure is only happening with file based config. for env vars we just skip applying the values if it fails validation

lib/config/apply-environment-overrides.js
```diff
@@ -0,0 +1,157 @@
...
+      targetConfig,
+      resolution.pathSegments,
+      coerceValue(rawValue, resolution.node, logger)
+    )
+
+    // A distributed-tracing sampler set to a string ('trace_id_ratio_based' or
+    // 'adaptive') is expanded into its object form from a second environment
+    // variable (e.g. ..._TRACE_ID_RATIO_BASED_RATIO). Delegate that to the
+    // sampler helper, which reads the second variable and rewrites the value.
+    if (resolution.node['x-newrelic-sampler'] === true) {
```
bizob2828, line 150: i get we used to have this before but can't we be more generic with setting env vars from diff data types?

lib/config/index.js
```diff
@@ -1,1849 +1,1974 @@
...
-  }
+  /**
+   * Create a new configuration instance. If a configuration object is passed in,
+   * it will be validated and used to return a new {@link Config} object without
+   * attempting to load values from the standard locations.
+   *
+   * When a base configuration object is not passed in, the standard config
+   * locations will be loaded in order (from a `newrelic.js` file):
+   *
+   * 1. The process's current working directory.
```
bizob2828: this order is not what's happening based on `find-config-file`

1. NEW_RELIC_HOME
2. cwd
3. HOME
4. node_modules
5. REPL main module

lib/config/samplers.js
```diff
@@ -247,5 +247,6 @@ function getNestedValue(obj, keys) {
 module.exports = {
   config,
```
bizob2828: this is no longer being used and the entire config above can be removed

lib/config/index.js
```diff
@@ -1,1849 +1,1974 @@
...
+      // The passed in configuration contains a key that is not defined in
+      // our default configuration, and we are not accepting such keys.
+      // So move on to the next.
+      return
+    }
 
-Config.prototype.getIPAddresses = function getIPAddresses() {
-  const addresses = Object.create(null)
-  const interfaces = os.networkInterfaces()
+    if (isValidDTSampler(key) === true) {
```
bizob2828: I'm unclear why this is getting checked since it's gettined checked in buildSamplers

lib/config/index.js
```diff
@@ -1,1849 +1,1974 @@
...
+   * Normalizes the application name on assignment, from whatever source (user
+   * config, environment, Azure APP_POOL_ID, server-side config). A string is
+   * split on `;`/`,` and trimmed into an array; arrays and null/undefined are
+   * stored as-is.
+   *
+   * @param {string|Array|null} value The application name(s).
+   */
+  set app_name(value) {
+    this.#appName = typeof value === 'string'
+      ? value.split(/[;,]/).map((name) => name.trim())
```
bizob2828, line 1951: this was only done before on env var. i think because you couldn't represent an array like. youcan in file.  I'm sure this is fine but just calling out this is diff behavior

lib/config/build-instrumentation-config.js
```diff
@@ -1,39 +0,0 @@
-/*
- * Copyright 2024 New Relic Corporation. All rights reserved.
- * SPDX-License-Identifier: Apache-2.0
- */
-
-'use strict'
```
bizob2828, line 6: this should be restored and used to dynamically build the json schema for instrumentation

lib/config/schema.js
```diff
@@ -0,0 +1,269 @@
...
+/**
+ * Loads the root schema and, lazily, compiles a reusable ajv validator. The
+ * root schema references each config block as `<block>.js`; every block
+ * schema in `schemas/` is registered with ajv so those refs resolve before root
+ * is compiled. Reading the directory rather than a hardcoded list means new
+ * block files are picked up automatically. Compiled once and cached.
+ *
+ * @returns {AgentConfigSchema} An object representing the schema.
+ */
+function build() {
```
bizob2828, line 45: I ran the benchmarks and it's ~29ms vs ~1.3ms for dynamically building this vs just building it and storing it that we can use. I think we should do this because that's a significant difference especially for folks in serverless.  We just wire up building out the static file in the release prep and it'll be included in the PR.

lib/config/schemas/transaction-tracer.js
```diff
@@ -0,0 +1,81 @@
...
+      additionalProperties: true
+    },
+    enabled: {
+      'x-newrelic-env-var': 'NEW_RELIC_TRACER_ENABLED',
+      type: 'boolean',
+      default: true,
+      description: "Whether to collect & submit slow transaction traces to New Relic. The instrumentation is loaded regardless of this setting, as it's necessary to gather metrics. Disable the agent to prevent the instrumentation from loading."
+    },
+    transaction_threshold: {
+      'x-newrelic-env-var': 'NEW_RELIC_TRACER_THRESHOLD',
```
bizob2828, line 47: The old had a `parseFloat` for env var.  So now `0.5` via env var is a string not number. The `lib/transaction/trace/aggregator` checks if it's a number so this will be an issue if you don't specify `apdex_f` as the value

lib/config/schemas/root.js
```diff
@@ -0,0 +1,373 @@
...
+        { type: 'object' },
+        { type: 'string' }
+      ],
+      default: {},
+      description: 'Label names and values applied to the data sent from this agent, as an object or a `;`-delimited `key:value` string. Label names and values are truncated to 255 characters and the set is capped at 64.'
+    },
+
+    license_key: {
+      type: 'string',
+      default: '',
```
bizob2828, line 144: should we enforce a minLength of 1 like it was in the fleet control schema, that way it'll fail if empty?

bizob2828 (changes requested):
This was quite a lot but a step in the right direction. Great work.  I have a few questions/comments/suggestions.

lib/config/schemas/agent-control.js
jsumners-nr: It's used when generating the rendered schema to omit internal properties:

https://github.com/jsumners-nr/node-newrelic/blob/109b1d644ddc4ed22cf5cf6a2086b7897b5ae5ba/.fleetControl/schemaGeneration/generate-schema.js#L41-L44

lib/config/schemas/README.md
jsumners-nr: We can, but I have purposefully skipped going through and adding description, or other documentation fields, in this PR. I think that is a prime task for a follow up.

lib/config/index.js
jsumners-nr: I see that I changed the behavior here. Previously, if config was attempted to be loaded from ESM we would hard crash the app and tell them "don't do that." Now, we fall back to the stub mode and agent doesn't collect any data. We currently have a thread about lenient configuration in Slack, and the folks participating are not fond of the agent crashing the app. What do you think we should do here?

lib/config/apply-environment-overrides.js
jsumners-nr: I don't understand the question.

lib/config/validate-config.js
jsumners-nr: We can log warnings, but what should we do about the configuration?

Yes, the environment handling is more lenient because we are able to coerce the data before we feed it into the validator.

lib/config/index.js
jsumners-nr: This is carried over from `main`:

```
On upstream/main:
- JSDoc order (index.js:1701–1707): cwd → main module → NEW_RELIC_HOME → HOME → above node_modules
- Actual order (getConfigFileLocations, index.js:50–57): NEW_RELIC_HOME → cwd → HOME → above node_modules → main module
```

lib/config/index.js
jsumners-nr: This centralizes the logic instead of making it specific to environment parsing.

lib/config/schema.js
jsumners-nr: We agreed that I will do this work in a follow up.

lib/config/schemas/root.js
jsumners-nr: If I recall correctly, I allowed this leniency because of tests.

https://github.com/jsumners-nr/node-newrelic/blob/109b1d644ddc4ed22cf5cf6a2086b7897b5ae5ba/lib/config/schemas/root.js#L27-L31

And here's the bot description of the decision I made that resulted in the current code:

> The chain of reasoning from earlier in this work:
>
> 1. The schema validates the config after env + defaults are merged, at agent runtime — not a static document check. So the schema constraints directly gate whether a `Config` can be constructed.
> 2. `license_key` defaults to `''` and `app_name` to `[]`. An agent legitimately starts up without a license key in many code paths (env-only setups, disabled/serverless modes, and — critically — the huge body of existing unit/integration tests that call `Config.initialize({})` or build configs without ever supplying a real key).
> 3. We deliberately relaxed both the required list and the per-field constraints. We set `required: []` (removed `app_name`/`license_key` from required) and dropped `minLength` from `license_key`, precisely so an empty/default value validates instead of throwing. We went through this exact tradeoff when the config unit suite was failing — the alternative (strict `required` + `minLength: 1`) broke a large number of historical tests that omit these values.

lib/config/schemas/README.md
bizob2828: that's fair, can you capture the follow ups somewhere so we don't lose track of them?

lib/config/apply-environment-overrides.js
bizob2828: we have bespoke logic in our schema specifically for the sampler nuances. is there some way to canonicalize it in case of a diff use case later? maybe not a big deal for now

lib/config/index.js
bizob2828: so we have a test

```
  await t.test('when requiring newrelic.js in ESM app', async () => {
    const { stdout, stderr } = await exec('node index.mjs', { cwd: path.join(__dirname, 'esm-js') })
    if (semver.satisfies(process.version, '>=20.19.0 <22 || >=22.12.0')) {
      match(stdout, 'Hello esm-test')
    } else {
      match(stderr, 'ERR_REQUIRE_ESM', 'should mention ERR_REQUIRE_ESM in error message')
    }
  })
```

which is no longer in play for this.  I think it's safe to remove this logic as it's no longer applicable since you can require esm

lib/config/index.js
bizob2828: right but you could have a situation where you have a `;` in an `app_name` but this is prob fine

bizob2828 (changes requested):
I think the change of needing to parse config to assign a logger but this now creates a logger right away is causing this issue:

```
(node:24383) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 end listeners added to [Logger]. MaxListeners is 10. Use emitter.setMaxListeners() to increase limit
```

```sh
node test/unit/config/config.test.js
node test/unit/agent/agent.test.js
```

The other changes look good. Can you log a ticket to compile the schema to a file and use that? we cannot ship this until that's done because of the cold start hit on compiling the schema

lib/logger.js
```diff
@@ -19,42 +19,46 @@ const logger = new Logger({
...
   const options = {
     name: 'newrelic',
     level: config.logging.level,
     enabled: config.logging.enabled,
     auditLogging: config.audit_log?.enabled === true
   }
-
-  // configure logger
+  // Apply the user supplied logging configuration to the logger instance:
   logger.configure(options)
```
bizob2828: you need a logger.unpipe(). this will help alleviate the leaking of event emitters I called out in my last round of review

PR #4294, opened 2026-09-21
refactor: Consolidated core subscriber configs into a single file

amychisholm03:
While working on the `crypto` refactor, I realized that the "simple" core instrumentations have identical `config.js` structures. We could just have a `core/config.js` file that does this work for us, instead of having a bunch of tiny near-identical config files. This PR implements that idea. 

Of course, for more complicated core libraries, like `http`, having more than one subscriber file (and thus breaking out of the current identical config structure) will likely be required, but in that case that library will just have their own `config.js` that will be added to `lib/subscriber-configs.js`.

I also renamed `child_process/index.js` and `dns/index.js` to just `child_process.js` and `dns.js` inside `lib/subscribers/core` because they don't really need a folder now. Again, `http` likely will need a folder, but I'm not sure if folders are necessary for these single-file instrumentations.


lib/subscribers/core/README.md
```diff
@@ -34,4 +34,10 @@ class FakeCoreSubscriber extends BaseCoreSubscriber {
 module.exports = FakeCoreSubscriber
 ```
 
+Add the module's name to `CORE_PACKAGES` in `config.js` so the subscriber is
+picked up. It resolves to `./core/<name>`, so a simple module lives at
+`core/<name>.js` and one that grows extra files can become `core/<name>/index.js`
+without a config change. A module needing more than one subscriber should get its
+own entry in `lib/subscriber-configs.js` instead.
```
jsumners-nr: I agree that `lib/subscribers/core/config.js` should export a structure that slots in with all of the other existing non-core configs. But I don't think we should complicate matters with "if A do B, but if C do D" as illustrated in the last sentence here.

lib/subscribers/core/config.js
```diff
@@ -0,0 +1,16 @@
...
+
+// Core modules that have been migrated to a subscriber. Each name resolves to
+// `./core/<name>`, so a subscriber can be a single file or a directory with an
+// `index.js`. A module needing more than one subscriber gets its own entry in
+// `lib/subscriber-configs.js` instead.
+const CORE_PACKAGES = ['child_process', 'dns']
+
+module.exports = Object.fromEntries(
+  CORE_PACKAGES.map((pkg) => [pkg, [{ path: `./core/${pkg}`, instrumentations: [] }]])
+)
```
jsumners-nr: Instead of being clever, which is leading to my previous comment, we should simply export a static object. This will be easier to read and maintain.

jsumners-nr (approved):
Looks good to me.

